PR #4, opened 2020-06-19
Make eslint configurable

harrysolovay:
In this fork, the `Eslint` construct accepts options (or undefined, in which case it falls back to the original config).

Here are the options:

```ts
interface EslintOptions {
  config: Linter.Config; // from definitely typed `@types/eslint`
  dependencies: Record<string, Semver>;
}
```

Here's an example of its usage:

```diff
import { Eslint } from './eslint';
import { JsiiProject } from './jsii-project';
import { Semver } from './semver';

const project = new JsiiProject({
+ eslint: false,
  name: 'projen',
  jsiiVersion: Semver.caret('1.5.0'),
  description: 'A new generation of project generators',
  repository: 'https://github.com/eladb/projen.git',
  authorName: 'Elad Ben-Israel',
  authorEmail: 'benisrae@amazon.com',
  stability: 'experimental',
  dependencies: {
    'constructs': Semver.pinned('2.0.1'),
    'yaml': Semver.caret('1.9.2'),
    'fs-extra': Semver.caret('9.0.1'),
  },
  peerDependencies: {
    constructs: Semver.pinned('2.0.1'),
  },
  devDependencies: {
    '@types/fs-extra': Semver.caret('8.0.1'),
  },
  bundledDependencies: [ 'yaml', 'fs-extra' ],
  projenDevDependency: false, // because I am projen
  workflowBootstrapSteps: [
    { run: 'yarn bootstrap' },
  ],
});
  
+ new Eslint(project, /* ESLint options */);

project.synth();
```


eladb:
@harrysolovay mind rebasing against master?

harrysolovay:
Will get around to this soon 👍

harrysolovay:
@eladb just updated. Also made sure it's working.

src/eslint.ts
```diff
@@ -3,14 +3,20 @@ import { JsonFile } from './json';
 import { NodeProject } from './node-project';
 import { Semver } from './semver';
 
-export class Eslint extends Construct {
+export interface EslintOptions {
```
eladb, line 6: please add docstrings

src/eslint.ts
```diff
@@ -3,14 +3,20 @@ import { JsonFile } from './json';
...
+  // (not "dev" dependencies, given that its relative to the construct, not the end workspace)
+  public readonly dependencies: Record<string, Semver>;
+  public readonly config: Record<string, any>;
 
-  constructor(project: NodeProject) {
+  constructor(project: NodeProject, options?: EslintOptions) {
     super(project, 'eslint');
 
-    project.addDevDependencies({
+    this.dependencies = options?.dependencies || {
```
eladb, line 19: Perhaps this should just be overrides?

```ts
this.dependencies = {
  // default deps
  ...options.dependencies
};
```

So people can override any default and add theirs

src/eslint.ts
```diff
@@ -3,14 +3,20 @@ import { JsonFile } from './json';
...
+}
 
-  public readonly rules: { [rule: string]: any[] };
+export class Eslint extends Construct {
+  // (not "dev" dependencies, given that its relative to the construct, not the end workspace)
+  public readonly dependencies: Record<string, Semver>;
+  public readonly config: Record<string, any>;
 
-  constructor(project: NodeProject) {
+  constructor(project: NodeProject, options?: EslintOptions) {
```
eladb, line 16: Saves some `?`s:

```suggestion
  constructor(project: NodeProject, options: EslintOptions = { }) {
```

src/eslint.ts
```diff
@@ -19,7 +25,74 @@ export class Eslint extends Construct {
       'eslint-import-resolver-typescript': Semver.caret('2.0.0'),
       'eslint-plugin-import': Semver.caret('2.20.2'),
       'json-schema': Semver.caret('0.2.5'), // required by @typescript-eslint/parser
-    });
+    }
+
+    this.config = options?.config || {
```
eladb, line 30:
```suggestion
    this.config = options?.config ?? {
```

src/eslint.ts
```diff
@@ -19,7 +25,74 @@ export class Eslint extends Construct {
...
+        },
+      },
+      ignorePatterns: [
+        '*.js',
+        '*.d.ts',
+        'node_modules/',
+        '*.generated.ts',
+        'coverage',
+      ],
+      rules: {
```
eladb, line 65: I feel it makes sense to also expose a `rules` option for just adding rules. Having to specify the entire configuration just to specify the rules is not ideal.

src/eslint.ts
harrysolovay: I did so for the options. I don't see other construct classes "tsdocked". Let me know if there are other places in need of commenting.

src/eslint.ts
harrysolovay: So, the ability to add deps without loosing the defaults?

src/eslint.ts
harrysolovay: Good call!

src/eslint.ts
harrysolovay: 👍

src/eslint.ts
harrysolovay: I took a slightly different approach to enabling this. Let me know if a specific `rules` option is preferable. For now, I've added an `overrides` option: a dictionary of fns which accept the previous value of a config field and return the updated. One could pass the following for `overrides`:

```ts
{
  rules: (previous) => {
    ...previous,
    quotes: [
      "error",
      "double",
      {
        "avoidEscape": true
      }
    ]
  },
}
```

harrysolovay:
Gonna take another crack at this later. In the meantime, please let me know how you feel about the approach wherein a dictionary of pure override fns can be supplied. This could work for dependencies as well. The options interface would become...

```ts
interface EslintOptions {
  // would specify a generic identity fn type... but JSII throws error
  overrides: Record<string, any>;
}
```

Thoughts?

eladb:
Closing for now. Feel free to pick up any time

harrysolovay:
Cool cool. @eladb please let me know your thoughts on the comments above.

PR #8, opened 2020-07-06
feat(ts): docgen / more docstrings

andrestone:
*Description of changes:*
- Simple docgen via typedoc for TS projects
- Refactored the jsii-specific `tsconfig.jest.json` settings to live inside `jsii-project`
- compilerOptions interface
- `charset` is [deprecated](https://www.typescriptlang.org/tsconfig#charset)
- fix for the `mkdirSync` fix (now recursive)

BREAKING CHANGE: `options` is now required for `TypescriptConfig`.

### DIFF
--- a/src/typescript.ts
+++ b/src/typescript.ts
@@ -5,6 +5,7 @@ import { Eslint } from './eslint';
 import { Semver } from './semver';
 import { Mergify, MergifyOptions } from './mergify';
 import { Construct } from 'constructs';
+import { TypedocDocgen } from './typescript-typedoc';
 
 export interface TypeScriptLibraryProjectOptions extends NodeProjectOptions {
   /**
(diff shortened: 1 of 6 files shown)


src/typescript.ts
```diff
@@ -43,11 +44,37 @@ export interface TypeScriptLibraryProjectOptions extends NodeProjectOptions {
...
+   * 
+   * @default 'API'
+   */
+  readonly docsDirectory?: string;
+
+  /**
+   * Custom TSConfig
+   * 
+   */
+  readonly tsconfigOptions?: TypescriptConfigOptions;
```
eladb, line 66: Let's rename to just `tsconfig`?

src/typescript.ts
```diff
@@ -115,16 +169,186 @@ export interface TypescriptConfigOptions {
    *
    * @default - see above
    */
-  readonly compilerOptions?: any;
+  readonly compilerOptions: TypescriptConfigCompilerOptions;
+}
+
+export interface TypescriptConfigCompilerOptions {
```
eladb, line 175:
```suggestion
export interface TypeScriptCompilerOptions {
```

src/version.ts
```diff
@@ -41,7 +41,7 @@ export class Version extends Construct {
     const versionFile = `${this.project.outdir}/${VERSION_FILE}`;
     if (!fs.existsSync(versionFile)) {
       if (!fs.existsSync(this.project.outdir)) {
-        fs.mkdirSync(this.project.outdir);
+        fs.mkdirSync(this.project.outdir, { recursive: true });
```
eladb, line 44: I think we need a newer node.js for this. Take a dependency on `fs-extra`?

PR #13, opened 2020-07-09
feat: lerna project

andrestone:
*Issue #, if available:*
- closes https://github.com/eladb/projen/issues/9

see README.

---


src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+    // New JsonFile for package.json
+    // ...
+
+
+  }
+
+  // new project with outdir on the package location, validate workspaces and locations using minimatch
+  public addPackage(name: string, location: string, project: NodeProject) {
+    // Run validation for this._options.rootScripts
+    const packageJson = Node.of(project).tryFindChild('package.json') as JsonFile; // This or expose props on NodeProject ?
```
eladb, line 29: Expose props

andrestone:
I will. It's that I needed to poke around a bit first in order to get a better feel.

src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+          process.stderr.write(`LernaProject: cannot continue as root script ${k} doesn't exist on package ${name}\n`);
+          process.exit(1);
+        }
+      }
+    }
+    // Run validations on the location using minimatch / glob patterns (location must match the glob pattern)
+
+    // We need to somehow change the outdir...
+    try {
+      project.synth({outdir: location});
```
eladb, line 42: `synth` should be done as part of the parent project's `synth()`. Otherwise, mutations to the sub-project will be lost.

src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+    } catch (e) {
+      process.stderr.write(`Error when synthesizing lerna package: ${e}\n`);
+      throw e;
+    }
+    this._packages.push(project);
+  }
+
+  public addBootstrappedPackage(name: string, location: string, npxCommand: string, createDir?: boolean) {
```
eladb, line 50: What's that? doc strings will help with the review

andrestone:
But if you could comment at least on the two points of the TODO, it would really help me organize my thoughts.

src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.addPackage(name, location, NodeProject.fromExternalScript(npxCommand, name, createDir))
+  }
+
+  public async addPackageFromTemplate(name: string, location: string, templateName: string) {
```
eladb, line 54: ?? Not sure I understand the use case / workflow here.

src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.addPackage(name, location, template.project as NodeProject);
+  }
+}
+
+export interface LernaProjectOptions extends NodeProjectOptions {
+
+  /**
+   * Packages to create
+   */
+  readonly packages: [{name: string, location: string, project: NodeProject}];
```
eladb, line 72: Feels like `location` should be optional and default to `packages/$name`, no?

src/lerna.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Scripts runnable via 'lerna run'.
+   * Projen will fail if any script is missing on any of the packages.
+   *
+   */
+  readonly rootScripts?: string[];
```
eladb, line 79: This should be have a similar API and funnel into the same path that manages scripts in `NodeProject`.

src/node-project.ts
```diff
@@ -367,7 +373,7 @@ export class NodeProject extends Project {
         antitamper: options.antitamper,
       });
 
-      if (options.releaseToNpm) {
+      if (options.releaseToNpm && !options.private) {
```
eladb, line 376: I thought we already had support for `private`. At any rate, shouldn't that also go into `package.json`?

src/node-project.ts
```diff
@@ -544,6 +550,14 @@ export class NodeProject extends Project {
       '.cache',
     );
   }
+
+  public static fromExternalScript(npxCommand: string, name: string, createDir?: boolean): NodeProject {
```
eladb, line 554: I am failing to understand what this is about

eladb:
> Create outdir parameter for Project.synth() / another solution for synthesizing to a different outdir

I think something like that would work, yes. Also see my comment about centralized synthesis.

> Interface with NodePorject use construct node to get the JsonFile object or expose NodeProject properties?

Yes, add properties to expose the various things. `package.json`, by the way, can already be controlled via `project.manifest` (bad name, yes).

> Logic for projects from scripts (should this belong to a separate PR? Could be generally used to bootstrap a simple NodeProject using the provided external script and projen overrides)

Yes, sounds like a separate PR, and I am not sure I fully understand the workflow around it, so worth a discussion before you dive into it. At any rate, I would not include that here.

src/lerna.ts
andrestone: Hmm... What if it isn't part of the root construct? I'm thinking of it like a `nestedStack`, which follows a different `synth` route. It makes it easier to manipulate the `outdir`. 

Another option would be having it as part of the tree but then I'm not sure about what the logic would be to make it synth to a different outdir (maybe extending a `FileBase` just for Lerna or go passing props up?).

As a third option would be having a `.projenrc.js` file created on the `location` and having projen to calling itself there. But this one feels pretty bad.

src/lerna.ts
andrestone: README incoming

andrestone:
> Yes, add properties to expose the various things. `package.json`, by the way, can already be controlled via `project.manifest` (bad name, yes).

Should I always try exposing a property as a first solution?

andrestone:
I have included a first README. I know this should be splitted in two PR's, please share your thoughts on if and where it's worth doing so.

src/lerna.ts
andrestone: The intention here is just to ensure the scripts exist in the packages. Is it worth it to have a method to add identical scripts to all child packages?

src/node-project.ts
andrestone: Moved to NodeProject (README).

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+
+#### Simple example - External Boilerplate
 
+Let's assume there isn't a Projen template to initialize a CDK construct package powered by jsii. Here is how 
+we could easily implement one:
 
+```typescript
+const { NodeProject } = require('projen');
+
+const cdkConstruct = NodeProject.fromExternalScript({
```
eladb, line 193: I don't like this api and I don't see how this is related to lerna. Can we please get rid of it for now?

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+be able to initialize your monorepo packages with this template, making small adjustments for each
+package as your monorepo grows:
+```typescript
+const { LernaProject, NodeProject } = require('projen');
+
+const lernaProject = new LernaProject({
+  /* ... */
+});
+
+const myService = NodeProject.fromTemplate({
```
eladb, line 223: Why is this needed?

eladb: Templates are only a way to bootstrap a projenrc file

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+const lernaProject = new LernaProject({
+  name: 'my-monorepo-project',
+  workspaces: {
+    packages: [
+      "packages/@web/*",
+      "packages/@rn/*",
+      "packages/@iac/*",
+      "packages/@services/*",
+    ],
+    nohoist: [],
```
eladb, line 272: Example?

I actually think this api is too low level. Can we provide some higher level support like `lerna.noHoist(package, dep)`

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+#### What about the packages?
+
+In order to add packages to your monorepo, you can pass the `packages` property:
+```typescript
+const { LernaProject, TypeScriptLibraryProject } = require('projen');
+
+const lernaProject = new LernaProject({
+  name: 'my-monorepo-project',
+  workspaces: {
+    packages: [
```
eladb, line 288: This feels redundant since we know exactly where are all the packages so we should generate this based
On the packages added to the project and not through a wildcard.

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+      "packages/@web/*",
+      "packages/@rn/*",
+      "packages/@iac/*",
+      "packages/@services/*",
+      "packages/*",
+    ],
+    nohoist: [],
+  packages: [
+    {
+      location: 'packages/@services',
```
eladb, line 298: This should be optional and default to `packages/PROJECTNAME`

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+`LernaProject` (before `synth()`):
+
+```typescript
+const { LernaProject, TypeScriptLibraryProject } = require('projen');
+
+const lernaProject = new LernaProject({
+  /* ... */
+});
+
+lernaProject.addPackage({
```
eladb, line 327: The minimum api should be 

```
lerna.addPackage(project);
```

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+  }),
+});
+
+lernaProject.synth();
+```
+>When omitted, `workspaces` will have a default packages pattern of `packages/*`. Likewise, `location` 
+>will default to `packages`. If you're willing to use the default location when adding packages and 
+>want to specify additional packages location patterns, don't forget to include `packages/*`.
+
+#### Using Projen templates and external scripts
```
eladb, line 341: Please get rid of this feature for now. Let's discuss later

README.md
```diff
@@ -177,10 +177,221 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+  location: 'packages/@web',
+  project: myLib,
+});
+
+lernaProject.synth()
+```
+
+More details on templates and external scripts [here](#templates-and-external-scripts). 
+
+#### Enforcing the use of root level scripts
```
eladb, line 372: Let's remove this feature for now and add later. Ok?

In the meantime users can easily achieve this by passing in the same set of scripts to all projects (through a local variable)

README.md
andrestone: Yep! It isn't really related, I just hijacked the PR to get your thoughts on it (sorry).

README.md
andrestone: This isn't needed, it's a related feature. I'll just discard it and add to another PR, as suggested.

README.md
andrestone: The idea here is to just have a way to validate package locations vs workspaces paths. How could we validate it without declared workspace patterns? Should we just add any arbitrary package location as a workspace?

README.md
andrestone: It is described like so.

andrestone:
- Requested
- Version file is now created during synth
- Preserve field functionality for `JsonFile` to support the above (array of fields not to overwrite on `synth`)
- `LernaVersion` to *optionally* get version for package.json via `lerna.json` and not create `.versionrc.json`
- Moved `main` key definition out of `NodeProject` (to jsii / typescript)
- Cleaned up package.json output a bit
- First iteration of a cdk8s monorepo mock `examples/lerna`

README.md
```diff
@@ -177,11 +177,93 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
 
+const lernaProject = new LernaProject({
+  /* ... */
+});
+
+lernaProject.addPackage(new TypeScriptLibraryProject({ 
+    name: 'geocoding-service', 
+    description: 'Geocoding Service' 
+  }),
+);
```
eladb, line 239: Style:

```suggestion
lernaProject.addPackage(new TypeScriptLibraryProject({ 
  name: 'geocoding-service', 
  description: 'Geocoding Service' 
}));
```

README.md
```diff
@@ -177,11 +177,93 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+  name: 'geocoding-service', 
+  description: 'Geocoding Service' 
+});
+
+const lernaProject = new LernaProject({
+  /* ... */
+});
+lernaProject.addPackage(myChildProject);
+lernaProject.noHoist('any-package', 'yaml') // generates 'any-package/yaml' nohoist entry
+lernaProject.noHoist(myChildProject, 'aws-sdk') // generates 'geocoding-service/aws-sdk'
```
eladb, line 262: Will it also include the '/**' for hoisting all transitive dependencies?

README.md
```diff
@@ -177,11 +177,93 @@ That's it. projen will auto-detect `bin/mycli` and will add it to your
...
+>will default to `packages`. If you're willing to use the default location when adding packages and 
+>want to specify additional packages location patterns, don't forget to include `packages/*`.
+
+You can use `lernaProject.noHoist(package, dep)` to tell yarn which packages can't be
+hoisted.
+
+```typescript
+const { LernaProject, TypeScriptLibraryProject } = require('projen');
+
+const myChildProject =  new TypeScriptLibraryProject({
```
eladb, line 252: Renamed:

```suggestion
const myChildProject =  new TypeScriptProject({ 
```

examples/lerna/.projenrc.ts
```diff
@@ -0,0 +1,139 @@
+import { LernaProject, Jest, JsiiProject, Semver, TypescriptConfig } from '../../lib';
+const common = require('./projen-common');
+
+// monorepo
+const cdk8sProject = new LernaProject({
+  name: 'cdk8s',
+	lernaVersioning: false,
```
eladb, line 7: indentation is off here

package.json
```diff
@@ -73,7 +71,8 @@
     "node": ">= 10.17.0"
   },
   "license": "Apache-2.0",
-  "version": "0.3.3",
+  "version": "0.3.2",
```
eladb, line 74: Seems like a downgrade

src/file.ts
```diff
@@ -63,6 +63,19 @@ export abstract class FileBase extends Construct {
       fs.chmodSync(filePath, '600')
     }
 
+    // Preserving existing field on JSON file (e.g version)
+    // not using JsonFile to avoid circular reference issue
+    if ((this as any).preserveJSONFields?.length) {
```
eladb, line 68: What's that?

src/jsii-project.ts
```diff
@@ -137,6 +137,8 @@ export class JsiiProject extends TypeScriptLibraryProject {
 
     this.addFields({ types: `${libdir}/index.d.ts` });
 
+    this.addFields({ main: `${libdir}/index.js` });
```
eladb, line 140: Let's move these two to the `TypeScriptProject` class.

src/json.ts
```diff
@@ -2,18 +2,35 @@ import { FileBase, FileBaseOptions } from './file';
...
+  /**
+   * Object to be parsed into JSON.
+   *
+   */
   readonly obj: any;
+
+  /**
+   * Fields to preserve from existing file.
+   */
+  readonly preserveJSONFields?: string[];
```
eladb, line 14: Mmmm. That's an interesting approach. Does it change how we manage versions? Perhaps requires a separate PR?

src/lerna-project.ts
```diff
@@ -0,0 +1,88 @@
+import { NodeProject, NodeProjectOptions } from './node-project';
+import { JsonFile } from './json';
+import { Semver } from './semver';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+export class LernaProject extends NodeProject {
+  public readonly packages: NodeProject[];
```
eladb, line 8: docstrings for all public APIs please (yes I know I don't have them everywhere, yet, but let's start).

src/node-project.ts
```diff
@@ -234,6 +240,16 @@ export interface NodeProjectOptions extends ProjectOptions, CommonOptions {
    * Additional entries to .npmignore
    */
   readonly npmignore?: string[];
+
+  /**
+   * Use lerna for versioning
+   */
+  readonly lernaVersioning?: boolean;
```
eladb, line 247: feels weird that `NodeProject` knows about lerna.

src/node-project.ts
```diff
@@ -286,21 +302,22 @@ export class NodeProject extends Project {
...
       },
       'bin': this.bin,
       'scripts': this.scripts,
-      'author': {
-        name: options.authorName,
-        email: options.authorEmail,
-        url: options.authorUrl,
-        organization: options.authorOrganization ?? false,
-      },
+      'author': options.authorName ?
```
eladb, line 314: How is this related to this PR?

src/node-project.ts
```diff
@@ -312,6 +329,7 @@ export class NodeProject extends Project {
 
     new JsonFile(this, 'package.json', {
       obj: this.manifest,
+      preserveJSONFields: ['version'], // this is managed by another script
```
eladb, line 332: Extract to a separate PR

src/project.ts
```diff
@@ -19,7 +19,7 @@ export class Project extends Construct {
     this.gitignore = new IgnoreFile(this, '.gitignore');
   }
 
-  public synth(): void {
-    Node.of(this).synthesize({ outdir: this.outdir });
+  public synth(outdir?: string): void {
+    Node.of(this).synthesize({ outdir: outdir || this.outdir });
```
eladb, line 23:
```suggestion
    Node.of(this).synthesize({ outdir: outdir ?? this.outdir });
```

src/typescript.ts
```diff
@@ -94,6 +94,8 @@ export class TypeScriptLibraryProject extends NodeProject {
     this.docgen = options.docgen;
     this.docsDirectory = options.docsDirectory || 'docs/';
 
+    this.addFields({ main: `${this.libdir}/index.js` });
```
eladb, line 97: the `.d.ts` should go here too and then it's not required in `JsiiProject` anymore

src/version.ts
```diff
@@ -38,14 +49,29 @@ export class Version extends Construct {
...
+        if (!fs.existsSync(path.join(this.project.outdir, this.versionFile))) {
+          return '0.0.0';
+        }
+        return JSON.parse(fs.readFileSync(path.join(this.project.outdir, this.versionFile), 'utf-8')).version;
+      },
+    })
+  }
+}
+
+export class LernaVersion extends Construct {
```
eladb, line 63: Let's have a separate discussion about versioning and bumping and releases....

templates/jsii/.template/template.js
```diff
@@ -0,0 +1,9 @@
+const { JsiiProject } = require('projen');
+
+const project = new JsiiProject({
```
eladb, line 3: please remove the templates

src/file.ts
andrestone: If I import `JsonFile` class to be used in the `if` condition (`this instanceof JsonFile`) I get this:
`TypeError: Class extends value undefined is not a constructor or null`

README.md
andrestone: Not by default, will update the README.

andrestone:
Okay.

Couple things that I needed to fix to have it working:

`generateSample()`'s current behaviour is creating files as a side effect of the class instantiation. Refactored it to happen on `project.synth()` so the files are generated according to the `outdir` of the synth session.

Likewise, `Version.current()` was creating an initial `.version.js` file when instanitiated, also refactored so it happens on synth using session `outdir`.

adamelmore:
Any plans to finish this out? I'd also like to see Lerna support and I'm happy to help if I can!

andrestone:
The project has evolved a lot since this PR was proposed.

It's probably better to start fresh.

adamelmore:
@andrestone care if I take this on, or did you still plan to?

andrestone:
Please do it!

PR #36, opened 2020-07-29
feat: do not depend on "constructs"

eladb:
As much as it's romantic, it didn't add much value to this project.

Implement a simple decentralized synthesis model where "Component" has a "synthesize" method
that accepts an output directory and can emit files into that directory.


andrestone:
Revolutions!

PR #65, opened 2020-08-15
Allow using default bootstrap steps in 3rd party projects

skorfmann:
Usually copy and paste would be good enough, but since this sets the explicit
projen version, it's probably the easiest to make the `DEFAULT_WORKFLOW_BOOTSTRAP`
accessible in other contexts as well.

### DIFF
--- a/src/node-project.ts
+++ b/src/node-project.ts
@@ -780,7 +780,7 @@ export interface PeerDependencyOptions {
   readonly pinnedDevDependency?: boolean;
 }
 
-const DEFAULT_WORKFLOW_BOOTSTRAP = [
+export const DEFAULT_WORKFLOW_BOOTSTRAP = [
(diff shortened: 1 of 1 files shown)


src/node-project.ts
```diff
@@ -780,7 +780,7 @@ export interface PeerDependencyOptions {
   readonly pinnedDevDependency?: boolean;
 }
 
-const DEFAULT_WORKFLOW_BOOTSTRAP = [
+export const DEFAULT_WORKFLOW_BOOTSTRAP = [
```
eladb, line 783: If we want to expose this, let's expose as a static const of `NodeProject`

eladb:
@skorfmann Want to peruse this or should we close for now?

eladb:
Already supported

PR #91, opened 2020-08-31
feat(node): automatic dependency management

### LINKED ISSUE #90: Permission denied on package.json
I'm adding projen to my CDK construct library using the `ConstructLibraryAws` project. As my construct includes a NodeJsFunction, it fails during tests as it can't access the package.json file. If I change the permissions it works until they are resetted by `npx projen`.

>    EACCES: permission denied, open '.../package.json'

eladb:
Introduce `devDeps`, `bundledDeps` and `deps` which are a simple list of module names without version numbers. When first introduced, `projen` will install the latest version (similarly to `yarn add`) and will record the installed version through a caret dependency in `package.json`.

If `package.json` already includes a version for a `dep`-type dependency, the `package.json` value will be preserved, which means that tools like `yarn upgrade` and *dependabot* will be able to update them.

- If `package.json` is removed and `pj` is executed, the latest versions of all modules will be used.
- `package.json` is no longer read-only (resolves #90) so it can be edited by tools (yarn, dependabot). Anti-tamper will ensure other areas of the file were not changed.
- No need to run `yarn install` after `npx projen` anymore since yarn is now executed automatically. So now `alias pj = 'npx projen'`.
- `bundledDeps` will update both `dependencies` and `bundledDependencies`.
- Added resiliency for when `package.json` is corrupted (esp. relevant in merge conflicts): just run `npx projen` and all generated files will be rewritten.
- You can only use `yarn add` to update an existing dependency. To add dependencies, you'll need to add it to the dependency list in `.projenrc.js`.

Example:

```js
bundledDeps: [
  'yaml',
  'fs-extra',
  'yargs',
  'decamelize',
  'glob',
  'inquirer',
],

devDeps: [
  '@types/fs-extra',
  '@types/yargs',
  '@types/glob',
  '@types/inquirer',
],
```

Missing features/tests:
- [x] Support versioning requirements and then that should be what's recorded in package.json (`deps: [ 'foo@^2' ]`).
- [x] Verify full backwards compatibility with projects that use the old-style dependencies
- [ ] Dependabot - if all deps are version-less, dependabot config should allow updates of package.json
- [x] `addDeps` - imperative API for adding named dependencies.
- [ ] Control version mode (pinned/tilde/caret).


skorfmann:
Sweet, that'll be a great improvement!

> * Support versioning requirements and then that should be what's recorded in package.json (`deps: [ 'foo@^2' ]`).

This would be really good to have, would solve a problem in http://cdk.tf/provider

Issue #124, opened 2020-09-28
Feature request: Support for private / non-default repositories


aripalo:
## Problem

Currently it seems that Projen only supports the public repositories (npmjs.com, pypi.org, maven central). 

This is controlled by `jsii-project.ts`:
- for [NPM in lines 299-300](https://github.com/eladb/projen/blob/ef47eebc3115a02a82c63fb546bb5f23ab2c58e0/src/jsii-project.ts#L299-L300):
   ```ts
   env: {
      NPM_TOKEN: '${{ secrets.NPM_TOKEN }}',
      NPM_DIST_TAG: this.npmDistTag,
   },
   ```
- for [Python in lines 400-401](https://github.com/eladb/projen/blob/ef47eebc3115a02a82c63fb546bb5f23ab2c58e0/src/jsii-project.ts#L400-L401):
   ```ts
   env: {
      TWINE_USERNAME: '${{ secrets.TWINE_USERNAME }}',
      TWINE_PASSWORD: '${{ secrets.TWINE_PASSWORD }}',
   },
   ```

## Proposed solution

Since `jsii-release` already supports setting [`NPM_REGISTRY`](https://github.com/aws/jsii-release#npm) and [`TWINE_REPOSITORY_URL`](https://github.com/aws/jsii-release#pypi) this probably could be managed by allowing one to set values in `.projenrc.js`:
- `options.nodejs.defaultRegistry = false` to enable `NPM_REGISTRY: '${{ secrets.NPM_REGISTRY }}',`
- `options.python.defaultRegistry = false` to enable `TWINE_REPOSITORY_URL: '${{ secrets.TWINE_REPOSITORY_URL }}',`

## Related AWS CodeArtifact issues

Then there's the issue of AWS CodeArtifact, which does not provide permanent token, but a temporary token which has a maximum duration of 12 hours (`43200` seconds). It would require having some extra step in the Github workflow which might not be that easy to support with projen 🤷‍♂️

eladb:
Sounds good! contributions are welcome

eladb:
Let's just add support for specifying the `npmRegistryUrl` and `twineRegistryUrl`. They are not secrets...

Regarding codeArtifact, given we have AWS credentials in a GitHub secret (which I am assuming are required for any solution), the workflow can do whatever it needs (like obtaining a token from aws).

JordanSinko:
@aripalo - Just an FYI... I merged in code to partially accommodate this request with #136 (NPM registry)

[...]

PR #140, opened 2020-10-02
feat(jsii): allowing twine registry to be set

JordanSinko:
Issue #, if available:
#124 

Description of changes:

This pull request will allow users to release to more than just the default twine repository such as CodeArtifact.


src/jsii-project.ts
```diff
@@ -113,6 +113,11 @@ export interface JsiiProjectOptions extends NodeProjectCommonOptions {
    * @default .compatignore
    */
   readonly compatIgnore?: string;
+
+  /**
+   * The registry url to use when releasing packages.
+   */
+  readonly twineRegistryUrl?: string;
```
eladb, line 120: Feels like this should be under `python` since it's only relevant when python is configured

eladb (changes requested):

src/jsii-project.ts
JordanSinko: Just the declaration of the type? Or also when setting said variable? Or both?

JordanSinko: I think I understand now. Are you talking about under `JsiiPythonTarget` @eladb ?

src/jsii-project.ts
```diff
@@ -130,6 +130,11 @@ export interface JsiiJavaTarget {
 export interface JsiiPythonTarget {
   readonly distName: string;
   readonly module: string;
+
+  /**
+   * The registry url to use when releasing packages.
+   */
+  readonly twineRegistryUrl?: string;
```
eladb, line 137:
```suggestion

  /**
   * The registry url to use when releasing packages.
   * @default - twine default
   */
  readonly twineRegistryUrl?: string;
```

API.md
```diff
@@ -2989,6 +2990,7 @@ Name | Type | Description
 -----|------|-------------
 **distName**🔹 | <code>string</code> | <span></span>
 **module**🔹 | <code>string</code> | <span></span>
+**twineRegistryUrl**?🔹 | <code>string</code> | The registry url to use when releasing packages.<br/>__*Optional*__
```
eladb, line 2993:
```suggestion
**twineRegistryUrl**?🔹 | <code>string</code> | The registry url to use when releasing packages.<br/>__*Default*__: twine default
```

eladb (approved):

PR #147, opened 2020-10-03
feat: add version pinning for CDK

hoegertn:
Add the option to pin the CDK version instead of using the latest.


src/awscdk-app-ts.ts
```diff
@@ -14,6 +14,13 @@ export interface AwsCdkTypeScriptAppOptions extends TypeScriptProjectOptions {
    */
   readonly cdkVersion: string;
 
+  /**
+   * Use pinned version instead of caret version for CDK.
```
eladb, line 18: Can you add a remark in the docstring that explains when someone would prefer to use pinned versions over caret versions?

eladb (changes requested):

src/awscdk-app-ts.ts
hoegertn: I added something but I am not sure if this is enough. Is there a reason you would not want to fix versions?

eladb (approved):

eladb (approved):

Issue #146, opened 2020-10-02
The Version component does not support alternate branch names for the default branch


hoegertn:
The scripts generated by the `Version` component assume the main branch to be called `master`. There should be a way to configure this, as new projects tend to use `main`.

pgollucci:
@mergifyio refresh

pgollucci:
https://github.com/projen/projen/pull/150

This can be closed right?

Chriscbr:
Yep, closing

PR #150, opened 2020-10-04
feat(version): enable master branch name to be configured

hass123uk:
Allow the user to define the name of their main/release branch.

*Issue #, if available:*
#146

*Description of changes:*
* Add new common node option that defines default release branch. 
  * Use this for the default value of the release branches.
  * Use for the value of the branch where version tags are pushed.


eladb (commented):
I think the GitHub workflows also reference 'master'. No?

hass123uk:
> I think the GitHub workflows also reference 'master'. No?

Yes, it takes in an array of release branches that default to [ 'master' ].
If we reuse this, would you, on release, push the code & tags to all branches within this array.

eladb:
> If we reuse this, would you, on release, push the code & tags to all branches within this array.

I think the other way around. I would expect the default value for `releaseBranches` to be aligned with the newly introduced `releaseBranch` so it works out of the box if I change the release branch to something that's not `master`.

hass123uk:
Perhaps the option should be called main-branch instead.

eladb (approved):

PR #155, opened 2020-10-05
feat(node): projen upgrade schedule

pahud:
This PR allows node projects to customize the projen upgrade schedule cron expressions.

Fixes eladb/projen#153


src/projen-upgrade.ts
```diff
@@ -42,7 +49,8 @@ export class ProjenUpgrade {
       const workflow = new GithubWorkflow(project, 'ProjenUpgrade');
 
       workflow.on({
-        schedule: [{ cron: '0 6 * * *' }], // 6am every day
+        schedule: options.autoUpgradeSchedule ? options.autoUpgradeSchedule.map( s => ({ cron: s }) ) :
+          [{ cron: '0 6 * * *' }], // 6am every day
```
eladb, line 53: Style:

```suggestion
        schedule: options.autoUpgradeSchedule 
          ? options.autoUpgradeSchedule.map(s => ({ cron: s }))
          : [{ cron: '0 6 * * *' }], // 6am every day
```

src/projen-upgrade.ts
```diff
@@ -20,6 +20,13 @@ export interface ProjenUpgradeOptions {
    * @default []
    */
   readonly labels?: string[];
+
+  /**
+   * Customize the projenUpgrade schedule in cron expression.
```
eladb, line 25: What's the rationale of supporting a list of CRON schedules instead of a just a single one?

eladb (changes requested):

src/projen-upgrade.ts
pahud: The [scheduled events](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events) from GH actions supports multiple schedule entries with cron expression. It allows users to define multiple cron entries like this. I believe most of the time single entry should be enough, but good to support multiple. What do you think?

```
on:
  schedule:
    # “At 06:00 on Sunday.” 
    - cron:  '0 6 * * 0'
    # “At 06:00 on day-of-month 1.”
    - cron:  '0 6 1 * *'
```

eladb: Okay cool

eladb (approved):

Issue #158, opened 2020-10-05
React project


eladb:
Reference from @harrysolovay: https://github.com/harrysolovay/react-templates-are-evil

eladb:
Resolved by #229

PR #167, opened 2020-10-07
feat(nextjs): add support for nextjs projects

Chriscbr:
*Issue #, if available:* relates to #158 

*Description of changes:* adds support for NextJS projects


Chriscbr:
This initial draft uses starter code from [next-learn-starter](https://github.com/vercel/next-learn-starter) under the MIT license, but we can cut it down to just a more minimal file if desired.

I named this `NextJSProject` as opposed to `ReactAppProject` or similar because I have this feeling that NextJS is a lot more batteries-included and opinionated than just a plain React template. I did look into creating a plain React project template, but turns out it's pretty nontrivial just because you need something that will handle all of the bundling and transpilation (i.e. babel and webpack), and those require several more config files. I imagine those are features users might eventually want options for, but this solution is at least an easier first step.

Some next steps are to add more properties / customizations (perhaps per the template referenced in #158) and also to add support for TypeScript versions of NextJS projects. Some suggested ways to go about this:

1. create a class `TypeScriptNextJSProject` that extends `TypeScriptAppProject`
2. create a class `NextJSTypeScriptProject` that extends `NextJSProject`
3. add a `typescript?` option as a parameter to `NextJSProject`

@eladb Does projen support multiple inheritance in any fashion, or is jsii a restrict us in that regard? It's a question that's been on my mind for a while, to be honest. I'm just wondering because JS projects do mix and match different libraries and frameworks etc., so I'm wondering what the best way to go about representing, for example, the 2x2 possible choices for (React, not-React) x (TypeScript, not-TypeScript) projects, as one example.

src/next.ts
```diff
@@ -0,0 +1,310 @@
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { Component } from './component';
+import { NodeProject, NodeProjectOptions } from './node-project';
+
+export interface NextJSProjectOptions extends NodeProjectOptions {
```
eladb, line 6: Rename to `NextJs` (jsii prefers no all caps abbreviations)

src/next.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * NextJS project.
+ *
+ * @pjid nextjs
+ */
+export class NextJSProject extends NodeProject {
+  /**
+   * The directory in which NextJS pages are declared.
+   */
+  public readonly pagesdir: string;
```
eladb, line 37:
```suggestion
  public readonly pagesDir: string;
```

src/next.ts
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * The directory in which app assets reside.
+   */
+  public readonly assetsdir: string;
```
eladb, line 42:
```suggestion
  public readonly assetsDir: string;
```

eladb (changes requested):
Great job looks awesome. A few minor comments

eladb:
> does projen support multiple inheritance in any fashion, or is jsii a restrict us in that regard? 

The way to implement this is through `Component` classes that can be composed together into a project.

eladb (approved):

PR #175, opened 2020-10-09
fix: paths are not compatible with windows

JordanSinko:
*Issue #, if available:*

*Description of changes:*

- Adjusted tmp paths in Json/Toml test paths to allow tests to pass with win32
- Adjusted bin path calculation to prevent backslashes in path with win32
- Adjusted types path calculation to prevent backslahes in path with win32


eladb (approved):
Thanks!

Issue #168, opened 2020-10-07
Feature Request: Initialize .projenrc.js with commented out project options


Chriscbr:
Quality of life / convenience feature.

The idea is when `.projenrc.js` gets generated, it should include a list of the available options out of the box, similar to how they are provided when you run `tsc --init` in a new repository, which produces:

```js
{
  "compilerOptions": {
    /* Visit https://aka.ms/tsconfig.json to read more about this file */

    /* Basic Options */
    // "incremental": true,                   /* Enable incremental compilation */
    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
    // "lib": [],                             /* Specify library files to be included in the compilation. */
    // "allowJs": true,                       /* Allow javascript files to be compiled. */
    ...
```

I see that jsii is being used to generate API docs right now -- maybe there's some way to hook into that or leverage it to perform this automatically?

eladb:
Yes. This idea came to mind and should be achievable.

pszabop:
Yes please.  As a new user I'm finding it difficult to figure out what I can do.  Any documentation help would be a plus.

PR #179, opened 2020-10-12
feat(awscdk-construct): documentation

mbonig:
*Issue #, if available:*

*Description of changes:*

Added first draft doc for using the AWS CDK construct library. Notes requested.


docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
+# AWS CDK - Construct library
+
+This projen type is used for building AWS CDK constructs using the JSII.
+
+Use this if you'd like to build and distribute CDK constructs for others to use.
+
+# Getting Started
```
eladb, line 7: why not use `projen new`?

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+
+Create a new instance of the project and set the appropriate fields.
+
+```typescript
+const project = new AwsCdkConstructLibrary({
+  authorAddress: "",
```
eladb, line 13: Instead of needing to maintain this, I rather we implement #168 and instruct people to use `npx projen new`

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+
+All are pretty standard setup and nothing CDK-specific at this point. The `keywords` automatically gets 'cdk' so you don't
+need to specify it. 
+
+### CDK Dependencies
+
+Next are getting CDK dependencies added:
+
+```typescript
+  entrypoint: 'lib/index.js',
```
eladb, line 90: How is `entrypoint` related to CDK dependencies? (also, I don't think it needs to be set up)

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+  cdkVersion: "1.67.0",
+  cdkDependencies: ["@aws-cdk/aws-ec2"],
+  cdkTestDependencies: ["@aws-cdk/assert"],
+```
+
+With the above example you'll get a package.json that looks like this:
```
eladb, line 96: I don't think it's very useful to show the synthesized output for everything. Just describe the behavior. In this case, I would write something like:

`cdkDependencies` will add both `dependencies` and `peerDependencies` to your `package.json` file with a **caret** semver requirement. You can set `cdkVersionPinning` to `true` to use a fixed version, but this means that any consumer of your library will have to use this exact CDK version.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+  },
+```
+
+Additionally, you can add CDK dependencies using the method:
+
+```typescript
+project.addCdkDependencies("aws-cdk/aws-sqs", "aws-cdk/aws-sns");
+```
+
+### JSII Packaging
```
eladb, line 113: This is not just packaging, it's also publishing.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+    mavenArtifactId: 'hello-jsii',
+    mavenGroupId: 'com.acme.hello'
+  },
+  python: {
+    distName: 'acme.hello-jsii',
+    module: 'acme.hello_jsii'
+  },
+```
+
+Will render out the following to the package.json:
```
eladb, line 133: Again, redundant. It also modifies the GitHub workflows.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+    },
+    "tsc": {
+      "outDir": "lib",
+      "rootDir": "src"
+    }
+  }
+```
+
+If you don't want to publish a particular package, do not include the `dotnet`, `java`, or `python` field.
+
```
eladb, line 163: In order to the github workflows to work, users need to add github secrets. This should be documented.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
+# AWS CDK - Construct library
```
eladb, line 1:
```suggestion
# AWS CDK Construct Library
```

docs/awscdk-construct.md
```diff
@@ -0,0 +1,186 @@
...
+
+This projen type is used for building AWS CDK constructs using the JSII.
```
eladb, line 3: Explain what's the value of using jsii and include some references to jsii docs.

jsii is spelled all lowercase BTW

eladb (changes requested):
Initial round of feedback

docs/awscdk-construct.md
mbonig: Yes, I'll add steps to use projen new instead of expecting this more manual approach.

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: Totally forgot mentioning this. Since most people probably didn't name their construct's file 'main.ts' I figured this would need to be overridden. I'll make that clearer.

docs/awscdk-construct.md
mbonig: Will do. I was putting the output because that usually helps me understand how properties transfer but I can remove it.

docs/awscdk-construct.md
mbonig: But only if the related workflows are setup, right?

eladb: No, workflows are automatically created as soon as you specify `python`, `java` etc. No additional setup is required (besides installing the secrets to your github repo).

mbonig:
I addressed all issues. I also added some additional coverage I didn't notice the first time around.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+Start like all projen projects:
+
+```sh
+npx projen new awscdk-construct
+```
+
+Review the resulting .projenrc.js file and make changes as needed. The following are some specific areas
+you may want to set explicitly.
+
+A `README.md` file is not created but encouraged. It is required for publishing your construct for Python (see [jsii publishing](#jsii-publishing)).
```
eladb, line 18: Should we bootstrap one?

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+
+All are pretty standard setup and nothing CDK-specific at this point. The `keywords` automatically gets 'cdk' so you don't
+need to specify it. 
+
+### CDK Dependencies
+
+Next are getting CDK dependencies added:
+
+```typescript
+  entrypoint: 'lib/our-s3-bucket.js',
```
eladb, line 42: Not sure what's up here? The practice that I'd recommend is leaving this as the default `lib/index.js` and adding this to `index.ts`:

```ts
export * from './our-s3-bucket';
```

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+  cdkVersion: "1.67.0",
+  cdkDependencies: ["@aws-cdk/aws-ec2"],
+  cdkTestDependencies: ["@aws-cdk/assert"],
+```
+
+The entrypoint can be set to point at your existing file, if not at `lib/index.js`. However, it's recommended that you
+rename your file to `index.js` and leave this un-set.
```
eladb, line 49: Remove

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+
+`cdkDependencies` will add both dependencies and peerDependencies to your package.json file with a caret semver 
+requirement (e.g. `^1.67.0`). You can set cdkVersionPinning to true to use a fixed version, but this means that any 
+consumer of your library will have to use this exact CDK version. Likewise, `cdkTestDependencies` will add dependencies 
+to the `devDependencies`. 
+
```
eladb, line 55: Explain why both deps and peer deps are required (this is actually resolved by npm@7)

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+```typescript
+project.addCdkDependencies("aws-cdk/aws-sqs", "aws-cdk/aws-sns");
+project.addCdkTestDependencies("aws-cdk/something-else");
+```
+
+The `@aws-cdk/assert` library is already added to the cdkTestDependencies for you.
+
+### jsii Publishing
+
+As this is a [jsii](https://github.com/aws/jsii) constructs, it will cross-compile to other languages.  You can set up
```
eladb, line 67:
```suggestion
As this is a [jsii](https://github.com/aws/jsii) project, it will cross-compile to other languages.  You can set up 
```

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+    distName: 'acme.hello-jsii',
+    module: 'acme.hello_jsii'
+  },
+```
+
+Each target requires a specific [Github project secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) 
+to be setup:
+
+* npm - NPM_TOKEN
+* .net - NUGET_API_KEY
```
eladb, line 90:
```suggestion
* .NET - NUGET_API_KEY
```

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+    mavenGroupId: 'com.acme.hello'
+  },
+  python: {
+    distName: 'acme.hello-jsii',
+    module: 'acme.hello_jsii'
+  },
+```
+
+Each target requires a specific [Github project secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) 
+to be setup:
```
eladb, line 87: Add a reference to [jsii-release](https://github.com/aws/jsii-release) which has detailed instructions on how to obtain these secrets.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+```sh
+npx projen new awscdk-construct
+```
+
+Review the resulting .projenrc.js file and make changes as needed. The following are some specific areas
+you may want to set explicitly.
+
+A `README.md` file is not created but encouraged. It is required for publishing your construct for Python (see [jsii publishing](#jsii-publishing)).
+
+### Standard Node Module Fields
```
eladb, line 20: Why H3 instead of H2?

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
+# AWS CDK Construct Library
+
+This projen type is used for building AWS CDK constructs using the jsii. The [jsii](https://github.com/aws/jsii) allows 
+you to write code once in Typescript and it will generate the Python, .net, and Java equivalents. Use this if you'd 
+like to build and distribute CDK constructs for others to use.
+
```
eladb, line 6: Add TOC

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+If you don't want to publish a particular package, do not include the `dotnet`, `java`, or `python` field.
+
+### Workflows
+
+Two workflows will be created as Github Actions:
+
+* The Build workflow - controlled by the `buildWorkflow` field. On a 'pull_request' or 'workflow_dispatch' the construct
+will be built and checked for anti-tamper (ensure no manual changes to generated files).
+* The Release workflow - controlled by the `releaseWorkflow` field. On a push to `master` (overridden at
+ `props.defaultReleaseBranch`) the construct is built, anti-tampered, version bumped with a commit, pushed back to git,
```
eladb, line 110: Replace the term "construct" across the doc with "construct library" or just "library"

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+
+Two workflows will be created as Github Actions:
+
+* The Build workflow - controlled by the `buildWorkflow` field. On a 'pull_request' or 'workflow_dispatch' the construct
+will be built and checked for anti-tamper (ensure no manual changes to generated files).
+* The Release workflow - controlled by the `releaseWorkflow` field. On a push to `master` (overridden at
+ `props.defaultReleaseBranch`) the construct is built, anti-tampered, version bumped with a commit, pushed back to git, 
+ and then published to the configured artifact repositories (e.g. npm, pypi)
+
+Additionally, a [dependabot](https://dependabot.com/) file will be added unless `dependabot` is set to 'false'.
```
eladb, line 113: This should be under "Dependencies", not "Workflows".

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+  name: 'cdk-watchful',
+  license: "MIT",
+  repository: "https://github.com/eladb/cdk-watchful.git",
+  keywords: ["cloudwatch", "monitoring"]
+```
+
+All are pretty standard setup and nothing CDK-specific at this point. The `keywords` automatically gets 'cdk' so you don't
+need to specify it. 
+
+### CDK Dependencies
```
eladb, line 37:
```suggestion
### Dependencies
```

Please also add a section about non-CDK dependencies. Specifically, the difference between jsii deps and non-jsii deps (which require bundling)

docs/awscdk-construct.md
```diff
@@ -0,0 +1,158 @@
...
+  catalog: {
+    announce: true,
+    twitter: '@yourhandle'
+  }
+```
+
+These values are optional but allow the construct catalog's Twitter account to mention your handle on the tweet. Setting `announce`
+to 'false' will stop all tweets about the construct. However, the construct will still be indexed.
+
+### docgen
```
eladb, line 129:
```suggestion
### API Documentation
```

eladb (changes requested):
Missing:

- Enabling API compatibility checks
- Release management
- Project structure (src, lib, test, resources,...)
- Workflow (what is the meaning of each command).

Some of those should be part of the node/typescript/jsii project docs, but since we don't have them lets try to cover all of it here and then we can move stuff around.

Would be nice to start structuring our docs according to project types and then we can just cross reference.

docs/awscdk-construct.md
mbonig: I added a Readme class.

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: I assume changes are needed if projen is to support the new npm@7 algorithm and not add the cdk deps to `dependencies`?

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: 👍

docs/awscdk-construct.md
mbonig: missed some cleanup... fixed now.

docs/awscdk-construct.md
mbonig: I think I got the right ones. Some places still seemed 'construct' was best.

docs/awscdk-construct.md
mbonig: 👍

mbonig:
> Missing:
> 
> * Enabling API compatibility checks

not sure what you're looking for here. More details about this? https://github.com/aws/jsii/tree/master/packages/jsii-diff

> * Release management

Added the Workflow section which should cover this, but let me know if there were other details you are thinking of.

> * Project structure (src, lib, test, resources,...)

Added. I didn't cover all the individual files, but added a comment about the resources directory.

> * Workflow (what is the meaning of each command).
> 

Added as `Scripts`

> Some of those should be part of the node/typescript/jsii project docs, but since we don't have them lets try to cover all of it here and then we can move stuff around.
> 
> Would be nice to start structuring our docs according to project types and then we can just cross reference.

I figured one thing that'll be useful is a breadcrumb at the top showing the hierarchy. Right now it doesn't link to any other docs but it should in the future (when more docs like this are built).

docs/awscdk-construct.md
```diff
@@ -0,0 +1,219 @@
...
+Start like all projen projects:
+
+```sh
+npx projen new awscdk-construct
+```
+
+Review the resulting .projenrc.js file and make changes as needed. The following are some specific areas
+you may want to set explicitly.
+
+## Standard Node Module Fields
```
eladb, line 37:
```suggestion
## Module Metadata
```

docs/awscdk-construct.md
```diff
@@ -0,0 +1,219 @@
...
+
+These fields are your basic Node module setup:
+
+```typescript
+  authorAddress: "benisrae@amazon.com",
```
eladb, line 42: Please normalize all quotes to single quotes (I know that `projen new` generates with double...)

docs/awscdk-construct.md
```diff
@@ -0,0 +1,219 @@
...
+
+```typescript
+  cdkVersion: "1.67.0",
+  cdkDependencies: ["@aws-cdk/aws-ec2"],
+  cdkTestDependencies: ["@aws-cdk/assert"],
+```
+
+`cdkDependencies` will add both dependencies and peerDependencies to your package.json file with a caret semver 
+requirement (e.g. `^1.67.0`). CDK dependencies must be both direct and peer dependencies, 
+see [this issue](https://github.com/aws/aws-cdk/issues/5064). You can set cdkVersionPinning to true to use a fixed
```
eladb, line 66:
```suggestion
see [this issue](https://github.com/aws/aws-cdk/issues/5064). You can set `cdkVersionPinning` to `true` to use a fixed 
```

docs/awscdk-construct.md
eladb: Yes, but it will take a while for users to migrate and this is a construct _library_, so we can't rely on people using npm@7.

Also, in CDK2.0, you'll only need a peer dependency.

docs/awscdk-construct.md
```diff
@@ -0,0 +1,219 @@
...
+
+Additional dependencies can be added through `deps`/`dependencies` for jsii-based libraries or 
+`bundledDeps`/`bundledDependencies` for libraries that are not built with the jsii. You can read more 
+[here](https://github.com/aws/jsii/blob/master/docs/configuration.md#dependency-considerations).
+
+A [dependabot](https://dependabot.com/) file will be added unless `dependabot` is set to 'false'.
+
+## jsii Publishing
+
+As this is a [jsii](https://github.com/aws/jsii) project, it will cross-compile to other languages.  You can set up
```
eladb, line 87:
```suggestion
As this is a [jsii project](./jsii.md), it will cross-compile to other languages.  You can set up 
```

docs/awscdk-construct.md
mbonig: k. I don't believe there is anything to add to this until we decide npm@7 is adopted enough.

mbonig:
Anything else I missed?

eladb (approved):

PR #181, opened 2020-10-12
fix(node): wrong npmIgnore field used

### LINKED ISSUE #139: NodeProjectCommonOptions.npmIgnore is not honored
[`NodeProjectCommonOptions.npmIgnore`](https://github.com/eladb/projen/blob/master/API.md#projen-nodeprojectoptions) is not honored and always generated.

```js
const { NodeProject } = require("projen");

const project = new NodeProject({
  name: "projen-test",
  npmIgnore: false,
});

project.synth();
```

```console
$ rm --force .npmignore

$ yarn projen
<snip>

$ cat .npmignore 
# ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
/.projenrc.js
/.versionrc.json
/.mergify.yml
```

`NodeProjectCommonOptions.npmIgnore` and `NodeProjectCommonOptions.npmignore` are somewhat tricky names...
https://github.com/eladb/projen/blob/2288f2bff06e601243d4e7ffdae92d4e286123f5/src/node-project.ts#L522-L524

It seems like this line should be `NodeProjectCommonOptions.npmIgnore`.

Chriscbr:
*Issue #, if available:* Closes #139 

*Description of changes:* Fixed npmIgnore to properly remove the .npmignore file if set to false.

It might be worth changing the name of one of these fields (which differ just by capitalization) to eliminate this kind of confusion. Perhaps change `npmignore` to `npmIgnorePatterns`? It's a little more verbose but not by much.

----


src/node-project.ts
```diff
@@ -585,7 +585,7 @@ export class NodeProject extends Project {
       readonly: false, // we want "yarn add" to work and we have anti-tamper
     });
 
-    if (options.npmignore ?? true) {
+    if (options.npmIgnore ?? true) {
```
eladb, line 588: How did it compile?

Chriscbr: There is one field named npmIgnore (capital i) from `NodeProjectCommonOptions`, and another field named npmignore (lowercase i) from `NodeProjectOptions` (which extends `NodeProjectCommonOptions`). They're different types, but the ?? operator doesn't seem to care, so both variations type check fine :-)

eladb: Let's fix this mess. Doesn't make sense to have two fields with the same name and different capitalization.

Perhaps `npmignoreEnabled` and `npmignore`?

And it feels that both should be in the common options.

eladb (approved):

Issue #182, opened 2020-10-12
Feature request: expose TypeScript project APIs for choosing eslint target directories


Chriscbr:
Right now `EslintOptions` is pretty bare bones, although it does have a `dirs` field:
https://github.com/eladb/projen/blob/e39f2b138d9591b4ece2ffd4fa6d0960a14f9709/src/eslint.ts#L6-L13

But currently the `dirs` values get hardcoded in `TypeScriptProject`, so anything other project types that extend `TypeScriptProject` must use these two directories which can't be changed. Moreover, if you create a new typescript-app project, disable jest, and remove the tests directory, eslint will complain to you because it is trying to lint the `testdir` directory, but finds nothing.

https://github.com/eladb/projen/blob/e39f2b138d9591b4ece2ffd4fa6d0960a14f9709/src/typescript.ts#L269-L274

We should add some APIs / options to the TypeScript class to allow this to be customized, and to handle the case when users do not want to have tests in their project (or if they want to just have tests intermingled in their `src` directory).

Issue #180, opened 2020-10-12
yarn build fails for default typescript-app project if eslint and jest are disabled


Chriscbr:
Steps to reproduce:
1. `npx projen new typescript-app`
2. Modify `.projenrc.js` to add `eslint: false` and `jest: false`
3. `yarn run build`

Chriscbr:
I can't seem to reproduce this issue (I should have included which build I was on, probably...), so I'm closing this.

PR #197, opened 2020-10-14
feat: typescript nextjs projects

Chriscbr:
*Issue #, if available:* N/A

*Description of changes:* Added a nextjs-ts project option which extends the typescript-app project class. Grouped some options together between the two nextjs project types, and implemented basic Next.js tsconfig project defaults so that both nextjs and nextjs-ts project types work out of the box.

----

Some notes:

I made a slight modification of TypeScriptConfig so that I can mark `tsconfig.json` as mutable, since the 'next build' command fails if it's immutable. The reason for this I believe is that the 'next' command tries to automatically change your tsconfig file if it sees that some options set incorrectly based on what Next.js's build system wants. I could send a feature request to the Next.js team to see if they can enable a CLI flag that skips this behavior, but for now this workaround makes the most sense.

There is a two-line file named `next-env.d.ts` that must be added to the root directory for TypeScript Next.js projects to successfully build. I included this in the PR, but it's worth suggesting that we should also eventually add an escape hatch so a user can omit or change the `next-env.d.ts` file if needed -- but I'm not sure what that mechanism should be, so I've left that unimplemented for now.

~Lastly, many Next.js projects don't have test directories from what I've seen, so I didn't include sample code for any. But this causes eslint to complain, and changing how the default eslint command behaves warrants a separate PR (see #182), and using `eslint: false` can lead to #180, so I ended up adding a small piece of cruft in the project constructor for now:~

```
this.addTestCommand('echo 0');
```

(I ended up overriding the default TypeScriptProject build command so this isn't an issue)

----


Chriscbr:
Not totally sure why this build is failing (works fine locally) -- is it not possible to override the docstring of class?

src/typescript.ts
```diff
@@ -503,6 +560,7 @@ export class TypescriptConfig {
   public readonly include: string[];
   public readonly exclude: string[];
   public readonly fileName: string;
+  public file: JsonFile;
```
eladb, line 563:
```suggestion
  public readonly file: JsonFile;
```

src/typescript.ts
```diff
@@ -380,6 +416,20 @@ export interface TypeScriptCompilerOptions {
    */
   readonly module?: string;
 
+  /**
+   * Determine how modules get resolved. Either "Node" for Node.js/io.js style resolution, or "Classic".
+   *
+   * @default 'node'
+   */
+  readonly moduleResolution?: string;
```
eladb, line 424: enum?

src/typescript.ts
```diff
@@ -365,6 +388,19 @@ export interface TypeScriptCompilerOptions {
...
+   * with transpileModule or @babel/plugin-transform-typescript) would be safe.
+   */
+  readonly isolatedModules?: boolean;
+
+  /**
+   * Support JSX in .tsx files: "react", "preserve", "react-native"
+   *
+   * @default undefined
+   */
+  readonly jsx?: string;
```
eladb, line 402: enum?

src/typescript.ts
```diff
@@ -365,6 +388,19 @@ export interface TypeScriptCompilerOptions {
    */
   readonly inlineSources?: boolean;
 
+  /**
+   * Perform additional checks to ensure that separate compilation (such as
+   * with transpileModule or @babel/plugin-transform-typescript) would be safe.
+   */
```
eladb, line 394: @default

src/next.ts
```diff
@@ -25,55 +21,239 @@ export interface NextJsProjectOptions extends NodeProjectOptions {
...
+        },
+        ...options.tsconfig,
+      },
+      ...options,
+
+      // never generate default TypeScript sample code, since this class provides its own
+      sampleCode: false,
+    });
+
+    this.srcDir = options.srcdir ?? 'pages';
```
eladb, line 154: Kind of don't like the inconsistency `srcDir` versus `srcdir`. Let's normalize to `srcdir`?

src/next.ts
```diff
@@ -25,55 +21,239 @@ export interface NextJsProjectOptions extends NodeProjectOptions {
...
-    this.addScripts({
-      dev: 'next dev',
-      build: 'next build',
-      server: 'next start',
+    // NextJS CLI commands, see: https://nextjs.org/docs/api-reference/cli
+    this.addScript('dev', 'next dev');
+    this.start?.addEntry('dev', {
+      desc: 'Starts the Next.js application in development mode',
+      category: StartEntryCategory.BUILD,
     });
```
eladb, line 169: You could reuse this with the `NextProject` using `Component`s. You can create a `NextComponent` and then compose it here and there.

eladb (changes requested):
minor-

eladb (approved):
🤗🤗🤗

PR #199, opened 2020-10-16
feat(makefile): add support for makefiles

eganjs:
*Issue #, if available:*
n/a

*Description of changes:*
Add class to generate simple Makefiles


src/makefile.ts
```diff
@@ -0,0 +1,55 @@
...
+  isPhony?: boolean;
+}
+
+export interface MakefileOptions extends FileBaseOptions {
+  all?: string[];
+  rules: Rule[];
+}
+
+export class Makefile extends FileBase {
+  readonly all?: string[];
```
eladb, line 17: Can we just model this as a built-in rule (defined in the ctor).

eladb (commented):
This is asking for some high level APIs, but those can be added later. For example, it would be nice if the prerequisites are not just strings but rather references to either existing files or artifacts of a rule. Things like that.

src/makefile.ts
```diff
@@ -0,0 +1,55 @@
+import { FileBase, FileBaseOptions, IResolver } from './file';
+import { Project } from './project';
+
+export interface Rule {
+  targets: string[];
+  prerequisites?: string[];
```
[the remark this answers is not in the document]
eganjs: I've based my terminology on this page from the Make manual https://www.gnu.org/software/make/manual/html_node/Rule-Introduction.html

src/makefile.ts
eganjs: I'm not sure I understand, what does "ctor" refer to?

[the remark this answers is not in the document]
eganjs: done

src/makefile.ts
```diff
@@ -0,0 +1,55 @@
...
+  recipe?: string[];
+  isPhony?: boolean;
+}
+
+export interface MakefileOptions extends FileBaseOptions {
+  all?: string[];
+  rules: Rule[];
+}
+
+export class Makefile extends FileBase {
```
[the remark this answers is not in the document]
eganjs: done

src/makefile.ts
```diff
@@ -0,0 +1,55 @@
...
+  readonly all?: string[];
+  readonly rules: Rule[];
```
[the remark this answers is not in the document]
eganjs: done

src/makefile.ts
```diff
@@ -0,0 +1,55 @@
+import { FileBase, FileBaseOptions, IResolver } from './file';
```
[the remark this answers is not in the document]
eganjs: done

src/makefile.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Project } from './project';
+
+export interface Rule {
+  targets: string[];
+  prerequisites?: string[];
+  recipe?: string[];
+  isPhony?: boolean;
```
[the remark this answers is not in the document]
eganjs: done

src/makefile.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Commands that are run (using prerequisites as inputs) to create a target.
+   */
+  readonly recipe?: string[];
+
+  /**
+   * Marks whether the target is phony.
+   *
+   * False by default.
```
eladb, line 28:
```suggestion
   * @default false
```

src/makefile.ts
```diff
@@ -0,0 +1,97 @@
...
+  readonly targets: string[];
+
+  /**
+   * Files that are used as inputs to create a target.
+   */
+  readonly prerequisites?: string[];
+
+  /**
+   * Commands that are run (using prerequisites as inputs) to create a target.
+   */
```
eladb, line 22: "@default" is required for all optional fields. In this case, you can do something like this:

```
@default []
```

src/makefile.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  constructor(project: Project, filePath: string, options: MakefileOptions) {
+    super(project, filePath, options);
+
+    if (!options.rules) {
+      throw new Error('"rules" cannot be undefined');
+    }
+
+    this.all = options.all;
+    this.rules = options.rules;
```
eladb, line 71: You can just implement `all` through `rules` I believe:

```suggestion
    const all = { phony: true, targets: ['all'], prerequisites: options.all };
    this.rules = [
       ...options.all ? all : [],
       ...options.rules,
    ];
```

src/makefile.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
```
eladb, line 73: Would be nice to also add imperative API:

```ts
makefile.addRule(rule);
makefile.addAll(targetName);
```

src/makefile.ts
eganjs: done

src/makefile.ts
eganjs: done

src/makefile.ts
eganjs: Done. I've used a builder like pattern so the calls can be chained. e.g.
```
makefile
  .addRule(rule1)
  .addRule(rule2);
```

src/makefile.ts
eganjs: done

eganjs: Also it turns out
```
.PHONY: all
all:
```
is a valid Makefile, so I've rewritten it to always add the all rule

eladb (approved):

PR #203, opened 2020-10-18
feat: README samples

mbonig:
*Issue #, if available:*

*Description of changes:*

Publishing to pypi requires a readme, so adding a simple README file to the jsii construct.


mbonig:
Since this itself is a jsii project I'm not sure how to get the build working. I think I could force it by making sure that the README.md doesn't get replaced but I'm not sure that's the right approach. Since a README.md file is expected to be manually maintained after the creation then I assume it can't be handled quite like the other files.

Taking suggestions.

Chriscbr:
Yeah, I've seen a similar approach used in `SampleCode` components in some of the projects - I think your solution makes natural sense (since I doubt most people would want to define their README through .projenrc.js). Maybe it's worthwhile to create a general component (maybe named `UserManagedFile` or `RawFile` or similar) for representing these kinds of files across the library, and have `Readme` extend that? Curious what y'all think.

eladb:
> Curious what y'all think.

I think it's a good idea. I don't like to duplicate this logic throughout. I'd call it `SampleFile` or something like that.

mbonig:
I refactored some functionality from the SampleCode class into a SampleFile abstract class. Wasn't loving the abstract class definition, feels like maybe this isn't an abstract class but something different so I'm open to suggestions.

src/typescript.ts
```diff
@@ -526,8 +526,18 @@ export class TypescriptConfig {
...
+  protected writeOnceFileContents(dir: string, filename: string, contents: string) {
+    if (fs.pathExistsSync(dir) && fs.readdirSync(dir).filter(x => x === filename)) {
+      return;
+    }
+    fs.mkdirpSync(dir);
+    fs.writeFileSync(path.join(dir, filename), contents);
+  }
+}
+
+class SampleCode extends SampleFile {
```
Chriscbr, line 539: I'm not sure if inheritance feels right here since this class is representing multiple files of sample code

src/readme.ts
```diff
@@ -0,0 +1,30 @@
...
+import { SampleFile } from './typescript';
+
+export interface ReadmeOptions extends FileBaseOptions {
+  /**
+   * The initial content of the README file.
+   *
+   * @default "# replace this"
+   */
+  readonly text?: string;
+}
```
Chriscbr, line 12: Looks like this isn't being used

src/typescript.ts
```diff
@@ -526,8 +526,18 @@ export class TypescriptConfig {
 }
 
 
-class SampleCode extends Component {
-  private readonly nodeProject: TypeScriptProject;
+export abstract class SampleFile extends Component {
```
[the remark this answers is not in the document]
mbonig: Moved SampleFile out to file.ts.

I learn towards keeping them separate but could be persuaded either way.

src/typescript.ts
```diff
@@ -526,8 +526,18 @@ export class TypescriptConfig {
...
+  protected writeOnceFileContents(dir: string, filename: string, contents: string) {
+    if (fs.pathExistsSync(dir) && fs.readdirSync(dir).filter(x => x === filename)) {
```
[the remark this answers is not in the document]
mbonig: Done

src/typescript.ts
mbonig: Those multiple files represent one "thing", the component and it's test, so it sorta feels appropriate to have them together. It doesn't feel like you'd want one without the other.

src/readme.ts
mbonig: you're right, a holdover from a previous impl. it's been removed.

mbonig:
Had to update this code a little around SampleFile. Didn't make sense to write the test if the code wasn't written.

src/file.ts
```diff
@@ -87,3 +88,22 @@ export interface IResolver {
 }
 
 
+export abstract class SampleFile extends Component {
```
eladb, line 91: Please separate to a different source file

eladb, line 91: Add docstrings to all public elements

src/file.ts
```diff
@@ -87,3 +88,22 @@ export interface IResolver {
...
+  /**
+   * A helper function that will write the file once and return if it was written or not.
+   * @param dir - the directory for the new file
+   * @param filename - the filename for the new file
+   * @param contents - the contents of the file to write
+   * @return boolean - whether a new file was written or not.
+   * @protected
+   */
+  protected writeOnceFileContents(dir: string, filename: string, contents: string) {
```
eladb, line 100: Shouldn't that be `synthesize()`

src/file.ts
eladb, line 91: I don't think this needs to be abstract. I'd expect the usage to be:

```ts
new SampleFile(project, 'README.me', {
  contents: 'bla bla read me'
};
```

And that's all. The class will take care of the rest

src/typescript.ts
eladb: I don't think `SampleCode` is needed as a class. Just use `SampleFile` here.

eladb (changes requested):

src/file.ts
mbonig: How would you expect me to handle the situation in the Typescript project where a sample hello.ts and hello.test.ts files are created. I assume each would become 

```
new SampleFile(project, 'src/hello.ts', ...);
new SampleFile(project, 'src/hello.test.ts', ...);
```

However, the test file shouldn't be touched if the hello.ts file wasn't modified (this was a scenario I discovered when this build was failing because the new hello.test.ts tests fail).

In that case, would this be good?

```
const sourceFile = new SampleFile(project, 'src/hello.ts', ...);
if (sourceFile.wasCreated){
  new SampleFile(project, 'src/hello.test.ts', ...);
}
```

src/readme.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Represents a README.md file. Unlike most other files this will only be created if the file doesn't already exist.
+ * You are expected to manage this file after creation.
+ */
+export class Readme extends Component {
+  private readonly text: string;
+
+  constructor(project: Project, text?: string) {
+    super(project);
+    this.text = text ?? '# jsii construct';
```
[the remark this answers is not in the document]
mbonig: how would I get the project's name? `Project` doesn't seem to have the field.

mbonig:
I reworked the code to remove the abstract class and just make it more direct.

However, there is an issue that needs to be worked out. I tried to simplify the code per your suggestions @eladb but I have a problem with the samplecode in the Typescript project. As the index.ts and hello.test.ts file are now separate components I don't know how to ensure that the hello.test.ts file doesn't get written if the index.ts file doesn't get written. This was one of the reasons why they were in the same file before. I like the separation now but I can't stop the hello.test.ts file from being generated (which is causing this build to break, since the test is added but fails)

Chriscbr:
I think the solution you proposed earlier sounded good!
```ts
if (sourceFile.created) {
  new SampleFile(project, 'src/hello.test.ts', ...);
}
```

src/file.ts
```diff
@@ -72,6 +64,14 @@ export abstract class FileBase extends Component {
       readonly: this.readonly,
     });
   }
+
```
eladb, line 67: Any reason this moved? Please avoid combining aesthetic changes and functional changes. Let's revert

src/file.ts
```diff
@@ -85,5 +85,3 @@ export interface IResolver {
    */
   resolve(value: any): any;
 }
-
```
eladb, line 88: Revert

src/jsii-project.ts
```diff
@@ -266,6 +267,8 @@ export class JsiiProject extends TypeScriptProject {
       new JsiiDocgen(this);
     }
 
+    new Readme(this, '# jsii construct');
```
eladb, line 270: Let's put it in `Project` instead of `JsiiProject`. I think all projects should have a README.

src/readme.ts
```diff
@@ -0,0 +1,14 @@
+import { Project } from './project';
+import { SampleFile } from './sample-file';
+
+/**
+ * Represents a README.md file. Unlike most other files this will only be created if the file doesn't already exist.
+ * You are expected to manage this file after creation.
+ *
+ * @param text - The initial contents of the README.md file. Defaults to '# replace this'
+ */
+export class Readme extends SampleFile {
```
eladb, line 10: Use the "Sample" prefix to signal this is a sample file.

```suggestion
export class SampleReadme extends SampleFile {
```

src/sample-file.ts
```diff
@@ -0,0 +1,63 @@
...
+  readonly contents: string;
+}
+
+/**
+ * Produces a file with the given contents but only once, if the file doesn't already exist.
+ * Use this for creating example code files or other resources.
+ */
+export class SampleFile extends Component {
+  private readonly filePath: string;
+  private options: SampleFileOptions;
```
eladb, line 25: Always prefer `readonly` if possible.

```suggestion
  private readonly options: SampleFileOptions;
```

src/sample-file.ts
```diff
@@ -0,0 +1,63 @@
...
+  public created: boolean = false;
```
eladb, line 26: Don't allow accessing this directly. If there is a use case of a getter, then add a getter.

```suggestion
  private created: boolean = false;
```

src/sample-file.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @param options - the options for the file.
+   */
+  constructor(project: Project, filePath: string, options: SampleFileOptions) {
+    super(project);
+    this.filePath = filePath;
+    this.options = options;
+  }
+
+
+  synthesize(outdir: string) {
```
eladb, line 41: Always include access modifiers:

```suggestion
  public synthesize(outdir: string) {
```

src/sample-file.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * A helper function that will write the file once and return if it was written or not.
+   * @param dir - the directory for the new file
+   * @param filename - the filename for the new file
+   * @param contents - the contents of the file to write
+   * @return boolean - whether a new file was written or not.
+   * @protected
+   */
+  protected writeOnceFileContents(dir: string, filename: string, contents: string) {
```
eladb, line 54: Why protected and not private? If there is no need to expose publicly (and protected is a type of public), keep private. We can always change later.

src/typescript.ts
```diff
@@ -590,9 +570,11 @@ export class TypeScriptAppProject extends TypeScriptProject {
 /**
  * @deprecated use `TypeScriptProject`
  */
-export class TypeScriptLibraryProject extends TypeScriptProject { };
+export class TypeScriptLibraryProject extends TypeScriptProject {
```
eladb, line 573: revert

src/typescript.ts
```diff
@@ -590,9 +570,11 @@ export class TypeScriptAppProject extends TypeScriptProject {
...
+};
 
 /**
  * @deprecated use TypeScriptProjectOptions
  */
-export interface TypeScriptLibraryProjectOptions extends TypeScriptProjectOptions { }
+export interface TypeScriptLibraryProjectOptions extends TypeScriptProjectOptions {
```
eladb, line 579: revert

src/util.ts
```diff
@@ -20,4 +20,8 @@ export function writeFile(filePath: string, data: any, options: { readonly?: boo
   if (options.readonly) {
     fs.chmodSync(filePath, '400');
   }
+}
+
+export function fileExists(filePath: string) {
```
eladb, line 25: Feels unnecessary.

src/typescript.ts
```diff
@@ -526,47 +525,28 @@ export class TypescriptConfig {
...
       '});',
-    ];
+    ].join('\n');
+    new SampleFile(project, path.join(project.srcdir, 'index.ts'), { contents: srcCode });
 
-    fs.mkdirpSync(testdir);
-    fs.writeFileSync(path.join(testdir, 'hello.test.ts'), testCode.join('\n'));
+    // so here's a problem. I don't want this file to be written if the above file was written.
+    // and I don't know how to add that dependency here...
+    new SampleFile(project, path.join(project.testdir, 'hello.test.ts'), { contents: testCode });
```
eladb, line 549: I think the correct way to model these two files (`src/index.ts` and `test/hello.test.ts`) is actually as `SampleDir`. If the directory exists, I wouldn't create anything:

```ts
new SampleDir(project, project.srcdir, {
  files: { 'index.ts': srcCode }
});

new SampleDir(project, project.testdir, {
  files: { 'hello.test.ts': testCode }
});
```

eladb (changes requested):

src/util.ts
pgollucci: I could see other things needing this, but if they do I can see them being in `util.ts`

src/typescript.ts
mbonig: If the directory existence is the only thing that's checked, multiple sample files couldn't be put in the same directory.

src/jsii-project.ts
mbonig: moved

src/sample-file.ts
mbonig: made private

src/typescript.ts
eladb: Why not?

mbonig: You said that if the directory exists, it wouldn't create anything. So, if I did:
```
new SampleDir(project, project.srcdir, {
  files: { 'index.ts': srcCode }
});
new SampleDir(project, project.srcdir, {
  files: { 'main.ts': srcCode }
});
```

index.ts would be created (assuming the directory doesn't exist) but then main.ts would never be created.

src/util.ts
mbonig: Added it here because it felt more utilitarian and universal than just the one class that was using it.

src/typescript.ts
eladb: But what's the motivation for not writing it like:

```ts
new SampleDir(project, project.srcdir, {
  files: {
    'index.ts': srcCode,
    'main.ts': srcCode,
  }
});
```

mbonig: Simply not looking at the interface close enough. That makes sense. Will make the change...

mbonig: k, I have reworked the code related to the typescript project's sample files into a new SampleDir class.

src/util.ts
mbonig: @eladb are you asking I refactor this into the one class that's calling it?

src/readme.ts
```diff
@@ -0,0 +1,14 @@
+import { Project } from './project';
+import { SampleFile } from './sample-file';
+
+/**
+ * Represents a README.md file. Unlike most other files this will only be created if the file doesn't already exist.
```
eladb, line 5:
```suggestion
 * Represents a README.md sample file.
```

src/sample-file.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * SampleDir options
+ */
+export interface SampleDirOptions {
+  /**
+   * The files to render into the directory
+   */
+  readonly files: SampleFiles;
```
eladb, line 81:
```suggestion
  readonly files: { [fileName: string]: string };
```

src/sample-file.ts
```diff
@@ -0,0 +1,107 @@
...
+    fs.mkdirpSync(dir);
+    fs.writeFileSync(fullFilename, contents);
+    return true;
+  }
+}
+
+/**
+ * Represents files and their string contents
+ */
+export interface SampleFiles {
```
eladb, line 67: This interface is not needed

src/sample-file.ts
```diff
@@ -0,0 +1,107 @@
...
+   */
+  readonly contents: string;
+}
+
+/**
+ * Produces a file with the given contents but only once, if the file doesn't already exist.
+ * Use this for creating example code files or other resources.
+ */
+export class SampleFile extends Component {
+  public created: boolean = false;
```
eladb, line 24: Why is this API needed again?

It also should be `readonly`

src/typescript.ts
```diff
@@ -634,47 +633,35 @@ export class TypescriptConfig {
 }
 
 
-class SampleCode extends Component {
-  private readonly nodeProject: TypeScriptProject;
-
+class SampleCode {
```
eladb, line 636: Should still inherit from `Component`:

```suggestion
class SampleCode extends Component {
```

src/util.ts
eladb: yes please

eladb (changes requested):

src/sample-file.ts
mbonig: removing it.

mbonig:
There are still an unresolved questions:

How can I access the project name to make that the default content of the readme file?

mbonig:
Also, I though inline interfaces, e.g.: `{ [fileName: string]: string };`, were undoable with jsii?

eladb:
> Also, I though inline interfaces, e.g.: { [fileName: string]: string };, were undoable with jsii?

jsii treats this syntax as a special case for a map of strings.

eladb:
> How can I access the project name to make that the default content of the readme file?

I think we should expose `project.name`

mbonig:
> > How can I access the project name to make that the default content of the readme file?
> 
> I think we should expose `project.name`

New issue?

src/sample-file.ts
```diff
@@ -0,0 +1,94 @@
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { Component } from './component';
+import { Project } from './project';
+
+/**
+ * Options for the SampleFile object.
+ *
+ * @field contents - The contents of the file.
```
eladb, line 9:
```suggestion
```

eladb (approved):

eladb:
Sadly I had to revert this because it broke `projen new` see https://github.com/projen/projen/commit/68b1e7faa2444e11a1e7bdf59cc1a8d6ccea293c

mbonig:
Submitted new PR.

PR #212, opened 2020-10-20
feat: projects from external modules (--from)

kcwinner:
*Issue #, if available:* [193](https://github.com/eladb/projen/issues/193)

*Description of changes:*
* Alters the command syntax slightly to make `PROJECT-TYPE` an optional positional parameter. Added examples
* Adds a new option for creating a new project from a remote module: `--from`
* Changes the `inventory.discover` method to accommodate external jsii type lookups
* Adds readme info on adding projects from external modules
* Fixes a few potential null/undefined errors in node-project

*Testing:*
Tested with [cdk-appsync-project](https://github.com/kcwinner/cdk-appsync-project). Works using local installs and via npm packages.


README.md
```diff
@@ -163,6 +163,37 @@ Adding new project types is as simple as submitting a pull request to this repo
 extends `projen.Project` (or one of it's derivatives). Projen automatically discovers project types so your
 type will immediately be available in `projen new`.
 
+### Projects in external modules
+projen is bundled with many project types out of the box, but it can also work with project types and components defined in external jsii modules (the reason we need jsii is because projen uses the jsii metadata to discover project types & options in projen new).
```
eladb, line 167: Wrap to 100 chars

README.md
```diff
@@ -163,6 +163,37 @@ Adding new project types is as simple as submitting a pull request to this repo
...
+
+Say we have a module in npm called projen-vue which includes a single project type for vue.js:
```
eladb, line 169:
```suggestion
Say we have a module in npm called `projen-vue` which includes a single project type for vue.js:
```

README.md
```diff
@@ -163,6 +163,37 @@ Adding new project types is as simple as submitting a pull request to this repo
...
+
+```bash
+$ npx projen new --from projen-vue
+```
+
+If the referenced module includes multiple project types, the type is required. Switches can also be used to specify initial values based on the project type APIs. You can also use any package syntax supported by [yarn add](https://classic.yarnpkg.com/en/docs/cli/add#toc-adding-dependencies) like `projen-vuejs@1.2.3`, `file:/path/to/local/folder`, `git@github.com/awesome/projen-vuejs#1.2.3`, etc.
+
+```bash
+$ npx projen new --from projen-vue vuejs-ts@^2 --name my-vuejs-sample
```
eladb, line 178: I think there's a mix up here:

`from` is supposed to be a package spec passed to `npm i` (or `yarn add`) and therefore can optionally include a version spec. The first positional is the project type (e.g `vuejs-ts`) and `--name` is just an optional property of the project type that can be passed to assign a value to this property (like all other props).

```suggestion
$ npx projen new --from projen-vue@^2 vuejs-ts --name my-vuejs-sample
```

src/cli/cmds/new.ts
```diff
@@ -7,12 +7,17 @@ import * as inventory from '../../inventory';
...
 
+const localNodeModules = './node_modules.local';
+
 class Command implements yargs.CommandModule {
-  public readonly command = 'new PROJECT-TYPE [OPTIONS]';
+  public readonly command = 'new [PROJECT-TYPE] [OPTIONS]';
   public readonly describe = 'Creates a new projen project';
 
   public builder(args: yargs.Argv) {
+    args.positional('PROJECT-TYPE', { describe: 'Required if --from is not specified', type: 'string' });
```
eladb, line 17:
```suggestion
    args.positional('PROJECT-TYPE', { describe: 'optional only when --from is used and there is a single project type in the external module', type: 'string' });
```

src/cli/cmds/new.ts
```diff
@@ -7,12 +7,17 @@ import * as inventory from '../../inventory';
...
     args.option('synth', { type: 'boolean', default: true, desc: 'Synthesize after creating .projenrc.js' });
+    args.option('from', { type: 'string', alias: 'f', desc: 'External jsii npm module to create project from' });
+    args.option('name', { type: 'string', alias: 'n', desc: 'Name of project type if module contains more than one project type' });
```
eladb, line 20: Remove this. See my explanation above 

```suggestion
```

src/cli/cmds/new.ts
```diff
@@ -7,12 +7,17 @@ import * as inventory from '../../inventory';
...
+
 class Command implements yargs.CommandModule {
-  public readonly command = 'new PROJECT-TYPE [OPTIONS]';
+  public readonly command = 'new [PROJECT-TYPE] [OPTIONS]';
   public readonly describe = 'Creates a new projen project';
 
   public builder(args: yargs.Argv) {
+    args.positional('PROJECT-TYPE', { describe: 'Required if --from is not specified', type: 'string' });
     args.option('synth', { type: 'boolean', default: true, desc: 'Synthesize after creating .projenrc.js' });
+    args.option('from', { type: 'string', alias: 'f', desc: 'External jsii npm module to create project from' });
```
eladb, line 19:
```suggestion
    args.option('from', { type: 'string', alias: 'f', desc: 'External jsii npm module to create project from. Supports any package spec supported by yarn (such as "my-pack@^2.0")' });
```

src/cli/cmds/new.ts
eladb, line 17: add an `.example()` for using `--from`

src/cli/cmds/new.ts
```diff
@@ -95,16 +96,17 @@ class Command implements yargs.CommandModule {
   }
 
   public async handler(args: any) {
-    console.log(`Invalid project type ${args.projectType}. Supported types:`);
-    for (const pjid of inventory.discover().map(x => x.pjid)) {
```
eladb, line 99: We've lost this friendly message. Right?

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
   }
 }
 
+function handleFromNPM(args: any) {
+  // fail if .projenrc.js already exists
```
eladb, line 150: Why again?

src/cli/cmds/new.ts
```diff
@@ -95,16 +96,17 @@ class Command implements yargs.CommandModule {
   }
 
   public async handler(args: any) {
-    console.log(`Invalid project type ${args.projectType}. Supported types:`);
-    for (const pjid of inventory.discover().map(x => x.pjid)) {
-      console.log(`  ${pjid}`);
+    if (args.from && args.from !== '') {
```
eladb, line 99: In javascript, an empty string is falsy:

```suggestion
    if (args.from) {
```

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+
+  const modulePath = args.from;
+  const moduleName = modulePath.split('/').slice(-1)[0].trim().split('@')[0].trim(); // Example: ./cdk-project/dist/js/cdk-project@1.0.0.jsii.tgz
+  const packageDir = addRemoteNpmModule(modulePath);
+
+  const externalJsii: { [name: string]: inventory.JsiiType } = fs.readJsonSync(path.join(packageDir, '.jsii')).types;
+
+  const projects = inventory.discoverRemote(externalJsii);
+  if (projects.length < 1) {
+    logging.error('No projects found in remote module');
```
eladb, line 161:
```suggestion
    logging.error(`No projects found in remote module ${moduleName}`);
```

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+    process.exit(1);
```
eladb, line 162: Throw an error instead of exiting here

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+  }
+
+  let type: inventory.ProjectType | undefined = projects[0];
+
+  if (projects.length > 1) {
+    const projectName = args.name;
+    if (!projectName) {
+      logging.error('Multiple projects found in package. Please specify a project name with --name option');
```
eladb, line 170: 1. Throw instead of .error()
2. Include the list of project types in the error.

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+  const projects = inventory.discoverRemote(externalJsii);
+  if (projects.length < 1) {
+    logging.error('No projects found in remote module');
+    process.exit(1);
+  }
+
+  let type: inventory.ProjectType | undefined = projects[0];
+
+  if (projects.length > 1) {
+    const projectName = args.name;
```
eladb, line 168: This should be the first positional argument which is project _type_ (such as `typesscript-app`), not the project name such as `my-awesome-app` (name is a just a prop of the project type).

See "mixup" above.

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+    if (!projectName) {
+      logging.error('Multiple projects found in package. Please specify a project name with --name option');
+      process.exit(1);
+    }
+
+    type = projects.find(project => project.typename === '');
```
eladb, line 174:
```suggestion
    type = projects.find(project => project.typename === projectType);
```

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+    if (!type) {
+      logging.error(`Project with name ${projectName} not found.`);
+      process.exit(1);
+    }
+  }
+
+  const params: any = {};
+  for (const [key, value] of Object.entries(args)) {
```
eladb, line 182: Please  extract this code from the normal code path and reuse here. Doesn't make sense to duplicate this mess :-)

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+  logging.info(`Created ${PROJEN_RC} for ${type.typename}`);
+
+  if (args.synth) {
+    synth();
+  }
+}
+
+function addRemoteNpmModule(module: string): string {
+  let modulePath = module;
+  const moduleName = module.split('/').slice(-1)[0].trim().split('@')[0].trim(); // Example: ./cdk-project/dist/js/cdk-project@1.0.0.jsii.tgz
```
eladb, line 222: DRY (don't repeat yourself)

src/cli/cmds/new.ts
```diff
@@ -144,4 +146,100 @@ function execOrUndefined(command: string): string | undefined {
...
+  const baseDir = process.cwd();
+
+  // Yarn fails to extract tgz if it contains @ https://github.com/yarnpkg/yarn/issues/6339
+  if (module.indexOf('.tgz') > -1) {
+    const npmPackOutput = execSync(`npm pack ${module}`, { stdio: ['inherit', 'pipe', 'ignore'] });
+
+    fs.mkdirpSync(path.join(baseDir, localNodeModules));
+    fs.moveSync(path.join(baseDir, npmPackOutput.toString('utf-8').trim()), path.join(baseDir, localNodeModules, npmPackOutput.toString('utf-8').trim()));
+    modulePath = path.join(baseDir, localNodeModules, npmPackOutput.toString('utf-8').trim());
+  }
```
eladb, line 232: Extract to a function and explain what's done here

src/inventory.ts
```diff
@@ -24,7 +24,7 @@ export interface ProjectType {
   docsurl: string;
 }
 
-interface JsiiType {
+export interface JsiiType {
```
eladb, line 27: Doesn't make sense to export this

src/node-project.ts
```diff
@@ -1109,7 +1109,7 @@ export class NodeProject extends Project {
 
     const readDeps = (user: Record<string, string>, current: Record<string, string>) => {
```
eladb, line 1110: What's the use case of `current` being undefined?

At any rate, fix the type info:

```suggestion
    const readDeps = (user: Record<string, string>, current?: Record<string, string>) => {
```

yarn.lock
```diff
@@ -744,14 +744,6 @@
     "@typescript-eslint/typescript-estree" "4.5.0"
     debug "^4.1.1"
 
-"@typescript-eslint/scope-manager@4.4.1":
```
eladb, line 747: Revert yarn lock change

src/inventory.ts
```diff
@@ -80,9 +80,32 @@ export function discover() {
 
   return result.sort((r1, r2) => r1.pjid.localeCompare(r2.pjid));
 }
-function discoverOptions(fqn: string): ProjectOption[] {
+
+export function discoverRemote(externalJsii: { [name: string]: JsiiType }) {
```
eladb, line 84: Not sure I understand why this is needed? Can't we just reuse the same code that we have for the built-in discovery?

eladb (changes requested):
This is an awesome contribution! Thanks so much for picking this up!

See my comment about the mixup with `--name`. I feel it's my fault for using that as an example in the RFC...

eladb:
okay, fixed the [rfc](https://github.com/eladb/projen/issues/193#issuecomment-712600369) to use a different prop other than `--name` and move the version spec from the project type to the module name. Sorry about that.

src/inventory.ts
kcwinner: This is used in `new.ts` under the handleFromNPM
`const externalJsiiTypes: { [name: string]: inventory.JsiiType } = fs.readJsonSync(path.join(packageDir, '.jsii')).types;`

src/node-project.ts
kcwinner: Made the change. Also, was never able to track down what is causing the issue but it's only on external projects as far as I can tell. I first noticed it before working on this and just trying to get projen to work with external packages in a different manner.

Screen Shot 2020-10-21 at 9 31 54 AM

src/cli/cmds/new.ts
kcwinner: Added this back in with a fallback to the `yargs.showHelp();` since we can get there by specifying nothing but there is no check on the (now) optional positional parameter of PROJECT-TYPE

src/cli/cmds/new.ts
kcwinner: All the local projects are handled in the builder for each command. Since we can't ( grab the package at build time we need to make sure to check before running, which is why I extracted it to a method.

It may make sense to have at the top of the handler here as well, though

src/cli/cmds/new.ts
kcwinner: Extracted to `buildProjenConfigParams`

src/cli/cmds/new.ts
kcwinner: I'll do whatever makes sense. Moved to the tope of the handler for now.

src/cli/cmds/new.ts
kcwinner: These messages look kind of cluttered since the help is output on a failure as well

Screen Shot 2020-10-21 at 1 32 49 PM

eladb: We should catch these exceptions at the program level and print only the message in nice bold red font.

eladb (changes requested):
Submitted a PR against your branch with some changes (https://github.com/kcwinner/projen/pull/1)

kcwinner:
> Submitted a PR against your branch with some changes ([kcwinner#1](https://github.com/kcwinner/projen/pull/1))

kcwinner:
> > Submitted a PR against your branch with some changes ([kcwinner#1](https://github.com/kcwinner/projen/pull/1))

Merged and will fix here

src/cli/cmds/new.ts
```diff
@@ -144,4 +133,135 @@ function execOrUndefined(command: string): string | undefined {
...
+ */
+function yarnAdd(baseDir: string, spec: string) {
+  const packageJsonPath = path.join(baseDir, 'package.json');
+  const packageJsonExisted = fs.existsSync(packageJsonPath);
+
+  // workaround: yarn fails to extract tgz if it contains '@' in the name, so we
+  // create a temp copy called pkg.tgz and install from there.
+  // see: https://github.com/yarnpkg/yarn/issues/6339
+  if (spec.endsWith('.tgz') && spec.includes('@')) {
+    const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'projen-'));
```
kcwinner, line 251: Should this get cleaned up? I used `node_modules.local` since it's relative to the project. Otherwise, the project gets stuff into something like `/var/folders/8k/2_wzn5qs4qz_qcw0ls5pv_ph0000gn/T/projen-iurcCi/pkg.tgz` and just sits on the machine. 

I'm not sure how common the local file use case is. It's obviously the best way for me to do most of the testing since it can iterate without publishing to NPM

src/cli/cmds/new.ts
```diff
@@ -144,4 +133,135 @@ function execOrUndefined(command: string): string | undefined {
...
+
+  // if user did not specify a type (and we know we have only one), the select it. otherwise, search by pjid.
+  const type = !requested ? projects[0] : projects.find(p => p.pjid === requested);
+  if (!type) {
+    throw new Error(`Project type ${requested} not found. Found ${types.join(',')}`);
+  }
+
+  // include a dev dependency for the external module
+  newProject(baseDir, type, args, {
+    devDeps: [spec],
```
kcwinner, line 211: This doesn't work for local packages. Working on a fix. Can either use the `PackageInfo` interface as before or a specific check right before this

src/inventory.ts
kcwinner: This has been reverted. No longer needed with your changes to discovery

kcwinner:
Other than not loving the temp package folder, everything else works after the merge with the slight dependency info change (as noted in the conversation inline).

I have not yet tested this with remote git urls (e.g. `git@github.com/awesome/projen-vuejs#1.2.3`). My main reason is that by default `.jsii` files are ignored (although this isn't specified by jsii anywhere). What is the likelihood of this scenario and support for it? I'm assuming it will need special consideration just as local packages do.

src/cli/cmds/new.ts
eladb: I think it's fine not to clean it up from /tmp

src/cli/cmds/new.ts
```diff
@@ -144,4 +133,145 @@ function execOrUndefined(command: string): string | undefined {
...
+  const packageJsonExisted = fs.existsSync(packageJsonPath);
+  let dependencyInfo = spec;
+
+  // workaround: yarn fails to extract tgz if it contains '@' in the name, so we
+  // create a temp copy called pkg.tgz and install from there.
+  // see: https://github.com/yarnpkg/yarn/issues/6339
+  if (spec.endsWith('.tgz') && spec.includes('@')) {
+    // if user passes in a file spec then we have to specify the project name and the package location
+    // (e.g foo@/var/folders/8k/qcw0ls5pv_ph0000gn/T/projen-RYurCw/pkg.tgz)
+    const moduleName = spec.split('/').slice(-1)[0].trim().split('@')[0].trim(); // Example: ./cdk-project/dist/js/cdk-project@1.0.0.jsii.tgz
```
eladb, line 255: What happens if the name of the file doesn't match this format? Any tarball can be used from yarn's perspective. 

The only way to really know the name is to unpack the file and read it's package json.

I guess we can unpack it to /tmp and then point yarn to this directory.

Wdty?

eladb (approved):

PR #228, opened 2020-10-25
fix: start menu does not respect package manager

### LINKED ISSUE #217: Should setting npm as the package manager still display Commands ("yarn run COMMAND"):
Currently the terminal after running projen with `packageManager: NodePackageManager.NPM` is still showing:

`Commands ("yarn run COMMAND"):`

instead of

`Commands ("npm run COMMAND"):`

eladb:
Add a `command` to each start menu entry in `package.json` and use it instead of assuming this is a yarn script. This opens the door for decoupling start menu from yarn/npm and allowing it to be used in other project types as an entry point.

This change also adds `ScriptOptions` to `node.addScript()` such that it is possible
to register a start menu entry together with adding scripts to the project.

Fixes #217

BREAKING CHANGE: `node.addScript(script, command)` now takes a single command instead of variable arguments. Use `node.addScript(script, [s1,s2].join(' && '))` to build a command from multiple statements.

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -36,7 +36,7 @@ const project = new JsiiProject({
 });
 
 // since this is projen, we need to compile before running projen, dah!
-project.addScript('projen', 'yarn compile', 'node .projenrc.js');
+project.addScript('projen', 'yarn compile && node .projenrc.js');
 
 project.gitignore.include('templates/**');
 
--- a/API.md
+++ b/API.md
@@ -73,6 +73,7 @@ Name|Description
 [NodeProjectOptions](#projen-nodeprojectoptions)|*No description*
 [PeerDependencyOptions](#projen-peerdependencyoptions)|*No description*
 [Rule](#projen-rule)|A Make rule.
+[ScriptOptions](#projen-scriptoptions)|Options for adding scripts.
 [StartEntryOptions](#projen-startentryoptions)|*No description*
 [StartOptions](#projen-startoptions)|*No description*
 [TomlFileOptions](#projen-tomlfileoptions)|*No description*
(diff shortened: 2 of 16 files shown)


eladb:
@Chriscbr @abelmokadem what do you think?

Chriscbr:
Registering start menu entries with addScript() LGTM!

> This opens the door for decoupling start menu from yarn/npm and allowing it to be used in other project types as an entry point.

 How do you mean - like project types in other languages, or ...? It seems like the information that is getting added in package.json is a bit redundant, at least in the current use case ("build" command refers to "yarn run build" which refers to "tsc && ...") My concern would be if this added layer of indirection later on leads to build logic seeping elsewhere, like...

```
"scripts": {
  "test": "yarn run lint && ...",
  "compile": "tsc && ..."
},
...
"start": {
  "build": {
    "desc": "Full release build (test+compile)",
    "command": "yarn run test && yarn run compile",
    "category": 0
  }
}
```

In that sense, I don't think we would want projen to accidentally develop its own build system logic. That said I probably have less experience using different build systems, so maybe what I'm describing isn't a big concern.

eladb:
I am actually thinking there is some value in making `projen` a programming language agnostic entry point for interacting with the project, similar to `npm run` or `yarn run`. Naturally not a build system but the front end for interacting with your project. This doesn't mean that node project types won't have all those commands available through npm scripts but I think we can offer a very attractive experience that spans beyond a specific language if projen takes that role.

To that end, decoupling the start menu from npm scripts (and moving it outside of package.json) can go a long way. 

The beauty is that this is all governed by our code so it won't incur additional setup overhead.

This is just the first step.

Chriscbr:
Ah gotcha, yeah this definitely makes sense as an intermediary step then. 👍

PR #230, opened 2020-10-26
feat: add support for docker-compose.yml

misterjoshua:
_Issue #, if applicable:_ None

_Description of changes:_
This adds a `DockerCompose` component that can synthesize docker-compose files.

I like to use docker-compose to set up local dev environments. I think it'd be pretty neat to put this stuff under projen's management.

The interface looks like this:

```ts
new DockerCompose(project, {
  services: {
    www: {
      image: 'nginx',
      volumes: [
        DockerCompose.bindVolume('./html', '/var/www/html'),
        DockerCompose.namedVolume('files', '/var/files'),
      ],
    },
  },
});
```

...and the component is smart about reusing named volumes across services. I have a Wordpress example that does this in one of the unit tests.


src/docker-compose.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    dc.addService('setup', {
+      image: 'alpine',
+      command: ['sh', '-c', 'chmod a+w -R /uploads'],
+      volumes: [
+        DockerCompose.namedVolume('uploads', '/uploads'),
+      ],
+    });
+
+    dc.addService('wordpress', {
+      dependsOn: ['db'],
```
eladb, line 311: Would have been cool if this was strongly coupled through a reference:

```ts
const db = dc.addService(...);
const wp = dc.addService('wordpress', {
  dependsOn: [db]
  ...
});

// or through an imperative api
wp.addDependsOn(db);
```

src/docker-compose.ts
```diff
@@ -0,0 +1,427 @@
...
+import { YamlFile } from './yaml';
+
+/**
+ * Props for DockerCompose
+ */
+export interface DockerComposeProps {
+  /**
+   * A name to add to the docker-compose.yml filename.
+   * @example 'myname' yields 'docker-compose.myname.yml'
+   * @default no name is added
```
eladb, line 13: Conventionally, if the default is not a specific value, we use "-" for the value:

```ts
@default - no name is added
```

src/docker-compose.ts
```diff
@@ -0,0 +1,427 @@
...
+  /**
+   * Create a port mapping
+   * @param publishedPort Published port number
+   * @param targetPort Container's port number
+   * @param protocol Port's protocol (Default: TCP)
+   */
+  static portMapping(
+    publishedPort: number,
+    targetPort: number,
+    protocol: DockerComposeProtocol = DockerComposeProtocol.TCP,
```
eladb, line 36: I'd move `protocol` into an options struct so it's extensible in the future. 

As a rule of thumb, no more than 2 positional arguments.

src/docker-compose.ts
```diff
@@ -0,0 +1,427 @@
...
+      this.addService(name, serviceDescription);
+    }
+  }
+
+  /**
+   * Add a service to the docker-compose file.
+   * @param name name of the service
+   * @param description a service description
+   */
+  addService(name: string, description: DockerComposeServiceDescription): void {
```
eladb, line 112: Specify access modifier:

```suggestion
  public addService(name: string, description: DockerComposeServiceDescription): void {
```

eladb (changes requested):
Great job! Looks awesome. A few minor comments.

misterjoshua:
Good ideas! I'll get to these changes tonight or tomorrow night. :)

misterjoshua:
@eladb I think I've got your review items handled. In addition to your comments, I fleshed out the imperative interfaces in general, fixed a few bugs, and added more tests.

eladb (approved):

PR #253, opened 2020-10-31
Test for AwsCDKTypeScriptApp

megaproaktiv:
Improves test coverage

### DIFF
--- a/test/awscdk-app-ts.test.ts
+++ b/test/awscdk-app-ts.test.ts
@@ -1,0 +1,85 @@
+import { AwsCdkTypeScriptApp, CdkApprovalLevel } from '../src/';
+import { synthSnapshot } from './util';
+
+test('sample src code synthesises correctly', () => {
+
+  const prj = new AwsCdkTypeScriptApp(
+    {
+      name: 'newcdkappproject',
+      cdkVersion: '1.2.3',
+    },
+  );
+  expect(prj.cdkConfig.context).toBeUndefined;
+  expect(prj.cdkConfig.requireApproval).toBeUndefined;
+  // there will be console outputs
+  console.log = jest.fn();
+  console.error = jest.fn();
+  const snap = synthSnapshot(prj);
+
+  expect(snap['.github/workflows/build.yml']).toBeDefined;
+  expect(snap['.eslintrc.json']).toBeDefined;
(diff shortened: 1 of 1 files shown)


test/awscdk-app-ts.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    },
+  );
+  expect(prj.cdkConfig.context).toBeUndefined;
+  expect(prj.cdkConfig.requireApproval).toBeUndefined;
+  // there will be console outputs
+  console.log = jest.fn();
+  console.error = jest.fn();
+  const snap = synthSnapshot(prj);
+
+  expect(snap['.github/workflows/build.yml']).toBeDefined;
```
eladb, line 19: Why not test against a snapshot instead of just undefined?

```
expect(snap).toMatchSnapshot()
```

Boom! It tests everything!

megaproaktiv: Do you mean creating a snapshot with all desired files? Cloud do that.
I just thought this way I am more able to test only certain aspects. Open to both possibilities.

eladb: I think it can't hurt and will give us a good way to ensure there are no unwanted regressions.

megaproaktiv: ok, just additional or what lines would you skip for that?

eladb: I think it the snapshot should cover all of these expectations, no?

megaproaktiv: Usually, when I  start fresh with test first, I tend to define separate tests. And i see that now afterwards its easier and has more coverage.

Just pushed changes. Would go for 100% coverage after first pr passed :)

eladb (approved):

eladb:
@megaproaktiv feel free to pick this up any time. Thanks!

megaproaktiv:
> @megaproaktiv feel free to pick this up any time. Thanks!

I would if i knew how to disable the tamper check.
With my old method of testing each file individually i think it would work. But with the snapshot locally it test and compiles ok, but here the Tamper Check stops the build.'
What to do?

Chriscbr:
How are you running the tests? Can you try merging/rebasing with the latest code from `master` and running `yarn run build` again from the root of your fork to see if that fixes it?

Issue #122, opened 2020-09-27
Feature request: Support custom licenses


aripalo:
## Problem

It seems that currently Projen supports following licenses:
- `Apache-2.0`
- `MIT`
- `MPL-2.0`

… but for internally distributed (within a company) constructs / projects one should be able to set the license to something else (proprietary license).

## Proposed solution

One idea could be allowing to specify the license as string:
```js
project.license = `COWBOY LICENSE
This is a custom proprietary license. 
Only people wearing cowboy hats can use this project. 
Beware of our scary legal team.`;
```

## Proposed Workaround

For a workaround one could use in `.projenrc.js`:
```js
project.gitignore.exclude('LICENSE');
```

… but that doesn't work since Projen applies `!/LICENSE` after that:
![Screenshot 2020-09-27 at 20 35 44](https://user-images.githubusercontent.com/679146/94371696-0aa34780-0101-11eb-9465-b0b69d054553.png)

So at least having some way of "forcing" manual `project.gitignore.exclude` operations to be appended at the end would solve the issue.

Chriscbr:
Maybe there could also be an option to support no license, and have it automatically set `private: "true"` in the package.json to avoid unintentionally publishing a package. (Or this flag could be provided by a separate option)

eladb:
Sounds good to me. Contributions are welcome

pgarbe:
Instead of copying all the license, can't we just download them from https://github.com/github/choosealicense.com ?

pgollucci:
Is anything else download by `projen` yet ?  Can we ,yes?  @eladb  thoughts?

eladb:
Technically, synthesis includes `yarn/npm install`, which implies we are downloading stuff, so I am quite comfortable with this...

jmourelos:
Hi! I just decided to give it a try with a very simple solution: a flag "licensed".

* If the flag is true (default value) the behavior is the current one.
* If the flag is false:
  - No LICENSE file is generated
  - No entry regarding LICENSE is added to .gitignore
  - package.json license is set to "UNLICENSED"

Any thoughts?

Chriscbr:
Could we make the features you listed in the "false" section to automatically happen if `license: 'UNLICENSED'` is provided, to eliminate the extra flag? I feel like this would be a logical default behavior as long as it's documented in the docstring

jmourelos:
[...]

PR #264, opened 2020-11-01
feat(license): adds some common licenses

pgollucci:


pgollucci:
or - https://github.com/projen/projen/issues/122

eladb (changes requested):
I tend to agree that perhaps it makes more sense to only include the common licenses and fall back to downloading the file if needed.

eladb (approved):

pgollucci:
?

Issue #74, opened 2020-08-19
Composable Github Workflows


skorfmann:
For custom requirements around Github build workflows, it'd be good to have a bit more flexibility when it comes to composing workflows. Right now it's pretty much take what the `NodeProject` provides, or build your own from scratch.

Things I'd like to do for a `JsiiProject` release workflow:

- Add an explicit step before `yarn bump`.
- Prepend some sort guard step, or maybe even an entire job, which decides if a release is actually required or not

I'm using a `prebump` script to make it work somehow for the time being. However, I think it would be beneficial to allow more flexibility by exposing the individual jobs as reusable components.

eladb:
Yes!

kcwinner:
What about something like this to make it easier to compose your own? This might be more functional than adding individual configurable "sections"

```
import { NodeProject } from './node-project';

export interface GithubActionStep {
  name?: string;
  uses?: string;
  run?: string;
  with?: {
    [name: string]: any
  }
}

export interface GithubAction {
  readonly steps: GithubActionStep[];
}

export class WorkflowBootstrapAction implements GithubAction {
  public readonly steps: GithubActionStep[];

  constructor(project: NodeProject) {
    const nodeVersion = !project.minNodeVersion ? [] : [
      {
        uses: 'actions/setup-node@v1',
        with: { 'node-version': project.nodeVersion },
      },
    ];

    this.steps = [
      // check out sources.
      { uses: 'actions/checkout@v2' },

      // use the correct node version
      ...nodeVersion,

      // bootstrap the repo
      ...project.bootstrapSteps,

      // first anti-tamper check (right after bootstrapping)
      // this will identify any non-committed files generated by projen
      ...project.workflowAntitamperSteps,
    ];
  }
}

export class CodeCovAction implements GithubAction {
  public readonly steps: GithubActionStep[];

  constructor(project: NodeProject) {
    this.steps = [
      {
        name: 'Upload coverage to Codecov',
        uses: 'codecov/codecov-action@v1',
        with: {
          token: '${{ secrets.CODECOV_TOKEN }}',
          directory: project.jest?.config?.coverageDirectory,
        },
      },
    ];
  }
}

export const SetGitIdentityActionStep: GithubActionStep = {
  name: 'Set git identity',
  run: [
    'git config user.name "Auto-bump"',
    'git config user.email "github-actions@github.com"',
  ].join('\n'),
}

export const UploadArtifactActionStep: GithubActionStep = {
  name: 'Upload artifact',
[...]

PR #266, opened 2020-11-02
feat: codecov.io

kcwinner:
Fixes # n/a

Description
* Moves `Jest` to node-project instead of typescript projects
* Adds an option for enabling a [codecov.io](https://codecov.io/) code coverage build step.
* Adds an option for the codecov token secret name
* Uses [codecov/codecov-action](https://github.com/codecov/codecov-action)
* Adds [coverageDirectory](https://jestjs.io/docs/en/configuration#coveragedirectory-string) to jest config
* Change react and nextjs project types to not create jest configs - they get created downstream by react-scripts, etc

Result:
```yaml
      ...
      - run: yarn run build
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v1
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          directory: coverage
      ...
```


src/node-project.ts
```diff
@@ -1107,6 +1116,21 @@ export class NodeProject extends Project {
       : [];
   }
 
+  /**
+   * Returns the set of steps to run codecoverage action in a github workflow.
+   */
+  public get workflowCodecoverageSteps(): any[] {
```
eladb, line 1122: Do we really need this API? Feels unneeded.

src/node-project.ts
```diff
@@ -1107,6 +1116,21 @@ export class NodeProject extends Project {
...
+    return [
+      {
+        name: 'Upload coverage to Codecov',
```
eladb, line 1125: There's some implicit coupling between where coverage output goes and where this action picks this up from. Ideally the `Jest` component should expose the output location for code coverage and this code should read it from there?

eladb (changes requested):

src/node-project.ts
kcwinner: Probably not? I only added it for consistency:
```
public get workflowBootstrapSteps() {
```
```
public get workflowAntitamperSteps(): any[] {
```

src/node-project.ts
kcwinner: Currently the jest configuration is only setup for projects that extend typescript and coverage directory is not a parameter.

Should we:
A. Add Jest component to `node-project` instead of `typescript`
B. Move this to `typescript` workflow
C. Make the steps options in `.projectrc.js` instead - then it's user defined
  * E.g. `...this.postBuildSteps ?? []`
D. Assume defaults

Jest [coverageDirectory](https://jestjs.io/docs/en/configuration#coveragedirectory-string) option

eladb: I am comfortable moving `Jest` to `NodeProject` and exposing `coverageDirectory` from `NodeProject`

kcwinner: Sounds good! It'll take me a bit longer so hopefully can finish up in the next few days then

eladb:
No worries. Thanks!

src/next.ts
```diff
@@ -43,6 +43,7 @@ export class NextJsProject extends NodeProject {
 
   constructor(options: NextJsProjectOptions) {
     super({
+      jest: false,
```
kcwinner, line 46: Needed because now `node-project` creates the jest config instead of `typescript`

src/react.ts
```diff
@@ -32,6 +32,7 @@ export class ReactProject extends NodeProject {
 
   constructor(options: ReactProjectOptions) {
     super({
+      jest: false,
```
kcwinner, line 35: Needed because now `node-project` creates the jest config instead of `typescript`

src/node-project.ts
```diff
@@ -173,6 +174,13 @@ export interface NodeProjectCommonOptions {
    */
   readonly buildWorkflow?: boolean;
 
+  /**
+   * Define a GitHub workflow step for sending
+   * code coverage metrics to https://codecov.io/
+   * @default false
+   */
+  readonly buildWorkflowCodeCoverage?: boolean;
```
kcwinner, line 182: Still not sold on this name. Should it just be `codeCov`? Similar to `antitamper`?

.npmignore
```diff
@@ -1,6 +1,8 @@
 # ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
 /.projenrc.js
 /.versionrc.json
+/coverage
+/test
```
kcwinner, line 5: Changed because of jest being moved to node-project

.gitignore
```diff
@@ -35,9 +35,9 @@ jspm_packages/
 .yarn-integrity
 # parcel-bundler cache (https://parceljs.org/)
 .cache
+/coverage
```
kcwinner, line 38: Changed because of jest being moved to node-project

package.json
```diff
@@ -15,14 +15,14 @@
     "no-changes": "(git log --oneline -1 | grep -q \"chore(release):\") && echo \"No changes to release.\"",
     "bump": "yarn run --silent no-changes || standard-version",
     "release": "yarn run --silent no-changes || (yarn run bump && git push --follow-tags origin master)",
+    "test": "jest --passWithNoTests --updateSnapshot && yarn run eslint",
```
kcwinner, line 18: Changed because of jest being moved to node-project

kcwinner:
Should be good for review. Had to change a few things because of antitamper with the new order of jest running

src/jest.ts
```diff
@@ -13,6 +12,12 @@ export interface JestOptions {
    */
   readonly coverage?: boolean;
 
+  /**
+   * The directory where Jest should output its coverage files.
+   * @default coverage
```
eladb, line 17:
```suggestion
   * @default "coverage"
```

src/jest.ts
```diff
@@ -59,44 +64,24 @@ export class Jest {
...
   private readonly ignorePatterns: string[];
+  private readonly project: NodeProject;
 
   constructor(project: NodeProject, options: JestOptions = { }) {
-    const version = options.jestVersion ?? DEFAULT_JEST_VERSION;
+    this.project = project;
 
-    project.addDevDependencies({ jest: version });
+    const version = options.jestVersion ?? DEFAULT_JEST_VERSION;
+    project.addDevDeps(`jest@${version}`);
```
eladb, line 73: I think that with the new "deps" mechanism we don't really need to specify a version by default:

```suggestion
    const version = options.jestVersion;
    const jestDep = version ? `jest@${version}` : 'jest';
    project.addDevDeps(jestDep);
```

This will have the effect of installing the latest jest version when the project is first created and then upgrades can be managed through `yarn upgrade`, dependabot, or any other tool

src/jest.ts
```diff
@@ -1,9 +1,8 @@
 import { NodeProject } from './node-project';
-import { Semver } from './semver';
 import { StartEntryCategory } from './start';
 import { TypescriptConfig } from './typescript';
 
-const DEFAULT_JEST_VERSION = Semver.caret('26.4.2');
+const DEFAULT_JEST_VERSION = '^26.4.2';
```
eladb, line 5: Not required any more:

```suggestion
```

src/jest.ts
```diff
@@ -34,7 +39,7 @@ export interface JestOptions {
    *
    * @default ^26.4.2
```
eladb, line 40:
```suggestion
   * @default - installs the latest jest version
```

src/jest.ts
```diff
@@ -131,4 +116,31 @@ export class Jest {
...
+    this.config.globals = {
+      'ts-jest': {
+        tsconfig: config.fileName,
+      },
+    };
+
+    // add relevant deps
+    this.project.addDevDeps(
+      '@types/jest@^26.0.7',
+      'ts-jest@^26.1.0',
```
eladb, line 143: Install latest:

```suggestion
      '@types/jest',
      'ts-jest',
```

src/node-project.ts
eladb: Yes, I like `codeCov` better.

src/node-project.ts
```diff
@@ -598,6 +618,16 @@ export class NodeProject extends Project {
    */
   public readonly runScriptCommand: string;
 
+  /**
+   * The directory in which .ts tests reside.
```
eladb, line 622:
```suggestion
   * The directory in which tests reside.
```

src/node-project.ts
```diff
@@ -598,6 +618,16 @@ export class NodeProject extends Project {
...
+   */
+  public readonly testdir: string;
+
+  /**
+   * The Jest configuration
```
eladb, line 627:
```suggestion
   * The Jest configuration (if enabled)
```

src/node-project.ts
```diff
@@ -752,12 +784,24 @@ export class NodeProject extends Project {
     this.antitamper = (options.buildWorkflow ?? true) && (options.antitamper ?? true);
     this.nodeVersion = options.workflowNodeVersion ?? this.minNodeVersion;
 
+    // configure jest if enabled
+    // must be before the build/release workflows
+    if (options.jest ?? true) {
+      this.jest = new Jest(this, {
+        ...options.jestOptions,
```
eladb, line 791: I was expecting testdir to be reflected somewhere in jest options, no?

src/node-project.ts
```diff
@@ -173,6 +174,13 @@ export interface NodeProjectCommonOptions {
    */
   readonly buildWorkflow?: boolean;
 
+  /**
+   * Define a GitHub workflow step for sending
+   * code coverage metrics to https://codecov.io/
```
eladb, line 179: We should mention in this description that a secret needs to be added to the repo in order to be able to submit code overage to codecov.io.

Which makes me think that maybe a better API for enabling code coverage would be to simply specify the secret name. This will signal to users that a secret is needed.

Instead of:

```ts
{ buildWorkflowCodeCoverage: true } // with implicit secret name
```

Users will simply do:

```ts
{ codeCovSecret: 'CODECOV_TOKEN' }
```

src/node-project.ts
eladb: `workflowBootstrapSteps` is needed somewhere which is why it's exposed.

src/node-project.ts
```diff
@@ -1117,6 +1162,22 @@ export class NodeProject extends Project {
...
+   * Returns the set of steps to run codecoverage action in a github workflow.
+   */
+  public get workflowCodecoverageSteps(): any[] {
+    return [
+      {
+        name: 'Upload coverage to Codecov',
+        uses: 'codecov/codecov-action@v1',
+        with: {
+          token: '${{ secrets.CODECOV_TOKEN }}',
+          directory: this.jest?.config?.coverageDirectory,
```
eladb, line 1175: I am assuming we should fail if `this.jest?.config?.coverageDirectory` is undefined, no?

eladb (changes requested):

src/node-project.ts
kcwinner: For public repositories a token is not required. 
I just changed to
```
/**
   * Define a GitHub workflow step for sending code coverage metrics to https://codecov.io/
   * Uses codecov/codecov-action@v1
   * A secret named 'CODECOV_TOKEN' is required for private repos
   * @default false
   */
  readonly codeCov?: boolean;
```

However, I do like representing it with `{ codeCovSecret: 'CODECOV_TOKEN' }` way better instead of assuming they will call it `CODECOV_TOKEN`. Only makes the public repo a bit more awkward. I can make that switch if we want

src/node-project.ts
kcwinner: This seems to relate to Issue #74 a little bit I think. I put an example of what things could look like. But perhaps these should be exposed instead for building composable workflows?

I'm happy to remove it, though

src/node-project.ts
kcwinner: That gets passed down via `project` unless we want it to be explicit

```
this.config = {
      clearMocks: true,
      collectCoverage: options.coverage ?? true,
      coverageDirectory: options.coverageDirectory ?? 'coverage',
      coveragePathIgnorePatterns: this.ignorePatterns,
      testPathIgnorePatterns: this.ignorePatterns,
      testMatch: [
        `**/${project.testdir}/**/*.js?(x)`,
      ],
    };
```

eladb: Feels slightly like the reverse direction. I rather this to be an explicit option for Jest and have the Project pass it down. No?

src/node-project.ts
eladb: Okay good to know about public repos. So I'd just introduce two options:

```
codeCov?: boolean;
codeCovTokenSecret?: string;
```

What we do in these cases to improve ergonomics is that if a user specifies a value to `codeCovTokenSecret` it implies that codeCov is true.

kcwinner: Added and cleaned up the condition logic

src/node-project.ts
kcwinner: Should the jest options `testdir` override the project level `testdir` then? Otherwise we can end up with a scenario like below. This causes tests to live in `test` directory but jest is configured with `__tests__` pattern so they never get read.

```
  testdir: 'test',
  codeCov: true,
  jestOptions: {
    testdir: '__tests__'
  }
```

Also - I found this (below) in the `AwsCdkTypeScriptApp`  class. Is there any specific reason for this or can I remove it? I don't see any other hardcoded references to `'test`' in the project type.
```
if (this.testdir !== 'test') {
  throw new Error('test sources are expected under the "test" directory');
}
```

src/github-actions.ts
```diff
@@ -0,0 +1,80 @@
+// import { NodeProject } from './node-project';
+
+export interface GithubActionStep {
```
eladb, line 3: Seems like this is unrelated.

src/node-project.ts
```diff
@@ -1411,6 +1465,19 @@ export interface NodeBuildWorkflowOptions {
...
+  /**
+   * Run codecoverage step
+   * Send to https://codecov.io/
+   * @default false
+   */
+  readonly codeCov?: boolean;
+
+  /**
+   * The secret name for the https://codecov.io/ token
+   * @default "CODECOV_TOKEN"
```
eladb, line 1478: Would that work for public repos? Wouldn't GitHub look up the secret?

I was thinking that this should have a default value of `undefined` and if it's defined, it will imply `codeCov: true`.

src/node-project.ts
```diff
@@ -1451,6 +1518,17 @@ export class NodeBuildWorkflow extends GithubWorkflow {
         // build (compile + test)
         { run: `${project.runScriptCommand} build` },
 
+        // run codecov if enabled or a secret token name is passed in
+        // AND jest must be configured
+        ...(options.codeCov || options.codeCovTokenSecret) && project.jest?.config ? [{
+          name: 'Upload coverage to Codecov',
+          uses: 'codecov/codecov-action@v1',
+          with: {
+            token: `\${{ secrets.${options.codeCovTokenSecret ?? 'CODECOV_TOKEN'} }}`,
```
eladb, line 1527: The token is only needed if `codeCovTokenSecret` is defined, no? Otherwise, GitHub will require that `CODECOV_TOKEN` exists in the repo, no?

src/node-project.ts
kcwinner: It still works even with a defined (but empty) secret. It could be confusing, though. 

https://github.com/kcwinner/cdk-appsync-transformer/runs/1355438476?check_suite_focus=true
https://codecov.io/github/kcwinner/cdk-appsync-transformer/commit/1f60a507fec0ebbbd97df2e5cbf159dd7cebbef1

package.json
```diff
@@ -159,27 +159,27 @@
...
+    "coverageDirectory": "coverage",
     "coveragePathIgnorePatterns": [
       "/node_modules/"
     ],
     "testPathIgnorePatterns": [
       "/node_modules/"
     ],
-    "preset": "ts-jest",
     "testMatch": [
-      "**/__tests__/**/*.ts?(x)",
```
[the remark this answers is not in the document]
kcwinner: Should this be supported alongside the testdir declaration?

Previously testmatch was the following no matter what value was supplied for `testdir`
```
"testMatch": [
      "**/__tests__/**/*.ts?(x)",
      "**/?(*.)+(spec|test).ts?(x)"
    ],
```

src/node-project.ts
kcwinner: Same as above

src/node-project.ts
```diff
@@ -598,6 +625,16 @@ export class NodeProject extends Project {
...
 
+  /**
+   * The directory in which tests reside.
+   */
+  public readonly testdir: string;
+
+  /**
+   * The Jest configuration (if enabled)
+   */
+  public jest?: Jest;
```
eladb, line 636:
```suggestion
  public readonly jest?: Jest;
```

src/node-project.ts
```diff
@@ -173,6 +174,20 @@ export interface NodeProjectCommonOptions {
...
+   * Uses codecov/codecov-action@v1
+   * A secret is required for private repos. Configured with @codeCovTokenSecret
+   * @default false
+   */
+  readonly codeCov?: boolean;
+
+  /**
+   * Define the secret name for a specified https://codecov.io/ token
+   * A secret is required to send coverage for private repositories
+   */
```
eladb, line 188:
```suggestion
  /**
   * Define the secret name for a specified https://codecov.io/ token
   * A secret is required to send coverage for private repositories
   * @default - if this option is not specified, only public repositories are supported
   */
```

src/node-project.ts
```diff
@@ -661,6 +698,9 @@ export class NodeProject extends Project {
       }
     }
 
+    this.testdir = options.testdir ?? 'test';
+    this.testdir = options.jestOptions?.testdir ?? options.testdir ?? 'test';
```
eladb, line 702: This is a bit error prone.

I think if testdir is specified both in `jestOptions` and in options, they should be the same. Otherwise, fail.

Also doesn't seem like we need both lines.

src/node-project.ts
eladb: the check in the awscdk typescript project encodes some hidden assumption, but for the life of me I can't recall what was it. Let's leave it please.

src/node-project.ts
```diff
@@ -1451,6 +1519,20 @@ export class NodeBuildWorkflow extends GithubWorkflow {
...
         { run: `${project.runScriptCommand} build` },
 
+        // run codecov if enabled or a secret token name is passed in
+        // AND jest must be configured
+        ...(options.codeCov || options.codeCovTokenSecret) && project.jest?.config ? [{
+          name: 'Upload coverage to Codecov',
+          uses: 'codecov/codecov-action@v1',
+          with: options.codeCovTokenSecret
+            ? {
+              token: `\${{ secrets.${options.codeCovTokenSecret} }}`,
```
eladb, line 1529: What happens here if this options is not defined?

eladb (commented):
Almost there...

package.json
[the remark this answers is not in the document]
eladb: Yes that's a good point. Just a regression of current behavior so let's call out in a breaking change clause in the PR description perhaps.

src/node-project.ts
kcwinner: It has to be defined at this point, yeah? Otherwise we omit it and just pass in directory. I tried to find a better way to write it but the yaml library doesn't ignore null/undefined and it just gets passed into the workflow, which github doesn't like.

```
with: options.codeCovTokenSecret
            ? {
              token: `\${{ secrets.${options.codeCovTokenSecret} }}`,
              directory: project.jest.config.coverageDirectory,
            } : {
              directory: project.jest.config.coverageDirectory,
            },
```

src/node-project.ts
kcwinner: Yeah my mistake on committing it. Was just supposed to be a part of this conversation
https://github.com/projen/projen/pull/266#discussion_r517429109

src/node-project.ts
eladb: okay now I see. sorry.

eladb (commented):
A few unresolved comments

kcwinner:
I think the last thing is figuring out how to handle the project `testdir` vs jestOptions `testdir`. Commented [here](https://github.com/projen/projen/pull/266#discussion_r518079677)

package.json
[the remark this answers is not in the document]
kcwinner: How is that annotated? Is it the same as in the CDK contributions?
```
BREAKING CHANGE: <describe what changed + link for details>
```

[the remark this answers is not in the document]
eladb: Yes

src/node-project.ts
eladb: Let's just require that `jestOptions.testdir` is the same as the project's:

```ts
const testdir = options.testdir ?? 'test';
if (jestOptions.testdir && jestOptions.testdir !== testdir) {
  throw new Error(`boom`);
}
```

src/typescript.ts
```diff
@@ -249,13 +229,7 @@ export class TypeScriptProject extends NodeProject {
         this.addTestCommand(`rm -fr ${this.libdir}/`);
```
kcwinner, line 229: Need to figure out a solution to this. With the change, if you compile before test, you get the following, which is out of order
```
"test": "jest --passWithNoTests --updateSnapshot && rm -fr lib/ && yarn run eslint",
```

src/typescript.ts
```diff
@@ -246,16 +226,13 @@ export class TypeScriptProject extends NodeProject {
       if (!compileBeforeTest) {
         // make sure to delete "lib" *before* running tests to ensure that
         // test code does not take a dependency on "lib" and instead on "src".
-        this.addTestCommand(`rm -fr ${this.libdir}/`);
-      }
+        this.addScript('test', `rm -fr ${this.libdir}/`);
```
kcwinner, line 229: @eladb Open to suggestions here. Since we need this to be at the beginning of the test command
It won't cause issues downstream but if something happens before this it will be dropped from the test command.

Chriscbr: Maybe there is a need for a `prependScriptCommand()` API of some sort?

A more hacky workaround could be to change the `deleteScript` API to return the current script(s) given the particular command, so that you could do something like:
```ts
const testCmds = deleteScript('test');
this.addScript('test', testCmds.concat([`rm -fr ${this.libdir}/`]).join('&&'));
```
But I don't think that's an especially healthy abstraction

package.json
```diff
@@ -159,27 +159,27 @@
...
-      "**/?(*.)+(spec|test).ts?(x)"
+      "**/test/?(*.)+(spec|test).ts?(x)"
```
eladb, line 173: I believe this shouldn't have the `**/` prefix. Otherwise, any subdirectory called `test/` could have test in it, and we just want tests to be under `test`. Also, it doesn't allow subdirectories under `test` to include tests, which is also something people use.

I think the correct pattern is:

```
test/**/?(.*)+(spec|test).ts?(x)
```

I am also thinking that perhaps we don't need to change this behavior in _this_ PR because it's not instrumental to enabling codecov.io and it will break users who decided to put tests under their `src/` directory.

When thinking about it, it's actually legitimate.

src/jest.ts
```diff
@@ -59,76 +69,94 @@ export class Jest {
...
     this.ignorePatterns = options.ignorePatterns ?? ['/node_modules/'];
 
+    const coverageDirectory = options.coverageDirectory ?? 'coverage';
     this.config = {
       clearMocks: true,
       collectCoverage: options.coverage ?? true,
+      coverageDirectory: coverageDirectory,
       coveragePathIgnorePatterns: this.ignorePatterns,
       testPathIgnorePatterns: this.ignorePatterns,
+      testMatch: [
```
eladb, line 92: I think we should just ignore the testdir here and just plug in:

```js
[
  '**/__tests__/**/*.js?(x)',	
  '**/?(*.)+(spec|test).js?(x)',
]
```

Which I believe is the default for jest.

src/jest.ts
```diff
@@ -59,76 +69,94 @@ export class Jest {
...
+  /**
+   * Configures jest with typescript options
+   * @param config The Jest Typescript Config
+   */
+  public addTypescriptOptions(config: TypescriptConfig) {
+    this.config.preset = 'ts-jest';
+
+    // only process .ts files
+    this.config.testMatch = [
+      `**/${this.testdir}/?(*.)+(spec|test).ts?(x)`,
```
eladb, line 127: Don't encode testdir into this pattern.

src/jest.ts
```diff
@@ -59,76 +69,94 @@ export class Jest {
   public readonly config: any;
 
   private readonly ignorePatterns: string[];
+  private readonly project: NodeProject;
+  private readonly testdir: string;
```
eladb, line 73: Remove

src/jest.ts
```diff
@@ -1,18 +1,28 @@
...
 import { TypescriptConfig } from './typescript';
 
-const DEFAULT_JEST_VERSION = Semver.caret('26.4.2');
 
 export interface JestOptions {
+  /**
+   * The directory in which tests reside.
+   * @default "test"
+   */
+  readonly testdir?: string;
```
eladb, line 12: Remove this please

src/typescript.ts
eladb: Yes, there's a bit of a modeling issue with those mutable scripts. Let's pick this up in a subsequent PR?

eladb (changes requested):
Thanks so much for bearing with me! I know it's been a bit of a back and forth with this PR.
I thought a little more about jest match patterns and I think we should just leave the current behavior untouched. I think it's very common in jest to just put tests everywhere. It's not ideal from a bundling perspective but it's also not the end of the world, and we can decide to tackle this more formally in a subsequent PR.

To my understanding, all the stuff related to `testdir` is actually not instrumental in order to support codecov, correct? If that is indeed the case, let's limit the scope of this change to what's needed for codecov.

kcwinner:
I agree about leaving the `testdir` out of the matching policy. I think people sprinkle them next to the they are testing and others don't.

kcwinner:
Reverted changes regarding Jest testdir. Reverted testMatch config changes.

eladb (approved):

PR #267, opened 2020-11-02
chore(cdk): remove deprecated Semver

hoegertn:
Removes the use of the deprecate Semver class and old-style dependency methods.

### DIFF
--- a/src/awscdk-app-ts.ts
+++ b/src/awscdk-app-ts.ts
@@ -22,6 +22,8 @@ export enum CdkApprovalLevel {
 export interface AwsCdkTypeScriptAppOptions extends TypeScriptProjectOptions {
   /**
    * AWS CDK version to use.
+   *
+   * @default 1.63.0
    */
   readonly cdkVersion: string;
 
--- a/src/awscdk-app-ts.ts
+++ b/src/awscdk-app-ts.ts
@@ -105,7 +105,7 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
     this.cdkVersion = options.cdkVersionPinning ? options.cdkVersion : `^${options.cdkVersion}`;
 
     // CLI
-    this.addDevDeps(`aws-cdk@${this.cdkVersion}`);
+    this.addDevDeps(this.formatModuleSpec('aws-cdk'));
 
     this.addCdkDependency('@aws-cdk/assert');
     this.addCdkDependency('@aws-cdk/core');
@@ -170,7 +170,11 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
    * @param modules The list of modules to depend on
    */
   public addCdkDependency(...modules: string[]) {
-    this.addDeps(...modules.map(m => `${m}@${this.cdkVersion}`));
+    this.addDeps(...modules.map(this.formatModuleSpec));
+  }
+
+  private formatModuleSpec(module: string): string {
+    return `${module}@${this.cdkVersion}`;
   }
 }
 
--- a/src/awscdk-app-ts.ts
+++ b/src/awscdk-app-ts.ts
@@ -2,7 +2,6 @@ import * as path from 'path';
 import * as fs from 'fs-extra';
 import { Component } from './component';
 import { JsonFile } from './json';
-import { Semver } from './semver';
 import { StartEntryCategory } from './start';
 import { TypeScriptAppProject, TypeScriptProjectOptions } from './typescript';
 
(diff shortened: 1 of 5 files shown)


src/awscdk-app-ts.ts
```diff
@@ -23,8 +22,6 @@ export enum CdkApprovalLevel {
 export interface AwsCdkTypeScriptAppOptions extends TypeScriptProjectOptions {
   /**
    * AWS CDK version to use.
-   *
-   * @default 1.63.0
    */
   readonly cdkVersion: string;
```
eladb, line 26: A required field with no "@default" means that it will impossible to run `projen new` without specifying this switch (we should probably [add a test](https://github.com/projen/projen/issues/268) that prevents this).

I understand putting "1.63.0" here is a bit weak, and I am wondering what is the ideal behavior. Here are a few options:

1. Create a [default macro] such as `$GIT_USER_NAME` and `$GIT_REMOTE` that will allow evaluating the latest version of an npm package during `projen new` and assigning that value here (e.g. `$NPM_MODULE_VERSION[aws-cdk]`).
2. Make `cdkVersion` an _optional_ attribute and call `addDep()` without a version specification - this means that projen will install the latest version of these modules and then the specific version will be managed in `package.json`. The problem with this approach is the fact that we really need to ensure that all CDK versions are synchronized, and one of the ways this is done is by controlling them through projen.
3. Determine the latest CDK version during _projen build_, including it as a resource in projen itself and using that as the default.

Personally, I am okay with leaving some static value for now (the current latest) and revisiting in a subsequent PR.

[default macro]: https://github.com/projen/projen/blob/a654767a7379309744dd73b48d7fd2f5b5bff67e/src/cli/cmds/new.ts#L116

eladb (changes requested):

src/awscdk-app-ts.ts
hoegertn: Already fixed as I read #268 

I wasn't aware of this and as it is not optional I was wondering what a default would be good for.

hoegertn: On a side note, with this Semver change is it really meant to be managed in package.json instead of projen? I thought the goal was to make the package.json throw-away capable.

eladb: In order to enable using dependency management tools like dependabot/ncu, we need specific versions to actually be managed inside package.json, so at the moment, it's "partially immutable". Versions are respected.

src/awscdk-construct.ts
```diff
@@ -137,10 +136,8 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
    */
   public addCdkDependencies(...deps: string[]) {
     // this ugliness will go away in cdk v2.0
-    for (const dep of deps) {
-      this.addPeerDependencies({ [dep]: this.version });
-      this.addDependencies({ [dep]: this.version });
-    }
+    this.addDeps(...deps.map(m => `${m}@${this.version}`));
```
eladb, line 139:
```suggestion
    this.addPeerDeps(...deps.map(m => `${m}@${this.version}`));
```

src/awscdk-construct.ts
```diff
@@ -137,10 +136,8 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+    this.addDeps(...deps.map(m => `${m}@${this.version}`));
```
eladb, line 140: Can we extract the logic that formats a CDK module spec into a private method:

```ts
private formatModuleSpec(module: string) {
  return `${module}@${this.version}`;
}
```

eladb (changes requested):

src/awscdk-construct.ts
hoegertn: Done

src/awscdk-app-ts.ts
```diff
@@ -171,10 +170,7 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
    * @param modules The list of modules to depend on
    */
   public addCdkDependency(...modules: string[]) {
-    for (const m of modules) {
-      // since synthesis runs at build time, CDK dependencies should be dev-dependencies
-      this.addDependencies({ [m]: this.cdkVersion });
-    }
+    this.addDeps(...modules.map(m => `${m}@${this.cdkVersion}`));
```
eladb, line 173: use `formatModuleSpec`

hoegertn: It is the only use case on the app project. So I have three options:

* Use the template string as currently
* Copy the method to this class and use it once
* Extract the method somewhere else and use it

Do you have a preference?

hoegertn: Ok it is used twice so it might be ok

eladb (approved):

PR #270, opened 2020-11-02
fix(cli): improve git config detection

hoegertn:
This PR improves the handling of the git default values by:

* Taking `--include` into account to resolve included git config files
* Using the `github.user` config for remote URL generation


src/cli/cmds/new.ts
```diff
@@ -111,12 +111,18 @@ function renderParams(params: any) {
 
 function processDefault(value: string) {
   const basedir = path.basename(process.cwd());
-  const userEmail = execOrUndefined('git config --get --global user.email') ?? 'user@domain.com';
+  const userEmail = execOrUndefined('git config --get --global --include user.email') ?? 'user@domain.com';
```
Chriscbr, line 114: Is `--include` the same as `--includes`? (https://git-scm.com/docs/git-config#Documentation/git-config.txt---no-includes)

eladb: Yes, sounds like this should be `--includes`:

```suggestion
  const userEmail = execOrUndefined('git config --get --global --includes user.email') ?? 'user@domain.com';
```

src/cli/cmds/new.ts
```diff
@@ -111,12 +111,18 @@ function renderParams(params: any) {
...
   switch (value) {
     case '$BASEDIR': return basedir;
-    case '$GIT_REMOTE': return execOrUndefined('git remote get-url origin') ?? `https://github.com/${userEmail?.split('@')[0]}/${basedir}.git`;
-    case '$GIT_USER_NAME': return execOrUndefined('git config --get --global user.name');
+    case '$GIT_REMOTE':
+      const origin = execOrUndefined('git remote get-url origin');
+      if (origin) {
+        return origin;
+      }
+      const slug = execOrUndefined('git config --get --global --include github.user') ?? userEmail?.split('@')[0];
```
eladb, line 123: Maybe extract to a function:

```suggestion
      const slug = gitConfigGet('github.user') ?? userEmail?.split('@')[0];
```

eladb (changes requested):

src/cli/cmds/new.ts
hoegertn: Done, the interesting thing is: It works also

eladb (approved):

PR #276, opened 2020-11-03
feat: README samples (again)

mbonig:
Re-adds the README file and sample file rework. Now uses the synth and _outdir instead of __dirname.


src/sample-file.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param dir directory to add files to. If directory already exists, nothing is added.
+   * @param options options for which files to create.
+   */
+  constructor(project: Project, dir: string, options: SampleDirOptions) {
+    super(project);
+    this.dir = dir;
+    this.options = options;
+  }
+
+  synthesize(_outdir: string) {
```
Chriscbr, line 88:
```suggestion
  public synthesize(outdir: string) {
```

mbonig:
do y'all care if I squash my commits?

src/sample-file.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param contents - the contents of the file to write
+   * @return boolean - whether a new file was written or not.
+   * @private
+   */
+  private writeOnceFileContents(dir: string, filename: string, contents: string) {
+    const fullFilename = path.join(dir, filename);
+    if (fs.existsSync(fullFilename)) {
+      return;
+    }
+    fs.mkdirpSync(dir);
```
eladb, line 54: please use `writeFile` from `util.ts` which takes care of creating the directory tree

src/sample-file.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  public synthesize(outdir: string) {
+    const fullOutdir = path.join(outdir, this.dir);
+    if (fs.pathExistsSync(fullOutdir)) {
+      return;
+    }
+
+    fs.mkdirpSync(fullOutdir);
+    for (const filename in this.options.files) {
+      fs.writeFileSync(path.join(fullOutdir, filename), this.options.files[filename]);
```
eladb, line 96: Please use writeFile from util.ts (and then mkdirp is not required)

eladb (changes requested):
Can you please confirm that you have manually tested that `projen new` works?

mbonig:
> Can you please confirm that you have manually tested that `projen new` works?

Yes. I went into `test/` and created a new directory, cd'd into that directory and then did `../../bin/projen new awscdk-app-ts`. The resulting code built as expected (projen runs without errors and a `yarn cdk synth` compiles without errors).

src/sample-file.ts
```diff
@@ -0,0 +1,98 @@
...
+   * @param contents - the contents of the file to write
+   * @return boolean - whether a new file was written or not.
+   * @private
+   */
+  private writeOnceFileContents(dir: string, filename: string, contents: string) {
+    const fullFilename = path.join(dir, filename);
+    if (fs.existsSync(fullFilename)) {
+      return;
+    }
+    writeFile(fullFilename, contents);
```
eladb, line 55: By default I think `writeFile` creates readonly files, so here you'd need to explicitly set:

```ts
writeFile(fullFilename, contents, { readonly: false });
```

eladb (changes requested):
files are written as read-only.

eladb (approved):

PR #279, opened 2020-11-03
feat(jest): test results processing

### LINKED ISSUE #278: feat(jest): allow test results processors + config
This is useful in CI tools where limited test result parsing tools are available. In our case, we're pretty much forced to use JUnit formatting, and so use [jest-junit](https://www.npmjs.com/package/jest-junit).

I'd love to be opinionated about the output directory, but unfortunately, all of the processor modules out there implement different config interfaces.

Here's what I had in mind in it's most basic form:

```javascript
// .projenrc.js
  ...
  jestOptions: {
    testResultsProcessor: "jest-junit",
    testResultsProcessorConfig: {
      "jest-junit": {
        outputDirectory: "./test-reports"
      }
    }
  }
  ...
}

// output package.json
{
  "jest": {
    "testResultsProcessor": "jest-junit",
    ...
  },
  "jest-junit": {
    outputDirectory: "./test-reports"
  }
}
```

I see processors eventually have their own opinionated configuration objects long term, which would eliminate the need for `testResultsProcessorConfig`.

gradybarrett:
First stab at the least opinionated implementation.

A possible next step would be a dedicated, opinionated JestJUnit config object

Fixes #278

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "eslint-import-resolver-typescript": "^2.3.0",
     "eslint-plugin-import": "^2.22.1",
     "jest": "^26.4.2",
+    "jest-junit": "^12",
     "jsii": "^1.11.0",
     "jsii-diff": "^1.11.0",
     "jsii-docgen": "^1.3.2",
@@ -172,6 +173,15 @@
     "testMatch": [
       "**/__tests__/**/*.ts?(x)",
       "**/?(*.)+(spec|test).ts?(x)"
+    ],
+    "reporters": [
+      "default",
+      [
+        "jest-junit",
+        {
+          "outputDirectory": "test-reports"
+        }
+      ]
     ],
     "preset": "ts-jest",
     "globals": {
--- a/src/jest.ts
+++ b/src/jest.ts
@@ -3,6 +3,7 @@ import { NodeProject } from './node-project';
 import { StartEntryCategory } from './start';
 import { TypescriptConfig } from './typescript';
 
+const DEFAULT_TEST_REPORTS_DIR = 'test-reports';
 
 export interface JestOptions {
   /**
@@ -34,6 +35,22 @@ export interface JestOptions {
   readonly typescript?: TypescriptConfig;
 
   /**
+   * Result processing with jest-junit.
+   *
+   * Output directory is `test-reports/`.
+   *
+   * @default true
+   */
+  readonly junitReporting?: boolean;
+
+  /**
+   * Preserve the default Jest reporter when additional reporters are added.
+   *
+   * @default true
+   */
+  readonly preserveDefaultReporters?: boolean;
+
+  /**
    * The version of jest to use.
    *
    * @default - installs the latest jest version
@@ -47,6 +64,8 @@ export interface CoverageThreshold {
   readonly lines?: number;
   readonly statements?: number;
 }
+
+type JestReporter = [string, {[key: string]: any}] | string;
 
 /**
  * Installs the following npm scripts:
@@ -64,6 +83,7 @@ export class Jest {
 
   private readonly ignorePatterns: string[];
   private readonly project: NodeProject;
+  private readonly reporters: JestReporter[];
 
   constructor(project: NodeProject, options: JestOptions = { }) {
     this.project = project;
@@ -75,6 +95,13 @@ export class Jest {
     this.ignorePatterns = options.ignorePatterns ?? ['/node_modules/'];
 
     const coverageDirectory = options.coverageDirectory ?? 'coverage';
+
+    this.reporters = [];
+
+    if (options.preserveDefaultReporters ?? true) {
+      this.reporters.unshift('default');
+    }
+
     this.config = {
       clearMocks: true,
       collectCoverage: options.coverage ?? true,
@@ -85,7 +112,25 @@ export class Jest {
         '**/__tests__/**/*.js?(x)',
         '**/?(*.)+(spec|test).js?(x)',
       ],
+      reporters: this.reporters,
     };
+
+    if (options.junitReporting ?? true) {
+      const reportsDir = DEFAULT_TEST_REPORTS_DIR;
+
+      this.addReporter([
+        'jest-junit',
+        { outputDirectory: reportsDir },
+      ]);
+
+      project.addDevDeps('jest-junit@^12');
+
+      project.gitignore.exclude(
+        '# jest-junit artifacts',
+        `/${reportsDir}/`,
+        'junit.xml',
+      );
+    }
 
     if (options.coverageThreshold) {
       this.config.coverageThreshold = {
@@ -152,5 +197,9 @@ export class Jest {
     });
 
     this.project.addScript('test:update', 'jest --updateSnapshot');
+  }
+
+  public addReporter(reporter: JestReporter) {
+    this.reporters.push(reporter);
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,9 @@ jspm_packages/
 .yarn-integrity
 # parcel-bundler cache (https://parceljs.org/)
 .cache
+# jest-junit artifacts
+/test-reports/
+junit.xml
 /coverage
 /lib
 /dist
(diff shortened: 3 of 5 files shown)


src/jest.ts
```diff
@@ -29,6 +29,20 @@ export interface JestOptions {
    */
   readonly typescript?: TypescriptConfig;
 
+  /**
+   * String pointing to a node module responsible for processing test results
+   *
+   * @example
+   *
+   * { testResultsProcessor: 'jest-junit' }
```
eladb, line 37:
```suggestion
   * @example "jest-junit"
```

eladb (changes requested):
Can we just be opinionated and simply add jest-junit support, enabled by default?

gradybarrett:
> Can we just be opinionated and simply add jest-junit support, enabled by default?

I didn't really consider that. I suppose we could. Let me switch this to reporters and do a little testing to make sure it works how I expect.

gradybarrett:
@eladb here's an interpretation of what you suggested. Let me know what you think.

cc: @Chriscbr

src/jest.ts
```diff
@@ -67,13 +77,33 @@ export class Jest {
...
 
+    if (options.junitReporting ?? true) {
+      const reportsDir = DEFAULT_TEST_REPORTS_DIR;
+
+      this.config.reporters.push([
+        'jest-junit',
+        { outputDirectory: reportsDir },
+      ]);
+
+      project.addDevDeps('jest-junit@^12');
```
Chriscbr, line 98: do we need a specific version here?

src/jest.ts
```diff
@@ -67,13 +77,33 @@ export class Jest {
...
     this.config = {
       clearMocks: true,
       collectCoverage: options.coverage ?? true,
       coveragePathIgnorePatterns: this.ignorePatterns,
       testPathIgnorePatterns: this.ignorePatterns,
+      reporters: reporters,
     };
 
+    if (options.junitReporting ?? true) {
+      const reportsDir = DEFAULT_TEST_REPORTS_DIR;
```
Chriscbr, line 91: Maybe specify this as configurable with a prop like `junitReportingDir`

src/jest.ts
```diff
@@ -67,13 +77,33 @@ export class Jest {
 
     this.ignorePatterns = options.ignorePatterns ?? ['/node_modules/'];
 
+    const reporters: Array<[string, {[key: string]: any}] | string> = ['default'];
```
Chriscbr, line 80: I think this should be a private field on the class. Also maybe we can add a method like `addReporter()` for it?

package.json
```diff
@@ -47,6 +47,7 @@
     "eslint-import-resolver-typescript": "^2.3.0",
     "eslint-plugin-import": "^2.22.1",
     "jest": "^26.4.2",
+    "jest-junit": "^12",
```
Chriscbr, line 50: I think unless there is a need for junit-formatted test reports in projen's own CI/CD processes, we should change the `.projenrc.js` for projen itself to disable junit reporting.

Chriscbr (changes requested):
Just a few comments, looks good overall! Personally I think it would make more sense having junit reporting off by default, but that's just me

src/jest.ts
Chriscbr: Actually this might be fine as is, since it looks like a lot of the other constructs in projen give specific versions by default (not sure if this is better or not, but hey)

src/jest.ts
gradybarrett: I was headed down that road when I stopped myself because it felt like I was going back towards what I had originally, just packaged differently. I'm still trying to figure out the opinionated vs configurable line here. I figured I'd at least tee that change up for the next person, unless you or Elad feels strongly that it should be configurable right now.

src/jest.ts
gradybarrett: I'm on the fence on this one. I suppose not, but in either case, someone can `project.addDevDeps('jest-junit@11')` after project initialization in order to override the version declared here. Since that's the case I think I'd rather see a known stable version declared here by default.

package.json
gradybarrett: > Just a few comments, looks good overall! Personally I think it would make more sense having junit reporting off by default, but that's just me

Hah, funny you should say that. I flip-flopped and committed about 3 times trying to decide how I felt about that. There's something nice about it just being on by default. If I need/want it, it's just there. It's not a particularly "expensive" operation, and I made sure the default reporter still runs.

I am, however, inclined to agree with you that generally, this might make more sense as an opt-in, rather than an opt-out feature.

gradybarrett: I figured I would see how you guys felt about it being on by default by committing the changes to the actual projen project. I agree that we should probably shut it off for projen specifically if we do leave it on by default.

gradybarrett: Switched back to disabled by default.

gradybarrett: I think I'm actually going to leave it the way it is. I feel like opt-out is actually consistent with how other configs are being handled so far in Projen, so I'd hate this to be one that's unexpectedly different.

It's also in line with 'coverageReporters' behavior, which gives you all 4 formats by default.

I'll turn it off for the actual projen project though.

src/jest.ts
gradybarrett: I did some research and testing on `reporters` vs `coverageReporters`, and unfortunately, the two do not overlap in the same way as `reporters` and `testResultsProcessor`, so that will need to be a separate feature. The new JestReporters class can, however, be reused for both, so adding coverage reporters functionality should be easy after this, if it's needed.

src/jest.ts
gradybarrett: Ha, ya, if nothing else, it's always good to be consistent. :)

src/jest.ts
```diff
@@ -44,6 +61,25 @@ export interface CoverageThreshold {
   readonly statements?: number;
 }
 
+type JestReporter = [string, {[key: string]: any}] | string;
```
gradybarrett, line 64: @eladb is this frowned upon?

I know JSII doesn't support type aliases, so I didn't export either of these. It just feels bad to have to repeat `[string, {[key: string]: any}] | string` in 3 places.

src/jest.ts
```diff
@@ -44,6 +61,31 @@ export interface CoverageThreshold {
...
+ * Array of Jest-compliant reporters.
+ *
+ * https://jestjs.io/docs/en/configuration#reporters-arraymodulename--modulename-options
+ *
+ * This also works with Jest `coverageReporters`, although preserving
+ * defaults is not supported for that config.
+ *
+ * https://jestjs.io/docs/en/configuration#coveragereporters-arraystring--string-options
+ */
+class JestReporters extends Array<JestReporter> {
```
Chriscbr, line 76: I see what you are going for here, but I'm not sure if extending an Array type is ideal from an abstraction/style standpoint. I feel like having the array as an instance property would have the upside of allowing other properties to be added in the future if needed (and allow the class to be exported by JSII potentially). But this is just me inferring from how I see code written in the aws-cdk (another project overseen by eladb), so I'm open to other options

src/jest.ts
Chriscbr: Another option here is to create a basic class called `JestReporter` with an optional `options` field for storing the reporter config. Then by adding a method named `synthesize()` or similar which returns the object in the correct jest format (an array of [string, object]), then you can put `reporters: () => this.reporters.map(x => x.synthesize())` in the config below and it should synthesize the results properly, I think.

Chriscbr (commented):
Just a few comments/suggestions

src/jest.ts
gradybarrett: I was trying to keep that extra code to a minimum by relying on types, but in the long term going that route probably makes objectifying a given config easier.

.projenrc.js
```diff
@@ -33,6 +33,10 @@ const project = new JsiiProject({
   releaseToNpm: true,
   minNodeVersion: '10.17.0',
   compileBeforeTest: true, // since we want to run the cli in tests
+
+  jestOptions: {
+    junitReporting: false,
```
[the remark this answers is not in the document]
Chriscbr: I think I suggested that earlier because I thought leaving it as default (`true`) would add extra test artifacts to projen. But I just noticed now that they would get gitignored anyway, so yeah in that case it doesn't matter (keeping it false might just shorten the build time slightly 🤷 ).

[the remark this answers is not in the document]
gradybarrett: Eh, I think the operation is pretty quick, so I'll remove this.

src/jest.ts
gradybarrett: Cool. I'll leave it the way it is then. It'll grow into whatever it needs to be long term.

src/jest.ts
gradybarrett: It does feel that way. I'm going to remove this and move it back down into the Jest class. It'll naturally get abstracted when it needs to be.

eladb (approved):
🎉

PR #281, opened 2020-11-06
docs(readme): clone url update

fongie:
Hi! Here's a *very* small PR to update the "git clone" url in the readme since projen now seems to live at `projen/projen` instead of `eladb/projen`!


Chriscbr (approved):

Chriscbr:
Thanks!

PR #282, opened 2020-11-07
feat(license): add flag to skip adding license

### LINKED ISSUE #122: Feature request: Support custom licenses

jmourelos:
* Small fix to util test function directorySnapshot: return was preventing
the function to transverse the whole tree.
* Added filtering to util test function synthSnapshot to limit size of the
snapshot

Fixes #122


test/util.ts
```diff
@@ -14,17 +14,17 @@ export function mkdtemp() {
   return fs.mkdtempSync(path.join(os.tmpdir(), 'projen-test-'));
 }
 
-export function synthSnapshot(project: Project) {
+export function synthSnapshot(project: Project, onlyIncludeFiles?: string[]) {
```
eladb, line 17:
```suggestion
export function synthSnapshot(project: Project, ...includeFiles?: string[]) {
```

jmourelos: Done.

eladb (approved):

PR #288, opened 2020-11-08
feat(eslint): add file extension and ignore pattern options

Chriscbr:
Add `fileExtensions` and `ignorePatterns` fields to eslintOptions, and add `eslintOptions` has a field to TypeScript projects.


src/typescript.ts
```diff
@@ -23,12 +23,17 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
...
   /**
-   *
    * Setup eslint.
    * @default true
    */
   readonly eslint?: boolean;
 
+  /**
+   * Eslint options
+   * @default - opinionated default options
```
pgollucci, line 33: Why not say the actual default ['.ts', '.tsx'],

Chriscbr: This includes all of the eslint options, not just the file extensions, so I don't think it would be practical to list all of nested defaults.

eladb (approved):

PR #289, opened 2020-11-08
feat: introduce CompositeProject

misterjoshua:
Introduces a `CompositeProject` type that allows multiple projects to be added with a relative directory and synthesized into a single tree.

```ts
const comp = new CompositeProject();
comp.addProject('packages/foo', new NodeProject(...));
comp.addProject('packages/bar', new NextjsProject(...));

comp.synth();
```

See also #284


misterjoshua:
@eladb @hoegertn Fresh PR for `CompositeProject` here.

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export class SubProjectComponent extends Component {
+  constructor(project: Project, private readonly options: SubProjectComponentOptions) {
+    super(project);
+  }
+
+  synthesize(outdir: string) {
+    this.options.subProject.synth(path.join(outdir, this.options.subPath));
+  }
+}
```
hoegertn, line 86: `postSynthesize` is missing, isn't it?

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Options for `CompositeProject`.
+ */
+export interface CompositeProjectOptions {
+  /**
+   * Declaratively define sub-projects by their sub paths.
+   * @example ```ts
+   * {
+   *   'packages/foo': new NodeProject({ name: 'foo' }),
```
hoegertn, line 13: How does this handle os-specific path delimiters?

src/composite-project.ts
misterjoshua: It is, but at least in the default implementation of `Project`, `synth()` calls its own `postSynthesize()`. I figured if I called it here, that a project's `postSynthesize()` would be called twice.

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Options for `SubProjectComponent`
+ * @internal
+ */
+export interface SubProjectComponentOptions {
+  /**
+   * Subpath of the sub-project.
+   */
+  readonly subPath: string;
```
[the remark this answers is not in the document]
misterjoshua: Makes sense.

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Adds a project as a sub-project at a sub path.
+   * @param subPath
+   * @param project
+   */
+  addProject(subPath: string, project: Project) {
+    if (this.projects[subPath]) {
+      throw new Error(`Cannot add project as the sub path ${subPath} is already in use`);
+    }
+
+    this.projects[subPath] = true;
```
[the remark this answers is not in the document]
misterjoshua: Yup, we should be able to.

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface SubProjectComponentOptions {
+  /**
+   * Subpath of the sub-project.
+   */
+  readonly subPath: string;
+
+  /**
+   * The sub-project.
+   */
+  readonly subProject: Project;
```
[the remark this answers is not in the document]
misterjoshua: Sounds good.

src/composite-project.ts
misterjoshua: Hmm. I think `CompositeProject.addProject()` has the same problem, in that it depends on the user to employ `path.join('a', 'b')` to handle os-specific path delims. The notation for typescript/javascript is definitely too exotic:

```ts
{
  [path.join('packages, 'foo')]: new NodeProject({ name: 'foo' }),
  [path.join('packages', 'bar')]: new NextjsProject({ name: 'bar' }),
}
```

src/composite-project.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Component representation of a sub-project.
+ * @internal
+ */
+export class SubProjectComponent extends Component {
+  constructor(project: Project, private readonly options: SubProjectComponentOptions) {
+    super(project);
+  }
```
misterjoshua, line 97: > Maybe we could also offer access to the parent root then for .github stuff etc

@hoegertn I was thinking that there's a design decision to be made about how a project can know the parent root so that it knows the right place to put the .github stuff. Given that a sub-project is instantiated before it knows that it's a sub-project, it seems to me that there's a need to inform the sub-project of that. I could call back to the project to inform it here, for instance.

src/composite-project.ts
misterjoshua: I've pushed up an alternative.

misterjoshua:
I'm starting to wonder if I like `ProjectComponent` more than everything else I've written here. 🤣 Well anyway, I'll leave this now for your guys' review.

src/composite-project.ts
```diff
@@ -0,0 +1,103 @@
...
+   * ]
+   * ```
+   */
+  projects?: CompositeProjectDeclarativeSubProject[];
+}
+
+/**
+ * Declares a sub-project of the composite project.
+ */
+export interface CompositeProjectDeclarativeSubProject {
```
eladb, line 27: Name is too long... Maybe `CompositeProjectChild`.

src/composite-project.ts
```diff
@@ -0,0 +1,103 @@
...
+      project: project,
+    });
+  }
+}
+
+/**
+ * Options for `ProjectComponent`
+ * @experimental
+ */
+export interface ProjectComponentOptions {
```
eladb, line 78: Can we leave this private for now? Feels like it could be trouble and I am not sure I fully understand the use case here.

```suggestion
interface ProjectComponentOptions {
```

eladb (changes requested):
Looks promising!

src/composite-project.ts
```diff
@@ -0,0 +1,103 @@
...
+   * ]
+   * ```
+   */
+  projects?: CompositeProjectChild[];
+}
+
+/**
+ * Declares a sub-project of the composite project.
+ */
+export interface CompositeProjectChild {
```
eladb, line 27: Now that it's private, can we just use `CompositeProjectOptions` and remove this type?

src/composite-project.ts
```diff
@@ -0,0 +1,103 @@
...
+   * [
+   *   {
+   *     path: path.join('packages', 'foo'),
+   *     project: new NodeProject({ name: 'foo' }),
+   *   },
+   *   ...
+   * ]
+   * ```
+   */
+  projects?: CompositeProjectChild[];
```
eladb, line 21:
```suggestion
  readonly projects?: CompositeProjectChild[];
```

eladb (changes requested):
Please add `composite-project.ts` to `index.ts`

src/composite-project.ts
misterjoshua: I'm sorry but I don't understand. I see that `CompositeProjectChild` and the private `ProjectComponentOptions` are the same. Is your suggestion here to eliminate the private one and re-use the remaining public type like in db183f2 that I've just pushed up?

test/composite-project.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  expect(fs.existsSync(path.join(tempDir, 'packages', 'foo', '.gitignore'))).toBeTruthy();
+});
+
+test('composing projects synthesizes to subdirs', () => {
+  // GIVEN
+  const comp = new CompositeProject();
+
+  // WHEN
+  comp.addProject('packages/foo', new NodeProject({ name: 'foo' }));
+  comp.addProject('packages/bar', new NodeProject({ name: 'bar' }));
```
Chriscbr, line 42: Would it make more sense to use path.join here (and in the last test) as well just for consistency?

misterjoshua: I've applied your suggestion. :)

src/composite-project.ts
eladb: yes :-D

misterjoshua: Sweet!

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+   * @example
+   * [
+   *   {
+   *     path: path.join('packages', 'foo'),
+   *     project: new NodeProject({ name: 'foo' }),
+   *   },
+   *   ...
+   * ]
+   */
+  readonly projects?: CompositeProjectChild[];
```
eladb, line 20:
```suggestion
  /**
   * Declaratively define sub-projects by their sub paths.
   * @default []
   * @example
   * [
   *   {
   *     path: path.join('packages', 'foo'),
   *     project: new NodeProject({ name: 'foo' }),
   *   },
   *   ...
   * ]
   */
  readonly projects?: CompositeProjectChild[];
```

src/composite-project.ts
```diff
@@ -0,0 +1,86 @@
...
+      this.addProject(subProject.path, subProject.project);
+    }
+  }
+
+  /**
+   * Adds a project as a sub-project at a sub path.
+   * @param subPath
+   * @param project
+   */
+  addProject(subPath: string, project: Project) {
```
eladb, line 61:
```suggestion
  addProject(subdir: string, project: Project) {
```

eladb (commented):
A couple of minor tweaks and we are ready

misterjoshua:
> A couple of minor tweaks and we are ready

@eladb Cool. The changes you requested are in now.

eladb (approved):

misterjoshua:
@eladb It looks like the build broke in master because after a version bump, NodeProject is looking for the not-yet-released projen version in npm during the tests. I could push up a new pr and change the NodeProject references to Project so this isn't so much a factor.

resistdesign:
Am I allowed to cheer for a PR? 😛  🎉

PR #291, opened 2020-11-08
fix(awscdk-construct): constructs 3.2.0 is required for latest CDK

### LINKED ISSUE #290: CDK depends on constructs@^3.2.0 but projen assumes constructs@^3.0.4
This breaks Nuget releasing aka donet packaging

https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1605

https://github.com/p6m7g8/p6-namer/pull/54/checks?check_run_id=1371474864#step:9:27

pgollucci:
Fixes #290 by updating `projen` constructs from 3.0.4 to 3.2.0


Chriscbr:
(Wasn't sure whether to leave this on the issue or here) Do we have any idea of what cause this breakage? Is this a package that will have to routinely be updated?

eladb (approved):

PR #294, opened 2020-11-09
fix/280-Render params changes for the comments broke the new --from method

### LINKED ISSUE #280: Try to write my own Construct
Hi Guys,

im trying to write my own construct and would be happy to update the docs about the process. Currently i just tried to create a jsii project and copied a bit of config from projen itself. Now i created a class in the typescript file that should extend an already active projen config. 

```js
import { AwsCdkTypeScriptApp } from 'projen';

/**
 * (experimental) AWS CDK app in TypeScript.
 *
 * @experimental
 * @pjid henry-cdk
 */
export class HenryCDK extends AwsCdkTypeScriptApp {

  constructor() {
    super({
      cdkVersion: '1.70.0',
      name: 'henry-cdk',
    });
  }
}
```

But when I try to compile the Project i just get:

`Unable to resolve type "projen.AwsCdkTypeScriptApp". It may be @internal or not exported from the module's entry point (as configured in "package.json" as "main").`

Isn't this the intended way of creating a custom construct? If not i'm also fine with understanding for of the concept. Also happy to deliver more configs from my project.

kcwinner:
Fixes #280 

* Makes a fix to the cli `new` renderParams. It was not properly outputting an array param

### DIFF
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -133,11 +133,7 @@ function renderParams(type: inventory.ProjectType, params: Record<string, any>, 
     const optionName = option.name;
     let paramRender;
     if (params[optionName] !== undefined) {
-      if (Array.isArray(params[optionName])) {
-        paramRender = `${optionName}: ${JSON.stringify(params[optionName])},`;
-      } else {
-        paramRender = `${optionName}: ${params[optionName]},`;
-      }
+      paramRender = `${optionName}: ${params[optionName]},`;
       optionsWithDefaults.push(optionName);
     } else {
       const defaultValue = option.default?.startsWith('-') ? undefined : (option.default ?? undefined);
@@ -276,7 +272,7 @@ function newProjectFromModule(baseDir: string, spec: string, args: any) {
 
   // include a dev dependency for the external module
   newProject(baseDir, type, args, {
-    devDeps: [specDependencyInfo],
+    devDeps: JSON.stringify([specDependencyInfo]),
   });
 }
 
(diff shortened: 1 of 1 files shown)


src/cli/cmds/new.ts
```diff
@@ -133,7 +133,11 @@ function renderParams(type: inventory.ProjectType, params: Record<string, any>,
     const optionName = option.name;
     let paramRender;
     if (params[optionName] !== undefined) {
-      paramRender = `${optionName}: ${params[optionName]},`;
+      if (Array.isArray(params[optionName])) {
+        paramRender = `${optionName}: ${JSON.stringify(params[optionName])},`;
```
kcwinner, line 137: An array of strings is gets output using `[].join(',') by default and we don't want that here
This is not perfect. I'm open to feedback. Just trying to get something out here for discussion.
This will break if us pass in 
```
[ '"someQuotedValue'" ]
```

Chriscbr: Gotcha, I see how if an would not be rendered correctly if one of the values in `params` was an array. However I'm wondering if really it should be specified earlier in the argument-processing that these args should already come passed in as strings. For example, if I wanted to specify dependencies through the CLI, I think the way this has to be done is:
```
npx projen new typescript --deps '["react", "react-dom"]'
```
In which case if I inspect `params` it contains:
```
{
  name: '"projen-test"',
  deps: '["react"]',
  allowLibraryDependencies: undefined,
...
```

So maybe a better fix could be to change the how the project gets constructed:
```
  // include a dev dependency for the external module
  newProject(baseDir, type, args, {
    devDeps: JSON.stringify([specDependencyInfo]),
  });
```

I'm not as familiar with --from, so can you try if that fix works? Otherwise I'm okay with your change as is

pgollucci: I like @Chriscbr cli approach.  Its consistent with the AWS CLI

kcwinner: Agreed. Way better. Making that change and will push

kcwinner: Updated

Chriscbr (approved):

PR #297, opened 2020-11-10
fix(cli): "new --from" scoped packages is broken

### LINKED ISSUE #296: Bugfix: Projen fails to discover jsii when running `projen new --from @.../...` for a scoped package
The `inventory.discover` method does not support scoped package discovery. 
Discovery expects the package to be something like `node_modules/my-project/.jsii`, however, scoped packages are `node_modules/@something/my-project/.jsii`

Taken from `inventory.ts`
```javascript
// read all .jsii manifests from all modules (incl. projen itself) and merge
  // them all into a single map of fqn->type.
  for (const dir of [...moduleDirs, PROJEN_MODULE_ROOT]) {
    const jsiiFile = path.join(dir, '.jsii');
    if (!fs.existsSync(jsiiFile)) { continue; } // no jsii manifest
    const manifest = fs.readJsonSync(jsiiFile);
    for (const [fqn, type] of Object.entries(manifest.types as JsiiTypes)) {
      jsii[fqn] = type;
    }
  }
```

kcwinner:
Fixes #296 

* Alters inventory discovery to check for child directories for scoped packages
* #280 - sets the new project `additionalProps` to be `Record<string, string>` in order to be in line with the 280 bugfix change from yesterday

### DIFF
--- a/src/inventory.ts
+++ b/src/inventory.ts
@@ -93,9 +93,10 @@ export function discover(...moduleDirs: string[]) {
     discoverJsii(dir);
 
     if (dir.includes('@') && fs.lstatSync(dir).isDirectory()) {
-      fs.readdirSync(dir).map(file => path.join(dir, file)).forEach(child => {
+      const childDirs = fs.readdirSync(dir).map(file => path.join(dir, file));
+      for (const child of childDirs) {
         discoverJsii(child);
-      });
+      }
     }
   }
 
--- a/src/inventory.ts
+++ b/src/inventory.ts
@@ -77,15 +77,25 @@ interface JsiiType {
 export function discover(...moduleDirs: string[]) {
   const jsii: JsiiTypes = {};
 
+  const discoverJsii = (dir: string) => {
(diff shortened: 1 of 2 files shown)


src/inventory.ts
```diff
@@ -77,16 +77,26 @@ interface JsiiType {
...
     }
+  };
+
+  // read all .jsii manifests from all modules (incl. projen itself) and merge
+  // them all into a single map of fqn->type.
+  for (const dir of [...moduleDirs, PROJEN_MODULE_ROOT]) {
+    discoverJsii(dir);
+
+    if (dir.includes('@') && fs.lstatSync(dir).isDirectory()) {
+      fs.readdirSync(dir).map(file => path.join(dir, file)).forEach(child => {
```
eladb, line 96: For readability, I rather use `for-of` instead of `.forEach()`. It's also less error prone when you try to `break` from a loop.

eladb (changes requested):

eladb (approved):

PR #300, opened 2020-11-11
fix: CompositeProject tests reference unreleased projen version after bump

misterjoshua:
This should fix the build in master when bumping versions after #289. `NodeProject` in the tests wanted to pull the not-yet-released projen packages after the bump. This fix addresses the issue by changing the tests not to depend on any project type that needs to know Projen's version.


misterjoshua:
@eladb I noticed the build failed in master from bumping versions after merging #289. I believe that this should fix it.

eladb (approved):

PR #301, opened 2020-11-11
feat: support aws-cdk constructs

eladb:
This adds support for using AWS CDK constructs within projen definitions:

    const aws = new AwsEnvironment(project, { env: 'aws://ACCOUNT/REGION' });
    new s3.Bucket(aws.stack, 'MyBucket');

During projen synth, the CDK output (the "cloud assembly") will be emitted into `.awscdk/account/region`
and a GitHub workflow will be added that will deploy these artifacts to the designated AWS environment.

TODO:

- [ ] Tests for GitHub workflow
- [ ] Support specifying CDK dependencies
- [ ] Documentation

Fixes #


pgollucci:
holy crap! Squeal. Horray! YAY Dance!

eladb:
> holy crap! Squeal. Horray! YAY Dance!

Feedback is welcome!

Chriscbr (commented):
Hmm, yeah this is a cute idea, using projen and cdk code together in the same file. A marriage I didn't see coming but now I can't imagine any other way. :-) A have a few questions (code comments are mostly nitpicks, no need to answer all) so otherwise the PR looks good to me!

----

Thoughts on how testing fits into this? I know the cdk team recommends people to write unit tests / snapshot tests for cdk (although some people might see it as an afterthought...). That said I think it would be weird for projen to have its own test suite. I'm guessing the idea would just be if you really want to test your constructs rigorously, you should do so in a separate package, and then just import that construct in .projenrc.js ?

Is there a future where you could have a CompositeProject with a AwsCdkConstructLib, ConstructLibCdk8s, or AwsCdkTypeScriptApp, and then reference constructs defined in those projects within the root .projenrc.js? I haven't used these projects as much, so this idea might not make sense in practice...

One more idea of something people could want in the future is something where the project (or a subproject) contains the code for an AWS lambda function, and then in .projenrc.js it will handle deploying that code and using CDK you can connect it to API Gateway etc. Maybe this is just duplicating behavior that is easier to achieve with AwsCdkTypeScriptApp, but figured I should mention it anyway. I also know this doesn't align with how some developers use CDK (defining separate repositories for lambda code and for CDK code).

Chriscbr:
I also had the thought that maybe there's a world where it makes sense for instantiating `AwsEnvironment` to automatically add some commands to your project -- a temporary way this could be added is by checking `if (project instanceof NodeProject)`. But I'm not sure in general if adding npm-based commands makes sense in the scope of non-node-based projects, so the idea needs a bit more grooming.

pgollucci:
> I also had the thought that maybe there's a world where it makes sense for instantiating `AwsEnvironment` to automatically add some commands to your project -- a temporary way this could be added is by checking `if (project instanceof NodeProject)`. But I'm not sure in general if adding npm-based commands makes sense in the scope of non-node-based projects, so the idea needs a bit more grooming.

This is the thing I'm trying to get away from is making Node the base it blocks the JSII of projen, and having projen generate projects like PythonProject.

pgollucci:
What am I missing here ?

```shell
gh repo clone pgollucci/projen
cd projen
gh pr checkout 301
yarn install
yarn build
yarn run v1.22.10
$ yarn run compile && yarn run test && yarn run package && mv README.md README.md.bak && cat README.md.bak | npx markmac > README.md && rm README.md.bak
$ jsii --silence-warnings=reserved-word --no-fix-peer-dependencies && jsii-docgen
[2020-11-28T12:45:49.114] [ERROR] jsii/compiler - Type model errors prevented the JSII assembly from being created
error JSII3000: Exported APIs cannot use un-exported type "projen.AwsCredentialsSecrets"
error JSII3000: Exported APIs cannot use un-exported type "projen.GithubAspect"
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

eladb:
@projen rebuild

eladb:
Closing for now

PR #302, opened 2020-11-12
fix(typescript): typescript uses semi colons

pgollucci:
- feat: README samples (#203)
- Revert "feat: README samples (#203)"
- fix(typescript): typescript uses semi colons


Chriscbr (approved):

eladb:
For bug fixes the title of the PR should describe the problem instead of the solution since it appears in the change log under "bug fixes".

PR #306, opened 2020-11-13
chore: increase default cdk version and replace deprecated methods

henrysachs:
I increased the cdk version to match the yesterday released one and also replaced the deprecated function call from `addPeerDependencies`. Just some cleanups that help me in my custom Projen module.


henrysachs:
if its fine for all i maybe take a look at other deprecated functions and try to convert them.

henrysachs:
i now went through all files and replaced deprecated methods without breaking compatibility. That's why i added string to the Type aswell as Semver.

henrysachs:
The License File Test seems to be a bit flaky isn't it? it seems to sometimes fail and sometimes pass. 🤔

src/cdk8s-construct.ts
```diff
@@ -25,10 +24,10 @@ export class ConstructLibraryCdk8s extends ConstructLibrary {
 
     const ver = options.cdk8sVersion;
 
-    this.addPeerDependencies({
-      'constructs': Semver.caret('2.0.2'),
-      'cdk8s': Semver.caret(ver),
-      'cdk8s-plus': Semver.caret(ver),
-    });
+    this.addPeerDeps(
+      'constructs@^2.0.2',
```
eladb, line 28: Let's also bump cdk8s to the latest version and I believe it now uses the same constructs major version as the CDK.

src/jsii-docgen.ts
```diff
@@ -10,7 +9,7 @@ import { StartEntryCategory } from './start';
  */
 export class JsiiDocgen {
   constructor(project: JsiiProject) {
-    project.addDevDependencies({ 'jsii-docgen': Semver.caret('1.3.2') });
+    project.addDevDeps('jsii-docgen@^1.3.2');
```
eladb, line 12: I don't think a version is needed here. Just omit the version

src/jsii-project.ts
```diff
@@ -249,15 +252,15 @@ export class JsiiProject extends TypeScriptProject {
...
 
-    this.addDevDependencies({
-      'jsii': jsiiVersion,
-      'jsii-diff': jsiiVersion,
-      'jsii-pacmak': jsiiVersion,
-      'jsii-release': Semver.caret('0.1.6'),
-      '@types/node': Semver.caret(minNodeVersion),
-    });
+    this.addDevDeps(
+      `jsii@^${jsiiVersion}`,
```
eladb, line 258: I think we can completely omit jsii version.

src/typescript-typedoc.ts
```diff
@@ -7,7 +6,7 @@ import { TypeScriptProject } from './typescript';
  */
 export class TypedocDocgen {
   constructor(project: TypeScriptProject) {
-    project.addDevDependencies({ typedoc: Semver.caret('0.17.8') });
+    project.addDevDeps('typedoc@^0.17.8');
```
eladb, line 9: Omit version

src/typescript.ts
```diff
@@ -26,7 +26,7 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
    * TypeScript version to use.
    * @default "^3.9.5"
    */
-  readonly typescriptVersion?: Semver;
+  readonly typescriptVersion?: string | Semver;
```
eladb, line 29: Unions are Not recommended in jsii

eladb (changes requested):
Generally with the  new "deps" options you don't really need a version specifier

henrysachs:
will now pick up the requested changed

src/typescript.ts
henrysachs: i now moved everything to string but wouldn't that create a breaking change for everyone using this property?

src/jsii-project.ts
henrysachs: I'll omitted the version but kept the parameter, if a user want to set it to their own fixed version.

henrysachs:
i now removed jsii version as a parameter completly and therefore omitted the version

src/jsii-project.ts
```diff
@@ -249,15 +245,13 @@ export class JsiiProject extends TypeScriptProject {
...
-      'jsii-pacmak': jsiiVersion,
-      'jsii-release': Semver.caret('0.1.6'),
-      '@types/node': Semver.caret(minNodeVersion),
-    });
+    this.addDevDeps(
+      'jsii',
+      'jsii-diff',
+      'jsii-pacmak',
+      'jsii-release',
+      `@types/node@^${minNodeVersion}`,
```
eladb, line 253: I suspect this is already covered by `TypeScriptProject` and therefore not required here anymore

src/node-project.ts
```diff
@@ -787,8 +787,8 @@ export class NodeProject extends Project {
 
     const projen = options.projenDevDependency ?? true;
     if (projen) {
-      const projenVersion = options.projenVersion ?? Semver.caret(PROJEN_VERSION);
-      this.addDevDependencies({ projen: projenVersion });
+      const projenVersion = options.projenVersion ?? `^${PROJEN_VERSION}`;
+      this.addDevDeps( `projen@${projenVersion}` );
```
eladb, line 791:
```suggestion
      this.addDevDeps(`projen@${projenVersion}`);
```

henrysachs:
removed the min node version from jsii

eladb (approved):

PR #312, opened 2020-11-16
fix: cdk app starter test code syntax and linting errors

misterjoshua:
This PR fixes syntax and linting errors in the starter code for `AwsCdkTypeScriptApp`.


eladb (approved):
thanks!

Issue #315, opened 2020-11-18
VS Code Recommended Extensions


henrysachs:
Would be nice to manage recommended Extensions with Projen: https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions 
just a json file with an array of strings.

eladb:
Yes!

pgollucci:
If https://github.com/projen/projen/pull/316 fixes this, then this can be closed. @eladb 

I'm glad I found #316.

PR #316, opened 2020-11-18
feat: Add VSCodeLaunchConfiguration

### LINKED ISSUE #314: CDK projects: configure VSCode debugger
Would be awesome if CDK projects will be configured by default with VSCode debugger.
Follow @tlakomy's blog: https://tlakomy.com/use-vscode-debugger-for-cdk

tlakomy:
Fixes #314


hoegertn:
Two proposals from my side. 1) Hide vscode settings behind a separate construct 2) add it to construct library too

henrysachs:
But wouldn't a special file type fit better so you could write a debugger config for each type of project?

henrysachs:
Or is construct === filetype?

hoegertn:
That is the same I think. I mean a class representing debugger configs

eladb:
Yes, this will likely be a bit more involved, but it will make the world a better place!

Let's create a `VsCode` class (extends `Component`) which has a method `addLaunchConfiguration()` which accepts a `LaunchConfiguration` struct that models the schema.

Then, we should probably add a `vscode` component to `NodeProject` so one can do `nodeProject.vscode.addLaunchConfiguration(...)`.

Then, we can create another component that represents `VsCodeCdkDebugger`. It should accept a `NodeProject` and then adds this launch configuration to it.

Then, we need to 

It should have a method like this:

```ts
launch.addConfiguration(...);
```

henrysachs:
Just a question here. Why would we extend component and not jsonfile?

hoegertn:
The VsCode class would HAVE a JsonFile and not BE one for two reasons.

1) Composition over inheritance
2) it might need to have other components later on like a second json file, or the proposed vscode extension file #315

henrysachs:
sounds great! and a Component can decide itself what types of files it generates?

henrysachs:
maybe i just wait for this PR to be merged and add the extension file afterwards :)

eladb:
https://github.com/projen/projen/issues/318

eladb:
Some ideas on how to codify these implicit assumptions: https://github.com/projen/projen/issues/318

eladb:
Please rebase (#319) and add this to the `VsCode` class (see how to `GitHub` stuff is organized as a reference).

tlakomy:
@eladb apologies for not getting back to you earlier, please take a look at my recent changes

src/vscode/launchConfiguration.ts
```diff
@@ -0,0 +1,33 @@
+import { FileBase } from '../file';
```
eladb, line 1: nit: rename to `launch-config.ts`

src/vscode/launchConfiguration.ts
```diff
@@ -0,0 +1,33 @@
...
+import { VsCode } from './vscode';
+
+/**
+ * Template for a VSCode Launch Configuration (useful for enabling in-editor debugger)
+ */
+export class VSCodeLaunchConfiguration extends FileBase {
```
eladb, line 7: I think the component should model the launch file and have an `addConfiguration()` method like so:

```ts
export class VsCodeLaunch extends Component { // the fact that it is a JsonFile is an impl. detail so don't extend

  private readonly content: any;

  constructor(vscode: VsCode) {
    super(vscode.project);

    this.content = {
      version: '0.2.0',
      configurations: []
    };
    
    new JsonFile(vscode.project, '.vscode/launch.json', {
      obj: this.content,
    });
  }

  public addConfiguration(cfg: VsCodeLaunchConfiguration) {
    this.content.configurations.push(cfg);
  }
}

export interface VsCodeLaunchConfiguration {
  readonly type: string;
  readonly request: ...
  // ... model it all
}
```

eladb (changes requested):

src/vscode/launchConfiguration.ts
tlakomy: I see, that makes sense 👍

src/vscode/launchConfiguration.ts
tlakomy: Done 👍

src/vscode/launchConfiguration.ts
tlakomy: Just pushed my recent changes, apparently there are quite a few optional fields in `launch.json` according to https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes

src/vscode/launch-config.ts
```diff
@@ -0,0 +1,67 @@
+import { Component } from '../component';
+import { JsonFile } from '../json';
+import { VsCode } from './vscode';
+
+interface VsCodeLaunchConfiguration {
```
eladb, line 5: Since this is a private type, move to the back of the file

src/vscode/launch-config.ts
```diff
@@ -0,0 +1,67 @@
...
+  version: string;
+  configurations: VsCodeLaunchConfigurationEntry[];
+}
+
+export interface Presentation {
```
eladb, line 10: Please add docstrings to all public APIs.

src/vscode/launch-config.ts
```diff
@@ -0,0 +1,67 @@
...
+      version: '0.2.0',
+      configurations: [],
+    };
+
+    new JsonFile(vscode.project, '.vscode/launch.json', {
+      obj: this.content,
+    });
+  }
+
+  public addConfiguration(cfg: VsCodeLaunchConfigurationEntry) {
```
eladb, line 64: docstrings

src/vscode/vscode.ts
```diff
@@ -1,8 +1,13 @@
...
+import { VsCodeLaunchConfig } from './launch-config';
 
 export class VsCode extends Component {
   constructor(project: Project) {
     super(project);
   }
-}
\ No newline at end of file
+
+  public addLaunchConfiguration() {
```
eladb, line 10: Since there is only a single launch config, I'd go with a lazy property instead of `addXxx()`:

```ts
public get launchConfiguration() {
  if (!this._launchConfig) {
    this._launchConfig = new VsCodeLaunchConfig(this);
  }

  return this._launchConfig;
}
```

eladb (changes requested):
Looks good. Minor notes.

Please re-request review when ready.

src/vscode/launch-config.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Options for a 'VsCodeLaunchConfigurationEntry'
+ * Source: https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes
+ */
+export interface VsCodeLaunchConfigurationEntry {
+  readonly type: string;
+  readonly request: string;
+  readonly name: string;
+  readonly args?: string[];
+  readonly debugServer?: number;
+  readonly internalConsoleOptions?:
```
eladb, line 38: Use an enum

eladb (approved):

PR #319, opened 2020-11-19
feat: consolidate github capabilities behind a centralized api

### LINKED ISSUE #318: Environment Assumptions
It will be useful to define what are the environment assumptions for `projen`. At the moment, some project types make implicit assumptions about the environment. For example, most project types assume the project is hosted on GitHub (they synthesize github workflows, etc). Similarly, we want to add some IDE support, and VSCode is becoming another implicit assumption.

It will definitely make it easier to just set those global assumptions like GitHub + VSCode, but it also limits users when they _want_ choice. Which is probably a good definition for what a healthy "opinionated behavior " should be -- stuff that people don't wan't to have a choice in. And that's obviously subjective measure. We hope that those opinions are good enough for enough users, and if it's not, we can change some of those decisions and release a major version.

It's also quite unhealthy to generalize at the first case. And at the moment it appears that the majority of the projen user base does use GitHub, and likely VSCode.

So for now, it feels like it's the pragmatic choice to take those global assumptions, but still interact with these services in a formal way, so we can evolve it later.

To that end, I think all the GitHub-specific capabilities should go under a `GitHub` component which will represent the various capabilities other components need when they use GitHub. For example, `github.addWorkflow()` or `github.addPullRequestTemplate()`.

Similarly, `VsCode` should be we implement capabilities related to `VsCode`, which other components can interact with.

And we will add a `.vscode` and `.github` _optional_ properties to `Project`:

```ts
class Project {
  public readonly github?: Github;
  public readonly vscode?: VsCode;
}
```

And those will initially be initialized at the `Project` level.

Consumers (components/projects) will interact with this API:

```ts
project.github?.addWorkflow(...);
project.vscode?.addExtension(...);
```

> Notice the `?.` which indicates that consumers acknowledge the fact that this can potentially be undefined.

eladb:
Move all GitHub and VsCode related components under a single API that can be accessed from the project level.

NOTE: API.md does not have support for submodules (yet) https://github.com/aws/jsii-docgen/issues/1 but all the GitHub component types are exported under the `github` submodule: `import github from 'projen'`.

Closes #318 

BREAKING CHANGE: there are multiple breaking changes in in how github components are used.


src/github/github.ts
```diff
@@ -0,0 +1,37 @@
+import { Component } from '../component';
+import { Project } from '../project';
+import { Dependabot, DependabotOptions } from './dependabot';
+import { Mergify, MergifyRule } from './mergify';
+import { PullRequestTemplate } from './pr-template';
+import { GithubWorkflow } from './workflows';
+
+export class GitHub extends Component {
+  private _mergify?: Mergify;
```
Chriscbr, line 9: nit: Is there a reason this private field begins with _ while others don't?

Chriscbr (commented):
Have you tried this with `CompositeProject`? Just curious if `GitHub` or `VsCode` will get generated by default for each subproject, or if it only gets generated in the root project. I think this PR can be merged as is, but it might be worth opening an issue later to track how that case should be handled.

LGTM 🙌

eladb:
Interesting point about `CompositeProject`. I am trying to think about how these should actually interact. In a sense, it's not going to be "transparent" for a sub-project to just add a github workflow to the parent project because it has to take into account it's local path and things like that. So perhaps the correct "right" behavior (for now at least) should be that these become `undefined` for sub projects. But I am not sure there's a clean way to do it like that, so I'll have to check.

eladb:
@Chriscbr see https://github.com/projen/projen/pull/332

src/github/github.ts
eladb: fixed

PR #324, opened 2020-11-20
fix(composite-project): prevent child NodeProjects failure during synth

### LINKED ISSUE #323: CompositeProject does not support NodeProject children
After generating a new "composite" project and adding a `NodeProject` to the projects list it fails to generate with error `ENOENT: no such file or directory, open '/Users/xxxxx/composite-test/package.json`

## Repo Steps:
1. `mkdir composite-test`
2. `cd composite-test`
3. `npx projen new composite` ... generates project
4. Edit `.projenrc.js` with following
```
const { CompositeProject, NodeProject } = require('projen');

const project = new CompositeProject({
  projects: [
		{
			path: 'packages/node-test',
			project: new NodeProject({
				name: 'node-test',
			})
		},
	]
});

project.synth();
```
5. Run `projen` => `ENOENT: no such file or directory, open '/Users/xxxxx/composite-test/package.json`

The `packages/node-test` package exists and has a `packages.json` as expected, but as expected there is no `package.json` in the root of composite project.

JeremyJonas:
The `printStartMenu` was based on cwd, which for composite projects doesn't work and should be based on `outdir` of child projects.

This fixes the issue by passing the `outdir` as `root` to `printStartMenu` during project.synth()

Fixes #323


Chriscbr (changes requested):
Thanks for this 🐛 fix!

eladb (approved):

PR #326, opened 2020-11-20
feat(nextjs): add tailwindcss option

### LINKED ISSUE #325: Add Tailwind CSS support
[Tailwind CSS](https://github.com/tailwindlabs/tailwindcss) is a pretty popular CSS utility framework, which is used in many JS projects. It makes sense to add an option to easily add support for it in the relevant project types.

TailwindCSS docs [here](https://tailwindcss.com/docs/installation) mention that there are two ways to install it - with PostCSS (recommended) and without. It probably makes most sense to use PostCSS as the opinionated default.

I think it makes sense to add Tailwind CSS as an option for `NextJs` and `React` project types, since these project types include sample code for serving web pages. But it could be added as an option for `Node` projects just as well.

It's worth noting that while NextJs supports PostCSS out of the box (so Tailwind will work with the default build script) - but this is not the case for React or node projects. To add Tailwind to these would require adding a CSS preprocessing build command which runs before the rest of the commands, so ideally something like #283 should be implemented first.

Chriscbr:
Closes #325

Adds Tailwind CSS integration to NextJs and NextJsTypeScript projects.

Included default options included are based on Tailwind CSS docs.


src/next.ts
```diff
@@ -472,3 +527,62 @@ class NextSampleCode extends Component {
     });
   }
 }
+
+export interface PostCssConfigOptions {
```
eladb, line 531: Move to a separate file

src/next.ts
```diff
@@ -472,3 +527,62 @@ class NextSampleCode extends Component {
...
+  /**
+   * @default "postcss.config.json"
+   */
+  readonly fileName?: string;
+}
+
+export class PostCssConfig {
```
eladb, line 538: Docstrings

Also rename to `PostCss`

src/next.ts
```diff
@@ -168,20 +196,36 @@ export interface NextComponentOptions {
...
+    this.tailwind = options.tailwind ?? true;
 
     project.addDeps('next', 'react', 'react-dom');
     if (this.typescript) {
       project.addDevDeps('@types/react', '@types/react-dom');
     }
+    if (this.tailwind) {
+      new PostCssConfig(project);
+      new TailwindConfig(project);
+      project.addDeps('tailwindcss', 'postcss', 'autoprefixer');
```
eladb, line 227: Move each dep to the specific component

src/next.ts
```diff
@@ -472,3 +527,62 @@ class NextSampleCode extends Component {
...
+}
+
+export interface TailwindConfigOptions {
+  /**
+   * @default "tailwind.config.json"
+   */
+  readonly fileName?: string;
+}
+
+export class TailwindConfig {
```
eladb, line 565: Move to a separate file.

I am okay with aggregating all these files under `src/web`

eladb (changes requested):

src/next.ts
Chriscbr: So my only objection here is that for Tailwind CSS, there are multiple ways to consume it ([docs](https://tailwindcss.com/docs/installation)). To summarize them:

- use it as a PostCSS plugin, requiring `tailwindcss, postcss, autoprefixer` dependencies (recommended by TailwindCSS). For NextJS, this works out of the box, and for other projects, a `tailwind build index.css -o tailwind.css` script needs to be added.
- use their CLI command, requiring the `tailwindcli` dependency, and requiring a `npx tailwindcss-cli@latest build -o tailwind.css` script to be added.
- referencing a CSS file from a CDN

Note that none of these choices have dependencies in common.

I'm implementing the first option as the default. But I don't want to lock users out of the others -- so I think `TailwindConfig` should only "model" the configuration file itself, and not the dependencies.

What I could do as a compromise is add a `tailwind` option to `PostCss`, and have that component keep the dependencies.

src/next.ts
Chriscbr: I wasn't sure which files you meant, so I just put these two components in `web.ts`.

src/next.ts
Chriscbr: done

Chriscbr:
I updated the implementation - in a nutshell, now `PostCss` has a `tailwind` option which if chosen, adds all necessary dependencies for Tailwind. But if some user wanted to just have a Tailwind config file without using PostCSS, they can also just use that component by itself, and it will not automatically add any dependencies (see the comment I made above for why). I think this is a good compromise, but I'm happy to do things another way.

src/next.ts
eladb: Sounds good. I think having each components expose these degrees of freedom while still encapsulating the details is the right way to go. I'd also be careful not to future-proof. Without taking an opinionated approach, it will be impossible to actually offer any abstraction.

My litmus test is usually what is the functional difference between the options. Sounds to me that there isn't much difference between these options.

src/next.ts
eladb: What I meant is:

```
web/tailwind.ts
web/next.ts
...
```

(basically start grouping domain-related classes in subdirectories/submodules -- like I did for `github`).

src/web.ts
```diff
@@ -0,0 +1,89 @@
+import { JsonFile } from './json';
+import { NodeProject } from './node-project';
+
+export interface PostCssOptions {
```
eladb, line 4: Split those to `web/postcss.ts` and `web/tailwind.ts`

Chriscbr:
Updated. Just note the change from:
```ts
const { ReactTypeScriptProject } = require('projen');

const project = new ReactTypeScriptProject({
```
to
```ts
const { web } = require('projen');

const project = new web.ReactTypeScriptProject({
```

eladb (approved):
👌👌🎉

PR #337, opened 2020-11-26
feat: projen tasks

eladb:
Tasks are a project-level feature to define a project command system backed by shell scripts. Tasks are synthesized into a manifest stored under `.projen/tasks.json` which is then consumed by the projen CLI. The `projen` CLI will automatically be extended based on the tasks defined for the project.

For example:

```ts
const task = project.addTask('hello');
task.exec('echo "hello, world!"');
```

Will result in a new CLI command called `hello` to be introduced:

```shell
$ projen hello
-----------------------------------------
hello: echo "hello, world!"
hello, world!
```

The `--help` option will show this new command:

```shell
$ projen --help
projen [command]

Commands:
  projen new [PROJECT-TYPE] [OPTIONS]  Creates a new projen project
  projen start                         Prints all project commands
  projen hello                       hello

Options:
  --version  Show version number                                         [boolean]
  --help     Show help                                                   [boolean]
```

The CLI also allows you to inspect the task:

```shell
$ projen hello -i
echo "hello, world!"
```

Here's a more elaborate example (from projen itself):

```ts
$ projen --help
projen [command]

Commands:
  projen new [PROJECT-TYPE] [OPTIONS]  Creates a new projen project
  projen start                         Prints all project commands
  projen compile                       Only compile
  projen test                          Run tests
  projen build                         Full release build (test+compile)
  projen bump                          Commits a bump to the package version based
                                       on conventional commits
  projen release                       Bumps version & push to master
  projen test:watch                    Run jest in watch mode
  projen test:update                   Update jest snapshots
  projen projen:upgrade                upgrades projen to the latest version
  projen watch                         Watch & compile in the background
  projen package                       Create an npm tarball
  projen eslint                        Runs eslint against the codebase
  projen compat                        Perform API compatibility check against
                                       latest version
  projen docgen                        Generate API.md from .jsii manifest
  projen readme-macros                 readme-macros

Options:
  --version  Show version number                                         [boolean]
  --help     Show help                                                   [boolean]
```

And then:

```shell
$ projen build --inspect
compile:
  jsii --silence-warnings=reserved-word --no-fix-peer-dependencies
  jsii-docgen
test:
  jest --passWithNoTests --updateSnapshot
  eslint:
    eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern src test
package:
  jsii-pacmak
readme-macros:
  mv README.md README.md.bak
  cat README.md.bak | markmac > README.md
  rm README.md.bak
```

Tasks can invoke each other as sub-tasks:

```ts
task.spawn(other_task);
```

All npm scripts from the `NodeProject` hierarchy were converted to use tasks instead of `addScript()` and `addScriptCommand()`.

This is (surprisingly) not a major breaking change. It likely has bugs, but from a project workflow perspective (as in using things like `yarn build`, it should continue to work (the npm script will delegate to the task).

BREAKING CHANGE: The `addScript()` method was renamed to `setScript()` and no longer supports adding start-menu items. To do that, just convert your code to use `project.addTask()`. It should be straightforward. `setScript()` can still be used to add just set npm scripts that are not advertised as tasks.
* The `addScriptCommand()` method was removed. Use `addTask()` instead.
* The `start` option and `Start` component have been removed since the start menu is now based on tasks.


src/cli/index.ts
```diff
@@ -1,14 +1,57 @@
...
+
+  const args = ya.argv;
+
+  // no command means just require .projenrc.js
+  if (args._.length === 0) {
+    synth();
+  }
+}
+
+function addBuildCommands(ya: yargs.Argv) {
```
eladb, line 23:
```suggestion
function addSequences(ya: yargs.Argv) {
```

Chriscbr:
At a high level I like this - it's pretty similar to what I imagined when you mentioned moving command logic outside of package.json. (I see there is still a "start" section - not sure if it can be removed, but it's not a big deal to me.)

That said, I'm a little iffy on expanding the scope of ownership of the library to handle build logic.

> In the roadmap, we will be able to also model cause->effect information and basically implement a makefile-like build system based on hash or timestamp checks. 

Is modeling dependency information and artifact-invalidation logic is something that should fall under projen? (If that is the goal, maybe this feature should be renamed to "trees" or "graphs" instead of "sequences"?)

Although I don't have a lot of hands-on experience, I know there are a *lot* of tools that try to solve this problem (bazel, cmake, gradle, gulp, grunt, ninja, ...). So, I'm just not sure if it might be more worthwhile to limit the scope of projen's build/script responsibilities to just being an "interface" for other build/script tools (without modeling dependencies/artifacts). That is, invest efforts in supporting existing build tools rather than creating a new build system from scratch.

Also, this PR makes it much harder for people to "eject" projen from their repo if they wanted to. We are essentially losing that feature with package.json "scripts" only referring to projen commands. I don't think this should be overlooked.

...That said, I have a lot less industry experience / might not have the right background knowledge to appreciate this feature (or might be overestimating the amount of complexity or optimization that actually exist in the build tools I mentioned), so take this with a heavy grain of salt. :-)  There could also be other benefits of this that I'm underestimating (like defining dependency-based build scripts between subprojects perhaps). What are your thoughts?

Also, interfacing with commands through "pj \<cmd\>" looks great / makes a lot of sense.

pgollucci:
I agree; however, I think the Level of Effort to migrate to or from projen is pretty low

pgollucci:
It builds over here.  Lets see if I can use it somewhere.

pgollucci:
So we're now committing to the `targets`? I'm not sure that was what was intended.  I can still work around this making a 'pre-build' 'hook' and then just making a wrapper to make sure they get called in sequence! ha, see what I did there!

The first thing I notice is I used to use `prependScriptCommand()` from #283, except I didn't b/c it never got merged, so I overrode build and I converted that to this

```ts
project.addSequence('build', {
  description: "builds it",
  shell: "bin/smilectl conf_merge && yarn run test && yarn run compile && yarn run synth",
});
```

```shell
pjp

🤖 Synthesizing project...
/Users/pgollucci/src/github.com/pgollucci/projen/lib/seqs/seq.js:160
            throw new Error(`duplicate sequence "${name}"`);
            ^

Error: duplicate sequence "build"
    at Sequences.addSequenceSpec (/Users/pgollucci/src/github.com/pgollucci/projen/lib/seqs/seq.js:160:19)
    at new Sequence (/Users/pgollucci/src/github.com/pgollucci/projen/lib/seqs/seq.js:27:23)
    at AwsCdkTypeScriptApp.addSequence (/Users/pgollucci/src/github.com/pgollucci/projen/lib/project.js:99:16)
    at AwsCdkTypeScriptApp.addSequence (/Users/pgollucci/src/github.com/pgollucci/projen/lib/node-project.js:454:27)
    at Object.<anonymous> (/Users/pgollucci/src/github.com/trexsolutions/smile-cdk/.projenrc.js:79:9)
    at Module._compile (node:internal/modules/cjs/loader:1102:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1131:10)
    at Module.load (node:internal/modules/cjs/loader:967:32)
    at Function.Module._load (node:internal/modules/cjs/loader:807:14)
    at Module.require (node:internal/modules/cjs/loader:991:19)
```

pgollucci:
So I've "converted" something over. `projen build` and `projen build -i` do the right thing. but `yarn build` which forwards to`projen build` does nothing.

```shell
yarn run build
yarn run v1.22.10
$ yarn run projen build
$ projen build
✨  Done in 0.50s.
```

eladb:
@Chriscbr good points about "projen ejection" and about leveraging existing build tools. I've been thinking about these topics, and I am trying to decide what to do. 

Perhaps the right direction is indeed to implement Sequences as an abstract interface with a pluggable backend (with a the current implementation as the base case perhaps), and allow build systems to be plugged-in. One such backing implementation could be to simply run those commands within npm scripts, to serve the "projen ejection" use case.

Naturally there is nothing the precludes simply using any build system behind those sequences...

Perhaps the correct abstraction, which is more inline with what this is currently about is to rename this to "Commands" and focus more on the CLI extension experience and less on the actual execution engine.

@pgollucci I'll have a look to see why this is not working as intended.

Let me revise and get back to you.

eladb:
@pgollucci wrote:

> So I've "converted" something over. projen build and projen build -i do the right thing. but yarn build which forwards toprojen build does nothing.

This is because you are using a local build of projen which gets overridden during synthesis when yarn installs the actual dependency. Simply run `yarn link projen` in order to restore the local link before `projen build` and you should be golden.

eladb:
@Chriscbr 

I understand there are many tools for implementing development workflows and builds but I'm a bit hesitant at the moment to take a strong dependency on one of them. And one of the goals of "tasks" is to be a unified command system for all projen projects. I am absolutely open if anyone is interested in offering a good execution engine for tasks, and we can easily switch up the underlying implementation.

I've also been thinking about the "eject" scenario and I think that we should formalize it. At the moment, it is merely a side-effect of how projen works, but it's not officially supported and requires manual tweaking. Let's add official support for ejecting projects (LOVE the term by the way!) which will perform the necessary changes. Namely: (1) make all files read-write; (2) remove projen markers; and (3) convert tasks to npm scripts.

The latter (convert tasks to npm scripts) can also be something we support as a switch if people are interested in "flattening" their tasks into npm scripts. We can do this quite easily today with the tasks model (simply synthesize the script content by concatenating all the commands in a task).

In the meantime, I am going to leave this like this (npm scripts delegate to projen commands), and see what people are thinking.

src/node-project.ts
```diff
@@ -455,6 +448,35 @@ export interface NodeProjectCommonOptions extends ProjectOptions {
    * @default - default content
    */
   readonly pullRequestTemplateContents?: string;
+
+  /**
+   * Determines how tasks are executed when invoked as npm scripts (yarn/npm run xyz).
+   */
```
eladb, line 454: Add default

src/node-project.ts
```diff
@@ -773,12 +811,11 @@ export class NodeProject extends Project {
     }
 
     if (options.start ?? true) {
```
eladb, line 814: just deprecate this

src/project.ts
```diff
@@ -175,39 +227,24 @@ export class Project {
...
       comp.postSynthesize();
     }
 
     // project-level hook
     this.postSynthesize();
 
-
     logging.info('Synthesis complete');
-
-    const start = this.components.find(c => c instanceof Start);
```
eladb, line 198: Yes, these were intentionally removed

src/tasks/task.ts
```diff
@@ -0,0 +1,168 @@
...
+   * @default {}
+   */
+  readonly env?: { [name: string]: string };
+
+  /**
+   * A shell command which determines if the steps of this task should be
+   * skipped. If the program exits with a zero exit code, steps will be skipped.
+   * A non-zero code means that task should be executed.
+   */
+  readonly skipIf?: string;
```
eladb, line 32: Not implemented yet

src/tasks/task.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * A task that can be performed on the project. Modeled as a series of shell
+ * commands and subtasks.
+ */
+export class Task {
+  /**
+   * The manifest file where all tasks are declared.
+   */
+  public static readonly MANIFEST_FILE = join(PROJEN_DIR, 'tasks.json');
```
eladb, line 49: Do we need this here?

src/tasks/task.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * The start menu category of the task.
+   */
+  public readonly category?: TaskCategory;
+
+  /**
+   * Project task list.
+   */
+  public readonly tasks: Tasks;
```
eladb, line 69: Make private

src/tasks/tasks.ts
```diff
@@ -0,0 +1,85 @@
...
+    this._tasks[name] = task;
+    return task;
+  }
+
+  /**
+   * Adds global environment.
+   * @param name Environment variable name
+   * @param value Value
+   */
+  public env(name: string, value: string) {
```
eladb, line 56: `addEnv`?

eladb:
@Chriscbr @pgollucci this is ready for another look if you are interested. I think the changes to `node-project.ts` can be a good way to see the full picture.

src/tasks/task.ts
Chriscbr: What is this field for?

Chriscbr (changes requested):
Cool! 👍  Yeah I don't think I would have suggested committing to any external build tool at this time (and even in the future, it might not be that advisable). And I agree, having project ejection as a separate operation is just as satisfactory. I think by keeping that as a feature, it might also have the side effect of keeping the complexity of this task system in check, in some respect.

.projenrc.js
```diff
@@ -32,20 +32,24 @@ const project = new JsiiProject({
   projenDevDependency: false, // because I am projen
   releaseToNpm: true,
   minNodeVersion: '10.17.0',
+  codeCov: true,
   compileBeforeTest: true, // since we want to run the cli in tests
+  npmTaskExecution: NpmTaskExecution.SHELL,
```
eladb, line 37: Add comment for why this is needed

package.json
```diff
@@ -10,22 +9,22 @@
...
-    "projen": "yarn compile && node .projenrc.js",
-    "no-changes": "(git log --oneline -1 | grep -q \"chore(release):\") && echo \"No changes to release.\"",
-    "bump": "yarn run --silent no-changes || standard-version",
-    "release": "yarn run --silent no-changes || (yarn run bump && git push --follow-tags origin master)",
+    "start": "yarn run projen start",
+    "projen": "yarn compile && node bin/projen",
+    "compile": "jsii --silence-warnings=reserved-word --no-fix-peer-dependencies && jsii-docgen",
     "test": "jest --passWithNoTests --updateSnapshot && yarn run eslint",
+    "build": "yarn run compile && yarn run test && yarn run package && yarn run readme-macros",
+    "bump": "standard-version",
```
eladb, line 17: No-changes is missing here

src/jsii-docgen.ts
```diff
@@ -10,11 +10,14 @@ import { StartEntryCategory } from './start';
...
-      startCategory: StartEntryCategory.RELEASE,
+
+    project.addTask('docgen', {
+      description: 'Generate API.md from .jsii manifest',
+      category: TaskCategory.RELEASE,
+      exec: 'jsii-docgen',
     });
-    project.addCompileCommand('jsii-docgen');
+
+    project.compileTask.exec('jsii-docgen');
```
eladb, line 20: Delegate to docgen instead of run manually

src/jsii-project.ts
```diff
@@ -174,14 +174,22 @@ export class JsiiProject extends TypeScriptProject {
       this.addFields({ deprecated: true });
     }
 
-    this.addScript('compat', `npx jsii-diff npm:$(node -p "require(\'./package.json\').name") -k --ignore-file ${compatIgnore} || (echo "\nUNEXPECTED BREAKING CHANGES: add keys such as \'removed:constructs.Node.of\' to ${compatIgnore} to skip.\n" && exit 1)`, {
-      startDesc: 'Perform API compatibility check against latest version',
-      startCategory: StartEntryCategory.RELEASE,
+    const compatCmd = this.addTask('compat', {
```
eladb, line 177:
```suggestion
    const compatTask = this.addTask('compat', {
```

src/node-project.ts
```diff
@@ -404,13 +404,6 @@ export interface NodeProjectCommonOptions extends ProjectOptions {
    */
   readonly start?: boolean;
 
-  /**
-   * Options for `yarn start`.
-   *
-   * @default - default options
-   */
-  readonly startOptions?: StartOptions;
```
eladb, line 412: Mention in breaking changes

src/node-project.ts
```diff
@@ -566,20 +589,16 @@ export enum AutoRelease {
...
-    { run: `yarn ${PROJEN_SCRIPT}` },
-  ];
-
   public readonly npmignore?: IgnoreFile;
   public readonly mergify?: Mergify;
   public readonly manifest: any;
   public readonly allowLibraryDependencies: boolean;
   public readonly entrypoint: string;
 
+  public readonly compileTask: Task;
```
eladb, line 598: Docstrings

src/node-project.ts
```diff
@@ -1099,38 +1120,30 @@ export class NodeProject extends Project {
...
-    this.start?.addEntry('test', {
-      desc: 'Run tests',
-      command: `${this.runScriptCommand} test`,
-      category: StartEntryCategory.TEST,
-    });
+  /**
+   * Adds commands to run as part of `yarn build`.
+   * @param commands The commands to add
+   */
+  public addBuildCommand(...commands: string[]) {
```
eladb, line 1143: Add @deprecation

src/project.ts
```diff
@@ -152,6 +177,23 @@ export class Project {
     return undefined;
   }
 
+  /**
+   * Finds a json file by name.
+   * @param filePath The file path.
+   */
+  public findJsonFile(filePath: string): JsonFile | undefined {
```
eladb, line 184: TryFind

src/projen-upgrade.ts
```diff
@@ -35,11 +35,14 @@ export class ProjenUpgrade {
...
+    const upgrade = project.addTask(script, {
+      description: 'upgrades projen to the latest version',
+      category: TaskCategory.MAINTAIN,
     });
 
+    upgrade.exec('yarn upgrade -L projen');
+    upgrade.exec('CI="" yarn projen');
+
     if (options.autoUpgradeSecret) {
       if (!project.github) {
```
eladb, line 44: Strong-reference task within workflow

src/tasks/runtime.ts
```diff
@@ -0,0 +1,105 @@
...
+import { join, resolve } from 'path';
+import * as chalk from 'chalk';
+import { TaskManifest, TaskSpec } from './model';
+import { Task } from './task';
+
+/**
+ * The runtime component of the tasks engine.
+ */
+export class TaskRuntime {
+  public readonly manifest: TaskManifest;
```
eladb, line 12: Docstrings

src/tasks/runtime.ts
```diff
@@ -0,0 +1,105 @@
...
+   * The tasks in this project.
+   */
+  public get tasks(): TaskSpec[] {
+    return Object.values(this.manifest.tasks ?? {});
+  }
+
+  /**
+   * Find a task by name, or `undefined` if not found.
+   */
+  public find(name: string): TaskSpec | undefined {
```
eladb, line 37: TryFind

src/tasks/runtime.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.run(task.subtask);
+      }
+
+      if (task.exec) {
+        const exec = task.exec;
+
+        if (firstCommandInSequence) {
+          console.log(`${chalk.magentaBright('-'.repeat(80))}`);
+          firstCommandInSequence = false;
+        }
```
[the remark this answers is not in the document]
eladb: Then it will have its own header. The logs currently won't reflect the hierarchy.

Let me see what I can do

src/tasks/runtime.ts
```diff
@@ -0,0 +1,105 @@
...
+        const result = spawnSync(exec, { cwd, shell: true, stdio: 'inherit', env });
+        if (result.status !== 0) {
+          console.log(chalk.red(`${name} failed in: "${exec}" at ${resolve(cwd)}`));
+          process.exit(1);
+        }
+      }
+    }
+  }
+
+  private renderRuntimeEnvironment(env: {[k: string]: string}, cwd: string) {
```
eladb, line 84: Docstring

src/tasks/task.ts
eladb: Not sure. Maybe I can remove. Let me check

src/typescript-typedoc.ts
```diff
@@ -7,9 +7,10 @@ import { TypeScriptProject } from './typescript';
 export class TypedocDocgen {
   constructor(project: TypeScriptProject) {
     project.addDevDeps('typedoc');
-    project.addScript('docgen', 'typedoc --out ' + project.docsDirectory, {
-      startDesc: `Generate TypeScript API reference ${project.docsDirectory}`,
-      startCategory: StartEntryCategory.RELEASE,
+    project.addTask('docgen', {
```
eladb, line 10: Shouldn't this also be part of "build"?

src/version.ts
```diff
@@ -17,18 +17,24 @@ export class Version extends Component {
...
+    const noChanges = 'git log --oneline -1 | grep -q "chore(release):"';
 
-    project.addScript('bump', `${project.runScriptCommand} --silent no-changes || standard-version`, {
-      startDesc: 'Commits a bump to the package version based on conventional commits',
-      startCategory: StartEntryCategory.RELEASE,
+    const bump = project.addTask('bump', {
+      description: 'Commits a bump to the package version based on conventional commits',
+      category: TaskCategory.RELEASE,
+      exec: 'standard-version',
+      skipIf: noChanges,
```
eladb, line 26: Implement

src/web/react.ts
```diff
@@ -149,24 +149,28 @@ export class ReactComponent extends Component {
...
+      description: 'Creates an optimized production build of your React application',
+      category: TaskCategory.BUILD,
+      exec: 'react-scripts build',
     });
 
-    project.addScript('eject', 'react-scripts eject', {
-      startDesc: 'Ejects your React application from react-scripts',
-      startCategory: StartEntryCategory.MISC,
+    project.addTask('eject', {
+      description: 'Ejects your React application from react-scripts',
```
eladb, line 165: Lol didn't know they have that

src/tasks/task.ts
eladb: It's needed in order to be able to resolve subtasks

src/tasks/runtime.ts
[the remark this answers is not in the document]
eladb: Fixed.

Chriscbr:
🎉

eladb:
> 🎉 

Thanks for all the feedback. Lmk what you think about the result.

Chriscbr:
It's great! Love how the printouts with chevrons look. I could nitpick a few things, like the name "spawn" feels a bit weird compared to "subtask", and giving RunTask a constructor with a side effect seems like nonstandard from a Java background (where having an execute() or run() method would be more common I think - also maybe rename it `TaskRunner`?)

But the rest of the changes makes sense. I think the real test would be seeing how well it plays with a module in another language, e.g. `PythonProject` (something I've been wanting to experiment with at some point, but it would take a bit of research).

pgollucci:
I understand that I can do `project.buildTask.spawn(foo) to` run something after the existing build but what about BEFORE?

pgollucci:
> I understand that I can do `project.buildTask.spawn(foo) to` run something after the existing build but what about BEFORE?

This one is going to take a minute for $work.

pgollucci:
The npx projen and simpler scripts in package.json is quite sweet

pgollucci:
It looks like yarn run became npx regardless of the `packageManager` setting, I assume that was intentional?

pgollucci:
Based on the `projen/projen` repo it looks like we are meant to commit `.projen/` dir. I don't have an issue with this, but it does open up the possibility we will write a secret to it which someone will then commit.

pgollucci:
> Based on the `projen/projen` repo it looks like we are meant to commit `.projen/` dir. I don't have an issue with this, but it does open up the possibility we will write a secret to it which someone will then commit.
I guess they would have already been in .projenrc.js so n/m

eladb:
@Chriscbr I tend to agree about `spawn`. I was looking for a verb and that's why `subtask` didn't work. How about `execTask()`?

Chriscbr:
@eladb I think they're equally good.

eladb:
> Based on the `projen/projen` repo it looks like we are meant to commit `.projen/` dir. I don't have an issue with this, but it does open up the possibility we will write a secret to it which someone will then commit.

Since projen manages your gitignore file, it should be included in your repo

PR #339, opened 2020-11-27
feat(cli): add switch for post-synthesis operations

### LINKED ISSUE #322: CLI option to disable installing dependencies
Right now, if you want to bootstrap a new project using `npx projen`, you can either use the `--no-synth` flag to just generate the `.projenrc.js` file, or you can generate all of the files and install all of the associated dependencies. There is no in between option.

It would be nice to be able to just generate the project-related files without also automatically running `yarn install` etc. to make it easier to test projen. I wanted to create a snapshot test which records a list of all of the files that are generated for the different project types, but this test ended up taking 4+ minutes on my local machine due to the time it takes to install dependencies of each project using yarn. If I remove the `postSynthesize()` part of `NodeProject` which installs all of the dependencies, then the test only takes a few seconds to run.

----

A couple implementation options that come to mind:
- make a general CLI flag called `post-synthesis` that disables the `postSynthesis` step
- extend `Project` to have a general `installDependencies` method that can be overwritten, refactor `NodeProject` to use it, and create a general CLI flag called `install-deps` that determines if this method gets called
- add an option to `NodeProject` called `installDeps` that specifically disables installing npm modules (this will automatically add an associated CLI flag)

Chriscbr:
Fixes #322 

- Add CLI options --post and --no-post to enable or disable post-synthesis operations (such as installing node dependencies) during `yarn projen` or `npx projen new <project>` commands.
- Update several tests to disable post-synthesis operations, reducing the total testing runtime from ~44s to ~11s (about 75%)


src/cli/cmds/new.ts
```diff
@@ -310,6 +310,7 @@ function newProject(baseDir: string, type: inventory.ProjectType, args: any, add
 
   // synthesize if synth is enabled (default).
   if (args.synth) {
+    process.env.POST_SYNTHESIS_ENABLED = args.post.toString();
```
eladb, line 313: Rename (and reverse logic) to `PROJEN_DISABLE_POST`

We want the lack of this defined to be the default behavior.

src/cli/index.ts
```diff
@@ -5,10 +5,12 @@ const args = yargs
   .commandDir('cmds')
   .recommendCommands()
   .wrap(yargs.terminalWidth())
+  .option('post', { type: 'boolean', default: true, desc: 'Run post-synthesis steps such as installing dependencies' })
```
eladb, line 8:
```suggestion
  .option('post', { type: 'boolean', default: true, desc: 'Run post-synthesis steps such as installing dependencies. Use --no-post to skip' })
```

test/toml.test.ts
```diff
@@ -16,7 +16,7 @@ test('toml object can be mutated before synthesis', () => {
     foo: 1234,
   };
 
-  const out = synthSnapshot(prj);
+  const out = synthSnapshot(prj, true);
```
eladb, line 19: Let's make `synthSnapshot` default to skipping post synthesis and add `synthSnapshotWithPost` for the cases where it's desired.

eladb (changes requested):

src/cli/cmds/new.ts
Chriscbr: Done

src/project.ts
```diff
@@ -185,12 +185,14 @@ export class Project {
       comp.synthesize();
     }
 
-    for (const comp of this.components) {
-      comp.postSynthesize();
-    }
+    if (process.env.PROJEN_DISABLE_POST !== 'true') {
```
eladb, line 188: I usually like `1` to also be a valid truthy value for boolean environment variables.

eladb (approved):

PR #343, opened 2020-11-30
feat: lint .projenrc.js

jogold:
Make the `.projenrc.js` file lintable.

Ensure it's created without linting errors by replacing double quotes
with single quotes.

Add support of eslint overrides.

Fix some defaults that were not JSON parsable


pgollucci:
:+1 Works for me

src/cli/cmds/new.ts
```diff
@@ -147,7 +147,7 @@ function renderParams(type: inventory.ProjectType, params: Record<string, string
       optionsWithDefaults.push(optionName);
     } else {
       const defaultValue = option.default?.startsWith('-') ? undefined : (option.default ?? undefined);
-      paramRender = `// ${optionName}: ${defaultValue},`;
+      paramRender = `// ${optionName}: ${defaultValue?.replace(/"/g, '\'')},`; // single quotes
```
eladb, line 150: what happens if the default value includes double quotes? Perhaps the right way to do this is that if the default value is a string, then just replace the surrounding quotes and not replace-all...

src/cli/cmds/new.ts
```diff
@@ -232,9 +232,9 @@ function commandLineToProps(type: inventory.ProjectType, argv: Record<string, un
           if (queue.length === 0) {
             let val = value;
 
-            // if this is a string, then quote it
+            // if this is a string, then single quote it
             if (val && typeof(val) === 'string') {
-              val = JSON.stringify(val);
+              val = JSON.stringify(val).replace(/"/g, '\'');
```
eladb, line 237: just replace the surrounding quotes in case the value itself includes quotes

src/eslint.ts
```diff
@@ -25,12 +26,29 @@ export interface EslintOptions {
   readonly ignorePatterns?: string[];
 }
 
+export interface EslintOverride {
```
eladb, line 29: docstrings

src/typescript.ts
```diff
@@ -215,6 +216,7 @@ export class TypeScriptProject extends NodeProject {
       const tsconfig = this.jest.generateTypescriptConfig({
         fileName: 'tsconfig.jest.json',
         include: [
+          PROJEN_RC,
```
eladb, line 219: Why do we need this here? PROJEN_RC is not expected to be compiled.

jogold: Since the eslint config specifies `parserOptions.project` all files that should be linted must be included in the project.

eladb (approved):

eladb (approved):

PR #360, opened 2020-12-10
feat(gitpod): Support Gitpod

pgollucci:
Some Project vs NodeProject issues need to be sorted before this
can be implimented in finalization, but this should allow use for now.


pgollucci:
This still needs tests but pending feedback first.  I addressed a few things from the previous version

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+        init: 'yarn install', // XXX: this has no answer at the project level yet, it also needs NodeProject for now
+        command: 'npx projen build', // XXX: project.projenCommand is on NodeProject for now
+      });
+    }
+
+    // This is what get synthesized
+    new YamlFile(project, GITPOD_FILE, {
+      obj: {
+        tasks: this.tasks,
+        docker: this.docker,
```
Chriscbr, line 214: If `options.docker` is `{ file: '.gitpod.Dockerfile' }`, then this gets synthesized as
```
docker:
  file: .gitpod.Dockerfile
```
when it should get synthesized as
```
image:
  file: .gitpod.Dockerfile
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    // Add user supplied tasks or produce the default .gitpod.yml
+    if (options?.tasks) {
+      this.addTasks(...options.tasks);
+    } else {
+      this.addTasks({
+        init: 'yarn install', // XXX: this has no answer at the project level yet, it also needs NodeProject for now
+        command: 'npx projen build', // XXX: project.projenCommand is on NodeProject for now
+      });
+    }
```
Chriscbr, line 208: These comments are valid concerns - I think for now it's fine if the default behavior (in the else branch) is to add no tasks, assuming that is allowed by the spec. If .gitpod.yml for some reason requires at least one task, then you could just put a simple "echo initialized" command instead, just so we aren't tied to any NodeProject specific details.

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+    });
+  }
+
+  /**
+   * Adds another task to the Gitpod configuration
+   * XXX: no way to clear out the default: yet
+   * @param task The additional tasks
+   */
+  public addTasks(...task: GitpodTask[]) {
+    this.tasks.push(...task);
```
Chriscbr, line 225:
```suggestion
  public addTasks(...tasks: GitpodTask[]) {
    this.tasks.push(...tasks);
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * Specify a custom Docker setup
+   *
+   * @param docker The docker configuration
+   */
+  public customDocker(docker: GitpodDocker) {
```
Chriscbr, line 233:
```suggestion
  public addCustomDocker(docker: GitpodDocker) {
```

src/project.ts
```diff
@@ -32,6 +33,14 @@ export interface ProjectOptions {
...
    */
   readonly outdir?: string;
+
+  /**
+   * Whether to enable `gitPod`
+   * It can be customized later if needed
+   *
+   * @default false
+   */
+  readonly gitPod?: boolean;
```
Chriscbr, line 43: I think this should be `gitpod`, along with the other occurrences in the file
```suggestion
  readonly gitpod?: boolean;
```

src/project.ts
```diff
@@ -32,6 +33,14 @@ export interface ProjectOptions {
    * @default "."
    */
   readonly outdir?: string;
+
+  /**
+   * Whether to enable `gitPod`
+   * It can be customized later if needed
```
Chriscbr, line 39:
```suggestion
   * Adds a gitpod configuration
```

src/project.ts
```diff
@@ -32,6 +33,14 @@ export interface ProjectOptions {
...
+   *
+   * @default false
```
Chriscbr, line 41: I'd be okay with enabling gitpod by default, but it's up to what you and others think.

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+   * We default to jsii/superchain fo the image, so this can be required
+   */
+  private docker: GitpodDocker;
+
+  /**
+   *
+   * @param project The project
+   * @param options The component configuration options for this project
+   */
+  constructor(project: Project, options?: GitpodOptions) {
```
Chriscbr, line 191: I think it's fine to omit the docstring for the constructor here.
```suggestion
  constructor(project: Project, options?: GitpodOptions) {
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * You can configure how the terminal should be opened relative to the previous task.
+   * @default
+   */
+  readonly openMode?: GitpodOpenMode;
+
+  /**
+   * You can configure where in the IDE the terminal should be opened
+   * @default BOTTOM
```
Chriscbr, line 125:
```suggestion
   * @default GitpodOpenIn.BOTTOM
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * A name for this
+   * @default
+   */
+  readonly name?: string;
+
+  /**
+   * You can configure how the terminal should be opened relative to the previous task.
+   * @default
```
Chriscbr, line 119: I found some more default information mentioned in https://github.com/gitpod-io/gitpod/blob/master/components/dashboard/public/schemas/gitpod-schema.json
```suggestion
   * @default GitpodOpenMode.TAB_AFTER
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+   * This must be defaulted per project
+   * @default []
+   */
+  readonly tasks: GitpodTask[];
+}
+
+/**
+ * The Gitpod component itself
+ */
+export class Gitpod extends Component {
```
Chriscbr, line 175: This description is kind of superfluous -- maybe just highlighting the output would make more sense?
```suggestion
/**
 * Creates a Gitpod configuration file
 */
export class Gitpod extends Component {
```

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * The List of tasks. The default will be present or a customization
+   */
+  private tasks = new Array<GitpodTask>();
+
+  /**
+   * We default to jsii/superchain fo the image, so this can be required
+   */
+  private docker: GitpodDocker;
```
Chriscbr, line 184: We could consider just omitting these docstrings, since these are private fields and I think their purpose should be clear from `GitpodOptions`.

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Splits and adds the terminal to the right
+   */
+  SPLIT_RIGHT = 'split-right',
+  /**
+   * Splits and adds the terminal to the left
+   */
+  SPLIT_LEFT = 'split-left',
+  /**
+   * Splits and adds the terminal to the tops
```
Chriscbr, line 63:
```suggestion
   * Splits and adds the terminal to the top
```

Chriscbr (changes requested):

.gitpod.yml
```diff
@@ -0,0 +1,7 @@
+# ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
+
+tasks:
+  - init: yarn install
+    command: npx projen build
+docker:
+  image: jsii/superchain
```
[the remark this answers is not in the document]
smacintyre: Yes, this should be `image:` [See to gitpod docs.](https://www.gitpod.io/docs/config-docker/)

Also, can we please default to `gitpod/workspace-full` -- it's the standard gitpod image. The `jsii/superchain` is not required unless you are actually developing CDK. And to the best of my knowledge it's not completely compatible with gitpod. My team exclusively develops in gitpod, using CDK, and have never run into any issues with the default image: `gitpod/workspace-full`. 

Also, It should seem odd for projen to default to CDK develop specific image. Projen should be more neutral.

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * https://www.gitpod.io/docs/configuration/
+ * https://www.gitpod.io/docs/config-start-tasks/
+ */
+
+/**
+ * If the standard Docker image provided by Gitpod does not include the tools
+ * you need for your project, you can provide a custom Docker image OR Dockerfile.
+ * https://hub.docker.com/r/gitpod/workspace-full/ is the default Gitpod image
```
smacintyre, line 18: Here is says `gitpod/workspace-full` is the default image. However in the code and other comments `jsii/superchain` is the default.

We should be keeping proejn application neutral and so use the standard default of `gitpod/workspace-full`

src/gitpod.ts
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * What can we configure for the GitPod component
+ */
+export interface GitpodOptions {
+  /**
+   * Optional Docker Configuration
+   * Defaults to https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile
+   * @default undefined
+   */
+  readonly docker?: GitpodDocker;
```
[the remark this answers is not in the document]
pgollucci: I'm torn here. Gitpod defaults to the full one.  Which is true if you have no .gitpod.yml or if you leave out the image/file fields.

Projen itself is defaulting it to jsiii/superchain. 

I'm not sure which makes the most sense to do.  I don't think we need jsii/superchain the giptod one has 90-100% of the deps, we can install anything missing in a `before`

.gitpod.yml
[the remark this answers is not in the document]
pgollucci: yep my goof

src/gitpod.ts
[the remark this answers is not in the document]
pgollucci: See below going with Gitpod defaults

src/gitpod.ts
pgollucci: It does not require it, but I like the echo initiliazed

src/gitpod.ts
pgollucci: fixed

src/gitpod.ts
pgollucci: done

src/gitpod.ts
pgollucci: fixed

src/gitpod.ts
pgollucci: agree. done.

src/gitpod.ts
pgollucci: done

src/gitpod.ts
pgollucci: done

src/project.ts
pgollucci: Formally its Gitpod so I guess that makes it one word in which case gitpod would be correct.

pgollucci: renamed

src/project.ts
pgollucci: Leaving as is for this pr, we can flip the default seperately

pgollucci:
I ran an integration test live in Gitpod on this PR works fine.  Though it does look like GitPod is going to move to GitLab eventually as GitHub brings on CodeSpaces.

pgollucci:
@Chriscbr @smacintyre  That's all the comments.

pgollucci:
Alright, what am I missing, if you call addTasks() or addCustomerDocker() they run fine, but when you call synth later the output is unchanged even though the project object's component gitpod is updated.  This looks similar to the mergify setup to.

Chriscbr (commented):
> Alright, what am I missing, if you call addTasks() or addCustomerDocker() they run fine, but when you call synth later the output is unchanged even though the project object's component gitpod is updated. This looks similar to the mergify setup to.

This is because of the way object references are handled in JavaScript - when you set `this.docker = ...` in `addCustomDocker()`, it changes which object `this.docker` points to, but the `YamlFile` keeps pointing to the old object.

One way to get around this is to ensure `YamlFile` unambiguously gets the latest value of `this.docker` each time, like this:

```ts
new YamlFile(project, GITPOD_FILE, {
  obj: {
    image: () => this.renderDockerImage(),
    tasks: this.tasks,
  },
});
```

with a corresponding function like:
```ts
private renderDockerImage() {
  if (this.docker?.image) {
    return this.docker.image;
  } else if (this.docker?.file) {
    return {
      file: this.docker.file,
    };
  } else {
    return undefined;
  }
}
```

pgollucci:
@eladb tests added.

Thanks @Chriscbr

test/gitpod.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+describe('gitpod image & file', () => {
+  test('given an image', () => {
+    const project = new Project({
+      outdir: tempDir,
+      gitpod: true,
+    });
+    if (project.gitpod) {
+      project.gitpod.addCustomDocker({ image: 'jsii/superchain' });
+    }
```
Chriscbr, line 58:
```suggestion
    project.gitpod?.addCustomDocker({ image: 'jsii/superchain' });
```
Same change can be made for the other tests below.

pgollucci: oh yeah duh.

Chriscbr (changes requested):
Can these comments be removed before merging?

#362 mentions possibly restructuring things to use a more abstract API, but I think this PR/implementation for Gitpod is good enough to publish - it's okay if things change later. I'm fine with merging after these changes. Feel free to shout out if you have other thoughts or comments @eladb

src/gitpod.ts
```diff
@@ -0,0 +1,235 @@
...
+   * @default
+   */
+  readonly before?: string;
+
+  /**
+   *
+   * The init property can be used to specify shell commands that should only be executed after a workspace
+   * was freshly cloned and needs to be initialized somehow. Such tasks are usually builds or downloading
+   * dependencies. Anything you only want to do once but not when you restart a workspace or start a snapshot.
+   * XXX: NodeProject
```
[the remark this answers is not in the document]
pgollucci: I agree

pgollucci:
@projen rebuild

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+ * you need for your project, you can provide a custom Docker image OR Dockerfile.
+ * https://hub.docker.com/r/gitpod/workspace-full/ is the default Gitpod image
+ */
+export interface GitpodDocker {
+
+  /**
+   *
+   * A publicly available image to use
+   *
+   * @default
```
eladb, line 26: Missing default description. Something like:

```
@default - uses the standard gitpod image (see [LINK])
```

By convention, if you don't have a specific value for default, use `-`

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+export interface GitpodTask {
+
+  /**
+   * Required. What to run
+   */
+  readonly command: string;
+
+  /**
+   * A name for this
+   * @default
```
eladb, line 113: Make sure to doc all defaults

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+   * if this is unset, so undefined here means `gitpod/workspace-full`
+   * @default undefined
+   */
+  readonly docker?: GitpodDocker;
+
+  /**
+   * This must be defaulted per project
+   * @default []
+   */
+  readonly tasks: GitpodTask[];
```
eladb, line 169:
```suggestion
  readonly tasks?: GitpodTask[];
```

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * The Gitpod component which emits .gitpod.yml
+ */
+export class Gitpod extends Component {
+  private tasks = new Array<GitpodTask>();
```
eladb, line 176:
```suggestion
  private readonly tasks = new Array<GitpodTask>();
```

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+  private docker: GitpodDocker | undefined;
```
eladb, line 177:
```suggestion
  private readonly docker: GitpodDocker | undefined;
```

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+        image: () => this.renderDockerImage(),
+        tasks: this.tasks,
+      },
+    });
+  }
+
+  /**
+   *
+   * Specify a customer docker setup
+   * @param docker
```
eladb, line 204: Missing docstrings

src/gitpod.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    this.docker = docker;
+  }
+
+  /**
+   * Adds another task to the Gitpod configuration
+   * @param tasks The additional tasks
+   */
+  public addTasks(...tasks: GitpodTask[]) {
+    this.tasks.unshift(...tasks);
```
eladb, line 219: Why unshift instead of push?

eladb (changes requested):
@pgollucci Looks awesome. A few minor comments and we are ready

src/gitpod.ts
pgollucci: Its b/c of the default echo Initialized.  This lets that remain the last statement.  Otherwise, It would be first, or we need change a default.  Or it doesn't matter

pgollucci:
@projen rebuild

pgollucci:
All feedback addressed again except for the `unshift` which I explained above.

src/gitpod.ts
eladb: That's weird because it will cause any `addTasks()` call to add tasks before the existing tasks.

Why do we need the "echo Initialized" default?

eladb (changes requested):

src/gitpod.ts
eladb: Ping @pgollucci

pgollucci: we don't.

pgollucci (commented):
@projen rebuild

pgollucci:
@projen rebuild

pgollucci:
unshift() has been addressed

eladb (approved):

pgollucci:
@Chriscbr ping, I think you need to give an approval

Chriscbr (approved):

PR #363, opened 2020-12-11
feat(jsii): support publishing maven artifacts to github

### LINKED ISSUE #355: Feature: Option for Github to store the maven package
The JSII https://github.com/aws/jsii-release#maven allows to specify GitHub as package registry for Maven packages. We easily could provide those option for projen too for the **JsiiProject** options . Should be fairly simple. I will start working on a PR if no one is faster any time soon ^^. 

Cheers Martin

mmuller88:
Fixes #355
* Add the envs necessary for releasing to github java artefact package https://github.com/aws/jsii-release#maven
(I will see where I have to provide more docu)


src/jsii-project.ts
```diff
@@ -357,6 +357,8 @@ export class JsiiProject extends TypeScriptProject {
             name: 'Release',
             run: 'npx -p jsii-release jsii-release-maven',
             env: {
+              MAVEN_SERVER_ID: '${{ secrets.MAVEN_SERVER_ID }}',
+              MAVEN_REPOSITORY_URL: '${{ secrets.MAVEN_REPOSITORY_URL }}',
```
eladb, line 361: Wondering if these two should just be under the `jsii.java` option and then hard coded here instead of a secret. Those secrets are such a pain to manage.

mmuller88: Yes was thinking similar and wanted to wait for feedback :). I think having them in the options woulds be cool.

mmuller88: @eladb  I will make that change request in a bit. Thank you for your feedback :)

mmuller88: I might have to give the private method publishToMaven() those properties in the options object.

src/jsii-project.ts
```diff
@@ -357,6 +361,8 @@ export class JsiiProject extends TypeScriptProject {
             name: 'Release',
             run: 'npx -p jsii-release jsii-release-maven',
             env: {
+              // MAVEN_SERVER_ID: '${{ secrets.MAVEN_SERVER_ID }}',
```
eladb, line 364: What's up here? I expected these to be fed from "options" now

eladb (changes requested):

src/jsii-project.ts
mmuller88: yeah that was my question if it is like that. Will adjust

eladb: Yap

mmuller88: done :)

docs/awscdk-construct.md
```diff
@@ -113,7 +115,7 @@ any number of jsii target languages.
 
 * npm - `NPM_TOKEN` ([docs](https://github.com/aws/jsii-release#npm))
 * .NET - `NUGET_API_KEY` ([docs](https://github.com/aws/jsii-release#nuget))
-* Java: `MAVEN_GPG_PRIVATE_KEY`, `MAVEN_GPG_PRIVATE_KEY_PASSPHRASE`, `MAVEN_PASSWORD`, `MAVEN_USERNAME`, `MAVEN_STAGING_PROFILE_ID` ([docs](https://github.com/aws/jsii-release#maven))
+* Java: `MAVEN_GPG_PRIVATE_KEY`, `MAVEN_GPG_PRIVATE_KEY_PASSPHRASE`, `MAVEN_PASSWORD`, `MAVEN_USERNAME`, `MAVEN_STAGING_PROFILE_ID`, `MAVEN_SERVER_ID`, `MAVEN_REPOSITORY_URL` ([docs](https://github.com/aws/jsii-release#maven))
```
eladb, line 118: I believe there is no need to upload these as secrets.

eladb (changes requested):

docs/awscdk-construct.md
mmuller88: reverted it :)

mmuller88:
Fixed some outdatd issues. happy new year :)

mmuller88:
@eladb i updated my fork with. I hope everything looks good now :)

src/jsii-project.ts
```diff
@@ -92,6 +92,8 @@ export interface JsiiJavaTarget {
   readonly javaPackage: string;
   readonly mavenGroupId: string;
   readonly mavenArtifactId: string;
+  readonly mavenServerId?: string;
```
eladb, line 95: Can we take the opportunity and add some docstrings here?

eladb (changes requested):
Please add unit test

src/jsii-project.ts
mmuller88: jup. in the aws-cdk main project is kind of check for that to enforce docstrings. I would like to have that :D

mmuller88:
@eladb I added it to the snapshot test. I think that was you mean with unit test right?

src/jsii-project.ts
```diff
@@ -92,6 +92,14 @@ export interface JsiiJavaTarget {
   readonly javaPackage: string;
   readonly mavenGroupId: string;
   readonly mavenArtifactId: string;
+  /**
+   * Used in maven settings for credential lookup (e.g. use github when publishing to GitHub). Defaults to ossrh for Maven Central.
+   */
```
eladb, line 97: Default should be explicitly marked like so:

```suggestion
  /**
   * Used in maven settings for credential lookup (e.g. use github when publishing to GitHub).
   * @default "ossrh" Defaults to Maven Central.
   */
```

src/jsii-project.ts
```diff
@@ -92,6 +92,14 @@ export interface JsiiJavaTarget {
...
+  readonly mavenServerId?: string;
+  /**
+   * Deployment repository when not deploying to Maven Central
```
eladb, line 100: Add "@default"

src/__tests__/integration/cdk-watchful.projenrc.js
```diff
@@ -48,6 +48,8 @@ const project = new ConstructLibraryAws({
     javaPackage: 'com.github.eladb.watchful',
     mavenGroupId: 'com.github.eladb',
     mavenArtifactId: 'cdk-watchful',
+    mavenServerId: 'github',
```
eladb, line 51: Instead of tweaking this integration test, I rather we add a more fine grained unit test to `jsii.test.ts`. Something like this should do:

```ts
test('maven repository options', () => {
  const p = new JsiiProject({
    outdir: mkdtemp(),
    name: 'testproject',
    mavenServerId: 'github',
    mavenRepositoryUrl: 'footer',
  });

  const workflow = synthSnapshot(p))['.github/workflows/xxxx'];
  // now assertions on the workflow
});
```

src/__tests__/__snapshots__/integ.test.ts.snap
```diff
@@ -1452,6 +1454,8 @@ project.synth();
           "maven": Object {
             "artifactId": "cdk-watchful",
             "groupId": "com.github.eladb",
+            "mavenServerId": "github",
+            "mavenRepositoryUrl": "https://maven.pkg.github.com/eladb",
```
eladb: Wait?! why are these going into `package.json`. I don't believe they are needed in there.

eladb (changes requested):
See comments

src/__tests__/__snapshots__/integ.test.ts.snap
mmuller88: At least for the first one following the instructions here https://github.com/aws/jsii-release#maven you have to save somehow if you want to use maven central or github and if github which repository Url. 

But yes we might could calculate mavenRepositoryUrl out of the choice if "github" and the existing repository name? that would be cool.

mmuller88: kind of

"https://maven.pkg.github.com/${{ github.repository }}"

src/__tests__/integration/cdk-watchful.projenrc.js
mmuller88: I think I get what you say and I try to add the test to jsii.test.ts but the running of the unit test is a bit strange. It seems I can not directly run it out of my VS Code with clicking on run there. It tries to run compiled js tests from the lib folder. 

And it happened two times now that the whole projen project erased itself from my local storage 0.o. I do use Ubuntu 20.04.

Can you please explain me how your workflow is to run tests from jsii.test.ts.

eladb: Sorry about that. My workflow is: start `yarn watch` and then run `yarn test:watch`. 

Then you have an interactive jest experience. You can press `p` to select which test to run and `u` to update snapshots.

mmuller88: oh wow thats a really cool way of unit testing. Basically on the fly <3. I added a unit test as you requested :).
Maybe not necesarry but the way of testing you describe here is very useful, do we have it in the docs somewhere? If not can we add it?

mmuller88:
@eladb  could you have a look please :)

src/jsii-project.ts
```diff
@@ -102,6 +102,18 @@ export interface JsiiJavaTarget {
...
+  /**
+   * Used in maven settings for credential lookup (e.g. use github when publishing to GitHub).
+   *
+   * @default "ossrh" Defaults to Maven Central.
+   */
+  readonly mavenServerId?: string;
+  /**
+   * Deployment repository when not deploying to Maven Central
+   *
+   * @default not set
```
eladb, line 114:
```suggestion
   * @default - not set
```

src/jsii-project.ts
```diff
@@ -208,10 +220,12 @@ export class JsiiProject extends TypeScriptProject {
         maven: {
           groupId: options.publishToMaven.mavenGroupId,
           artifactId: options.publishToMaven.mavenArtifactId,
+          serverId: options.publishToMaven.mavenServerId,
+          repositoryUrl: options.publishToMaven.mavenRepositoryUrl,
```
eladb, line 224: Why are those here? They are not needed/supported by jsii-pacmak:

https://aws.github.io/jsii/dev-guide/configuration/targets/java/

src/jsii-project.ts
```diff
@@ -329,7 +343,7 @@ export class JsiiProject extends TypeScriptProject {
     });
   }
 
-  private publishToMaven() {
+  private publishToMaven(java: any) {
```
eladb, line 346: Keep a minimal surface area if possible:

```suggestion
  private publishToMaven(serverId: string | undefined, repositoryUrl: string | undefined) {
```

eladb (changes requested):

mmuller88:
Hi @eladb  thank you for your patient. And sorry for the late response. The week was a bit rough :)

eladb (approved):

Issue #356, opened 2020-12-07
Tasks: allow passing arguments to tasks


eladb:
Example:

```ts
const hello = project.addTask('hello');
project.exec(`echo Hello, $@!`);
```

```shell
$ pj hello "world"
Hello, world!
```

### Questions

1. The `$@` notation is just an example. Perhaps it would be easiest to support this by injecting the arguments into environment variable(s).
2. Would be amazing if tasks could leverage yargs' engine to parse switches.
3. Should tasks also "register" positional arguments and switches for yargs?

pgollucci:
Another use case
`yarn deploy --profile damadden88`

mmuller88:
Yep and that is actually a regression because that was possible before. The workaround now could be:
```
project.setScript('cdkDeploy', 'cdk deploy');
```
thank you can do again:
```
yarn cdkDeploy --profile damadden88
```

asmajlovicmars:
Oh, thank you all! This is amazing, and exactly what I was looking for! My case is deploying multiple stacks, therefore (for now): 

`project.setScript('deployall', 'cdk deploy --all');`

eladb:
In the meantime, you can simply add a task:

```ts
project.addTask('deploy-all', {
  exec: 'cdk deploy --all'
});
```

And then both of these will work:

```shell
$ pj deploy-all
# or
$ yarn deploy-all
```

asmajlovicmars:
> ```shell
> pj
> ```

Thank you @eladb! This is amazing

bigkraig:
Also poses a problem running `bump` with flags for things like `--prerelease`.

Chriscbr:
Maybe as an escape hatch for all tasks, we could allow extra arguments to be passed automatically after a "--", as in:

```ts
const deploy = project.addTask('deploy');
deploy.exec(`cdk deploy`);
```

```shell
$ pj deploy -- --all
... runs 'cdk deploy --all'
```

And meanwhile, more fancy argument-insertion could be allowed like so:

```ts
const lint = project.addTask('lint');
lint.exec(`eslint --ext .ts,.tsx $@ src tests`);
```

```shell
$ pj lint --fix -- build-tools
... runs 'eslint --ext .ts,.tsx --fix src tests build-tools'
```

eladb:
I am good with this proposal. Only wondering how this will be have in case the task executes more then one command. Which command will accept the additional args (that's the main reason I haven't done this like this yet).

wchaws:
[...]

PR #368, opened 2020-12-14
chore(node-project): adds `gh` cli rebuild example to doc string

pgollucci:


pgollucci (commented):
@projen rebuild

pgollucci:
Live exercise of comment above

src/node-project.ts
```diff
@@ -470,6 +470,7 @@ export interface NodeProjectCommonOptions extends ProjectOptions {
   /**
    * The pull request bot command to use in order to trigger a rebuild and
    * commit of the contents of the branch. The command must be prefixed by "@projen", e.g. "@projen rebuild"
+   * `gh pr review $pr --comment -b "@projen rebuild"`
```
eladb, line 473: Maybe we we can add a task to `NodeProject` that will do that:

```ts
project.addTask('pr:rebuild', { exec: 'gh pr review --comment ... ' });
```

eladb (approved):

src/node-project.ts
pgollucci: Sure.

pgollucci: That likely depends on #356 for $pr unless we grab it from .git/ if someone does gh pr checkout

pgollucci (commented):
@projen rebuild

PR #372, opened 2020-12-16
chore: upgrade jsii-docgen

Chriscbr:
Upgrades jsii-docgen and adds missing docs.


eladb (approved):
nice! docs looks great with submodules.

Copy @rix0rrr

PR #377, opened 2020-12-17
fix: tasks not executing on Windows due to invalid PATH values

### LINKED ISSUE #370: Invalid tag name "$PATH'": Tags may not have any characters that URIComponent encodes
On windows (I know, I'm sorry), `npx projen` works fine and generates all content. As well as the command `npx projen start`. Other commands such as `npx projen compile` or `npx projen build` will fail with the following error

```bash
C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\node_modules\npm-package-arg\npa.js:296
      throw invalidTagName(spec)
      ^

Error: Invalid tag name "$PATH'": Tags may not have any characters that encodeURIComponent encodes.
    at invalidTagName (C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\node_modules\npm-package-arg\npa.js:91:15)
    at resolve (C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\node_modules\npm-package-arg\npa.js:81:12)
    at npa (C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\node_modules\npm-package-arg\npa.js:52:10)
    at Function.parseArgs (C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\node_modules\libnpx\parse-args.js:48:23)
    at Object.<anonymous> (C:\Users\codyj\AppData\Roaming\npm\node_modules\npm\bin\npx-cli.js:8:9)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
}
🤖 Cannot read property 'toString' of null
```

The code throwing the error:

```javascript
function fromRegistry (res) {
  res.registry = true
  const spec = res.rawSpec === '' ? 'latest' : res.rawSpec
  // no save spec for registry components as we save based on the fetched
  // version, not on the argument so this can't compute that.
  res.saveSpec = null
  res.fetchSpec = spec
  if (!semver) semver = require('semver')
  const version = semver.valid(spec, true)
  const range = semver.validRange(spec, true)
  if (version) {
    res.type = 'version'
  } else if (range) {
    res.type = 'range'
  } else {
    if (encodeURIComponent(spec) !== spec) {
      throw invalidTagName(spec)
    }
    res.type = 'tag'
  }
  return res
}

function invalidTagName (name) {
  const err = new Error(`Invalid tag name "${name}": Tags may not have any characters that encodeURIComponent encodes.`)
  err.code = 'EINVALIDTAGNAME'
  return err
}
```

To make sure the issue wasn't with npx or node I ran `npx create-react-app .` and `yarn start` with no issues

Chriscbr:
Fixes #370


.projen/tasks.json
```diff
@@ -169,7 +169,7 @@
     }
   },
   "env": {
-    "PATH": "$(npx -c 'echo $PATH')"
+    "PATH": "$(node -e \"console.log(process.env.PATH)\")"
```
eladb, line 172: How does this work? The idea here was to add `node_modules/.bin` to the path so installed tools can be found. This seems to just reflect the same path back.

Should it be `npx node -e ...` or something like that?

eladb (changes requested):

src/node-project.ts
```diff
@@ -1225,7 +1225,7 @@ export class NodeProject extends Project {
     // run "projen"
     install.push({
       name: 'Synthesize project files',
-      run: this.projenCommand,
```
eladb, line 1228: This shouldn't be required if the path would have included the node module's bin directory

eladb (changes requested):

.projen/tasks.json
Chriscbr: Yeah, I was been wrestling with this for a bit. I think `npx node -e ...` might work though (on both windows and unix, that is). Thanks for the suggestion!

src/node-project.ts
Chriscbr: undid this change

Chriscbr:
Alright I've tested this change on a windows machine to validate the fix works. The steps were:

- created test node project
- synthesized contents
- validated that "npx projen task / yarn run task" fails
- "yarn link projen" to use the fixed version of projen
- synthesize contents again (env.PATH in tasks.json updates)
- validated that "npx projen task / yarn run task" works now

I'll make note here that projen still has several other issues on Windows that should eventually get looked at. For one, files added to .gitignore get windows-style paths (which is not allowed). And a lot of the unit tests in projen fail. But these were beyond the scope of this PR.

eladb (approved):

PR #380, opened 2020-12-17
fix(awscdk-app-ts): cdkVersion has an @default, make it true and optional

pgollucci:
- chore(release): 0.6.11
- chore: renames `projen.sh` -> `projen.bash` since its bash not sh script
- chore(release): 0.6.12
- chore(release): 0.6.13
- chore(release): 0.6.14
- fix(awscdk-app-ts): cdkVersion has an @default, make it true and optional


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### 0.6.15 (2020-12-17)
```
eladb, line 7: Revert this file please

eladb (changes requested):
This is intentional. Required Properties with @default are automatically populated when a new project is created. 

We want CDK version to be explicit. 

Closing for now.

PR #385, opened 2020-12-21
fix(cli): new --from leads to invalid projenrc

### LINKED ISSUE #384: New external projects lead to invalid .projenrc.js file
When I create a new project from an external npm package, the generated .projenrc.js file does not contain quotes for string values, e.g. cdkVersion and fails

```
/foo/.projenrc.js:4
  cdkVersion: 1.73.0,
                  ^^

SyntaxError: Unexpected number
```

The file looks like this:
```
const project = new TaimosCdkApp({
  cdkVersion: 1.73.0,
  devDeps: ["@taimos/projen"],
  name: foo,
```

How to reproduce:

```
# Optional run in container: docker run -it node:14 /bin/sh
npx projen new --from @taimos/projen taimos-cdk-app --synth false
cat .projenrc.js
```

kcwinner:
Fixes #384

Description:
Modifies the `newProjectFromModule` function to pass default arg props to the `newProject` function. This aligns the module method more closely to the regular cli method and fixes the invalid string values.

I toyed with other alternatives but this *should* help in other regression issues down the line...


src/cli/cmds/new.ts
```diff
@@ -285,6 +285,20 @@ async function newProjectFromModule(baseDir: string, spec: string, args: any) {
     throw new Error(`Project type ${requested} not found. Found ${types.join(',')}`);
   }
 
+  for (const option of type.options ?? []) {
+    if (option.type !== 'string' && option.type !== 'number' && option.type !== 'boolean') {
```
eladb, line 289: Looks like duplicated code... can we reuse?

kcwinner: I looked at extracting to an `isNonPrimitive` function. It causes the command builder to throw an error at because it can't infer that `option.type` is `"string", "number", or "boolean"`
```
cargs.option(option.switch, {
  group: required ? 'Required:' : 'Optional:',
  type: option.type,
  description: desc.join(' '),
  default: defaultValue,
  required,
});
            ```

eladb: Ok. Let's merge this and I'll see what I can do.

eladb (approved):

PR #389, opened 2020-12-21
- fix(node-project): removes now duplicate `node_modules` from `.gitignore`

pgollucci:
- fix(node-project): removes now duplicate `node_modules` from `.gitignore`


Chriscbr (approved):

eladb (commented):
@pgollucci @Chriscbr 

Doesn't feel right. Instead can we just have the GitIgnore class dedup at runtime?

pgollucci:
Sure, but why wouldn't you keep this change ?

eladb:
From a modeling perspective, node_modules is a "node project" concern. I understand that the fact that projen itself uses node "leaks" this concern to all projen projects but i rather have both components add that to gitignore and have it dedupped at that level.

PR #390, opened 2020-12-22
feat: project dependencies

eladb:
All types of projects have dependencies. This is an attempt to create a general model for project dependencies. It's API is available under `project.deps` and is initially modeled roughly against the npm dependency model. The `NodeProject` dependency APIs should remain the same (besides removal of deprecated APIs).

See `docs/deps.md` for more details.

BREAKING CHANGE: all deprecated `dependencies` APIs have been removed.

```
warn - PROP projen.NodeProjectCommonOptions.bundledDependencies: has been removed [removed:projen.NodeProjectCommonOptions.bundledDependencies]
warn - PROP projen.NodeProjectCommonOptions.dependencies: has been removed [removed:projen.NodeProjectCommonOptions.dependencies]
warn - PROP projen.NodeProjectCommonOptions.devDependencies: has been removed [removed:projen.NodeProjectCommonOptions.devDependencies]
warn - PROP projen.NodeProjectCommonOptions.peerDependencies: has been removed [removed:projen.NodeProjectCommonOptions.peerDependencies]
warn - PROP projen.NodeProjectOptions.bundledDependencies: has been removed [removed:projen.NodeProjectOptions.bundledDependencies]
warn - PROP projen.NodeProjectOptions.dependencies: has been removed [removed:projen.NodeProjectOptions.dependencies]
warn - PROP projen.NodeProjectOptions.devDependencies: has been removed [removed:projen.NodeProjectOptions.devDependencies]
warn - PROP projen.NodeProjectOptions.peerDependencies: has been removed [removed:projen.NodeProjectOptions.peerDependencies]
```


Chriscbr (commented):
Trying to model package dependencies in a language-agnostic way seems tricky, but very cool. 😄 

What kinds of features do you have in mind that could use ProjectType? Can we deprecate allowLibraryDependencies, or add some logic  for adding yarn.lock to .gitignore iff it is a library for instance?

Would there be any value in trying to model where each dependency is from? (e.g. whether it's from a public or private package manager/repository, or whether it is a path to the dependency on the local system, or a link to a GitHub repo of the dep directly)

Some package managers like cargo also allow specifying if a dependency is OS-specific, or to be downloaded/installed with subsets of features. But it's probably OK if we don't model those now

eladb:
@Chriscbr some responses:

> What kinds of features do you have in mind that could use ProjectType? Can we deprecate allowLibraryDependencies, or add some logic for adding yarn.lock to .gitignore iff it is a library for instance?

Yes, this is the idea behind projectType. At the moment, `Dependencies` doesn't have too much dependency logic in it (most of it is left in `NodeProject`, but the idea is that we will be able to implement some of these at the generic level.

> Would there be any value in trying to model where each dependency is from? (e.g. whether it's from a public or private package manager/repository, or whether it is a path to the dependency on the local system, or a link to a GitHub repo of the dep directly)

Yes, the first version basically just extracts the dependencies model from `NodeProject` but eventually these are the things that we should be able to model.

> Some package managers like cargo also allow specifying if a dependency is OS-specific, or to be downloaded/installed with subsets of features. But it's probably OK if we don't model those now.

Yes, next steps.

> Maybe logic for specifying ranges of dependency versions (e.g. >= 3.1, < 4.0 in python) should be handled separately / per language project type?

I've been thinking to standardize around the Server syntax and format it to e.g. Python when rendered.

pgollucci:
Is like Yarn 2.x Plug N play .pnp.js ?

PR #391, opened 2020-12-22
feat: DevContainer component

### LINKED ISSUE #362: Feature: GitHub codespaces / devcontainer.json support
See: https://github.com/features/codespaces

AFAIK this similar to Gitpod, just with slight different in features/performance/etc.

GitHub codespaces was also recently featured in the recent GitHub keynote - it gives a quick idea of the use case (see 7:13): https://youtu.be/2m9nUP-e8Co?t=433

Full reference: https://code.visualstudio.com/docs/remote/devcontainerjson-reference

Example configuration:
https://github.com/aws/aws-cdk/blob/master/.devcontainer.json

Chriscbr:
Fixes #362 

BREAKING CHANGES:
- Deprecated `GitpodDocker` for `DevEnvironmentDockerImage`.
- Renamed `gitpod.addCustomDocker()` to `gitpod.addDockerImage`
- Changed `gitpod.addTasks` to take `Task` instances as inputs. Tasks can be created via `const task = project.addTask('name', { ...options })`.

Other changes:

- Creates an abstract interface for configuring container-based developer environments like Gitpod and GitHub Codespaces.
- Adds `addTaskWithOptions`, `addPorts`, `addPrebuilds`, and `addVscodeExtensions` methods to `Gitpod`.
- Adds the `DevContainer` component which emits `.devcontainer.json` files.
- Fix bug where `YAMLFile` can synthesize incorrectly because `JsonFileOptions` get ignored.
- Refactored several unit tests


pgollucci:
Gitpod doesn't actually use Vscode, it uses Thea.

Chriscbr:
> Gitpod doesn't actually use Vscode, it uses Thea.

Thanks for pointing that out, I wasn't aware. I think I can still leave the API name `addVscodeExtensions()` as is since the GitHub page for Theia refers to them that way ("supports VS Code Extension protocol").

Chriscbr:
Just tested that this works with GitHub codespaces.  On a separate branch I enabled `devContainer: true` and added this snippet to `.projenrc.js`, and the task ran & extension was installed.

```ts
const workspaceSetup = project.addTask('workspace:setup');
workspaceSetup.exec('yarn install');
workspaceSetup.spawn(project.buildTask);

project.devContainer.addTasks(workspaceSetup);
project.devContainer.addVscodeExtensions('dbaeumer.vscode-eslint');
```

src/devcontainer.ts
```diff
@@ -0,0 +1,201 @@
...
+   * @example '.gitpod.Docker'
+   */
+  readonly file?: string;
+}
+
+/**
+ * Abstract interface for container-based IDE tools, such as Gitpod and
+ * GitHub Codespaces.
+ */
+export abstract class WorkspaceConfig extends Component {
```
eladb, line 58: I am not sure about this name. First, we usually try to refrain from using `Config` when we model things in the CDK because this is not a model of the configuration, it's a model of the actual workspace. The configuration is the artifact. (this is why the AWS CDK has a `Stack` class instead of a `Template` class.).

How about calling this `DevEnvironment`?

src/devcontainer.ts
```diff
@@ -0,0 +1,201 @@
+import { Component } from './component';
```
eladb, line 1: Can you move `devcontainer.ts` under `vscode/` and only leave the abstract types here?

eladb (changes requested):

smacintyre:
@Chriscbr I just noticed this, looks cool. From the Gitpod side, would it be possible to include [prebuilds](https://www.gitpod.io/docs/prebuilds/#configure-the-github-app)? I didn't realize there was this effort to merge the two and need prebuilds and the VS Code extension features from Gitpod so I had created #417

Chriscbr:
@smacintyre No problem, I've gone ahead and added prebuilds. :)

src/gitpod.ts
```diff
@@ -153,87 +126,291 @@ export interface GitpodTask {
...
+/**
+ * Options for an exposed port on Gitpod
+ */
+export interface GitpodPort {
+  /**
+   * A port that should be exposed (forwarded) from the container.
+   *
+   * @example "8080"
+   */
+  readonly port?: string;
```
eladb, line 177: Why string?

src/project.ts
```diff
@@ -124,6 +131,13 @@ export class Project {
   public readonly gitpod: Gitpod | undefined;
 
   /**
+   * Access for .devcontainer.json (used for GitHub Codespaces)
+   *
+   * This will be `undefined` if devContainer boolean is false
+   */
+  public readonly devContainer: DevContainer | undefined;
```
eladb, line 138: Shouldn't that be part of the `project.GitHub` api?

eladb (approved):
A was expecting to see gitpod and devcontainer at the Project level replaced by the abstract api. I'd like us to slowly move towards a place where the Project API is mostly agnostic to the specific tools people/concrete projects decide to use.

Examples: tasks, deps.

Chriscbr:
@eladb Yeah I'm not totally satisfied with this design... It's hard to choose the right abstraction sometimes. devcontainer.json for instance can be used for GitHub Codespaces, but it also can be used directly with the VS Code "Remote Containers" extension. It probably would be OK under the `project.Vscode` namespace.

But taking a step back, I agree with your broader point. One logical conclusion of that could be creating a `project.devenv` space, similar to `project.GitHub`, which manages _all_ "development environment" options, and put Gitpod and VS Code escape-hatch access inside that (with room for more, e.g. [.editorconfig files](https://editorconfig.org/) and [sublime workspaces](https://www.sublimetext.com/docs/3/projects.html) maybe). And APIs like `addExtensions` or `setIndentSize` would automatically dispatch to those different implementations under the hood. And the task system should play nicely with them as well, in order to generate e.g. `.vscode/tasks.json`.

src/gitpod.ts
Chriscbr: Sorry, I should have made this clearer in the docs. Gitpod allows you to specify individual ports **or** ranges of ports ("3000-3999").

PR #403, opened 2020-12-26
fix: defaults the `new` repository name to `cwd` vs `cwd`/..

pgollucci:
Previous: cwd: /path/org/repo => org
Now:      cwd: /path/org/repo => repo


src/cli/cmds/new.ts
```diff
@@ -437,7 +444,7 @@ async function askAboutGit(): Promise<boolean> {
             type: 'input',
             name: 'gitProjectName',
             message: 'What would you like to name it?',
-            default: path.basename(path.dirname(process.cwd())),
+            default: () => repoName(),
```
eladb, line 447: why lazy?

```suggestion
            default: repoName(),
```

eladb (changes requested):

src/cli/cmds/new.ts
pgollucci: brain fart, IDK

eladb (approved):

PR #411, opened 2020-12-27
chore: projen release workflow broken

### LINKED ISSUE #410: Release workflow is broken

eladb:
Since the integration test snapshots include the synthesized package.json, which includes the projen
version itself as a devDependency, when projen itself is bumped as part of the projen release, the snapshot fails.

To fix this, we replace the version of the `projen` devDependency in the synthesized `package.json` file in with `999.999.999`.

Fixes #410


pgollucci:
belated 👍

PR #413, opened 2020-12-27
chore: dont auto merge core pull requests

pgollucci:
Thus mergify will not auto-merge this.

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -109,6 +109,7 @@ project.github.addMergifyRules({
   },
   conditions: [
     'label=contribution/core',
+    'label!=do-not-merge',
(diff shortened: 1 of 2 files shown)


.projenrc.js
```diff
@@ -109,6 +109,7 @@ project.github.addMergifyRules({
   },
   conditions: [
     'label=contribution/core',
+    'label!=do-not-merge',
```
eladb, line 112: Let's just remove the `Label auto-merge for core` rule from mergify. It's good enough that I can just apply a label so that auto-merge will kick in.

eladb (changes requested):

eladb (approved):

PR #414, opened 2020-12-27
feat: expose all `SampleFile` options for `SampleReadme` in the constructor

### LINKED ISSUE #407: SampleReadme doesn't allow contents to be changed

pgollucci:
This will let things like `p6-projen-project-awesome-list` set their own `readme.md`
`filename` and `contents`.

Closes #407


test/samplereadme.test.ts
```diff
@@ -36,9 +20,10 @@ describe('SampleReadProps', () => {
 
   test('customized w/ default SampleReadme', () => {
```
Chriscbr, line 21: Can we omit this test? I'm not sure what different behavior it is testing from the test 'customized SampleReadme' below

test/samplereadme.test.ts
```diff
@@ -1,28 +1,12 @@
-import * as path from 'path';
```
Chriscbr, line 1: nit: Can we rename this file to `readme.test.ts` to make it clear it corresponds to the `readme.ts` source file? Or rename them `sample-readme(.test).ts`?

Chriscbr (changes requested):

test/samplereadme.test.ts
pgollucci: It proves the default and back compat. That said, I can remove it.

test/samplereadme.test.ts
pgollucci: sure, I stuck with the original name of which I am not the creator

Chriscbr (approved):

Issue #415, opened 2020-12-27
unable to evaluate environment variable PATH=$(npx node -e "console.log(process.env.PATH)"): unknown error


pgollucci:
[](url)

pgollucci:
![init_zsh_—_Untitled__Workspace_](https://user-images.githubusercontent.com/34295/103181750-a9f7e800-4872-11eb-8aec-7fb29feb59c3.png)

pgollucci:
This is a node 15.5.0 issue.

pgollucci:
```shell
for ver in $(nodenv versions --bare); do
echo -n "==[$ver]==>: "; nodenv local $ver; npx node -e "console.log('msg')"; done           
==[10.23.0]==>: msg
==[11.15.0]==>: msg
==[12.20.0]==>: msg
==[13.14.0]==>: msg
==[14.13.1]==>: msg
==[14.15.3]==>: msg
==[15.0.1]==>: msg
==[15.1.0]==>: msg
==[15.2.0]==>: msg
==[15.3.0]==>: msg
==[15.4.0]==>: msg
==[15.5.0]==>: sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `node -e console.log('msg')'
npm ERR! code 2
npm ERR! path /Users/pgollucci/src/github.com/p6m7g8/p6-namer
npm ERR! command failed
npm ERR! command sh -c node -e console.log('msg')

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/pgollucci/.npm/_logs/2020-12-28T00_17_45_457Z-debug.log
```

pgollucci:
See #416

PR #416, opened 2020-12-28
chore: note that node 15.5.0 is broken currently

pgollucci:


Chriscbr:
I think we can avoid restricting the node versions and resolve the bug in #415 instead by changing the line here:

https://github.com/projen/projen/blob/af3162ed18083270ea9e98e49b44b84d401fbd4f/src/node-project.ts#L736

to use `$(npx -c 'node -e "console.log(process.env.PATH)"')` (suggested by @eladb )

eladb:
Superseded by #418

pgollucci:
Works for me. I didn't look into the bug.  Just reported it.

PR #422, opened 2020-12-28
feat: project name

eladb:
The `name` field is now required for any `Project`, and can be read from `project.name`.

Related: #367

BREAKING CHANGE: the `name` option is now required for all project types.


eladb:
@projen rebuild

PR #424, opened 2020-12-29
feat: parallel tasks

eladb:
Allow a step in a task to spawn multiple subtasks in parallel.

Attempting to implement parallel test execution in github workflows.

WIP


src/tasks/parallel-exec.ts
```diff
@@ -0,0 +1,33 @@
+import { spawn } from 'child_process';
+import { TaskRuntime } from './runtime';
+
+const tasks = process.argv.slice(2);
```
Chriscbr, line 4: Maybe add a comment to emphasize first two elements are 'node' and 'runner'

src/jest.ts
```diff
@@ -698,5 +703,36 @@ export class Jest {
...
+        const test = this.project.addTask(`test:${path.basename(file)}`, {
+          exec: `jest ${jestOpts.join(' ')} --no-coverage ${file}`,
+          unlisted: true,
+        });
+
+        tests.push(test);
+      }
+    }
+
+    this.project.testTask.parallel(tests);
```
Chriscbr, line 724: Just curious, doesn't jest already run tests in parallel by default?

src/jest.ts
```diff
@@ -542,6 +544,7 @@ export class Jest {
   private readonly reporters: JestReporter[];
   private readonly jestConfig?: JestConfigOptions;
   private readonly typescriptConfig?: TypescriptConfigOptions;
+  private readonly testMatch: string[];
```
Chriscbr, line 547: nit:
```suggestion
  private readonly testPatterns: string[];
```

package.json
```diff
@@ -20,6 +20,29 @@
     "release": "/bin/bash ./projen.bash release",
     "test:watch": "/bin/bash ./projen.bash test:watch",
     "test:update": "/bin/bash ./projen.bash test:update",
+    "test:deps.test.ts": "/bin/bash ./projen.bash test:deps.test.ts",
```
Chriscbr, line 23: Make sure these are removed

Chriscbr:
Also, thoughts on supporting a fluent interface for tasks? So we could write

```ts
const task = project.addTask('my-workflow').spawn('task1').spawn('task2').say('All done!');
```

src/jest.ts
eladb: Yes it does, but I actually want those tests to run as parallel jobs in the GitHub workflow... Stay tuned. I think that when executed from the command line, we will still just use jest as-is.

eladb:
> Also, thoughts on supporting a fluent interface for tasks?

I am not a huge fan of fluent interfaces because they don't translate well to all languages and I want to keep projen open to supporting other languages in projenrc.

eladb:
Closing for now

PR #431, opened 2020-12-30
feat: make `projen start` interactive by default

### LINKED ISSUE #429: `projen start` should be interactive by default (make --interactive true)
I think it would be nice to have `projen start` be interactive in a way that you could select a task and press enter to run it instead of only showing the options.

skyrpex:
Closes #429


eladb (approved):
🙏🙏

PR #435, opened 2020-12-31
feat: Make IgnoreFile stateful to allow for overrides

abelmokadem:
I faced an issue that did not allow me to override the gitignore file so that it would ignore the projen generated files. This is due to how the IgnoreFile currently works. With the modification I made, we can allow users to override the IgnoreFile anywhere down the road. 

---


Chriscbr:
Thanks for writing up this contribution. :-) I think it would be nice to also have some options for exploring

src/ignore-file.ts
```diff
@@ -15,24 +19,21 @@ export class IgnoreFile extends FileBase {
    * @param patterns Patterns to exclude from git commits.
    */
   public exclude(...patterns: string[]) {
-    this.excludes.push(...patterns);
+    patterns.forEach(pattern => this.patterns[pattern.replace(/^\!/, '')] = IgnorePatternType.EXCLUDE)
```
Chriscbr, line 22: What is the purpose of removing the exclamation marks here?

src/ignore-file.ts
```diff
@@ -1,10 +1,14 @@
 import { FileBase, IResolver } from './file';
 import { Project } from './project';
-import { dedupArray } from './util';
+
+export enum IgnorePatternType {
+    INCLUDE,
+    EXCLUDE
+}
```
Chriscbr, line 7: Since these are technically only used internally, is it possible we could leave this private?

Chriscbr (commented):
Could we also add a unit test to show what happens if you exclude a pattern and then include it (or vice versa)?

src/ignore-file.ts
abelmokadem: Good question. I can imagine that someone would think that they have to use the question mark in order to include a file. I could leave the logic out, but then it's up to the user to not make the mistake.

Without removing the question mark you get the following result:
```
exclude('/my-file.js') // this.patterns = {'/my-file.js': 'EXCLUDE'}
include('!/my-file.js') // this.patterns = {'/my-file.js': 'EXCLUDE', '!/my-file.js': 'INCLUDE'}
```

If you remove the question mark you get:
```
exclude('/my-file.js') // this.patterns = {'/my-file.js': 'EXCLUDE'}
include('!/my-file.js') // this.patterns = {'/my-file.js': 'INCLUDE'}
```

src/ignore-file.ts
abelmokadem: I can make this private by not exporting the enum right?

abelmokadem:
@Chriscbr I have made the enum private and made the test cases more explicit. I also noticed a small flaw in the first test case. The results were already sorted so there was no way to know if the test was failing or not.

abelmokadem:
Test against my own project. I'm now able to ignore files that were generated by projen 😁

Chriscbr (approved):

PR #445, opened 2021-01-04
feat(eslint): devdirs

eladb:
Allow specifying `devdirs` which is a list of directories that should be linted but also allow importing deps from `devDependencies`.

Previously these were hard-coded to `test` and `build-tools`.

---


eladb:
@projen rebuild

PR #446, opened 2021-01-04
fix(node): move typescript options from node project

eladb:
Remove `libdir`, `srcdir` and `testdir` from `NodeProject` and move them to specific projects (typescript, nextjs, react).

BREAKING CHANGE: The `libdir`, `srcdir` and `testdir` options are no longer available in `NodeProject`. Instead they are available in specific project types like typescript, nextjs and react.

---


eladb:
@projen rebuild

Issue #448, opened 2021-01-04
Projen broken on Windows


john-tipper:
The new changes to the npm command cli have broken development on Windows. This is related to #418.

```
$ npx projen build
''node' is not recognized as an internal or external command,
operable program or batch file.
🤖 unable to evaluate environment variable PATH=$(npx -c 'node -e "console.log(process.env.PATH)"'): unknown error
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Chriscbr:
Fixed by #451

PR #451, opened 2021-01-04
fix(cli): broken on Windows

john-tipper:
Fix issue on Windows (#448 ) where single quotes are not recognised when setting PATH via shell command, so use double quotes and escape existing inner double quotes.

---


eladb:
@projen rebuild

john-tipper:
Hi @eladb, please can I ask for some help here? The new CLI format where everything runs through proven (rather than defining `yarn ...` commands) is new to me, so I wasn't sure of the function of `tasks.json`. I thought it was like a template file, so I edited it to put the correct quoting in, but I guess this failure means that I'm not supposed to edit that file - sorry!

eladb (approved):

PR #456, opened 2021-01-04
fix(ignore): "include" entries should always follow "exclude" entries

### LINKED ISSUE #453: projenrc.gitignore entries ordering unexpected, exclamation entries (!) incorrectly changed
Fist off. Love the project, really appreciate the work being done here. 

Version: Projen 6.14.9

I'm evaluating projen for a CDK-Construct Lib project. Due to project requirements I have some .gitignore entries that conflict with the project type defaults. I found the `gitginore` property in the `.projenrc.js` file which I've modified as follows:
```
  gitignore: [
    '!lib/',
    '!OtherFileNotToIgnore*'
]
```

**Adding entries to this array I would naturally expect them to be appended to the END of the .gitignore file, since ordering for .gitignore is relevant.** 
**Currently these entries are injected into a random location in the centre of the .gitignore file.**

It seems like this is cause by the entries being added in the base [node-project.ts](https://github.com/projen/projen/blob/392e458303e255d257268c53c4e9b5022972e0cc/src/node-project.ts#L471):
```    
this.addDefaultGitIgnore();
    if (options.gitignore?.length) {
      for (const i of options.gitignore) {
        this.gitignore.exclude(i);
      }
    }
```

After which other projects that extend node-project e.g. [typescript-project](https://github.com/projen/projen/blob/392e458303e255d257268c53c4e9b5022972e0cc/src/typescript.ts#L252)
attach on their ignores. 

In addition, it seems like the `!` is being stripped from the entries in the projenrc.js.gitignore property due to the [replace pattern](https://github.com/projen/projen/blob/392e458303e255d257268c53c4e9b5022972e0cc/src/ignore-file.ts#L21). This is definitely causing usage behaviour to be extremely confusing. 

Should the IngoreFile.Exclude/Include be called depending if the projenrc.js.gitignore array entries contain an `!` or not explicitly?

- Is my assumption that the entries should be appended at the end of the file wrong?
- Is my assumption that if I add an entry with an exclamation (!) in the projenrc.js.gitignore property it should be passed through to the .gitignore file wrong?

If these assumptions aren't wrong, I'll be open to looking into a PR.

eladb:
Update `IgnoreFile` to always put entries that begin with `!` after "exclude" entries, regardless of whether they've been added with `include()` or `exclude()`. Also introduce `addPatterns()` as a more general API.

The `ignorefile.removePatterns()` API can be used to remove previously ignored entries.

ignore file entries are now also sorted (per category).

Fixes #453
Fixes #456
Superseds #435 

---


Chriscbr (commented):
This is totally reasonable, but I'm wondering per an earlier issue (#435) if projen puts "!/test" in gitignore, then this change is (in some sense?) a regression since there is no way for the user to override this (because the include will always be respected). Is that intended behavior?

eladb:
> This is totally reasonable, but I'm wondering per an earlier issue (#435) if projen puts "!/test" in gitignore, then this change is (in some sense?) a regression since there is no way for the user to override this (because the include will always be respected). Is that intended behavior?

Damn, I haven't seen this one. I am wondering if the right approach is to support explicit overrides. Maybe something like: `gitignore.remove('!/test')`.

WDTY?

Chriscbr:
cc @abelmokadem I'm curious what your use case was / what you were trying to override

`remove` sounds fine to me though

abelmokadem:
That was indeed an alternative implementation that would serve the same purpose. `.remove` would also do the job.

src/__tests__/__snapshots__/integ.test.ts.snap
```diff
@@ -475,74 +475,74 @@ jobs:
...
-.eslintcache
 # Output of 'npm pack'
-*.tgz
+# Runtime data
+# TypeScript cache
 # Yarn Integrity file
-.yarn-integrity
+# jest-junit artifacts
+# nyc test coverage
 # parcel-bundler cache (https://parceljs.org/)
```
Chriscbr, line 491: I guess all of the comments got hoisted up, lol. Are ya sure about this? 😅  I think you can still assume iterating over a set goes in insertion order

eladb: Yak! Let me see what I can do. Good catch.

eladb: I think I'll just filter out the comments from .gitignore. You think everyone will hate me?

Chriscbr: Filtering sounds OK to me.

PR #458, opened 2021-01-04
feat: allow text file to be marked as executable

### LINKED ISSUE #454: Executable File
Is there a way to make an output from `projen` be marked as executable? Specifically, I would like to generate a bash script with my `projen` project and give it `+x`. 

If this doesn't exist and would be useful, I would be happy to try to submit a PR.

jakepearson:
Add optional field to FileBase that will allow it to be marked as executable. It merges with the value of readonly to produce the correct file permissions
Fixes #454

---


smashedtoatoms (approved):

src/file.ts
```diff
@@ -27,6 +27,13 @@ export interface FileBaseOptions {
    * @default true
    */
   readonly readonly?: boolean;
+
+  /**
+   * Whether the generated file should be readonly.
```
Chriscbr, line 32: Needs to be updated

Chriscbr (changes requested):

src/file.ts
jakepearson: Oops.

src/util.ts
```diff
@@ -26,17 +26,45 @@ export function execOrUndefined(command: string, options?: child_process.ExecSyn
...
+  /**
+   * Whether the generated file should be readonly.
+   *
+   * @default true
+   */
+  readonly?: boolean;
+}
+
+export function getFilePermissions(options: WriteFileOptions): string {
+  const readonly = options.readonly ?? true;
```
Chriscbr, line 46: I just realized we need to be careful since the are a few other places where `writeFile` is used in the repo, and they're expecting the file to be writeable afterwards. Could you update this (and the docstring) so the default for `readonly` is false?

pgollucci: Would it make more sense to expose the full `mode` ?

Chriscbr: I'm not against it, but for now this helper method does a good job keeping the chmod details in a single abstraction. We can open it up more later if needed

jakepearson: @Chriscbr, I updated the default value for `readonly` as you suggested. Thanks for looking at this.

Chriscbr (approved):

PR #459, opened 2021-01-04
feat(cli): improved bootstrapping

eladb:
Looking towards supporting multiple languages for projenrc, we need to give more control to the project classes (over the cli) when it comes to bootstrapping new projects.

This changes how `new` is implemented such that synthesis is performed by calling `.synth()` on a project object, and not by using the CLI's synth code path, which is tailored for the javascript version of .projenrc.

For example, it will now be possible to implement a `JavaProject` which will synthesize a java directory structure and a manifest that will take a java dependency on "projen" from maven. It will also be configured to read the projenrc from a java file within the source tree.

This solves the bootstrapping problem for compiled languages: if projenrc need to be compiled in order to be synthesized, we will need the full toolchain and its configuration in order to compile and execute it. If we want to be able to use any (supported) language we want for projenrc, we need to first synthesize the toolchain environment, and that's exactly what projects are doing.

---

### DIFF
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -1,5 +1,6 @@
 import * as os from 'os';
 import * as path from 'path';
+import * as vm from 'vm';
 import * as fs from 'fs-extra';
 import * as inquirer from 'inquirer';
 import * as yargs from 'yargs';
@@ -8,7 +9,6 @@ import * as inventory from '../../inventory';
 import * as logging from '../../logging';
 import { exec, execOrUndefined } from '../../util';
 import { tryProcessMacro } from '../macros';
-import { synth } from '../synth';
 
 class Command implements yargs.CommandModule {
   public readonly command = 'new [PROJECT-TYPE] [OPTIONS]';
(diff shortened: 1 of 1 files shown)


src/cli/cmds/new.ts
```diff
@@ -87,35 +87,82 @@ class Command implements yargs.CommandModule {
...
+
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const module = require(mod);
+  const ctx = vm.createContext(module);
+
+  process.env.PROJEN_DISABLE_POST = (!opts.post).toString();
+  vm.runInContext([
+    newProjectCode,
+    opts.synth ? 'project.synth();' : '',
+  ].join('\n'), ctx);
```
Chriscbr, line 145: Does this interoperate with stdout/stderr, i.e. will things still be logged?

src/cli/cmds/new.ts
```diff
@@ -87,35 +87,82 @@ class Command implements yargs.CommandModule {
...
+  comments: boolean;
+
+  /**
+   * Should we call `project.synth()` or instantiate the project (could still
+   * have side-effects) and render the .projenrc file.
+   */
+  synth: boolean;
+
+  /**
+   * Should we execuite post synthesis hooks? (usually package manager install).
```
Chriscbr, line 118:
```suggestion
   * Should we execute post synthesis hooks? (usually package manager install).
```

src/cli/cmds/new.ts
eladb: Yes I don't see a reason why not

PR #461, opened 2021-01-05
feat(jest): coverage text output and watch ignore patterns

eladb:
Allow disabling the text output option for jest coverage to improve test output readability. This is okay in case coverage is collected and calculated through some other form.

This change also adds a "watch ignore pattern" for `src/**` when tests are executed from `lib/`.

Enable this for projen.

---


eladb:
@projen rebuild

PR #462, opened 2021-01-05
feat: do not produce empty `deps.json` and `tasks.json` files

eladb:
To support this, if `FileBase.synthesizeContent()` returns `undefined`, the file write is skipped.

---


eladb:
@projen rebuild

PR #465, opened 2021-01-06
fix(java): corrupted java directory name

thomasklinger1234:
## Summary

This PR removes the usage of `...` causing the package name to be split into its characters when being processed by `path.join`. This would create `src/test/java/o/r/g/a/c/m/e` from `org.acme` in the sample package. 

Screenshot: 

Screenshot 2021-01-06 at 20 33 15

---


eladb (approved):
oops Thanks!

eladb:
@thomasklinger1234 changed the "fix" title to describe the bug and not the solution (this goes to the changelog).

PR #470, opened 2021-01-07
feat(java): projenrc.java

eladb:
If `projenrcJava` is set to `true` for java projects (or `--projenrc-java` switch is provided in `projen new`, the java project will support writing your projenrc file in java (yeah!).

In this case, the projenrc file will be located under `src/test/java/projenrc.java` and `projen synth` will take care to compile it and execute it with a dependency on `com.github.eladb/projen`.

Next steps: 

1. The CLI should auto-detect the `synth` task and defer to it when `pj` is executed without a subcommand. This will allow projects to customize how synth is performed for each language.
2. Currently `projen new` will still generate a `.projenrc.js` file, but this should be moved from the CLI to the library and then `java.Projenrc` should generate the Java equivalent in the right location when the project is bootstrapped. That should be fun.

---


docs/java.md
```diff
@@ -119,6 +119,53 @@ Where `groupID` and `artifactId` are the Maven coordinates and `version` is the
...
+        project.synth();
+    }
+}
+```
+
+In order to synthesize, run: `pj synth`, which will compile your test code and
+execute this program.
+
+By default, `projenrc.java` is placed under the `test` scope (and
+`com.github.eladb/projen` test dependency is added). This ensures that
```
hoegertn, line 165: Above it says `org.projen/projen`

eladb: The group id is `com.github.eladb` but the java package is `org.projen`. I was just lazy to create a new group ID for projen at the moment, but we can do that later.

PR #473, opened 2021-01-08
fix(eslint): lintProjenRc not working with compiled jest tests

### LINKED ISSUE #472: lintProjenRc fails when compiled tests are enabled
Currently enabling lintProjenRc doesn't work when both of these conditions are true in a TypeScript project:

- compiled tests are enabled (e.g.by changing `testdir` to a subdirectory of the src directory)
- tsconfig is not managed by projen (e.g. it's generated by JSII, so `disableTsconfig` is enabled)

The root cause is that in this scenario, eslint expects the default `tsconfig.json` to include .projenrc.js in the "include" section, but it is not included (because `tsconfig.json` is not created by projen / it is created by jsii). Normally .projenrc.js would just be added to `tsconfig.jest.json`, but compiled tests causes this file to not be created/synthesized.

Chriscbr:
Fixes #472

- Changes eslint's default behavior in TypeScript projects so that by default it creates its own tsconfig file, independent of the one used by jest (`tsconfig.jest.json`).
- Exposes `eslintOptions` on JsiiProjectOptions since JsiiProjectOptions doesn't extend TypescriptProjectOptions for some reason. (?)
- Deletes tsconfig.jest.ts from projen since it is not currently being generated/used.

---


.projenrc.js
```diff
@@ -49,18 +49,22 @@ const project = new JsiiProject({
 
   // makes it very hard to iterate with jest --watch
   jestOptions: {
-    coverageText: false
+    coverageText: false,
+  },
+
+  eslintOptions: {
+    lintProjenRc: true,
```
eladb, line 56: Shouldn't that be the default?

src/eslint.ts
```diff
@@ -309,4 +310,31 @@ export class Eslint extends Component {
...
     this.ignorePatterns.push(pattern);
   }
+
+  /**
+   * Generates a custom tsconfig for eslint using passed in tsconfig options.
+   * Files must satisfy an include pattern (and not satisfy any exclude patterns)
+   * in order for them to be linted.
+   * @param options TypescriptConfigOptions
+   */
+  public generateTypescriptConfig(options: TypescriptConfigOptions) {
```
eladb, line 320: Looking at this code I don't really understand what additional logic is added here compared to just `new TypescriptConfig()`

src/typescript.ts
```diff
@@ -341,13 +336,31 @@ export class TypeScriptProject extends NodeProject {
...
         devdirs: [this.testdir, 'build-tools'],
         lintProjenRc: compiledTests ? false : true,
         fileExtensions: ['.ts', '.tsx'],
         ...options.eslintOptions,
       });
+
+      // eslint will only lint files that are included in `tsconfig.include`.
+      // if tsconfig is disabled (for example if it is generated by jsii) we cannot
+      // modify the include/exclude items, so we create a separate tsconfig instead
+      if (options.disableTsconfig) {
```
eladb, line 350: Does this mean that if we use jsii then projenrc is not linted?

eladb (changes requested):

src/eslint.ts
Chriscbr: Ah you're right -- I originally wrote it this way for parity with a similar method in the `Jest` class, but that's not really needed here. Will fix.

On one hand I'd like to refactor this further so that, say, `tsconfig.jest.json` is managed by `Jest` and `tsconfig.eslint.json` is managed by `Eslint` (so that changes to "include"/"exclude" options are kept in sync easier). But this breaks OOP principles by leaking a lot of Typescript logic elsewhere, so I'd have to think about it more.

.projenrc.js
Chriscbr: It is usually the default, although there was a line of code added that makes the default "false" if compiled tests are being used (I'm not totally sure why)

https://github.com/projen/projen/blame/6e17704cf0b49c6d40a8913ddf2072e498b03574/src/typescript.ts#L347

I'll remove that line, can easily revert if something comes to your mind

src/typescript.ts
Chriscbr: As long as the eslintrc's `parserOptions.project` refers to the tsconfig generated by jsii, yes -- it will not get linted. But we can change it to refer to any tsconfig file we want. (Hence in this PR I create a new tsconfig file for it to refer to).

[Before projen v0.13.0-ish, eslint was getting synthesized with `parserOptions.project: "tsconfig.jest.json"`, so this was not an issue. But since we moved projen tests to `src/__tests__`, it's now synthesized with `parserOptions.project: "tsconfig.json"` by default, hence the bug.]

eladb (approved):

eladb:
@projen rebuild

src/typescript.ts
```diff
@@ -341,13 +336,31 @@ export class TypeScriptProject extends NodeProject {
...
-        tsconfigPath: `./${eslintTsConfig}`,
+        tsconfigPath: options.disableTsconfig ? './tsconfig.eslint.json' : './tsconfig.json',
         dirs: [this.srcdir],
         devdirs: [this.testdir, 'build-tools'],
-        lintProjenRc: compiledTests ? false : true,
         fileExtensions: ['.ts', '.tsx'],
         ...options.eslintOptions,
       });
+
+      // eslint only lints files that are included in the tsconfig it refers to.
```
eladb, line 346:
```suggestion
      // eslint only lints files included in the tsconfig it refers to.
```

eladb (approved):

PR #476, opened 2021-01-08
Fix bug in jest collectCoverageFrom type

### LINKED ISSUE #474: (bug) jest collectCoverageFrom is boolean not string[]
I believe this should this be a `string[]` ?

https://github.com/projen/projen/blob/30047506f5fd1a60bccc507a64acc03fd82fbb14/src/jest.ts#L48-L52

bigkraig:
closes #474

---


Chriscbr (approved):
Thanks for the bug fix!

@eladb Is it OK to exclude this as a breaking change? This API/interface isn't even valid as a boolean here

eladb:
> 
> @eladb Is it OK to exclude this as a breaking change? This API/interface isn't even valid as a boolean here
> 
> 
> 
Yes. If the fix breaks an api that could have never worked, it's not a breaking change because you wouldn't break anyone by definition.

PR #477, opened 2021-01-08
fix: bump task fails on subprojects

bigkraig:
When running `yarn bump` on a subproject, standard-version executes projen in the subproject path. As the .projenrc.js only exists on the root project, this fails. This patch changes to the relative path of the root project before running projen.

---


src/version.ts
```diff
@@ -48,14 +49,18 @@ export class Version extends Component {
     project.npmignore?.exclude('/.versionrc.json');
     project.gitignore.include(VERSION_FILE);
 
+    let projenCommand = project.package.projenCommand;
+    if (project.outdir !== project.root.outdir) {
```
pgollucci, line 53: I think project.parent != null might be better

bigkraig: done 👍🏽

src/version.ts
```diff
@@ -48,14 +49,18 @@ export class Version extends Component {
     project.npmignore?.exclude('/.versionrc.json');
     project.gitignore.include(VERSION_FILE);
 
+    let projenCommand = project.package.projenCommand;
+    if (project.parent != null) {
```
eladb, line 53:
```suggestion
    if (project.parent) {
```

eladb (approved):

PR #486, opened 2021-01-13
feat: add line endings to files

skyrpex:
Adds a line ending to all generated files (text files, ignore files, JSON files, etc). Without those, git diffs may look strange, linting programs may complain, and other tools like npm or yarn may cause inconsistencies.

---


src/util.ts
```diff
@@ -62,7 +62,7 @@ export function writeFile(filePath: string, data: any, options: WriteFileOptions
   }
 
   fs.mkdirpSync(path.dirname(filePath));
-  fs.writeFileSync(filePath, data);
+  fs.writeFileSync(filePath, `${data}\n`);
```
eladb, line 65: Technically `data` could be something other than a string. Maybe add the new line only if it is a string?

eladb (changes requested):

src/util.ts
skyrpex: Makes sense!

eladb (approved):

skyrpex:
@eladb I noticed that YAML files have now two line endings. Should we fix that before merging?

eladb:
> @eladb I noticed that YAML files have now two line endings. Should we fix that before merging?

Yes

src/__tests__/ignore-file.test.ts
```diff
@@ -163,5 +163,5 @@ test('comments are filtered out', () => {
 
 // parses file contents without 'Generated by...' spiel
 function splitAndIgnoreMarker(fileContents: string) {
-  return fileContents.split('\n').slice(1);
+  return fileContents.split('\n').slice(1, -1);
```
eladb, line 166: Would it make more sense to just filter out empty lines instead of implicitly removing the last line (which feels error-prone)?

src/github/workflows.ts
```diff
@@ -35,7 +35,8 @@ export class GithubWorkflow extends FileBase {
 
     return [
       `# ${FileBase.PROJEN_MARKER}`,
-      YAML.stringify(workflow),
+      // remove the last character, which is an EOL
+      YAML.stringify(workflow).slice(0, -1),
```
eladb, line 39: I don't like this. It's too implicit and there's a good chance we will trip on this in the future. Perhaps each file type can add this newline explicitly instead of implementing it in `writeFile()`. 

What do you think?

skyrpex: Yeah, I agree. That's the best way to implement to proceed. Will amend later, though!

skyrpex:
This ended up being much cleaner that way.

Issue #452, opened 2021-01-04
Reliance on bash `rm` command breaks Windows


john-tipper:
The use of `rm -fr lib/` as the first step within the `test` script of generated projects breaks on Windows. Whilst it is possible to work around this with some hackery (e.g. defining a different test script and using that), it would be nice to strip out the OS dependency.

I think the issue is here: 

https://github.com/projen/projen/blob/e9ca1aa5d92fea64e16e4be8bc6b4a9fd3d2a662/src/typescript.ts#L306

We could do something like use https://github.com/shelljs/shx as a way of using portable Shell commands (I'm sure others may think of other options).  Are there any objections to doing this please? If so, does the library get declared as a dependency of Projen, or a peer dependency?

I'm happy to submit a PR, but I want to make sure others are happy with this approach before I do.

Chriscbr:
I'm not familiar with the portable shell command ecosystem, but the idea sounds reasonable to me.

> If so, does the library get declared as a dependency of Projen, or a peer dependency?

I think shx would be added as a dev dependency because it's only used for running development scripts/tasks (I think this logic applies both for projects using projen, and for projen itself)? But I'm not totally sure.

rpawlaszek:
I have the same issue on Windows with rm -fr. From my side the proposition is to use [rimraf](https://www.npmjs.com/package/rimraf) and have it as a global dependency for projen.

john-tipper:
Yes, the reason I went with [shx](https://www.npmjs.com/package/shx) in #540 was because there are other breaking commands other than `rm` that are used with Projen, such as `mv` and `mkdir`.

pgollucci:
I think you need it in regular deps for when external projects happen.

eladb:
I believe this is resolved by #781

PR #487, opened 2021-01-13
fix: unit tests failing on Windows

Chriscbr:
There was around 30+ unit tests failing on Windows due to:

- line ending differences between windows and posix
- path separator differences between windows and posix
- echo "foo" would include the quotes in stdout in windows
- paths not being quoted in `exec` calls, causing paths with spaces to get messed up

There are still some tests failing due for other reasons that need to be resolved in follow up pull requests, e.g.: tests that refer to $PWD or the pwd command (some potential fixes are discussed in #452).

Also snapshot tests seem to get deleted after the tests run (not entirely sure why), so I wasn't able to verify if these were getting generated with differences or not.

---


eladb (approved):

john-tipper:
Is this still a draft PR or is it good to go please?

Chriscbr:
> Is this still a draft PR or is it good to go please?

This might still be valid, but it was written with a much older version of projen. I'll try update it this weekend.

Chriscbr:
@projen rebuild

Chriscbr:
For whatever reason Windows doesn't like the way we were changing directories before each unit test in `java-project.test.ts` (all of them failed like the screenshot) so in this last commit I just changed the fixture to use the 'outdir' field.

Screenshot 2021-02-26 011929

PR #493, opened 2021-01-15
docs: add RFC for component levels

Chriscbr:
[Rendered page](https://github.com/Chriscbr/projen/blob/docs/rfc-component-levels/rfcs/001-component-levels.md)

I'm not actually looking to merge this in, but I thought it made more sense to post this as a pull request than an issue since it's easier to comment on specific sections / keep track of changes with the GitHub PR interface.

This is just a high level sketch of the first direction I thought of for improving how components compose in projen and improving the escape hatch situation. I'm not particularly committed to anything in the proposal, so feel free to post any thoughts or criticisms (especially high level stuff)! 🙌 

Credit to @eladb for seeding several of these ideas.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #495, opened 2021-01-16
feat: standard escape hatches

Chriscbr:
Adds standard escape hatches for all object-based files (JSON, YAML, TOML) in the form of `addOverride(path, value)` and `addDeletionOverride(path)`. All you need to do is directly access the file, for example:

```ts
project.tryFindObjectFile('.github/workflows/release.yml').addOverride('jobs.build.env', {
  CI: true,
  myKey: 'myValue',
});
```

Most of the code for `addOverride` and its test cases were adopted from the aws-cdk project [here](https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/core/lib/cfn-resource.ts).

`GithubWorkflow` had to be refactored in order for it to work with the new escape hatch methods (since they all exist in the `ObjectFile` class).

Also I've added the capability to include "generated by projen" markers for XML files.

BREAKING CHANGE: `GithubWorkflow` no longer extends `FileBase`. Use `workflow.file` instead to access file properties.
* **project** `project.tryFindJsonFile()` is now deprecated. Use `tryFindObjectFile()` instead.
* **project** The `obj` field on files is now private. Use `file.addOverride()` to make changes instead.

---


Chriscbr:
Though I initially included this in an RFC yesterday (#493), I think it's ultimately independent from the multi-tiered components concept, so it can probably be extracted out of the RFC and implemented sooner since we can already get immediate benefits. I also feel like this API is pretty robust since it was already done by the CDK / the parallels are more similar. But we can definitely put this on hold if we want to consider other kinds of escape hatch mechanisms.

I think one direction that would be natural after this PR is to begin deprecating raw object fields from being public (e.g. `public manifest: any` or `public config: any` in the `NodePackage` and `Jest` components respectively), and instead provide direct access to the underlying file, through which these escape hatches can be accessed.

Chriscbr:
@projen rebuild

src/object-file.ts
```diff
@@ -30,6 +30,11 @@ export abstract class ObjectFile extends FileBase {
    */
   public readonly obj: object;
```
eladb, line 31: Can we make this `private`? This is eventually our goal. Otherwise `package.maniest.obj` is a leak.

src/object-file.ts
```diff
@@ -53,6 +124,71 @@ export abstract class ObjectFile extends FileBase {
...
+
     return resolved ? JSON.stringify(resolved, undefined, 2) : undefined;
   }
 }
+
+/**
+ * Merges `source` into `target`, overriding any existing values.
+ * `undefined`s will cause a value to be deleted.
+ */
+function deepMerge(target: any, ...sources: any[]) {
```
eladb, line 139: Don't we have that function in `util` or somewhere?

src/project.ts
```diff
@@ -287,6 +289,40 @@ export class Project {
     return file;
   }
 
+  /**
+   * Finds a yaml file by name.
+   * @param filePath The file path.
+   */
+  public tryFindYamlFile(filePath: string): YamlFile | undefined {
```
eladb, line 296: Why do we need to distinguish between file types here? Can we just offer `tryFindObjectFile`?

eladb (changes requested):
Please add documentation under `docs/`

src/object-file.ts
Chriscbr: Yes, although there is subtle difference between the two versions. The `deepMerge(base, override)` in `util` is purely constructive, so when it's given `{ key: undefined }` in `override`, it essentially says "there is nothing to merge, so I'll keep the existing value of `key` from `base`".

But the deepMerge I am adding is destructive (in order to support `addDeletionOverride`), so when it sees `{ key: undefined }` in `override` it essentially says "this key should be deleted from `base`". (And if the object is empty as a result, the object will be deleted). For example:

```ts
deepMerge({ a: "1" }, { a: undefined }) // interpreted as merging the right obj onto the left
```

produces `{ a: "1" }` for the constructive version, and `{}`  for destructive version.

----

I did consider just relying on this destructive version everywhere. But with the way options currently pass through constructors, using this implementation would lead to confusing behavior for users. For example, consider passing in this options to `ReactTypescriptProject`...

```
tsconfig: {
  compilerOptions: {
    module: undefined,
  }
},   
```

tsconfig.compilerOptions.module defaults to "CommonJS" in `TypeScriptProject`, but `ReactTypescriptProject` overrides this with "esnext". So when this user option gets deepMerge'd in the first constructor, the "module" field gets removed (kind of like how `addOverride` operates with `undefined` values); but when the parent class constructor gets called, the "CommonJS" default gets [added back](https://github.com/projen/projen/blob/ea15a203aaec945ea7edb346a2c333c5e787d8d2/src/typescript.ts#L249-L250)! So the user setting it to "undefined" changed the value from "esnext" to "CommonJS" (!?)

There isn't any simple way to fix this independently, so it makes sense to keep the prior behavior (where the user providing `undefined` will set the field to the correct default value, which here would be "esnext"), and leave the two versions of `deepMerge` separate.

src/project.ts
Chriscbr: Sure. I think for consistency it makes sense to deprecate tryFindJsonFile then.

src/object-file.ts
eladb: Let's add an option to the util one that enables the destructive behavior?

src/project.ts
eladb: 👍

src/object-file.ts
Chriscbr: OK. I made non-destructive the default behavior since that's what's wanted in most other places.

src/object-file.ts
Chriscbr: Done.

src/object-file.ts
```diff
@@ -23,12 +25,23 @@ export interface ObjectFileOptions extends FileBaseOptions {
...
  * Represents an Object file.
  */
-export abstract class ObjectFile extends FileBase {
+export abstract class ObjectFile extends FileBase implements IMarkableFile {
   /**
    * The output object. This object can be mutated until the project is
    * synthesized.
    */
-  public readonly obj: object;
+  private readonly obj: object;
```
eladb, line 33: Announce as a breaking change

eladb (approved):
Looks great!

Chriscbr:
OK, there's a small issue where _if_ you enable the "marker" option for pom.xml (currently off by default - I might as well change that right?), the build fails, because pom.xml doesn't allow a \ comment at the top, for whatever reason. It looks like inserting it at the end fixes it though.

src/object-file.ts
Chriscbr: Done.

Chriscbr:
@eladb Build passes, but needs manual merge due to changes to github workflows

Issue #499, opened 2021-01-18
react-ts: Error: EACCES: permission denied, open '/path/to/tsconfig.json'


flyingImer:
This happens when at least `react-scripts build` and `react-scripts start`. It is because when running these commands, they are updating `tsconfig.json` field `jsx` from the projen generated `react` to `react-jsx`, but the `tsconfig.json` file is set to read-only.

After manually changing `tsconfig.json` with write permission e.g. `chmod 600 tsconfig.json`, it starts working.

- [x]  I, the requester, plan to submit a PR to fix it.

PR #500, opened 2021-01-18
fix(react-ts): `react-scripts build` altering `tsconfig.json` for React 17 conflicts anit-tamper

flyingImer:
fixes: #499

Tl;dr
Commands like `react-scripts build` or `react-scripts start` will fail due to this file being immutable. 

Longer version
These commands updates `tsconfig.json` field `jsx` from `react` to `react-jsx`, but currently the generated `tsconfig.json` is read-only, as a result, fails the workflow.

---


Chriscbr (approved):
@eladb LGTM

eladb (changes requested):
1. Please update PR title to reflect the bug that is fixed.
2. Do we expect `react-scripts build` to actually mutate the file? Because this will cause anti-tamper to fail.

Chriscbr:
> 2\. Do we expect `react-scripts build` to actually mutate the file? Because this will cause anti-tamper to fail.

Yeah, it does mutate tsconfig in order to set up certain default and certain required options. This is the script that gets run:

https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/utils/verifyTypeScriptSetup.js

But I think this should not cause problems as long as the user commits any changes that it makes to the tsconfig (i.e. by running `react-scripts build` once locally), since at the end of the script, it only writes to the file if it calculates there are differences. That's what I expect, anyways. :-)

flyingImer:
> 2. Do we expect `react-scripts build` to actually mutate the file? Because this will cause anti-tamper to fail.

@eladb that is a problem. Not sure what's the proper way to get around. Any thought?

Chriscbr:
@eladb oh wait, I see what you mean 😅 . Even if the changes to tsconfig are commit, the github workflow would rerun `npx projen`, and the changes applied by react-scripts would get overwritten, causing anti-tamper to fail - I forgot about that.

For now, I think the best fix is just to apply the changes that are being suggested by the CLI:

> compilerOptions.jsx must be react-jsx (to support the new JSX transform in React 17)

To avoid issues like this again (since tsconfig is basically a living standard), I think we'd have to add some more general mechanisms in projen to handle other tools modifying projen-generated files, or just mark the tsconfig file as only-generate-once for these project types.

@flyingImer : Could you update this PR to add the new jsx options described in the [tsconfig handbook](https://www.typescriptlang.org/tsconfig#jsx) to the corresponding enum [here](https://github.com/projen/projen/blob/82d03122800524c91c872a6e5e229a2683bf4789/src/typescript.ts#L433), and update react.ts [here](https://github.com/projen/projen/blob/82d03122800524c91c872a6e5e229a2683bf4789/src/web/react.ts#L101) to use the `react-jsx` option? That should stop react-scripts from making changes to the default tsconfig for now.

flyingImer:
@Chriscbr yes, I was about to propose the same as a mid-term solution. Will do.

flyingImer:
@Chriscbr @eladb new commits coming in, what do you think?

flyingImer:
@projen rebuild

eladb (approved):

PR #505, opened 2021-01-18
feat(java): generate projenrc in java

eladb:
From now, java projects will always generate a projenrc file in java based on the options passed to `projen new` (and default values for mandatory arguments).

To allow non-javascript projects to be synthesized idiomatically, the CLI will execute a task named `synth` if it exists instead of the default javascript synthesis flow. In a subsequent change, all project types will always include a `synth` task and this will be the default way for projects to be synthesized.

---


eladb:
@projen rebuild

src/java/java-project.ts
```diff
@@ -137,8 +137,11 @@ export class JavaProject extends Project {
     this.distdir = options.distdir ?? 'dist/java';
     this.pom = new Pom(this, options);
 
-    if (options.projenrcJava ?? false) {
-      this.projenrc = new ProjenrcJava(this, this.pom, options.projenrcJavaOptions);
+    if (options.projenrcJava ?? true) {
+      this.projenrc = new ProjenrcJava(this, this.pom, {
+        ...options.projenrcJavaOptions,
+        initializationOptions: options,
+      });
```
Chriscbr, line 144: It's like ProjenrcOptions and ProjectOptions are inverses of each other haha 😄

Chriscbr (commented):
I'm starting to see a little of what you mean by code generation. I feel like some of this code should really be handled in a separate module/package, or by something like jsii-rosetta (although it might not be compatible enough for our use case). 

I wonder if there's any way we could create a test fixture/mock/thing that mimics an external module to simulate that functionality of our library for integration tests.

Also, we should make sure the "synth" task in `awscdk-app-ts` gets renamed so it doesn't break when adding synth tasks to non-java project types.

eladb:
> I'm starting to see a little of what you mean by code generation. I feel like some of this code should really be handled in a separate module/package, or by something like jsii-rosetta (although it might not be compatible enough for our use case). 
> 

Yes I agree. I wanted to land this and then refactor. Perhaps we can use `jsii-rosetta` here. I'll explore this.

> 
> 
> I wonder if there's any way we could create a test fixture/mock/thing that mimics an external module to simulate that functionality of our library for integration tests.
> 
> 
Not sure I completely follow. 

> 
> Also, we should make sure the "synth" task in `awscdk-app-ts` gets renamed so it doesn't break when adding synth tasks to non-java project types.

Yes that's a good point. Although one may argue there is philosophically no difference between projen synth and CDK synth. Maybe they are the same thing? (Mu ha ha ha ha)

src/cli/index.ts
```diff
@@ -10,7 +11,9 @@ async function main() {
   const ya = yargs;
   ya.commandDir('cmds');
 
-  discoverTaskCommands(ya);
+  const workdir = '.';
+  const runtime = new TaskRuntime(workdir);
+  discoverTaskCommands(runtime, ya);
```
eladb, line 16:
```suggestion
  const runtime = new TaskRuntime('.');
  discoverTaskCommands(runtime, ya);
```

eladb:
@projen rebuild

Chriscbr:
> Not sure I completely follow.

Ah, I was just noticing that in projen code pertaining to project-bootstrapping or the CLI (not necessarily in this PR), I think there are some code paths that only apply when using external modules. (i.e. what if the user synthesizes a new project with the '--from' flag). I wasn't sure if we are currently testing those much.

eladb:
@projen rebuild

eladb:
> Ah, I was just noticing that in projen code pertaining to project-bootstrapping or the CLI (not necessarily in this PR), I think there are some code paths that only apply when using external modules. (i.e. what if the user synthesizes a new project with the '--from' flag). I wasn't sure if we are currently testing those much.

Yes, likely requires some more testing to validate that I am doing the right thing with external modules here.

PR #506, opened 2021-01-19
fix(eslint): task fails due to test directory missing from config

Chriscbr:
Fixes a regression causing the eslint task to fail in some typescript projects.

---


src/typescript.ts
```diff
@@ -336,31 +336,25 @@ export class TypeScriptProject extends NodeProject {
...
-            PROJEN_RC,
-            `${this.srcdir}/**/*.ts`,
-            `${this.testdir}/**/*.ts`,
-          ],
-          exclude: [
-            'node_modules',
-          ],
-          compilerOptions: compilerOptionDefaults,
-        });
-      }
```
eladb, line 363: I like it better. Less conditions!

eladb (approved):

PR #510, opened 2021-01-20
feat(node): npm publish configuration

### LINKED ISSUE #192: Specifying 'publishConfig' in package.json
Use case: Building AWS CDK JSII construct and publishing to company internal Artifactory repository.

It would be good to have a possibility to specify this in `package.json`:

```json
  "publishConfig": {
    "registry": "https://company.com/api/npm/company-local/"
  },
```

eladb:
Renders `publishConfig` inside `package.json` based on `npmAccess` (new), `npmDistTag` and `npmRegistryUrl`. The `npmRegistry` option is deprecated (but still supported).

Supersedes #317
Resolves #192

---


eladb:
@henrysachs  @gradybarrett, could you please take a look to make sure I didn't miss anything?

src/node-package.ts
```diff
@@ -286,28 +339,55 @@ export class NodePackage extends Component {
...
+
+  /**
+   * npm registry (e.g. `https://registry.npmjs.org`). Use `npmRegistryHost` to get just the host name.
+   */
+  public readonly npmRegistryUrl: string;
+
+  /**
+   * The npm registry host (e.g. `registry.npmjs.org`).
+   */
+  public readonly npmRegistry: string;
```
gradybarrett, line 355: Is referring to just the registry host as npmRegistry a convention thing? If not, any specific reason this isn't more specific (npmRegistryHost)?

src/node-package.ts
```diff
@@ -739,6 +848,15 @@ export class NodePackage extends Component {
     writeFile(root, JSON.stringify(pkg, undefined, 2));
   }
 
+  private renderPublishConfig() {
+    // omit values if they are the same as the npm defaults
+    return resolveJson({
+      registry: this.npmRegistryUrl !== DEFAULT_NPM_REGISTRY_URL ? this.npmRegistryUrl : undefined,
```
gradybarrett, line 854: If the options are ultimately omitted, then why the defaults?

gradybarrett (approved):
Awesome! 🎉

Left a couple non-blocking questions. Otherwise LGTM!

gradybarrett:
> Left a ~~couple~~ non-blocking question~~s~~. Otherwise LGTM!

😆

src/node-package.ts
Chriscbr: I think this is just for backwards compatability, since `options.npmRegistry` is now being parsed by `NodePackage` instead of `NodeProject`.

eladb: Yes that's the reason

PR #513, opened 2021-01-21
feat(jsii): Golang support (without publishing yet)

eladb:
Adds `publishToGo` option to jsii projects which renders the `go` target in the jsii config. This will cause the build output to include a `dist/go` artifact that can then be published to a git repository.

Once `jsii-release` supports go, we need to add the relevant workflow steps to actually publish.

---


eladb:
@projen rebuild

PR #514, opened 2021-01-24
feat: python projects

### LINKED ISSUE #359: RFC: Support for Python project
Hi,

I would love to have a Python project and a Python CDK app project as types for projen.

We should also look at https://github.com/aws/aws-cdk/issues/9135 for the file structure.

Chriscbr:
Closes #359 

Introduces a new python project type which can be created via `npx projen new python`. See docs/python.md for more information.

---


src/__tests__/__snapshots__/new.test.ts.snap
```diff
@@ -813,6 +813,39 @@ project.synth();
 "
 `;
 
+exports[`projen new python 1`] = `
+"const { python } = require('projen');
+
+const project = new python.PythonProject({
+  jsiiFqn: \\"projen.python.PythonProject\\",
+  name: 'my-project',
+  pythonPath: '/Users/rybickic/.pyenv/shims/python',
```
Chriscbr, line 822: Need to fix this. That said, I _would_ like to keep the behavior of filling in the pythonPath automatically at runtime

src/python/requirements-file.ts
```diff
@@ -0,0 +1,62 @@
...
+import { FileBase, IResolver } from '../file';
+import { Project } from '../project';
+import { toPythonVersionRange } from '../util/semver';
+
+export interface RequirementsFileOptions {
+  /**
+   * Accepts a function that resolves to an list of packages that should get included.
+   * @internal
+   */
+  readonly _lazyPackages?: any;
```
Chriscbr, line 11: Trying to think of a better solution for this

src/python/python-project.ts
```diff
@@ -0,0 +1,268 @@
...
+   * Pytest component.
+   */
+  public readonly pytest?: Pytest;
+
+  constructor(options: PythonProjectOptions) {
+    super(options);
+
+    if (!PYTHON_PROJECT_NAME_REGEX.test(options.name)) {
+      throw new Error('Python projects must only consist of alphanumeric characters, hyphens, and underscores.');
+    }
```
Chriscbr, line 149: Could remove this. The issue is if I don't, imports in sample code based on the module name require special syntax e.g. if it has dashes, instead of:

```python
import python_code
```
I'd need
```python
__import__('python-code')
```
which most nobody uses since they just name modules sensibly. So I think I'll just keep this.

src/python/python-project.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  /**
+   * Convert an arbitrary string to a valid module filename.
+   *
+   * Replaces hyphens with underscores.
+   *
+   * @param name project name
+   */
+  private safeName(name: string) {
+    return name.replace('-', '_');
```
Chriscbr, line 227: Update this to also replace periods with underscores

Chriscbr (commented):
I wanted to also have some kind of task that would activate the virtual environment in the developer's shell, but tasks currently just run in a child process IIRC so it's yet not possible to do that. But I don't think it's strictly needed for a MVP.

docs/python.md
```diff
@@ -0,0 +1,144 @@
...
+
+Before creating a new project, make sure you have the version of Python you want
+to use set up in your terminal. If running `which python` on UNIX/macOS or
+`Get-Command python` on Windows does not print the path to the python version
+you want to use, you can specify the right path when setting up the project.
+
+To create a new Python project, use `projen new python`:
+
+```shell
+$ projen new python --name=my-project --python-path=/usr/bin/python
```
eladb, line 11:
```suggestion
$ projen new python --name=my-project [--python-path=/usr/bin/python]
```

docs/python.md
```diff
@@ -0,0 +1,144 @@
...
+  jsiiFqn: "projen.python.PythonProject",
+  name: 'my-project',
+  pythonPath: '/usr/bin/python',
+});
+
+project.synth();
+```
+
+> At this point, projenrc is in JavaScript, but in the future we plan to allow
+> specifying your project definitions in python.
```
eladb, line 47: yass sooner rather then later. happy to help out with that

docs/python.md
```diff
@@ -0,0 +1,144 @@
...
+to re-synthesize your project. The following sections describe the various
+features of your project.
+
+The following sections describe the various features of Python projects.
+
+## Managing environments, dependencies, and packaging
+
+Every Python project must have a component for managing/installing dependencies,
+a component for managing the Python virtual environment, and if it is a library,
+a component for managing packaging the library. Some components satisfy multiple
```
eladb, line 59:
```suggestion
a component for packaging the library. Some components satisfy multiple
```

docs/python.md
```diff
@@ -0,0 +1,144 @@
...
+1. Runtime dependencies (or just "dependencies").
+2. Development dependencies
+
+You can define dependencies when defining the project itself:
+
+```ts
+const project = new python.PythonProject({
+  deps: [
+    'Django@3.1.5',
+    'aws-cdk.core@*',
```
eladb, line 96: The "*" shouldn't be required:

```suggestion
    'aws-cdk.core', // implies "@*"
```

docs/python.md
```diff
@@ -0,0 +1,144 @@
...
+
+Notice the syntax for dependencies:
+
+```text
+<module>[@version]
+```
+
+Where `module` is the module name and `version` is the [semantic version
+requirement](https://semver.org) for the dependency. The semver syntax will be
+converted to the appropriate syntax in synthesized files. For example,
```
eladb, line 118:
```suggestion
converted to the appropriate dependency manager syntax in synthesized files. For example,
```

src/__tests__/new.test.ts
```diff
@@ -14,7 +14,7 @@ for (const type of inventory.discover()) {
       const projectdir = createProjectDir(outdir);
 
       // execute `projen new PJID --no-synth` in the project directory
-      execProjenCLI(projectdir, ['new', '--no-synth', type.pjid]);
+      execProjenCLI(projectdir, ['new', '--no-synth', '--python-path=/usr/bin/python', type.pjid]);
```
eladb, line 17: This looks odd here for the general case. Why is this required?

src/python/pip.ts
```diff
@@ -0,0 +1,74 @@
...
+import { DependencyType } from '../deps';
+import { Task, TaskCategory } from '../tasks';
+import { exec } from '../util';
+import { IPythonDeps } from './python-deps';
+import { PythonProject } from './python-project';
+import { RequirementsFile } from './requirements-file';
+
+export interface PipOptions {}
+
+export class Pip extends Component implements IPythonDeps {
```
eladb, line 11: docstrings

src/python/python-deps.ts
```diff
@@ -0,0 +1,20 @@
+export interface IPythonDeps {
+  /**
```
eladb, line 2: Feels like `installTask` should also be part of this

src/python/python-project.ts
eladb: Sounds good

src/python/python-project.ts
```diff
@@ -0,0 +1,508 @@
...
+   * @param spec Format `<module>@<semver>`
+   */
+  public addDevDependency(spec: string) {
+    return this.depsManager.addDevDependency(spec);
+  }
+
+  public postSynthesize() {
+    super.postSynthesize();
+
+    this.envManager.setupEnvironment();
```
eladb, line 505: Perhaps `envManager` and `depsManager` should just be `Component`s and then they can just implement `postSynthesize()` on their own, no?

src/python/requirements-file.ts
eladb: maybe:

```ts
packageProvider: IPackageProvider;
```

And then:

```ts
interface IPackageProvider {
  getPackages(): ...
}
```

Or something like that.

eladb (changes requested):
Looks great!

Feels like we should find a more robust pattern for "toolchain selection". It's a repeating pattern and maybe we can do better than a set of hard-coded booleans (which is what we've done so far, I know...)

I'll try to sketch something and share.

src/__tests__/new.test.ts
Chriscbr: pythonPath is a required parameter that gets auto-filled into .projenrc.js with the value of $(which python). Since this depends on your environment, I added this to make the Python .projenrc.js snapshot deterministic. I'll add a conditional to help make it clear that it isn't needed for all projects.

src/python/python-deps.ts
Chriscbr: Sure, definitely.

src/python/python-project.ts
Chriscbr: I considered that initially, but then the execution order becomes ambiguous (it's not really transparent which component's `postSynthesize()` would get called first, so `installDeps()` could run before `setupEnvironment()` and then modules could get installed to your global python installation instead of the virtual environment).

Chriscbr:
> Feels like we should find a more robust pattern for "toolchain selection". It's a repeating pattern and maybe we can do better than a set of hard-coded booleans (which is what we've done so far, I know...)
> 
> I'll try to sketch something and share.

Cool, I'm totally down. 👍

src/python/requirements-file.ts
Chriscbr: Hmm, yeah that's an idea. Object oriented design patterns to the rescue 😅

src/python/python-project.ts
```diff
@@ -0,0 +1,492 @@
...
+   *
+   * @default - true if the project type is library
+   */
+  readonly setuptools?: boolean;
+
+  /**
+   * Setuptools options
+   * @default - defaults
+   */
+  readonly setuptoolsOptions?: SetuptoolsOptions;
```
Chriscbr, line 145: Substitute this with a more general `packagingOptions` field, and eliminate any of the xxxOptions fields that aren't needed

src/__tests__/new.test.ts
eladb: Thinking about this some more- I think python path should probably just be based on the system definition and not baked into the project because there might be differences between developer environments and CI environments

Chriscbr: Hmm... I think you're right. Where you have your python installation really shouldn't be tied to the project at all. I'll make the assumption that consumers should have the version of python that they want to use accessible via the `python` command and they should adjust their PATH accordingly.

src/cli/macros.ts
```diff
@@ -18,6 +19,8 @@ export function tryProcessMacro(macro: string) {
 
     case '$GIT_USER_NAME': return getFromGitConfig('user.name') ?? 'user';
     case '$GIT_USER_EMAIL': return resolveEmail();
+    case '$PYTHON_PATH': return resolvePython();
```
eladb, line 22: Is this still needed?

src/python/poetry.ts
```diff
@@ -0,0 +1,301 @@
...
+    // declare the python versions for which the package is compatible
+    this.addDependency('python@^3.6');
+
+    this.packageTask = project.addTask('package', {
+      description: 'Creates source archive and wheel for distribution.',
+      category: TaskCategory.RELEASE,
+      exec: 'poetry build',
+    });
+
+    this.uploadTestTask = project.addTask('upload:test', {
```
eladb, line 46: should this be called `publish` instead of "upload"?

src/python/python-project.ts
eladb: Okay. sounds good

eladb (commented):
Looks good. See minor comment.

src/cli/macros.ts
Chriscbr: Good catch

src/python/poetry.ts
Chriscbr: Yep, updating

Chriscbr:
Updated the docs and PR description.

cc @garnaat Feel free to take a look. The docs rendered [here](https://github.com/Chriscbr/projen/blob/feat/python-project/docs/python.md) give an overview of what we have so far.

There are still some high level features like github workflows and configuration through projenrc.py left, but I'll leave those for a later PR.

I think the mental model I've captured here for several of the python tools is a good enough start, but like @eladb mentioned, there might be a better way to structure the APIs for choosing between these tools. 

As far as what the components actually do under the hood, most of it I feel OK with, although I wonder if `Pip` should be doing more -- like perhaps freezing the requirements and automatically diff'ing the results with the previous version and showing the changes to the user etc. But I don't have much python dependency management experience to build off of, so I'm mostly reading docs online to see what the options are. I just know we want the different language project types to feel similar to use.

Anyway, I'm looking forward to having this merged. :D

eladb (approved):

PR #517, opened 2021-02-02
feat: change default branch to "main"

campionfellin:
Change the default branch from `master` to `main`.

BREAKING CHANGE: Projects created with `projen` will now default to using the `main` branch.

---

I think this will introduce breaking changes for projects using `projen`. Making changes here I think will be more impactful. Rather than leaving each repository to choose to migrate `master` to `main`, we can now default `projen` projects to `main`.

After merging this into `master`, change main `projen` project branch to `main` and copy branch over.

Signed-off-by: campionfellin <campionfellin@gmail.com>

---


src/gitpod.ts
```diff
@@ -198,10 +198,10 @@ export interface GitpodPort {
  */
 export interface GitpodPrebuilds {
   /**
-   * Enable for the master/default branch
+   * Enable for the main/default branch
    * @default true
    */
-  readonly master?: boolean;
```
Chriscbr, line 204: This field was taken from a schema here: https://github.com/gitpod-io/gitpod/blob/master/components/dashboard/public/schemas/gitpod-schema.json

Since "main" might not be valid, I think this change should be reverted

Chriscbr (changes requested):
I support changing the default to `main`. I agree it will probably break consumers, but prioritizing it sooner is better than later.

Also normally we don't include version bumps in PRs, and it looks like the bump-workflow automatically got ran on your branch - I'm guessing that should be reverted (you might have to disable the workflow on your fork)

CHANGELOG.md
```diff
@@ -2,6 +2,128 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [0.16.0](https://github.com/projen/projen/compare/v0.5.2...v0.16.0) (2021-02-02)
```
eladb, line 5: revert

package.json
```diff
@@ -91,7 +91,7 @@
   },
   "main": "lib/index.js",
   "license": "Apache-2.0",
-  "version": "0.15.14",
+  "version": "0.16.0",
```
eladb, line 94: revert

src/__tests__/dev-env.test.ts
```diff
@@ -182,7 +182,7 @@ test('gitpod prebuilds config', () => {
 
   // WHEN
   project.gitpod?.addPrebuilds({
-    master: true,
```
eladb, line 185: revert

eladb (commented):
see comments + add a "BREAKING CHANGE" section in PR desc.

src/gitpod.ts
campionfellin: Thanks for the callout! Reverted

eladb (approved):

PR #524, opened 2021-02-03
feat: change default release branch to “main” (phase 1)

eladb:
Reintroduce #517 after it has been reverted to avoid unexpected breaking change.

We want to change the default value of `defaultReleaseBranch` from `master` to `main`.

Since this is a breaking change that will cause CI workflows not to be triggered by projects that still use `master` as their default branch, we will do this in two phases:

1. Phase 1 (this PR): make `defaultReleaseBranch` required and error out if it’s not set with an appropriate message. This will intentionally break CI for current projen users and require them to explicitly indicate if their release branch is “master” or “main” (those users will most likely need to set the value to “master”). New projects will be bootstrapped with a value of `main` here.
2. Phase 2 (future PR): change back `defaultReleaseBranch` to be optional with a default value of `main`.

Co-authored-by: campionfellin <campionfellin@gmail.com>

BREAKING CHANGE: as we migrate the default value of `defaultReleaseBranch` from `master` to `main`, this option will be required and an error will be printed if the value is not set. We will make this optional in a future release.

---


eladb:
Copy @campionfellin @Chriscbr

eladb:
@campionfellin I decided to use a slightly more aggressive migration strategy (i.e. require that users explicitly set `defaultReleaseBranch` as an initial phase), because otherwise they wouldn't know that this has changed and their CI releases will just not be triggered.

campionfellin:
Good call! Strategy seems great to me, and thanks for putting the comment in the code as well.

eladb:
Let's see how it goes...

PR #528, opened 2021-02-06
fix(cli): error when the given task doesn't exist

### LINKED ISSUE #484: Return error if you try running a task that doesn't exist
e.g. "Error: <task name> not found"

Right now if I try running a task that doesn't exist in my project (e.g. `npx projen floop`) then nothing gets printed, and the error code returned is 0 (obtained via `echo $?`).

Chriscbr:
Closes #484

When you run a task that doesn't exist, it will now raise an error, showing the CLI help printout and the message "Unknown command: foobar" with the error code 1.

Existing functionality that I verified as left intact:
- if the invalid task name is similar to an existing task name, the CLI will still show the error message "Did you mean <name>?" (this is existing behavior provided by yargs)
- providing no task still runs the 'default' synthesis task

---

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -16,6 +16,7 @@ async function main() {
   discoverTaskCommands(runtime, ya);
 
   ya.recommendCommands();
+  ya.strictCommands();
   ya.wrap(yargs.terminalWidth());
   ya.option('post', { type: 'boolean', default: true, desc: 'Run post-synthesis steps such as installing dependencies. Use --no-post to skip' });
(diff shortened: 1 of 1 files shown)


Chriscbr:
@eladb Feel free to edit the description to mark as breaking if needed. (Hypothetically, if someone had a github workflow that ran invalid projen tasks, such a workflow would now fail.)

eladb (approved):
Sounds like this is not a breaking change (given the previous behavior was broken)

PR #540, opened 2021-02-15
#452 Fix broken Windows shell commands by running through shx.

john-tipper:
---

[shx](https://github.com/shelljs/shx) is added as a dev dependency in order for it to be available for use in scripts. Linux script commnds (e.g. `rm -fr ...`) are modified to run through `shx` in order to be supported on Windows.

### DIFF
--- a/src/typescript.ts
+++ b/src/typescript.ts
@@ -197,10 +197,11 @@ export class TypeScriptProject extends NodeProject {
         category: TaskCategory.RELEASE,
       });
 
-      this.packageTask.exec('rm -fr dist');
-      this.packageTask.exec('mkdir -p dist/js');
+      this.addDevDeps('shx');
+      this.packageTask.exec('shx rm -fr dist');
+      this.packageTask.exec('shx mkdir -p dist/js');
       this.packageTask.exec(`${this.package.packageManager} pack`);
-      this.packageTask.exec('mv *.tgz dist/js/');
+      this.packageTask.exec('shx mv *.tgz dist/js/');
 
       this.buildTask.spawn(this.packageTask);
     }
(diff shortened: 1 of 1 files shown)


eladb (commented):
I am wondering if we can auto-discover these commands in task.exec() and add shx automatically?

john-tipper:
What mechanism have you got in mind for the auto-discovery?

eladb:
> What mechanism have you got in mind for the auto-discovery?

For example we can have a list of supported commands in shx and when the runtime identifies that we are on windows, it will pass the exec command through it.

john-tipper:
In the spirit of YAGNI, what benefit does this (OS discovery) give us versus just passing the currently broken `rm` and `mv` commands through `shx` for all OS?

john-tipper:
Having given it some further thought, I don't disagree with the idea of supporting other commands per se, but I think determining whether a command matches something in a defined list of supported shx commands could be a pain.  Whilst it's trivial to do a replace in the string of a matching command (e.g. `rm` -> `shx rm`), there are edge cases around what happens if a command is wrapped in quotes, for instance:

`find /some/path --exec bash -c 'rm {}' \;`

(Apologies for the badly formatted `find` command, I'm just trying to illustrate that a user might supply a command that may be difficult to parse).  Also, we need to consider what happens if the user is running inside Windows but using something like a GitBash shell.

My instinct is to make this the responsibility of the person writing the command(s) to ensure OS compatibility as the user themselves requires it and just wrap the commands that are inside projen with shx.  Just my 2c, though.

eladb:
This feels too error prone and easy to forget to do. If there's a way we can encapsulate this behavior behind the task runner, I rather we take that path.

eladb:
Closing in the meantime, feel free to open when you'd like to continue working on this. 🙏

iwt-janbrauer:
Hm, why not run all commands passed to `Task.exec` through `shx`? If a command is not supported by shx, it will error. I would rather optimize for users of projen, in this case Windows users, rather than optimize for implementors of new projen features.

john-tipper:
In the meantime @iwt-janbrauer, here's what I use in my `.projenrc.js` to allow projen to run on Windows, where I overwrite the `test` task:

```
const test = project.addTask('test', {
  category: '10.test',
  description: 'Run tests',
  exec: 'shx rm -fr lib/',
});
test.spawn(project.tasks.tryFind('test:compile'));
test.exec('jest --passWithNoTests --all');
test.spawn(project.tasks.tryFind('eslint'));

```

Issue #569, opened 2021-02-24
projen fails to create projects from external modules due to require failure


kcwinner:
When trying to create a new project from an external module projen fails with an error like: 
```
% npx projen new --from cdk-appsync-project
...
...

Error: Cannot find module 'cdk-appsync-project'
Require stack:
- /Users/ken/.npm/_npx/51135/lib/node_modules/projen/lib/cli/cmds/new.js
...
...
```

This is due to the `const module = require(mod);` portion of `createProject` method in `new.ts`. The module needs to be something like:
```
// External projects need to load the module from the modules directory
  const mod = opts.type.moduleName !== 'projen'
    ? path.join(process.cwd(), 'node_modules', opts.type.moduleName)
    : '../../index';
```

PR incoming

PR #570, opened 2021-02-24
fix: createProject now requires external modules from the correct path/directory

kcwinner:
Fixes: #569 

---


src/cli/cmds/new.ts
```diff
@@ -131,7 +131,10 @@ interface CreateProjectOptions {
  * type, so we can easily support multiple languages of projenrc.
  */
 function createProject(opts: CreateProjectOptions) {
-  const mod = opts.type.moduleName !== 'projen' ? opts.type.moduleName : '../../index';
+  // External projects need to load the module from the modules directory
+  const mod = opts.type.moduleName !== 'projen'
+    ? path.join(process.cwd(), 'node_modules', opts.type.moduleName)
+    : '../../index';
```
eladb, line 137: I think we should use `require.resolve()` to let node determine where the module is:

```ts
const mod = opts.type.moduleName !== 'projen'
    ? path.dirname(require.resolve(path.join(opts.type.moduleName, 'package.json'), { paths: [process.cwd()] }))
    : '../../index';
```

Also, can you please try/catch the `require.resolve()` call and throw a nice error that says that the external module could not be found.

kcwinner: Done!

eladb (approved):

PR #581, opened 2021-03-01
feat(awscdk-construct): cdkDependenciesAsDeps

### LINKED ISSUE #577: Maybe rethink `cdkDependencies` in awscdk-construct
Hi there and thanks for the great project!

We're currently building https://github.com/superluminar-io/super-eks/ with it and are very happy.

Unfortunately we're either doing smth wrong or there might be a conceptual problem in the way `cdkDependencies` are handled.
The Problem (and solution i think) is described in this blogpost https://dev.to/udondan/correctly-defining-dependencies-in-l3-cdk-constructs-45p , basically: defining cdk deps as real `dependencies` puts limitations on your downward consumers, therefore they should be added to `peerDependencies` and `devDependencies`, as this enables to use them during development but allows consumers to use their own specified version.

If this is correct thinking on my side, this line https://github.com/projen/projen/blob/main/src/awscdk-construct.ts#L150 would probably has to be changed? Or is there another way that I sould be using? Also the comment (and therefore reasoning) above the method is probably outdated, since npm 7 is actually installing `peerDependencies` (https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies)

If you're agreeing with me I can submit a PR if you'd like.
Thanks again!

Miradorn:
Allows disabling the default behavior in which CDK dependencies are also added as normal dependencies (and not just peer dependencies).

Resolves #577

---

### DIFF
--- a/API.md
+++ b/API.md
@@ -431,7 +431,7 @@ new AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
 
 Name | Type | Description 
 -----|------|-------------
-**cdkDependenciesAsDeps**🔹 | <code>boolean</code> | Wether to add `cdkDependencies` to `dependencies`.
+**cdkDependenciesAsDeps**🔹 | <code>boolean</code> | Whether CDK dependencies are added as normal dependencies (and peer dependencies).
 **version**🔹 | <code>string</code> | The target CDK version for this library.
 
 ### Methods
--- a/API.md
+++ b/API.md
@@ -420,6 +420,7 @@ new AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
   * **cdkVersion** (<code>string</code>)  Minimum target version this library is tested against. 
   * **cdkAssert** (<code>boolean</code>)  Install the @aws-cdk/assert library? __*Default*__: true
   * **cdkDependencies** (<code>Array<string></code>)  Which AWS CDK modules (those that start with "@aws-cdk/") does this library require when consumed? __*Optional*__
+  * **cdkDependenciesAsDeps** (<code>boolean</code>)  If this is enabled (default), all modules declared in `cdkDependencies` will be also added as normal `dependencies` (as well as `peerDependencies`). __*Default*__: true
   * **cdkTestDependencies** (<code>Array<string></code>)  AWS CDK modules required for testing. __*Optional*__
   * **cdkVersionPinning** (<code>boolean</code>)  Use pinned version instead of caret version for CDK. __*Default*__: false
 
@@ -430,6 +431,7 @@ new AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
 
 Name | Type | Description 
 -----|------|-------------
+**cdkDependenciesAsDeps**🔹 | <code>boolean</code> | Wether to add `cdkDependencies` to `dependencies`.
 **version**🔹 | <code>string</code> | The target CDK version for this library.
 
 ### Methods
@@ -439,12 +441,13 @@ Name | Type | Description
 
 Adds CDK modules as runtime dependencies.
 
-Modules are currently added with a caret CDK version both as "dependencies"
+Modules are currently by default added with a caret CDK version both as "dependencies"
 and "peerDependencies". This is because currently npm would not
 automatically install peer dependencies that are not declared as concerete
 dependencies by the consumer, so this is a little npm "hack" so that
 consumers will not need to depend on them directly if they don't interact
 with them.
+See `cdkDependenciesAsDeps` for changing the default behavior.
 
 ```ts
 addCdkDependencies(...deps: string[]): void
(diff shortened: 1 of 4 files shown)


src/awscdk-construct.ts
```diff
@@ -27,6 +27,17 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    */
   readonly cdkDependencies?: string[];
 
+
+  /** Should `cdkDependencies` be added to package.json `dependencies`.
+   *
```
eladb, line 32:
```suggestion
  /** 
   * If this is enabled (default), all modules declared in `cdkDependencies` will be also added as normal `dependencies` (as well as `peerDependencies`).
   *
```

src/awscdk-construct.ts
```diff
@@ -27,6 +27,17 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
...
+   * This is to ensure that downstream consumers actually have your cdk Dependencies installed
+   * when using npm < 7 or yarn, since theese don't install peerDependencies by default.
+   * If `false` `cdkDependencies` will be added to `devDependencies` to ensure they are present during development.
+   *
+   * @default true
+   */
+  readonly addCdkDependenciesToDeps?: boolean;
```
eladb, line 39: Rename to: `cdkDependenciesAsDeps`

eladb (changes requested):

API.md
```diff
@@ -418,6 +418,7 @@ new AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
   * **sampleCode** (<code>boolean</code>)  Generate one-time sample in `src/` and `test/` if there are no files there. __*Default*__: true
   * **catalog** (<code>[Catalog](#projen-catalog)</code>)  Libraries will be picked up by the construct catalog when they are published to npm as jsii modules and will be published under:. __*Default*__: new version will be announced
   * **cdkVersion** (<code>string</code>)  Minimum target version this library is tested against. 
+  * **addCdkDependenciesToDeps** (<code>boolean</code>)  Should `cdkDependencies` be added to package.json `dependencies`. __*Default*__: true
```
eladb, line 421: Rebuild

src/awscdk-construct.ts
```diff
@@ -27,6 +27,18 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    */
   readonly cdkDependencies?: string[];
 
+
+  /** 
+   * If this is enabled (default), all modules declared in `cdkDependencies` will be also added as normal `dependencies` (as well as `peerDependencies`).
+   *
+   * This is to ensure that downstream consumers actually have your cdk Dependencies installed
+   * when using npm < 7 or yarn, since theese don't install peerDependencies by default.
+   * If `false` `cdkDependencies` will be added to `devDependencies` to ensure they are present during development.
```
eladb, line 36:
```suggestion
   * This is to ensure that downstream consumers actually have your CDK dependencies installed
   * when using npm < 7 or yarn, 
where peer dependencies are not automatically installed. 
   * If this is disabled, `cdkDependencies` will be added to `devDependencies` to ensure they are present during development.
```

src/awscdk-construct.ts
```diff
@@ -112,6 +124,8 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
    */
   public readonly version: string;
 
+  public readonly cdkDependenciesAsDeps: boolean
```
eladb, line 127: Docstring please

src/awscdk-construct.ts
```diff
@@ -27,6 +27,18 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
...
+  /**
+   * If this is enabled (default), all modules declared in `cdkDependencies` will be also added as normal `dependencies` (as well as `peerDependencies`).
+   *
+   * This is to ensure that downstream consumers actually have your CDK dependencies installed
+   * when using npm < 7 or yarn, 
+where peer dependencies are not automatically installed. 
+   * If this is disabled, `cdkDependencies` will be added to `devDependencies` to ensure they are present during development.
+   *
+   * @default true
+   */
```
eladb, line 39: Wrap:

```suggestion
  /**
   * If this is enabled (default), all modules declared in `cdkDependencies` will be also added as
   * normal `dependencies` (as well as `peerDependencies`).
   *
   * This is to ensure that downstream consumers actually have your CDK dependencies installed
   * when using npm < 7 or yarn, where peer dependencies are not automatically installed.
   * If this is disabled, `cdkDependencies` will be added to `devDependencies` to ensure
   * they are present during development.
   *
   * @default true
   */
```

src/awscdk-construct.ts
```diff
@@ -112,6 +124,11 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
    */
   public readonly version: string;
 
+  /**
+   * Wether to add `cdkDependencies` to `dependencies`.
+   */
```
eladb, line 129:
```suggestion
  /**
   * Whether CDK dependencies are added as normal dependencies (and peer dependencies).
   */
```

eladb:
@projen rebuild

eladb (approved):

hoegertn:
@Miradorn if using the new flag and installing CDK as devDeps and peerDeps, shouldn't the devDeps be pinned to guarantee a minimal version and the peerDeps are `^` 

Currently both are unpinned.

PR #608, opened 2021-03-12
fix(node-project): fix projenVersion type to correctly use a string

skyrpex:
The NodeProject class expects a Semver object in `projenVersion`, but it's treated like a string, causing errors when stringifying to JSON files. I just corrected the type, which looks like it should be just string.

---


src/node-project.ts
```diff
@@ -35,9 +34,9 @@ export interface NodeProjectOptions extends ProjectOptions, NodePackageOptions {
   /**
    * Version of projen to install.
    *
-   * @default Semver.latest()
+   * Defaults to the latest version.
```
eladb, line 37:
```suggestion
   * @default - Defaults to the latest version.
```

eladb (commented):
Minor

eladb (approved):

PR #611, opened 2021-03-12
feat(typescript): TypeScriptProject now merges tsconfig prop into jest and eslint config

misterjoshua:
This PR changes `TypeScriptProject` so that it merges its `tsconfig` option
with the jest and eslint tsconfigs. Before, tsconfig changes had to be
duplicated in both the `tsconfigs` option and `jestOptions.typescriptConfig`
and there was no access to add includes or excludes to the eslint tsconfig.

---


misterjoshua:
@Chriscbr Here's that PR I promised in Slack.

Chriscbr (approved):
Looks good.

I think the control flow for jest might be a bit more confusing now because a merge happens once when declaring jest options, then again in the generateTypescriptConfig function. Ultimately I think a lot of the logic needs to be refactored at some point, but I'm okay with adopting this fix for now.

eladb (approved):

PR #602, opened 2021-03-09
fix(node): security vulnerability in rebuild-bot

eladb:
To address a security issue, the `@projen rebuild` workflow will no longer be triggered by a comment on the pull request. This commit removes the `issue_comment` trigger from rebuild-bot workflow so that this workflow can only be executed manually by administrators.

Additional details will be provided at a future date.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #624, opened 2021-03-17
feat(node): mutable pull request builds

eladb:
Introduce "mutablke builds". When enabled (by default) it will cause PR builds to automatically push any changes to synthesized files or snapshots to the PR branch. This only works for branches and will fail for forks (similar behavior to the anti-tamper check).

See attached documentation.

This means that if a PR is submitted and requires updates to generated files, the PR branch will automatically get updated. It also implies that anti-tamper check is disabled for PR builds.

This is done as part of the PR build which means that if the build passes and changes are pushed successfully, the PR check will be successful (doing that as a result of a different workflow will not trigger the PR build).

Related to #602 and #604

---


RomainMuller (approved):

pgollucci:
There is a bug here.  This doesn't work with forks.  https://github.com/projen/projen/pull/684

STR:
```sh
gh repo fork projen/projen --clone --remote
cd projen
git checkout -b demo
echo >> README.md
git add -A .
git commit -m "foo"
git push
gh pr create -a user -f
git checkout main
```

PR #632, opened 2021-03-18
fix(node): removeScript does not remove scripts associated with tasks

### LINKED ISSUE #631: .removeScript does not remove NPM script
Hello I am working on converting an existing AWS CDK Construct Library to use Projen however I would like to remove some of the auto generated npm scripts. For example I created a AwsCdkConstructLibrary using Projen and some scripts were auto generated like bump, release, compat, and so on but I already have existing shell scripts to do those actions and favor my scripts at the moment (in the future I may use the generated scripts). I see the API has a .removeScript function however after synthesizing, the scripts will not be removed. I could get around this with the escape hatches but that means I would get rid of it in the tasks.json and package.json manually which will clutter my projenrc file. 

To summarize I am trying to use the .removeScript function on an  AwsCdkConstructLibrary created by projen but the scripts will not be removed as I believe the function should. Based off of this https://github.com/projen/projen/blob/main/docs/awscdk-construct.md it seems that an AwsCdkConstructLibrary is a NodeProject so im not sure why the .removeScript function is not working. I am using yarn as the default to run the scripts, although im not sure if that has anything to do with it.

Chriscbr:
Fixes #631.

Deprecates the `hasScript` method on node projects (it still partially works, but `setScript` and `removeScript` APIs now use the [override escape hatch](https://github.com/projen/projen/blob/main/docs/escape-hatches.md) which is designed to be write-only, so only scripts added via the task system will be detected).

---


src/__tests__/__snapshots__/integ.test.ts.snap
```diff
@@ -2719,7 +2719,7 @@ project.synth();
       "url": "https://github.com/awslabs/cdk8s.git",
     },
     "scripts": Object {
-      "build": "npx projen build",
```
eladb, line 2722: What happened here?

src/node-package.ts
```diff
@@ -361,10 +361,10 @@ export class NodePackage extends Component {
 
   private readonly keywords: Set<string> = new Set();
   private readonly bin: Record<string, string> = {};
-  private readonly scripts: Record<string, string[]> = {};
   private readonly engines: Record<string, string> = {};
   private readonly peerDependencyOptions: PeerDependencyOptions;
   private _renderedDeps?: NpmDependencies;
+  private file: JsonFile;
```
eladb, line 367:
```suggestion
  private readonly file: JsonFile;
```

eladb (changes requested):

src/__tests__/__snapshots__/integ.test.ts.snap
Chriscbr: This is part of the snapshot test for cdk8s-projenrc.js. You can see [here](https://github.com/projen/projen/blob/05025555ef8448978450241a36b4bdfe48536101/src/__tests__/integration/cdk8s/cdk8s.projenrc.js#L52) it calls a shared "fixup" function, which calls `project.setScript('build', 'yarn compile');`. From my perspective, "yarn compile" is the correct output -- setScript was just ignored before because a corresponding task named "build" was taking higher precedence.

src/node-package.ts
Chriscbr: Done.

eladb (approved):

PR #634, opened 2021-03-18
feat: git tag-based versioning

eladb:
To avoid the need to mutate the default branch on releases (which may be blocked by branch protection), projen will now use git tags to manage project versions.

This is achieved by creating a GitHub Release as part of the "release" workflow (executed per commit, manually or on schedule). The release includes a change log generated by standard-version and the artifacts from the `dist/` directory. We also added a synchronization mechanism in the release workflow which will skip a release if there are new commits in the release branch. This protects against a race condition that can happen if multiple commits are being merged into the release branch at the same time.

BREAKING CHANGE: version numbers are now maintained through git tags instead of `version.json`. This file should be deleted alongside the `CHANGELOG.md` file. Change log entries are now included inside a GitHub Release.

---


misterjoshua:
Just a small comment from me. I think the CHANGELOG files are a good thing. They're very handy for users to keep up to date with a project. I read the CDK one every time a release goes out. Haha.

Is there any possibility of creating a scheme to merge the CHANGELOG from a PR to the main branch prior to the release? Perhaps the release workflow could even be triggered by that merge and the tag pushed back after it succeeds.

The way I see it, this approach is just a stone's throw away from supporting a release branch strategy if the user so chose.

Just food for thought.

gradybarrett:
I agree with the idea of a changelog file being a good thing. In practice what I end up reading mostly is the Github release notes - at least for Github projects. But at work we use Bitbucket, so a changelog file is all we've got.

What about annotating tags with the changelog data? Currently, I think standard-version puts in a note to reference the changelog.md file. If folks wanted a changelog, then we could generate one on the fly from the tags.

src/node-package.ts
```diff
@@ -405,17 +410,16 @@ export class NodePackage extends Component {
       engines: () => this.renderEngines(),
       main: this.entrypoint !== '' ? this.entrypoint : undefined,
       license: () => this.license ?? UNLICENSED,
-      version: '0.0.0',
       homepage: options.homepage,
       publishConfig: () => this.renderPublishConfig(),
+      version: current.version ?? '0.0.0',
```
pgollucci, line 415: Is the 0.0.0 here to prevent something like 99.99.99 by accident?

eladb: Yap

PR #637, opened 2021-03-19
feat: duck-typed resolvable objects

### LINKED ISSUE #636: feat: resolvable objects
## Feature

It would be nice to be able to pass a resolvable object for lazy resolution.

```typescript
new pj.JsonFile(this.project, this.fileName, {
  marker: true,
  obj: {
    someKey: new BufferedContent(...),
  },
});
```

Right now the resolve function only resolves fundamental JS types.

https://github.com/projen/projen/blob/05025555ef8448978450241a36b4bdfe48536101/src/_resolve.ts#L30-L33

##  Use Case

In this case, `BufferedContent` is an object used to read code from a file. That code is then included in a configuration file. This is useful for things like declaring Auth0 hook scripts for upload through an SPI.

## Workaround

```typescript
const bufferedContent = new BufferedContent(...);

new pj.JsonFile(this.project, this.fileName, {
  marker: true,
  obj: {
    someKey: bufferedContent.resolveContent.bind(bufferedContent),
  },
});
```

gradybarrett:
It's nice to be able to just pass a resolvable object that the resolve function knows how to lazily evaluate.

The resolve function already knows how to handle functions, but being able to pass an object means we can pass all of its context along with the resolve function without having to employ our venerable old friend `bind`.

I'm unsure of how to accomplish this in a strictly-typed manner without things becoming overly complicated, or overhauling the resolve function. I don't think either of those is desirable at this point.

I've also added an `IResolvable` interface, which at this point is only used in a loose type guard.

Closes #636.

---


eladb (changes requested):
Can we change this to `toJSON` instead? After 4 years with this model in the CDK, I'm realizing that this is basically the same protocol as JavaScript's `toJSON`.

gradybarrett:
> Can we change this to `toJSON` instead? After 4 years with this model in the CDK, I'm realizing that this is basically the same protocol as JavaScript's `toJSON`.

Great point! Updated.

gradybarrett:
@eladb any other concerns?

gradybarrett:
@eladb just checking in on this. If it's a no-go I'll close this out.

eladb (approved):

Issue #639, opened 2021-03-23
bug: node projects broken with `jsii/superchain`


misterjoshua:
After https://github.com/aws/jsii/pull/2718, builds are failing with "yarn: command not found". It looks like when GitHub runs a job step, the GitHub actions aren't hitting the `.bash_profile` where NVM and the correct node version would be set up.

I tried changing `workflowContainerImage` to an old copy of `jsii/superchain` (I pushed it up to `wheatstalk/jsii-superchain:working`) and most of the build works, but releasing to NPM fails because the superchain image is hard-coded in the release workflow.

I believe this is being tracked on the JSII side here: https://github.com/aws/jsii/issues/2734

misterjoshua:
It looks like https://github.com/aws/jsii/pull/2737 reverted the changes that broke projen and it looks like the new `jsii/superchain` image has been pushed up.

PR #640, opened 2021-03-23
feat: allow users to customize the publisher workflow image

misterjoshua:
This PR allows the user to customize the release workflow image as it was previously hard-coded.

#639 prompted this PR as I needed a short-term workaround for the `jsii/superchain` regression.

---

### DIFF
--- a/src/__tests__/node-project.test.ts
+++ b/src/__tests__/node-project.test.ts
@@ -304,6 +304,19 @@ test('buildWorkflowMutable will push changes to PR branches', () => {
   expect(workflow.jobs.build.steps).toMatchSnapshot();
 });
 
+test('workflowContainerImage propagates to the release workflow', () => {
+  // WHEN
+  const project = new TestNodeProject({
+    workflowContainerImage: 'some-other-workflow-image',
+    releaseToNpm: true,
+  });
+
+  // THEN
+  const workflowYaml = synthSnapshot(project)['.github/workflows/release.yml'];
+  const workflow = yaml.parse(workflowYaml);
+  expect(workflow.jobs.release_npm.container.image).toEqual('some-other-workflow-image');
+});
+
 function packageJson(project: Project) {
   return synthSnapshot(project)['package.json'];
 }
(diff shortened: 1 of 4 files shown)


Chriscbr (approved):
LGTM

eladb (changes requested):
Can this be just `workflowImage`?

misterjoshua:
> Can this be just `workflowImage`?

I couldn't see the details of the review clearly - I've pushed up some changes. Is this what you had in mind?

src/__tests__/node-project.test.ts
```diff
@@ -304,6 +304,19 @@ test('buildWorkflowMutable will push changes to PR branches', () => {
   expect(workflow.jobs.build.steps).toMatchSnapshot();
 });
 
+test('workflowImage propagates to the release workflow', () => {
+  // WHEN
+  const project = new TestNodeProject({
+    workflowContainerImage: 'some-other-workflow-image',
```
eladb, line 310:
```suggestion
    workflowImage: 'some-other-workflow-image',
```

eladb (approved):

eladb (changes requested):
Wait... I misunderstood this PR...

What's the use case for customizing only the image used to call `jsii-release-npm`? What is the motivation?

eladb:
To work around #639 you should be able to use escape hatches, no?

misterjoshua:
> To work around #639 you should be able to use escape hatches, no?

No obvious ways how jump out at me, but I may be missing some context. What escape hatches would have even been available to substitute the faulty job container in that npm publish step?

At the time, none of the node commands or `jsii-release-npm` worked in the superchain container. It seemed pretty reasonable to look to work around the issue by replacing the bad container image with a known good one. But, while the build job container was easy to change by project prop, it didn't seem like there was any straightforward way to patch the publishing job container since it was hard-coded.

But maybe there's something I don't understand here?

eladb:
You should be able to to find the workflow file using `tryFindObjectFile` and then add an override.

See https://github.com/projen/projen/blob/main/docs/escape-hatches.md and let me know if that makes sense.

misterjoshua:
> You should be able to to find the workflow file using `tryFindObjectFile` and then add an override.
> 
> See https://github.com/projen/projen/blob/main/docs/escape-hatches.md and let me know if that makes sense.

@eladb Ah. I see now. I hadn't discovered these methods. I looked in all the wrong places too. Lol. Yeah, with these methods, I don't think this PR is strictly necessary.

PR #649, opened 2021-03-26
feat(python): pipenv support

ArmaanT:
This PR adds the ability for python projects to use Pipenv to manage virtual environments and dependencies. I also cleaned up a bit of duplicate code in the python tests.
Alongside the unit tests, I've tested this PR out locally by creating a small python project managed by projen.

---

### DIFF
--- a/src/python/pipenv.ts
+++ b/src/python/pipenv.ts
@@ -1,0 +1,162 @@
+import { Component } from '../component';
+import { DependencyType } from '../deps';
+import { Task, TaskCategory } from '../tasks';
+import { TomlFile } from '../toml';
+import { exec, execOrUndefined } from '../util';
+import { IPythonDeps } from './python-deps';
+import { IPythonEnv } from './python-env';
+import { PythonProject } from './python-project';
+
+
+export interface PipenvOptions {
+  /**
+   * Options to pass through to the Pipfile
+   */
+  readonly pipfileOptions?: PipenvPipfileOptionsWithoutDeps;
+}
+/**
+ * Manage project dependencies and virtual environments through the
+ * Pipenv CLI tool.
+ */
+export class Pipenv extends Component implements IPythonDeps, IPythonEnv {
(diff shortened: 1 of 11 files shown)


src/python/pipenv.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  private synthDevDependencies() {
+    const dependencies: { [key: string]: any } = {};
+    for (const pkg of this.project.deps.all) {
+      if ([DependencyType.DEVENV].includes(pkg.type)) {
+        dependencies[pkg.name] = pkg.version;
+      }
+    }
+    return dependencies;
+  }
```
Chriscbr, line 62: Can you add some unit tests that check `Pipfile` synthesizes correctly when the user adds dependencies with different kinds of  (e.g. `deps: ['flask', 'isort@^5.8.0', 'test@0.0.1']`)? When I tried this out, it seemed to cause issues because it looks like Pipefile expects packages in a specific format (see [here](https://github.com/pypa/pipfile#pipfile)).

We have a utility function [here](https://github.com/projen/projen/blob/b9152631d7f79d83df175cda497fe3f01cb3b46f/src/util/semver.ts#L52) that converts strings like '*' or '^6.0.3' into traditional python version strings, like '>=0.0.0' and '>=6.0.3, <7.0.0' respectively; this might help.

Chriscbr (changes requested):
Looks great! See comments below

eladb:
@ArmaanT would you be able to address @Chriscbr's feedback?

ArmaanT:
Sorry about that, I had a busy few weeks. I can definitely make those changes in the next week or so.

eladb:
Any updates? Should we close for now until you are ready to pick this up?

ArmaanT:
Yeah, I think closing for now makes sense. I have a couple of questions about PEP 508 markers that I think should be resolved before merging this. I'll make an issue for that discussion.

PR #650, opened 2021-03-26
feat: IniFile component

### LINKED ISSUE #646: Add component for .ini files
.ini is used for some Python configuration files. A new component should be functionally similar to TomlFile, YamlFile, etc.

ini files are used in the Python ecosystem for `setup.cfg` (used by setuptools) and `tox.ini` (used by tox).

Chriscbr:
Closes #646 

---

### DIFF
--- a/src/__tests__/ini.test.ts
+++ b/src/__tests__/ini.test.ts
@@ -1,0 +1,38 @@
+import * as INI from 'ini';
+import { IniFile } from '..';
+import { synthSnapshot, TestProject } from './util';
+
+test('ini object can be mutated before synthesis', () => {
+  const prj = new TestProject();
+
+  const obj: any = {
+    hello: 'world',
+  };
+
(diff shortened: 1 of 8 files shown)


src/__tests__/ini.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    anotherField: { foo: '1234' },
+  });
+});
+
+test('ini file can contain projen marker', () => {
+  const prj = new TestProject();
+
+  const obj: any = {};
+
+  new IniFile(prj, 'my/ini/file-marker.ini', { obj, marker: true });
```
eladb, line 31: I feel like this should be the default, no?

Chriscbr: You mean for `marker: true`? I agree. But the default should be the same across all file components, right? Makes more sense to change that in separate PR imo

eladb (approved):

Issue #651, opened 2021-03-27
Bug: projen:upgrade fails when using npm instead of yarn


seeebiii:
# Expected

I wanted to run `projen:upgrade` to upgrade the `projen` version in my repository, e.g. from `0.16.38` to `0.17.17` (latest version) in my `package.json`.

# Actual

The following error occurs:

```bash
npx projen projen:upgrade
🤖 projen:upgrade | yarn upgrade -L projen
yarn upgrade v1.22.10
error No lockfile in this directory. Run `yarn install` to generate one.
info Visit https://yarnpkg.com/en/docs/cli/upgrade for documentation about this command.
🤖 Task "projen:upgrade" failed when executing "yarn upgrade -L projen" (cwd: [deleted])
```

I'm using `packageManager: NodePackageManager.NPM` in my project. That's why there's no `yarn` lockfile. You can look at the full [`.projenrc.js`](https://github.com/seeebiii/ses-verify-identities/blob/main/.projenrc.js) if you like.

Do you have any idea how to fix it besides switching to `yarn`? 😄 

If you need anything else from my side, please let me know :)

# Versions

Projen: 0.17.17 (installed globally on my system via `npm i -g`)
Node: v14.15.4
MacOS: 11.2

thtliife:
I have the same behaviour occurring using ppm (`packageManager: NodePackageManager.PNPM`)
It seems the projen cli is hardcoded to use yarn for a few things through [the yarnAdd function](https://github.com/projen/projen/blob/125166e7975158252a62a9b3c36c8dbc49f62013/src/cli/cmds/new.ts#L418)

Would be aweome to have a cli option to specify the preferred package manager, and then have the generated .projenrc.js reflect that

pgollucci:
We don't need an option.  The code is buggy.  It needs to call the wrapper.  This is a good-first-issue @eladb

eladb:
I believe this is resolved by https://github.com/projen/projen/pull/692

PR #652, opened 2021-03-29
feat(node): dependencies upgrade workflow

iliapolo:
Dependency upgrades in projen are currently managed by dependabot, which creates a PR for each package upgrade separately. This poses two issues:

- Creates clutter and possibly high maintenance load if PR's cannot be automatically merged.
- Prevents updates of packages that need to be upgraded in unity (for example `@aw-cdk`).

This PR introduces an alternative by using a GitHub workflow that executes `npm-check-updates` to upgrade `package.json` files, and `yarn` to synchronize the lock file.

Another problem with the current setup is that automatic merges are accomplished by two distinct mergify rules. 

- `auto-merge` label + build passes CI.
- PR's from dependabot + build passes CI.

This approach breaks down if branch protection rules require a PR approval. Since this is the best practice we want to encourage, we need to come up with a different merge mechanism. 

With this PR, dependency update PR's will be automatically approved by a different github workflow (`auto-approve.yml`).
This means we can remove the aforementioned mergify rules, and use the standard rule of merging approved PR's that pass CI build.

The auto approval workflow is enabled by default since it can utilize the default github token available to workflows, without requiring an additional projen specific token.

Example PR: https://github.com/iliapolo/constructs/pull/4

---


src/projen-upgrade.ts
```diff
@@ -2,30 +2,21 @@ import { NodeProject } from './node-project';
...
 
 export interface ProjenUpgradeOptions {
-  /**
-   * The secret name which contains a GitHub Access Token with `repo` and
-   * `workflow` permissions. This token is used to submit the upgrade pull
-   * request, which will likely include workflow updates.
-   *
-   * @default - auto-upgrade is disabled
-   */
-  readonly autoUpgradeSecret?: string;
```
iliapolo, line 12: Instead, I exposed `projenSecret` on `NodeProject` and reused it.

src/node-project.ts
```diff
@@ -1085,3 +1045,73 @@ export interface NodeWorkflowSteps {
...
+  public static githubActions(options: AutoUpgradeDependenciesOptions = {}): DependenciesUpgrade {
+    return new DependenciesUpgrade('github-actions', options);
+  }
+
+  /**
+   * Use Dependabot (with custom options) to upgrade dependencies.
+   *
+   * @param options The options.
+   */
+  public static dependabot(options: DependabotOptions = {}): DependenciesUpgrade {
```
iliapolo, line 1105: Not in love with this API, happy to accept alternatives :)

src/node-project.ts
```diff
@@ -1085,3 +1045,73 @@ export interface NodeWorkflowSteps {
...
+
+  /**
+   * Use dependabot (with the default options) to upgrade dependencies.
+   */
+  public static readonly DEPENDABOT: DependenciesUpgrade = new DependenciesUpgrade('dependabot');
+
+  /**
+   * Use GitHub actions (with the default options) to upgrade dependencies.
+   */
+  public static readonly GITHUB_ACTIONS: DependenciesUpgrade = new DependenciesUpgrade('github-actions');
```
iliapolo, line 1089: Just some sugar to provide an enum like experience for the common use-case.

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,17 @@
...
+name: auto-approve
+on:
+  pull_request:
+    types:
+      - labeled
+      - opened
+      - unlabeled
+jobs:
+  approve:
+    if: contains(github.event.pull_request.labels.*.name, 'pr/auto-approve')
```
eladb, line 12: I don't see a huge value in the `pr/` prefix here.

```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'auto-approve')
```

eladb, line 12: Are we at all concerned that this basically allows maintainers to approve their own PRs even if "required for administrators" is set. @RomainMuller will likely have a strong opinion here.

src/__tests__/node-project.test.ts
```diff
@@ -1,13 +1,46 @@
...
 import * as logging from '../logging';
 import { NodePackage, NpmAccess } from '../node-package';
 import { Project } from '../project';
 import { mkdtemp, synthSnapshot, TestProject } from './util';
 
 logging.disable();
 
+describe('dependencies upgrade', () => {
+
+  test('defaults to github actions when a projen secret is defined', () => {
```
eladb, line 14:
```suggestion
  test('defaults to a github workflow when a projen secret is defined', () => {
```

src/github/auto-approve.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Options for 'AutoApprove'
+ */
+export interface AutoApproveOptions {
+
+  /**
+   * PR's assigned with this label will be auto-approved.
+   *
+   * @default pr/auto-approve
```
eladb, line 12: Defaults should be JSON encoded (and also, please change to `auto-approve`)

```suggestion
   * @default "auto-approve"
```

src/github/auto-approve.ts
```diff
@@ -0,0 +1,54 @@
...
+  public readonly label: string;
+
+  constructor(project: Project, options: AutoApproveOptions = {}) {
+    super(project);
+
+    if (!project.github) {
+      throw new Error('GitHub must be configured to enable auto approvals');
+    }
+
+    this.label = options.label ?? 'pr/auto-approve';
```
eladb, line 34:
```suggestion
    this.label = options.label ?? 'auto-approve';
```

src/github/auto-approve.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const autoApprove = project.github.addWorkflow('auto-approve');
+    autoApprove.on({
+      pull_request: { types: ['labeled', 'opened', 'unlabeled'] },
+    });
+    autoApprove.addJobs({
+      approve: {
+        'if': `contains(github.event.pull_request.labels.*.name, '${this.label}')`,
```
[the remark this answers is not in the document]
eladb: Mmm.. curious - what will cause this it fail if this is the same user?

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+   * enabling this will throw an error, as auto approval cannot be accomplished.
+   *
+   * @default - true if auto-approve workflow is configured for the project, false otherwise.
+   */
+  readonly autoApprove?: boolean;
+
+  /**
+   * List of package names to ignore during the upgrade.
+   *
+   * @default - Only projen itself is ignored.
```
eladb, line 29:
```suggestion
   * @default - `projen` is ignored since there is a separate workflow to upgrade it.
```

Which makes me wonder - with the combination of this and the self-mutation feature, do we really need a separate upgrade workflow for projen? Feel like we can just use this workflow to upgrade projen itself, and let the self-mutation workflow update the branch if there are any synth changes, no?

src/github/auto-approve.ts
```diff
@@ -0,0 +1,54 @@
...
+   *
+   * @default pr/auto-approve
+   */
+  readonly label?: string;
+}
+
+/**
+ * Configures a GitHub workflow that auto-approves PR's that are assigned with a specific label.
+ */
+export class AutoApprove extends Component {
```
eladb, line 20: I am wondering how this will work with the self-mutation workflow (the current version and the future version which will submit a PR against your branch). 

I am thinking that maybe initially we can avoid auto-approve in the dependency upgrade workflows, so that if there are any synth changes that result in the update, a human will need to review them. WDYT?

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
+import { Component } from '../component';
+import { NodeProject } from '../node-project';
+
+/**
+ * Options for `AutoUpgradeDependencies`.
+ */
+export interface AutoUpgradeDependenciesOptions {
```
eladb, line 7: Rename to `UpgradeDependencies` instead of `AutoUpgradeDependencies` - the "auto" part is optional.

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * How often to check for new versions and raise pull requests for version updates.
+ */
+export enum AutoUpgradeDependenciesSchedule {
+  /**
+   * At 00:00.
+   */
+  DAILY = 'daily',
```
eladb, line 42: NEVER is missing

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    const labels: string[] = [];
+
+    if ((options.autoApprove ?? true) && project.autoApprove) {
+      labels.push(project.autoApprove.label);
+    }
+
+    const ignore = ['projen', ...(options.ignore ?? [])];
+
+    const dependencies = project.github.addWorkflow('auto-upgrade-dependencies');
```
eladb, line 88:
```suggestion
    const dependencies = project.github.addWorkflow('upgrade-dependencies');
```

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+            uses: 'actions/checkout@v2',
+          },
+          ...project.installWorkflowSteps,
+          {
+            name: 'Upgrade package.json dependencies',
+            run: `npx npm-check-updates --upgrade --reject '${ignore.join(',')}'`,
+          },
+          {
+            name: 'Upgrade yarn.lock dependencies',
+            run: `yarn install && yarn upgrade --pattern !(${ignore.join('|')})`,
```
eladb, line 109: We should use `npm` if the project is configured to use npm, no?

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+          },
+          {
+            name: 'Create Pull Request',
+            uses: 'peter-evans/create-pull-request@v3',
+            with: {
+              'token': `\${{ secrets.${project.projenSecret} }}`,
```
eladb, line 115: Is this really needed? I suspect a scheduled task should have a read/write token by default.

eladb (changes requested):
High level comment is whether we really need the projen upgrade workflow any more.

.github/workflows/auto-approve.yml
iliapolo: See https://github.com/projen/projen/pull/652#discussion_r603120017

src/__tests__/integration/cdk-watchful.projenrc.js
```diff
@@ -17,7 +17,7 @@ const project = new ConstructLibraryAws({
   },
 
   // creates PRs for projen upgrades
-  projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',
+  projenSecret: 'PROJEN_GITHUB_TOKEN',
```
pgollucci, line 20: Why did this get renamed again ?

src/github/auto-upgrade-dependencies.ts
pgollucci: tough question.

src/github/auto-upgrade-dependencies.ts
pgollucci: https://github.com/projen/projen/issues/651

src/github/index.ts
```diff
@@ -1,8 +1,10 @@
 export * from './github';
 
 export * from './dependabot';
-export * from './workflows';
 export * from './mergify';
 export * from './pr-template';
 export * from './mergify';
-export * from './auto-merge';
\ No newline at end of file
+export * from './workflows';
```
pgollucci, line 7: Make a workflows dir ?
workflows/auto-merge
worksflow/foo
worksflows/....
?

eladb:
@iliapolo any chance we can get this merged? It will quickly get stale and hard to maintain this branch...

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+   * @default - Only projen itself is ignored.
+   */
+  readonly ignore?: string[];
+
+}
+
+/**
+ * How often to check for new versions and raise pull requests for version updates.
+ */
+export enum AutoUpgradeDependenciesSchedule {
```
RomainMuller, line 38: I don't like how all of these options are at midnight UTC. You'd usually want to randomise that a bit to spread the load over (free) CI environments... As a basic courtesy.

At least, there should be an option to provide one's own POSIX cron expression.

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+        'runs-on': 'ubuntu-latest',
+        'steps': [
+          {
+            name: 'Checkout',
+            uses: 'actions/checkout@v2',
+          },
+          ...project.installWorkflowSteps,
+          {
+            name: 'Upgrade package.json dependencies',
+            run: `npx npm-check-updates --upgrade --reject '${ignore.join(',')}'`,
```
RomainMuller, line 105: Suggest making the upgrade target explicit (valid values are: `'minor' | 'patch' | 'latest'`).

```suggestion
            run: `npx npm-check-updates --upgrade --target 'minor' --reject '${ignore.join(',')}'`,
```

1. If the environment has `npm@^7`, this will actually not work, as it'll hang on a prompt asking if it's okay to install `npm-check-updates` before proceeding
2. Since you control the project settings, why not add `npm-check-updates` to the `devDependencies` of the project, and open with a dependency install stage? This has the added benefit of allowing the customer to control what version of `npm-check-updates` is being used (you may get nasty surprises otherwise, as `npx` will just get whatever the `latest` is... it may be incompatible with your invocations!
3. There are missing features here:
   1. I need to be able to major-pin certain dependencies (typically, I'd want this workflow NOT to upgrade any of my non-`dev` dependencies to a new major line -- and also `@types/node` should remain on the lowest supported `node` major). I'd be tempted to make this `pinToMajor = ['@types/node']`, perhaps with a way to opt out of the `dependencies` security.
   2. I need to be able to minor-pin certain dependencies (some packages do not honour SemVer, and so you wanna make sure they only get patch-upgraded... for example: `typescript`). I'd be tempted to make this `pinToMinor=['typescript']`
   3. This is not `yarn workspaces`/mono-repos aware. This may be fine (does `projen` support this use-case anyway?)

src/github/auto-upgrade-dependencies.ts
RomainMuller, line 109: This is precisely why I'd have named this `yarn-upgrade`, and would have checked that the project uses `yarn` before configuring.

src/github/auto-upgrade-dependencies.ts
RomainMuller, line 115: Check out the (relatively new) [`permissions`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions) configuration...

Since this workflow triggers either on a `cron` or `workflow_dispatch`, it can be seen as only touching code that has been accepted into the default branch already (i.e: is "trusted")... You could grant the workflow extended permissions so you no longer have to provision a PAT for this kind of use-case.

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+          {
+            name: 'Create Pull Request',
+            uses: 'peter-evans/create-pull-request@v3',
+            with: {
+              'token': `\${{ secrets.${project.projenSecret} }}`,
+              'commit-message': 'Upgrade dependencies',
+              'branch': 'github-actions/dependencies',
+              'labels': labels.join(','),
+              'title': 'chore(deps): Upgrade dependencies',
+              'body': 'Upgrade dependencies\n\n------\n\n*Automatically created by projen via GitHubActions*',
```
RomainMuller, line 120: Those should be made user-configurable (`labels` obviously is already).

I would also not name the default branch `github-actions/dependencies` -- it's way too generic and could likely end up being shared by several dependency-management actions, which would not be good... At the very least, you'll want this to include the name of the workflow.

src/github/auto-upgrade-dependencies.ts
```diff
@@ -0,0 +1,143 @@
...
+      },
+    });
+  }
+
+  private scheduleToCron(schedule: AutoUpgradeDependenciesSchedule): string {
+
+    switch (schedule) {
+      case AutoUpgradeDependenciesSchedule.DAILY:
+        return '0 0 * * *';
+      case AutoUpgradeDependenciesSchedule.WEEKDAY:
```
RomainMuller, line 133: If you have `WEEKDAY` you also need `WEEKEND`. And then again... Are these defined "universally"? I'd be tempted to say a weekend is Saturday & Sunday... But folks in Israel may be tempted to say it's Friday & Saturday... Depending on *why* you want to refer to the week-end here (you likely mean "days when nobody is in the office").

This looks like trying too hard to me, and I'd kinda feel like you might want to allow people to craft their own CRON expression, possibly with factories for the "basic" cases such as daily, weekly, and monthly.

src/node-project.ts
```diff
@@ -1085,3 +1067,73 @@ export interface NodeWorkflowSteps {
...
+      case 'disabled':
+        break;
+      case 'dependabot':
+        this.bindDependabot(project);
+        break;
+      case 'github-actions':
+        this.bindGitHubActions(project);
+        break;
+      default:
+        throw new Error(`Unsupported dependency upgrade type: ${this.type}`);
```
RomainMuller, line 1127: ❤️

src/node-project.ts
```diff
@@ -1085,3 +1067,73 @@ export interface NodeWorkflowSteps {
...
+    }
+  }
+
+  private bindDependabot(project: NodeProject) {
+    project.github?.addDependabot((this.options ?? {}) as DependabotOptions);
+  }
+
+  private bindGitHubActions(project: NodeProject) {
+    new AutoUpgradeDependencies(project, (this.options ?? {}) as AutoUpgradeDependenciesOptions);
+  }
```
RomainMuller, line 1137: I'm not entirely sure these are necessary, as they are 2/3rd boilerplate... The one-liner content could probably have been directly in the `switch` within `bind`... But since they're already broken out... I guess it's not bad!

eladb (commented):
Looking forward for this!

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+            uses: 'actions/upload-artifact@v2',
+            with: { name: patchArtifact, path: patchFile },
+          },
+        ],
+      };
+
+      const createPR: any = {
+        'needs': 'upgrade-dependencies',
+        'name': 'Create Pull Request',
+        'permissions': 'write-all',
```
eladb, line 175: Can we scope this down?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+                '*Automatically created by projen via GitHubActions*',
+              ].join('\n'),
+            },
+          },
+        ],
+      };
+
+      if (project.buildWorkflowJobId) {
+        createPR.steps.push({
+          name: 'Update status check',
```
eladb, line 220: Can this be reused with the node build workflow somehow?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+                + `"head_sha":"${branchName}",`
+                + '"status":"completed",'
+                + '"conclusion":"\${{ needs.upgrade-dependencies.outputs.build-conclusion }}",'
+                + '"output":{'
+                  + `"title":"Created via the ${workflowName} workflow.",`
+                  + '"summary":"Build log is available here: ${{ needs.upgrade-dependencies.outputs.build-log }}"'
+                + '}'
+              + '}\'',
+          env: {
+            GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}',
```
eladb, line 237: Is this needed?

src/node-project.ts
```diff
@@ -1182,3 +1201,46 @@ export interface NodeWorkflowSteps {
...
+  public static dependabot(options: DependabotOptions = {}) {
+    return new DependenciesUpgrade((project: NodeProject) => {
+      project.github?.addDependabot(options);
+    }, options.ignoreProjen);
+  }
+
+  /**
+   * Upgrade via github actions.
+   */
+  public static githubActions(options: UpgradeDependenciesOptions = {}) {
```
eladb, line 1227: The name `githubActions` is a bit ambiguous. It could be interpreted as "only upgrade GitHub actions". It's also not very accurate because the actual tool is not actions.

Maybe the name of the class could be different- something like `DependencyUpgradeMechanism` and then instead of githubActions we can use `ncu`?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * Schedule to run on.
+   *
+   * @default UpgradeDependenciesSchedule.DAILY
+   */
+  readonly schedule?: UpgradeDependenciesSchedule;
+
+  /**
+   * Which secret to use when creating the PR.
```
eladb, line 34: Describe the difference between using the default token and a user supplied token

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+   *
+   * @default - default github token.
+   */
+  readonly secret?: string;
+
+  /**
+   * Labels to apply on the PR.
+   *
+   * @default - no labels.
```
eladb, line 43:
```suggestion
   * @default [] no labels.
```

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+   */
+  readonly labels?: string[];
+
+  /**
+   * Execute 'build' after the upgrade.
```
eladb, line 48: Add some details on what that means

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * List of package names to include during the upgrade.
+   *
+   * @default - Everything is included.
+   */
+  readonly include?: string[];
+
+  /**
+   * Include a github workflow for creating PR's that upgrades the
```
eladb, line 76: What happens if this is `false`? Does that imply dependabot? Does that imply that upgrades can be done manually?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+   *
+   * @default - default options.
+   */
+  readonly workflowOptions?: UpgradeDependenciesWorkflowOptions;
+
+  /**
+   * The name of the task that will be created.
+   * This will also be the workflow name.
+   *
+   * @default 'upgrade-dependencies'.
```
eladb, line 94:
```suggestion
   * @default "upgrade-dependencies"
```

Defaults should be JSON

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+   */
+  readonly taskName?: string;
+
+  /**
+   * Whether or not to ignore projen upgrades.
+   *
+   * @default true
+   */
+  readonly ignoreProjen?: boolean;
```
eladb, line 103: Why not just make this the default for 'exclude'?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+
+}
+
+/**
+ * How often to check for new versions and raise pull requests for version upgrades.
+ */
+export class UpgradeDependenciesSchedule {
```
eladb, line 110: Style: I prefer to put all these auxiliary types _after_ the main options and class.

.projen/tasks.json
```diff
@@ -133,19 +133,6 @@
         }
       ]
     },
-    "projen:upgrade": {
```
eladb, line 136: I expected to see the new task here

src/node-project.ts
```diff
@@ -752,25 +758,38 @@ export class NodeProject extends Project {
...
-      autoUpgradeSchedule: options.projenUpgradeSchedule,
-      labels: (projenAutoMerge && this.autoMerge?.autoMergeLabel)
-        ? [this.autoMerge.autoMergeLabel]
-        : [],
-    });
+    const defaultDependenciesUpgrade = (options.dependabot ?? true) ? DependenciesUpgrade.dependabot() : DependenciesUpgrade.githubActions();
+    const dependenciesUpgrade = options.depsUpgrade ?? defaultDependenciesUpgrade;
+    dependenciesUpgrade.bind(this);
+
+    if (dependenciesUpgrade.ignoresProjen && this.package.packageName !== 'projen') {
```
eladb, line 775: Why is this only used if ignoreProjen is true?

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,368 @@
...
+      jobs[upgrade.jobId] = upgrade.job;
+      jobs[pr.jobId] = pr.job;
+
+      this.workflow.addJobs(jobs);
+    }
+  }
+
+  private createUpgrade(task: Task): Upgrade {
+
+    const project = this.project as NodeProject;
```
eladb, line 209: I'd keep `nodeProject` as a member (otherwise this can break in the future)

eladb (approved):
See some comments. Added `do-not-merge`

eladb (changes requested):
No do not merge so changed review verdict to avoid auto-merge

src/github/auto-upgrade-dependencies.ts
iliapolo: Exposed a way to provide a custom cron expression 👍

src/github/auto-upgrade-dependencies.ts
iliapolo: - This isn't mono-repo aware, projen doesn't really support it anyway. 

- Regarding specific minor-pins - yes, feature needed. I'll add in a subsequent PR. This currently only addresses the naive case. 

- `npm-check-updates` now added to `devDependencies` 👍

src/github/auto-upgrade-dependencies.ts
iliapolo: Yeap, added workflow name.

src/github/auto-upgrade-dependencies.ts
iliapolo: Added custom cron expression support - this should suffice for now.

src/node-project.ts
iliapolo: Simplified by having the static factories pass a `binder` function

src/github/auto-upgrade-dependencies.ts
iliapolo: Yeap - using the default token with `write-all` permissions.

src/github/auto-upgrade-dependencies.ts
iliapolo: Added support for all package managers

src/github/auto-merge.ts
```diff
@@ -33,7 +25,6 @@ export interface AutoMergeOptions {
  * the PR to be merged.
  */
 export class AutoMerge extends Component {
-  public readonly autoMergeLabel: string;
```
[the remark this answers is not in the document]
iliapolo: Yeap

src/node-package.ts
```diff
@@ -716,7 +716,7 @@ export class NodePackage extends Component {
     this.addEngine('node', nodeVersion);
   }
 
-  private renderInstallCommand(frozen: boolean) {
+  public renderInstallCommand(frozen: boolean) {
```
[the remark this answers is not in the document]
iliapolo: It avoid duplicating the exact same logic in the upgrade workflow. Since this method doesn't expose anything inherently private, figured its harmless.

src/node-project.ts
```diff
@@ -180,15 +178,6 @@ export interface NodeProjectOptions extends ProjectOptions, NodePackageOptions {
...
   readonly mergifyOptions?: MergifyOptions;
 
-  /**
-   * Automatically merge PRs that build successfully and have this label.
-   *
-   * To disable, set this value to an empty string.
-   *
-   * @default "auto-merge"
-   */
-  readonly mergifyAutoMergeLabel?: string;
```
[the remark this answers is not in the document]
iliapolo: yes

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+      ncuCommand.push(`--filter=${options.include.join(',')}`);
+    }
+
+    task.exec(ncuCommand.join(' '));
+    task.exec(project.package.renderInstallCommand(false));
+
+    // then we define a github workflow if possible
+    if (project.github) {
+
+      const schedule = options.schedule ?? UpgradeDependenciesSchedule.DAILY;
```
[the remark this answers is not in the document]
iliapolo: Not sure which part you mean here exactly. I've refactored the code since this comment so hopefully its more readable and clean now.

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+      this.workflow = project.github.addWorkflow(workflowName);
+      const triggers: any = { workflow_dispatch: {} };
+      if (schedule !== UpgradeDependenciesSchedule.NEVER) {
+        triggers.schedule = [{ cron: this.scheduleToCron(schedule) }];
+      }
+      this.workflow.on(triggers);
+
+      const upgradeAndBuild = {
+        'name': 'Upgrade & Build',
+        'permissions': 'read-all',
```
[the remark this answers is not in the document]
iliapolo: added

src/upgrade-dependencies.ts
```diff
@@ -0,0 +1,272 @@
...
+          },
+          {
+            name: 'Build',
+            id: buildStepId,
+            run: [
+              `${project.runTaskCommand(project.buildTask)} && ${this.setOutput(buildConclusionOutputName, 'success')} || ${this.setOutput(buildConclusionOutputName, 'failure')}`,
+              `${this.setOutput(buildLogOutputName, 'https://github.com/${REPO}/actions/runs/${{ github.run_id }}')}`,
+            ].join('\n'),
+            env: {
+              GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}',
```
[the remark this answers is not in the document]
iliapolo: Its not, letfovers.

src/upgrade-dependencies.ts
iliapolo: I think so, let me try.

src/upgrade-dependencies.ts
iliapolo: Yes - its used in the authentication header of the `curl` command.

src/upgrade-dependencies.ts
iliapolo: With a little work, yes - but i'll skip it for now.

package.json
```diff
@@ -18,7 +18,6 @@
     "build": "/bin/bash ./projen.bash build",
     "test:watch": "/bin/bash ./projen.bash test:watch",
     "test:update": "/bin/bash ./projen.bash test:update",
-    "projen:upgrade": "/bin/bash ./projen.bash projen:upgrade",
```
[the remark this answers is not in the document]
iliapolo: What's the meaning of a `projen:upgrade` task in projen itself? It uses its own source code.

src/node-project.ts
iliapolo: I've been struggling with this :\ 

`ncu` also feels funky cause its an implementation detail - if we switch to `npm update` would we have to change the name? 

I'll change to `DependencyUpgradeMechanism.githubWorkflow` and let me know what you think.

src/upgrade-dependencies.ts
iliapolo: Added

src/upgrade-dependencies.ts
iliapolo: Added

src/upgrade-dependencies.ts
iliapolo: It simply creates a local projen task and upgrades can be done manually by running `yarn upgrade-dependencies`. It's not related to dependabot. Added to comment.

src/upgrade-dependencies.ts
iliapolo: So that when a user explicitly sets `exclude` - they are not implicitly including `projen`. This seemed simpler, and is consistent with how the dependabot component configures it.

src/upgrade-dependencies.ts
iliapolo: 👍

.projen/tasks.json
iliapolo: No - since projen itself doesn't need a `projen:upgrade` task..unless i'm missing something - this never belonged here.

src/upgrade-dependencies.ts
iliapolo: 👍

src/node-project.ts
iliapolo: Because otherwise projen is already being taken care of. Note the code below only configures a projen upgrade.

src/upgrade-dependencies.ts
iliapolo: Yes, scoped down to: 

- `content` (push branch)
- `pull-request` (create pr)
- `checks` (update status check - if needed)

iliapolo:
@eladb Addressed all your comments.

Note that I changed the default `dependabot` to `false` - so you'll see an `upgrade-dependencies.yml` workflow here instead. 

Capturing the outcome after this PR:

- When `dependabot` is `true` - use dependabot.
- When `dependabot` is `false` - use new workflow. This might be a little surprising - should we switch to not use anything at all? 
- When `dependabot` is `undefined` - users get the new workflow.

In all cases users get a projen upgrade workflow like they have now in case a projen secret is configured.

.projen/tasks.json
eladb: I meant the "upgrade-dependencies" task

eladb (approved):
Ship it 🛳️

PR #653, opened 2021-03-29
feat: include projen markers in synthesized files by default

Chriscbr:
Changes the default behavior for structured files generated by projen (`JsonFile`, `YamlFile`, etc.) so that including a "Generated by projen" marker is the default behavior.

This can be disabled with the `{ marker: false }` option when you are creating a file object. If it has already been created, it can be disabled with `someFile.marker = false`.

This change does not affect text files or sample files generated by projen.

The `IMarkableFile` interface has also been removed.

---


src/object-file.ts
```diff
@@ -36,7 +42,7 @@ export abstract class ObjectFile extends FileBase implements IMarkableFile {
    * Indicates if the projen marker JSON-comment will be added to the output
    * object.
    */
-  public readonly marker: boolean;
```
Chriscbr, line 39: I removed the "readonly" field so people can change/disable the markers for built-in components if needed (as an escape hatch). Otherwise there's no way to remove them.

There is some precedent for this (see this mutable `readonly` field): https://github.com/projen/projen/blob/539fa2c9331bd337019938adacc09e841363adff/src/jsii-project.ts#L294-L297

To resolve a JSII issue this caused, I also dropped the IMarkableFile interface.

Chriscbr (commented):
Totally open to thoughts on this, and whether it's even necessary (not totally sure if this is better/worse for the UX). I think it could be a healthy default since including the projen tag will cause the files to get cleaned up if they're no longer needed. But some programs do complain about these comments (e.g. eslint will error if you include PROJEN_TAG).

src/__tests__/dev-env.test.ts
```diff
@@ -82,7 +83,10 @@ describe('dev environment docker options', () => {
     expect(gitpodSnapshot).toContain('image: jsii/superchain');
 
     const devContainerSnapshot = synthSnapshot(project)[DEVCONTAINER_FILE];
-    expect(devContainerSnapshot).toStrictEqual({ image: 'jsii/uberchain' });
+    expect(devContainerSnapshot).toStrictEqual({
+      '//': FileBase.PROJEN_MARKER,
+      'image': 'jsii/uberchain',
+    });
```
Chriscbr, line 89: An alternative to adding the projen marker here is to use jest's `toMatchObject`, whose semantics lets us check that the object on the left side has _at least_ the properties/values of the object on the right side.

src/object-file.ts
eladb: Cool

eladb (approved):

pgollucci:
This has downstream effects on nextjs et al

info  - Checking validity of types  
Warning: Your PostCSS configuration defines a field which is not supported (`//`). Please remove this configuration value.
Warning: Your PostCSS configuration defines a field which is not supported (`//`). Please remove this configuration value.

pgollucci:
@Chriscbr

Chriscbr:
@pgollucci Thanks for catching this. I'll submit a PR to fix this

PR #681, opened 2021-04-13
chore(cdk8s-construct): beta3 to beta10

pgollucci:


.devcontainer.json
```diff
@@ -1,4 +1,3 @@
 {
-  "postCreateCommand": "( npx projen devenv:setup )",
```
eladb, line 2: This still looks fishy

eladb (changes requested):

.devcontainer.json
pgollucci: I agree; I have the same sha's as the source.

PR #682, opened 2021-04-13
chore(cdk8s-construct): beta3 to beta10

pgollucci:


pgollucci:
I reforked it; seemed to fix it. 3rd time is the charm

eladb (approved):

PR #697, opened 2021-04-17
fix(nextjs): honor the default of tailwind: true

pgollucci:
if options tailwind: true wasn't present tailwind was off


Chriscbr (approved):

eladb (approved):
Would be nice to add some tests for nextjs components...

pgollucci:
I'm getting there...........

PR #704, opened 2021-04-20
fix(nextjs)!: Only tailwind.config.js/tailwind.js are valid config names

pgollucci:
BREAKING CHANGE: TailwindConfigOptions no longer accepts any fileName

 https://tailwindcss.com/docs/installation
 https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/tailwindcss-intellisense/README.md

### DIFF
--- a/API.md
+++ b/API.md
@@ -82,7 +82,7 @@ Name|Description
 [web.ReactProject](#projen-web-reactproject)|React project without TypeScript.
 [web.ReactTypeDef](#projen-web-reacttypedef)|*No description*
 [web.ReactTypeScriptProject](#projen-web-reacttypescriptproject)|React project with TypeScript.
-[web.TailwindConfig](#projen-web-tailwindconfig)|Declares a Tailwind CSS configuration file.
+[web.TailwindConfig](#projen-web-tailwindconfig)|Declares a Tailwind CSS JS configuration file.
 
 
 **Structs**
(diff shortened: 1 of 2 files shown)


pgollucci:
Extension__Tailwind_CSS_IntelliSense_—_p6m7g8_github_io

They do not agree.  What is in this PR is a subset of this list.

eladb (commented):
@Chriscbr thoughts?

Chriscbr:
I'm not really clear on this change -- what is the benefit by restricting the filename to an enum?

Also not sure about the change to using a sample file. I think there could still be value in providing a higher-level API for configuring one's tailwind configuration (e.g. for adding plugins), and if the file isn't managed that would not be possible.

What do you think? @pgollucci

pgollucci:
Oh yeah, SampleFile was just a quick flip.  Definitely in favor of the API.

The point of this PR is not every filename works.  Only select ones.  Which names work actually vary by thing; however, the two above are supported by ALL.

tailwind
postcss
intellinse
...

I'm not tied to this PR don't force it.  I spent the better part of a day debugging why tailwind intellisense didn't work.

My recently stint of nextjs prs is all stemming from that.  The funny part is after you do _ALL_ that, you have to set something in .vscode/settings.json for it to work in react aka strings

```js
  "editor.quickSuggestions": {
    "strings": true
  },
```

un related but recommended to prevent double runs and errors in non css files.
```js
  "css.validate": false,
```

eladb (changes requested):
Tests?

eladb:
Updates?

eladb:
Closing. Reopen when ready to continue

PR #707, opened 2021-04-21
fix(release): unable to upload all files to release

eladb:
Since `dist/` may contain a large number of files (esp. for java), we decided to disable this feature for now.

---


Chriscbr:
@eladb is the size of the files or the number of files the issue? Is it possible we could zip the java artifacts together? (Or is this something that should really be automated by jsii?)

eladb:
Yes, I think we should probably zip `dist/` and then we can upload. It's the number of files.

Issue #696, opened 2021-04-17
(NodeProject) projen silently fails to remove .mergify.yml when setting `mergify: false`


ilyasotkov:
### Steps to reproduce

In an empty directory:

1. `npx projen new node`.
2. In `.projenrc.js`, add `mergify: false` to `NodeProject` options, then run `npx projen`.

### Observed behavior

`.mergify.yml` is still there

### Expected behavior

`.mergify.yml` is removed

### Other notes

I also tried setting e.g. `npmignoreEnabled: false` and this works as expected. I also tried manually removing `.mergify.yml` but it's still generated by projen despite `mergify: false`.

I can reproduce this in GitHub Codespaces using `node:14.16.1-buster-slim` Docker image.

ilyasotkov:
I ran this through a debugger and discovered that `.mergify.yml` is being generated here

https://github.com/projen/projen/blob/b9104a6550b0c0685b43cf21e372843329116970/src/github/dependabot.ts#L168-L184

and that setting `dependabot: false` in `NodeProject` options stops  `.mergify.yml` from being generated.

pgollucci:
Tagging myself @pgollucci

ilyasotkov:
Closing, fixed by #741

PR #714, opened 2021-04-21
fix: honor mergify: false when dependabot: true

pgollucci:
Fixes: https://github.com/projen/projen/issues/696

Its perfectably legit to have dependabot on but not mergify, or
mergify to not be involved in that flow.

I will add a test, but am I blind, I don't see _any_ dependabot, mergify
tests??

### DIFF
--- a/src/node-project.ts
+++ b/src/node-project.ts
@@ -709,10 +709,10 @@ export class NodeProject extends Project {
       });
 
       this.npmignore?.exclude('/.mergify.yml');
-    }
 
-    if (options.dependabot ?? true) {
-      this.github?.addDependabot(options.dependabotOptions);
+      if (options.dependabot ?? true) {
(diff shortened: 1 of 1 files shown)


src/node-project.ts
```diff
@@ -709,10 +709,10 @@ export class NodeProject extends Project {
       });
 
       this.npmignore?.exclude('/.mergify.yml');
-    }
 
-    if (options.dependabot ?? true) {
```
eladb, line 714: It seems like this implies that if mergify is `false` dependabot is never enabled, no?

eladb (changes requested):

src/node-project.ts
pgollucci: I will take another look. I didn't check too closely before.

eladb:
@RomainMuller #741 supersedes this PR, correct? Can we close this one?

PR #727, opened 2021-04-25
feat: commitlint

Hunter-Thompson:
It would be great to be able to enforce conventional commits on a project. 

The maintainer of a project can disable pushing to the main branch and squash merge a PR with a conventional title, this works well with opensource projects since the contributions are done via forks.

But when the project is private and contributors are not used to following the convention, it helps to enforce it using husky hooks.

Usage:

```
git init
../projen/bin/projen new typescript --commitlint=true
```

Husky Git hooks are only enabled if a `.git` folder exists within the repo.

Usage without Husky : 

```
../projen/bin/projen new typescript --commitlint=true
echo 'foo' | commitlint
```

 
---


eladb:
@Hunter-Thompson thanks for this contribution. We will be able to look into it in the coming weeks!

pgollucci:
Is this different from semantic/conventional commits?  Or I've just realized that its an integration and not in gh worksflows so this moves into projen world.

I support this.

Hunter-Thompson:
> Is this different from semantic/conventional commits? Or I've just realized that its an integration and not in gh worksflows so this moves into projen world.
> 
> I support this.

User has the ability to enforce conventional commits based on a default config which is `@commitlint/config-conventional`, or they can add their own config.

If it's a Github repo, Husky will also be added, and Husky will create Githooks which will enforce conventional commits. It does that by running `commitlint` before every commit, if the commit follows the config provided, it will commit, otherwise it will throw an error.

PR #742, opened 2021-04-28
fix(awscdk-lib): incorrect version of constructs used for CDK 2.0

RomainMuller:
CDK 2.0 pre-releases depend on constructs@10, and not constructs@3. This
selects the version of constructs to be registered based on the CDK major
version being used.

Additionally, a new `constructsVersion` option enables user-defined
versions to be used instead of the default/inferred version.

---

### DIFF
--- a/src/awscdk-construct.ts
+++ b/src/awscdk-construct.ts
@@ -16,7 +16,7 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    * the default value depends on the configured `cdkVersion`:
    *
    * - For CDK 1.x, the default is "3.2.27"
-   * - For CDK 2.x, the default is "10.0.0"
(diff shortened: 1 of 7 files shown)


src/awscdk-construct.ts
```diff
@@ -11,6 +11,20 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    */
   readonly cdkVersion: string;
 
+  /**
+   * Minimum target version of constructs being tested against. If not provided,
+   * the default value depends on the configured `cdkVersion`:
+   *
+   * - For CDK 1.x, the default is "3.2.27"
+   * - For CDK 2.x, the default is "10.0.0"
```
eladb, line 19: I think the default for 2.x should be 10.0.0.5 (because of the mishap we had this week)

eladb (approved):

PR #743, opened 2021-04-28
feat: mark generated files with the linguist-generated

### LINKED ISSUE #629: Set GitHub-linguist attributes for generated files
To improve the PR review experience, it would be nice if `projen` maintained a `.gitattributes` file that enabled the `linguist-generated` attribute on files it is in control of (this can include `yarn.lock` files, even through "technically" not generated by `projen` - they are not covered by the default *generated* behavior).

The advantage of this are:
- Diff on those files will be hidden by default on PRs (you can still click-to-load if you want to control them)
- They get excluded from repo statistics (so the "hand-written" code - which relates to maintenance cost / quality - is better reflected)

Ref: https://github.com/github/linguist/blob/master/docs/overrides.md

RomainMuller:
Using github-linguist overrides to exclude generated files from
repository statistics (hence providing a better sense of the actual
human-maintained code size), and improves the PR review experience by
hiding generated files' diff by default.

Fixes #629

---


src/node-package.ts
```diff
@@ -390,6 +390,10 @@ export class NodePackage extends Component {
     this.packageManager = options.packageManager ?? NodePackageManager.YARN;
     this.entrypoint = options.entrypoint ?? 'lib/index.js';
 
+    if (this.packageManager === NodePackageManager.YARN) {
```
eladb, line 393: What about package-lock.json?

eladb (approved):

PR #744, opened 2021-04-28
feat(typescript): projenrc in typescript

### LINKED ISSUE #671: projenrcJava option is misleading/wrongly documented
Trying out projen in my company's ecosystem, and I love the capabilities, but I hit an early snag that caught me a bit by surprise for Java projects.

Documentation in the projen repository highlights that changes to the project structure can be authored in the .projenrc.js file, but I found when using minimal options for Java projects, once the project is synthesized once, it no longer uses this file, and instead defaults to using the src/test/projenrc.java file.  No updates in the .projenrc.js file will ever be reflected in the project, which frustrated me for quite a while, until I found the --projenrc-java false option.

Shouldn't this flag should be false by default?  Otherwise Java projects operate differently than other projects out of the box, and might surprise (and possibly deter users trying to follow the documentation).  Until I found this option, I was convinced that Java project support wasn't working properly.

eladb:
TypeScript projects (and derivatives) now support `--projenrc-ts` which means that their projenrc file is written in TypeScript instead of JavaScript.

To enable this change, we needed to complete the work started to enable Java projenrc files, in which the responsibility of generating the rc file have moved from the CLI to the project type. To that end, the CLI will inject a special field (`__new__`) which will include the FQN of the project type and the CLI arguments, so that it will be possible to generate the rc file. Relevant code was extracted from the CLI and reused both for JavaScript rc files and TypeScript rc files. Additionally, the `default` task is added to tell `projen` what to do when it is executed without arguments.

Resolves #671 (no `.projenrc.js` file for java projects anymore, which was misleading)

---


src/python/python-project.ts
```diff
@@ -248,6 +249,10 @@ export class PythonProject extends Project {
     }
 
     this.addDefaultGitIgnore();
+
+    // python currently only supports projenrc in java (would be great to
+    // support .projenrc.py of course).
+    new Projenrc(this);
```
Chriscbr, line 255: Is this meant to say javascript, not java?

With this PR, the python project default task fails because `node .projenrc.js` cannot find module 'projen'. In previous versions, `node_modules` is created with a local copy of projen, but with these changes, `node_modules` isn't created anymore, hence the error.

I don't think this happy path is critical since we should be using .projenrc.py soon, but worth mentioning in case this is breaking anyone...

eladb: Good catch [can we add a test to protect us against such regression in the future?]. The workaround, if someone hits this, is to manually install projen through `npm install projen` and then `projen` should work.

PR #747, opened 2021-04-29
feat: only show featured options in new projenrc files by default

### LINKED ISSUE #746: --no-comments flag is ignored
Prior to v0.17.70, this would generate a .projenrc.js file with no commented options. But now, commented options are generated when you first run `npx projen <project-type> --no-comments`.

Chriscbr:
Showing only featured options is now the default for new projen projects. Options for projects can be annotated as featured by including the @featured tag in the docstring.

Also, fixes #746.

---


src/cli/cmds/new.ts
```diff
@@ -290,7 +290,7 @@ async function newProject(baseDir: string, type: inventory.ProjectType, args: an
     dir: baseDir,
     type,
     params: props,
-    comments: args.comments,
+    comments: args.comments ? ProjectOptionsVerbosity.FEATURED : ProjectOptionsVerbosity.NONE,
```
Chriscbr, line 293: I added an enum to represent all three ways of rendering the commented options (all options, featured options, or none options), but it might be superfluous given the "comments" flag has to be boolean (in this code I am doing the conversion) for usage in the CLI.

Maybe we should just scrap this enum and the code which renders all of the options altogether?

src/node-package.ts
```diff
@@ -166,6 +170,7 @@ export interface NodePackageOptions {
    * The Node Package Manager used to execute scripts
    *
    * @default NodePackageManager.YARN
+   * @featured
```
eladb, line 173:
```suggestion

```

src/node-package.ts
```diff
@@ -228,7 +233,10 @@ export interface NodePackageOptions {
   /**
    * License's SPDX identifier.
    * See https://github.com/projen/projen/tree/master/license-text for a list of supported licenses.
+   * Use the `licensed` option if you want to no license to be specified.
+   *
    * @default "Apache-2.0"
+   * @featured
```
eladb, line 239:
```suggestion

```

src/typescript.ts
```diff
@@ -76,6 +76,7 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
 
   /**
    * Custom TSConfig
+   * @featured
```
eladb, line 79:
```suggestion

```

src/javascript/render-options.ts
```diff
@@ -1,6 +1,27 @@
 import * as inventory from '../inventory';
 
 const PROJEN_NEW = '__new__';
+const TAB = makePadding(2);
+
+/**
+ * Choices for how to display commented out options.
+ */
+export enum ProjectOptionsVerbosity {
```
eladb, line 9:
```suggestion
export enum OptionHints {
```

eladb (commented):
Yey!

Chriscbr:
Comments have been addressed

API.md
```diff
@@ -175,6 +175,7 @@ Name|Description
 [java.PomOptions](#projen-java-pomoptions)|Options for `Pom`.
 [java.ProjenrcOptions](#projen-java-projenrcoptions)|Options for `Projenrc`.
 [javascript.ProjenrcOptions](#projen-javascript-projenrcoptions)|*No description*
+[javascript.RenderProjectOptions](#projen-javascript-renderprojectoptions)|Options for `renderProjectOptions`.
```
eladb, line 178: This should be updated?

src/javascript/projenrc.ts
```diff
@@ -14,23 +14,23 @@ export interface ProjenrcOptions {
 
   /**
    * Include commented out properties.
-   * @default true
+   * @default ProjectOptionsVerbosity.FEATURED
```
eladb, line 17:
```suggestion
   * @default OptionHints.FEATURED
```

src/javascript/render-options.ts
```diff
@@ -18,9 +39,9 @@ export interface RenderProjectOptions {
 
   /**
    * Include commented out options.
-   * @default false
+   * @default ProjectOptionsVerbosity.FEATURED
```
eladb, line 42:
```suggestion
   * @default OptionHints.FEATURED
```

API.md
Chriscbr: Agh, I keep assuming `yarn build` will update my API.md for me. On it

src/javascript/render-options.ts
```diff
@@ -96,21 +115,15 @@ export function renderJavaScriptOptions(opts: RenderProjectOptions) {
...
+      renders[optionName] = `${optionName}: ${js},`;
       optionsWithDefaults.push(optionName);
     } else {
       const defaultValue = option.default?.startsWith('-') ? undefined : (option.default ?? undefined);
-      paramRender = `// ${optionName}: ${defaultValue?.replace(/"(.+)"/, '\'$1\'')},`; // single quotes
-
-      const parentModule = option.parent;
-      optionsByModule[parentModule] = optionsByModule[parentModule] ?? [];
-      optionsByModule[parentModule].push(option);
+      renders[optionName] = `// ${optionName}: ${defaultValue?.replace(/"(.+)"/, '\'$1\'')},`; // single quotes
```
eladb, line 125: Can we reuse the `.replace(...)` code between here and line 120?

src/typescript/projenrc.ts
```diff
@@ -13,23 +13,23 @@ export interface ProjenrcOptions {
 
   /**
    * Include commented out properties.
-   * @default true
+   * @default ProjectOptionsVerbosity.FEATURED
```
eladb, line 16:
```suggestion
   * @default OptionHints.FEATURED
```

API.md
```diff
@@ -9748,11 +9750,27 @@ Name | Type | Description
 
 Name | Type | Description 
 -----|------|-------------
-**comments**?🔹 | <code>boolean</code> | Include commented out properties.<br/>__*Default*__: true
+**comments**?🔹 | <code>[javascript.OptionHints](#projen-javascript-optionhints)</code> | Include commented out properties.<br/>__*Default*__: ProjectOptionsVerbosity.FEATURED
 **filename**?🔹 | <code>string</code> | The name of the projenrc file.<br/>__*Default*__: ".projenrc.js"
 
 
 
+## struct RenderProjectOptions 🔹 <a id="projen-javascript-renderprojectoptions"></a>
```
eladb, line 9758: Where is this coming from?

src/javascript/index.ts
```diff
@@ -1 +1,2 @@
-export * from './projenrc';
\ No newline at end of file
+export * from './projenrc';
+export * from './render-options';
```
eladb, line 2: The `render-options` file should not be exported here - it contains internal stuff. If it's needed from another file inside the project, just reference it directly and not through the `javascript` directory

API.md
eladb: It's supposed to!

Chriscbr: Err, you're right lol. Misread which line this was referring to.

API.md
Chriscbr: Hmmm so I put `OptionHints` in `render-options.ts`, but this file wasn't being exported, so after exporting it this interface also got exported. I can move OptionHints somewhere else if we don't want to expose RenderProjectOptions.

src/javascript/index.ts
Chriscbr: Ah cool 👍

API.md
eladb: Yes. Will likely need to go somewhere else and then will also need a new name such as `NewProjectOptionHints`

src/javascript/index.ts
Chriscbr: So I couldn't just reference OptionHints directly from `render-options` because `OptionHints` is getting used in several exported interfaces (and `render-options` isn't exported), which is a jsii error.

I also couldn't move it to javascript/projenrc.ts because this caused a runtime dependency cycle.

So I've moved it to it's own file in the root directory for now.

src/javascript/render-options.ts
Chriscbr: Done - I created a helper

API.md
Chriscbr: Just saw this -- yep this sounds better, changing

Chriscbr:
Have to run, but can address any other issues this evening

src/javascript/projenrc.ts
```diff
@@ -14,23 +14,23 @@ export interface ProjenrcOptions {
 
   /**
    * Include commented out properties.
-   * @default true
+   * @default NewProjectOptionHints.FEATURED
    */
-  readonly comments?: boolean;
```
eladb, line 19: so... I think this option should be passed as part of the `project.newProject` field and not through the normal options route (similar to `jsiiFqn`) because we only care about this when we create new projects. Does that make sense?

eladb (changes requested):

src/javascript/projenrc.ts
Chriscbr: I think so - just had to spend a little time to wrap my head around how the bootstapping works. There's a bit of overlap between some of the interfaces (like `NewProject`, `ProjenNew`, and `RenderProjectOptions`) but I wasn't sure if it would be right to merge them or not somehow

Chriscbr:
The updated changes now fix #746.

src/javascript/projenrc.ts
```diff
@@ -55,7 +46,7 @@ export class Projenrc extends Component {
     const className = parts.slice(1).join('.');
 
     const js = renderJavaScriptOptions({
-      comments: this.comments,
+      comments: bootstrap.comments,
```
eladb, line 49: 👍

eladb (approved):
Great job. That's exactly what I had in mind :-)

PR #755, opened 2021-05-04
feat(ts): add prettier option for formatting

MrArnoldPalmer:
Adds a `prettier` option to `EsLintOptions` to enable using prettier
code formatting embedded into the `eslint --fix` task. This will cause
code linted with eslint to format using the default prettier
configuration.

Additionally removes all eslint errors related to style handled by
prettier formatting when enabled.

---


MrArnoldPalmer:
We want to use prettier on a couple of repos and integrating it with eslint seems to the be most straightforward option, though this could be added as a separate component. Since eslint is a test step and the CI steps add commits to PRs post build with any changes, this should allow users to submit code with their own styles for line-lengths, quotes, etc and these would automatically be normalized.

Tested this by creating a new react-ts project and by adding `prettier: true` to the eslint options in the projen repo. If we want to enable prettier in projen itself we can by adding that option. Targetting files for formatting besides those that are linted, IE projen managed files, was hard for me to figure out as a new contributor.

eladb (changes requested):
1. Unit test?
2. Enable for projen?

MrArnoldPalmer:
@eladb you mean format projen source files with prettier? I'm fine with that but it includes a ton of changes, I'd say lets do it in a separate PR?

eladb (approved):

PR #756, opened 2021-05-04
fix(awscdk-app-ts): use posix path for app in cdk.json file

mmcculloh-dms:
When multiple people are working in a mixed OS team on a single CDK project, `cdk.json` will constantly have changes to the path used for the app due to it switching between `/` and `\`. As far as I know there's no reason this path can't just always use `/` since windows supports it (in this context, atleast).

---


eladb (approved):

mmcculloh-dms:
@Mergifyio update

Issue #754, opened 2021-05-03
web projects: "build" tasks do not run projen


eladb:
The various web projects' "build" task overrides the default build task from `Projen` which means that now `projen` is not executed during build (or at all).

I am wondering if we need a more robust model to manage these tasks, but as an interim solution, maybe we can add a switch that will allow disabling the execution of `compile` and `test` as part of `build` (at the `NodeProject` level) and then the web projects will be able to _append_ instead of overriding.

It will also be good to now allow creating new tasks if a task already exist (`addTask` should fail) and add a `deleteTask` to allow explicitly deleting existing tasks if that's the intent.

eladb:
@Chriscbr what do you think?

Chriscbr:
I see, so the current way web projects override the build task somewhat nullifies the rest of the library's effort to compose the `build` task out of smaller pieces -- so changing it to some kind of "append" operation would fix that. I agree.

> It will also be good to now allow creating new tasks if a task already exist (`addTask` should fail) and add a `deleteTask` to allow explicitly deleting existing tasks if that's the intent.

Sounds good. Depending on the implementation, like if it were possible to `deleteTask('build')` and subsequently `addTask('build')`, I think it's possible to have weird sequencing effects where if a component adds a command to the `build` task, and the user deletes and re-adds the task, then it could appear as though the component isn't working. But I suppose this kind of self-sabotaging is already possible with `reset()` anyway.

PR #757, opened 2021-05-04
feat(tasks): add removeTask method

mmcculloh-dms:
Simple method. Unsure of the general philosophy of projen on whether or not this should throw if you try to delete something that doesn't exist. I went with my personal choice, which is to handle that. I would be happy to change it if needed.

Also, chose `remove` instead of `delete` since `remove` is already used for other methods (e.g. `removeScript`)

Addresses a missing functionality, and was mentioned in #754 as well

---


Chriscbr (changes requested):
Thanks for the PR! 😄 

re: throwing -- I could see it going either way.  It's also possible we could return the deleted `Task` if a task is found (and return undefined otherwise).

I think it might be useful to add a test case to check the behavior when a task is removed and then re-added.

@eladb should anything be done to handle other components "depending" on particular tasks? ex. the github workflow `build.yml` would fail if the `build` task is removed (since the task's name is inserted as a string). Kind of minor edge case.

mmcculloh-dms:
Thanks @Chriscbr, made some changes to address your feedback.

> It's also possible we could return the deleted Task if a task is found (and return undefined otherwise).

I'm down for that, I think it's more useful than simple boolean.

> should anything be done to handle other components "depending" on particular tasks?

I know this isn't what you were referring to, but you made me realize that this should probably throw when trying to remove a task that another task depends on. Added handling of that as well.

eladb (approved):
Looks great

mmcculloh-dms:
@Mergifyio update

PR #764, opened 2021-05-06
fix(bump): invalid version selected for multiple tags

### LINKED ISSUE #762: bump: wrong version is selected if there are two versions on the same commit
When resolving the latest version, if the last tag is on the same commit as the previous tag, then the wrong version is chosen.

Repro:

```bash
$ git tag v1.0.0
$ git tag v1.0.1
$ yarn bump
✔ bumping version in package.json from 0.0.0 to 1.0.1
```

Since both tags `v1.0.0` and `v1.0.1` are on the same commit, the `git describe` returns only the first one lexicographically, which will likely not be the one we want.

eladb:
If there are multiple version tags on the same commit, `git describe` would return the first tag which is lexicographically first. This is not what we want, so we now use `git tag` with a `version:refname` sort order in order to query the repo for the latest version (for real).

Fixes #762

---

### DIFF
--- a/src/version.ts
+++ b/src/version.ts
@@ -31,7 +31,7 @@ export class Version extends Component {
       '-c "versionsort.suffix=-"', // makes sure pre-release versions are listed after the primary version
       'tag',
       '--sort="-version:refname"', // sort as versions and not lexicographically
-      '--list "v*"', // only tags that start with "v"
+      '--list="v*"', // only tags that start with "v"
     ].join(' ');
 
     this.bumpTask.exec(`${listGitTags} | head -n1 > ${versionFile}`);
--- a/src/version.ts
+++ b/src/version.ts
@@ -26,10 +26,15 @@ export class Version extends Component {
       condition: changesSinceLastRelease,
     });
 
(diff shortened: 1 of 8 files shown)


src/version.ts
```diff
@@ -26,10 +26,15 @@ export class Version extends Component {
...
-    // the default for "standard-version" is to find a tag across the entire
-    // repository but we want to select the last tag accessible from the
-    // *current branch*.
-    this.bumpTask.exec(`git describe --tags --match="v*" --first-parent --abbrev=0 > ${versionFile}`);
+    const listGitTags = [
+      'git',
+      '-c "versionsort.suffix=-"', // makes sure pre-release versions are listed after the primary version
+      'tag',
+      '--sort="-version:refname"', // sort as versions and not lexicographically
+      '--list "v*"', // only tags that start with "v"
```
RomainMuller, line 34: Just so the long form parameters are consistently using an `=` delimiter :)


```suggestion
      '--list="v*"', // only tags that start with "v"
```

RomainMuller (approved):

PR #770, opened 2021-05-08
feat(cli): allow enum-based CLI arguments

Chriscbr:
---


src/javascript/render-options.ts
```diff
@@ -189,6 +194,31 @@ function renderCommentedOptionsInOrder(renders: Record<string, string>, options:
...
+/**
+ * Renders a CLI argument as a basic JavaScript value. It must either be a
+ * string, number, boolean, or enum.
+ *
+ * Returns a string and the name of any needed imports if needed as an
+ * object in the form { js, import }.
+ */
+function renderArgAsJavaScript(arg: any, option: inventory.ProjectOption) {
+  // devDeps added as an exception to handle bootstrapping projects from external modules
+  if (['string', 'number', 'boolean'].includes(option.type) || option.name === 'devDeps') {
```
Chriscbr, line 206: This is referencing the line of code here:

https://github.com/projen/projen/blob/46fa460ada2d23e01493f481aa994ba394c45051/src/cli/cmds/new.ts#L269-L273

The `--from cdk-appsync-project` unit test wouldn't pass unless I added this exception.

eladb (changes requested):
This is very cool!

A few notes:
1. Should we include the enum options in `--help`?
2. We need to add support for `projenrc.java`...

I am wondering if we need some form of a "compliance test" that will go through all projenrc types and make sure we don't miss future projenrc languages in the future if we add things like this.

eladb:
@Chriscbr let me know when this is ready for another review.

Chriscbr:
I didn't test if this implementation will handle enums that are defined without strings (i.e. `enum Foo { BAR }` instead of `enum Foo { BAR = 'bar' }`), but I don't think we have many enums like that so I don't think it should be an issue.

eladb:
> I didn't test if this implementation will handle enums that are defined without strings (i.e. `enum Foo { BAR }` instead of `enum Foo { BAR = 'bar' }`), but I don't think we have many enums like that so it shouldn't be an issue.

Are the names of the CLI switches derived from the member name or from the member value? I believe they should be derived from the member name, no?

eladb (approved):

Chriscbr:
Yeah, you're right. From the CLI standpoint, `yargs` just gives us a string passed by the user such as "lib" (which we expect is the member name), and in `Projenrc` it gets translated based on the jsii manifest into "ProjectType.LIB". So at no point does the underlying member value have an affect in this process.

I also did a quick manual test where I changed some member values, and the CLI still worked as expected.

PR #779, opened 2021-05-11
feat: generate tasks for deploy groups

hoegertn:
This feature add the new option `deployGroups` to any AwsCdkTypeeScriptApp.

It allows to defined named glob patterns that result in tasks to deploy/diff/destroy these stacks.

```ts
{
  prod: 'stacks-prod-*'
}
```

results in `deploy:prod`, `diff:prod`, and `destroy:prod` tasks that target all stacks starting with `stacks-prod-`

---


hoegertn:
I am happy to support this too. But I see the `*` a lot in the wild.

src/awscdk-app-ts.ts
```diff
@@ -66,6 +66,18 @@ export interface AwsCdkTypeScriptAppOptions extends TypeScriptProjectOptions {
...
+   * Projen will create deploy/diff/destroy tasks for these groups.
+   *
+   * The key is the name for the task and the value is the glob pattern to use.
+   *
+   * @example
+   * {
+   *   prod: 'stacks-prod-*',
+   * }
+   */
+  readonly deployGroups?: { [groupName: string]: string };
```
eladb, line 80: Let's rename this from "deploy groups" to "stack groups" since this is not only for deployment.

I'd also change the api to allow passing an array of patterns:

```ts
{
  prod: [ "prod-*", "production-*" ],
  // ...
}
```

src/awscdk-app-ts.ts
```diff
@@ -147,6 +159,16 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
       exec: 'cdk diff',
     });
 
+    if (options.deployGroups) {
+      for (const groupName in options.deployGroups) {
+        if (Object.prototype.hasOwnProperty.call(options.deployGroups, groupName)) {
```
eladb, line 164: Why is this needed?

src/awscdk-app-ts.ts
```diff
@@ -203,6 +225,26 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
   private formatModuleSpec(module: string): string {
     return `${module}@${this.cdkVersion}`;
   }
+
+  private addDeployTasks(groupName: string, globPattern: string) {
+    this.addTask(`deploy:${groupName}`, {
```
eladb, line 230: Validate group name does not have unwanted chars like spaces.

Probably `/[a-z0-9\-_]+/i`

eladb (changes requested):
Add unit tests please

src/awscdk-app-ts.ts
hoegertn: Can I specify multiple patterns on a `cdk deploy`?

src/awscdk-app-ts.ts
hoegertn: I learned this as best practice for `for..in` to make sure you iterate only over real fields and not internal stuff of the object. It is also what VSCode generates for loops. But I can get rid of it if you like.

src/awscdk-app-ts.ts
hoegertn: Shouldn't this be part of the Task class to validate that the name is valid?

eladb:
Closing for now. Reopen when ready to continue

Issue #498, opened 2021-01-17
Not working on Windows


davidsteed:
Get the following - which I think is related to the paths not being correct
```
npm ERR! code EPERM
npm ERR! syscall rename
npm ERR! path C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\.xmlbuilder2.DELETE\node_modules\@types
npm ERR! dest C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\xmlbuilder2\node_modules\@types
npm ERR! errno -4048
npm ERR! Error: EPERM: operation not permitted, rename 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\.xmlbuilder2.DELETE\node_modules\@types' -> 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\xmlbuilder2\node_modules\@types'
npm ERR!  [OperationalError: EPERM: operation not permitted, rename 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\.xmlbuilder2.DELETE\node_modules\@types' -> 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\xmlbuilder2\node_modules\@types'] {
npm ERR!   cause: [Error: EPERM: operation not permitted, rename 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\.xmlbuilder2.DELETE\node_modules\@types' -> 'C:\Users\david\AppData\Roaming\npm-cache\_npx\10136\node_modules\projen\node_modules\xmlbuilder2\node_modules\@types'] {
npm ERR!     errno: -4048,
npm ERR!     code: 'EPERM',
npm ERR!     syscall: 'rename',
npm ERR!     path: 'C:\\Users\\david\\AppData\\Roaming\\npm-cache\\_npx\\10136\\node_modules\\projen\\node_modules\\.xmlbuilder2.DELETE\\node_modules\\@types',
npm ERR!     dest: 'C:\\Users\\david\\AppData\\Roaming\\npm-cache\\_npx\\10136\\node_modules\\projen\\node_modules\\xmlbuilder2\\node_modules\\@types'
npm ERR!   },
[...]

PR #781, opened 2021-05-12
fix: some tasks fail on Windows due to missing shell commands

bracki:
This PR tries to address #452 by wrapping the commands 'rm', 'mkdir' and 'mv' with `shx`.
A whitelist of commands is maintained in the task runtime.

@eladb Is this what you had in mind when commenting on #540?


src/tasks/runtime.ts
```diff
@@ -87,10 +87,29 @@ class RunTask {
       }
 
       if (step.exec) {
-        const command = step.exec;
+        let command = '';
+        let hasError = false;
+        const cmd = step.exec.split(' ')[0];
+        if (['mkdir', 'mv', 'rm'].includes(cmd)) {
```
eladb, line 93: Yes!

1. Can we only do this on Windows?
2. Is there a way to ask shx which commands it supports?

bracki: You mean only prepend `shx` if `OS == Windows`? Sure. But that means not rendering to .projen/tasks.json, correct?

There's now white list, only a blacklist. So either we keep it that way, or pipe everything on Windows through 'shx'.

eladb: Will it be safe to simply run all commands through `shx` on windows?

bracki: I don't know. I believe so. But we can keep the whitelist.

eladb: Can you check maybe. If we can just pass through to shx for windows that would be perfect.

bracki:
@eladb I also tried to "refactor" by extracting the code, but it becomes gnarly so I'd rather let it be the way it is.

src/tasks/runtime.ts
bracki: This would be the way to find it out: https://github.com/shelljs/shx/blob/7877b59122f6bb4c0c03bc7b7663a5969a735aef/src/shx.js#L80

The alternative would be to replace `task.exec("rm -rf foo")` with `shell.rm('rf', 'foo')`. And do this for everything. The only thing that wouldn't be portable immediately is stuff like `git ls-files | head -n1`. 

See https://documentup.com/shelljs/shelljs#examples for examples.

eladb (approved):

bracki:
Oh man, this extremely frustrating. 
Basically got hit by
https://github.com/projen/projen/issues/498 after applying the change from https://github.com/projen/projen/pull/540#issuecomment-839557669.

We ditched projen because we are on a mixed team with mostly Windows developers. Overall kind of sad, because projen would help so much, especially inexperienced TypeScript/Javascript folks on Windows.

Issue #800, opened 2021-05-18
upgrade-deps build fails with "Your lockfile needs to be updated"


eladb:
The new workflow which was introduced in #652 fails in the "Build" step when dependencies are updated. This is because `projen build` will execute `yarn install --frozen-lockfile` after the `package.json` was changed, so we get this error:

```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

See example [here](https://github.com/cdklabs/cdk-triggers/runs/2608296205?check_suite_focus=true)

lym953:
I still see this error when a workflow runs `npx projen`. Could anyone help with this?
https://github.com/DataDog/datadog-cdk-constructs/actions/runs/13527637337/job/38014208796?pr=362

mrgrain:
> I still see this error when a workflow runs `npx projen`. Could anyone help with this?
> https://github.com/DataDog/datadog-cdk-constructs/actions/runs/13527637337/job/38014208796?pr=362

Pretty straight forward: Dependabot is attempting to upgrade a dependency these you have explicitly declared here: 

https://github.com/DataDog/datadog-cdk-constructs/blob/6814eead55bfc5d17a01a19aa16fd0bc03ae062c/.projenrc.js#L28

Projen protects you from this unintentional change.

lym953:
@mrgrain That answers my question. Thanks!

PR #801, opened 2021-05-18
fix(deps): frozen installation fails during ugprade workflow

iliapolo:
Fixes https://github.com/projen/projen/issues/800.

The problem was that projen itself can change `package.json` files due to its own requirements. 

> For example https://github.com/cdklabs/cdk-triggers/blob/main/.projenrc.js#L6

Which causes the `package.json` file to change after running `yarn install` - and a subsequent frozen installation would fail. 

This PR uses projen itself to install the dependencies and update the lock file, so this happens after all `package.json` mutations have happened. 

---


src/upgrade-dependencies.ts
```diff
@@ -118,7 +118,11 @@ export class UpgradeDependencies extends Component {
     }
 
     task.exec(ncuCommand.join(' '));
-    task.exec(this._project.package.renderInstallCommand(false));
+
+    // we use projen to install the dependencies and thus update the lock file,
+    // so we need to make it think its not running inside CI.
+    task.exec(`CI=0 ${this._project.projenCommand}`);
```
eladb, line 124:
```suggestion
    task.exec(`${this._project.projenCommand}`, {
      env: { CI: '0' }
    });
```

iliapolo: Cool

src/upgrade-dependencies.ts
```diff
@@ -118,7 +118,11 @@ export class UpgradeDependencies extends Component {
     }
 
     task.exec(ncuCommand.join(' '));
-    task.exec(this._project.package.renderInstallCommand(false));
```
eladb, line 121: Does that mean we can make this private again?

iliapolo: Yes

eladb (approved):

Issue #805, opened 2021-05-18
bug: tests failing for upgrade-dependencies.test.ts


Hunter-Thompson:
https://github.com/projen/projen/blob/2c99dc6540bb39dda1856a94421d0d9e22b43774/src/__tests__/upgrade-dependencies.test.ts#L7-L10

We are creating a test project, which by default creates the `.github/workflows/upgrade-dependencies.yml` file, we are also creating a new instance of `UpgradeDependencies`, which causes the test to fail with the following error because the file already exists. Im assuming.

```
Summary of all failing tests
FAIL lib/__tests__/upgrade-dependencies.test.js
  ● default options

    there is already a file under .github/workflows/upgrade-dependencies.yml

      80 |     const existing = project.root.tryFindFile(this.absolutePath);
      81 |     if (existing && existing !== this) {
    > 82 |       throw new Error(`there is already a file under ${path.relative(project.root.outdir, this.absolutePath)}`);
         |             ^
      83 |     }
      84 |
      85 |     const gitignore = options.editGitignore ?? true;

      at new FileBase (src/file.ts:82:13)
      at new ObjectFile (src/object-file.ts:58:5)
      at new YamlFile (src/yaml.ts:16:5)
      at new GithubWorkflow (src/github/workflows.ts:15:17)
      at GitHub.addWorkflow (src/github/github.ts:47:12)
      at UpgradeDependencies.createWorkflow (src/upgrade-dependencies.ts:133:29)
      at new UpgradeDependencies (src/upgrade-dependencies.ts:100:28)
      at Object.<anonymous>.test (src/__tests__/upgrade-dependencies.test.ts:10:3)

  ● custom options

    there is already a file under .github/workflows/upgrade-dependencies.yml

      80 |     const existing = project.root.tryFindFile(this.absolutePath);
      81 |     if (existing && existing !== this) {
    > 82 |       throw new Error(`there is already a file under ${path.relative(project.root.outdir, this.absolutePath)}`);
         |             ^
      83 |     }
      84 |
      85 |     const gitignore = options.editGitignore ?? true;

      at new FileBase (src/file.ts:82:13)
      at new ObjectFile (src/object-file.ts:58:5)
      at new YamlFile (src/yaml.ts:16:5)
      at new GithubWorkflow (src/github/workflows.ts:15:17)
      at GitHub.addWorkflow (src/github/github.ts:47:12)
      at UpgradeDependencies.createWorkflow (src/upgrade-dependencies.ts:133:29)
[...]

PR #806, opened 2021-05-18
feat: codeowners file

Hunter-Thompson:
Gives the ability to add a CODEOWNERS file on the repo.

```
  codeOwners: {
    owners: [{
      owners: [
        'Hunter-Thompson',
      ],
      patterns: '*'
    }]
  }


project.github.codeOwners.addOwners({
  owners: [
    'Raoul-Duke',
  ],
  patterns: '*.js'
})
```

---


Hunter-Thompson:
Builds failing due to #805

eladb (commented):
Wouldn't it makes sense to provide a higher level API than just an array of strings?

hoegertn:
Additionally, the default value doe snot really make sense imho. So maybe it would be enough to have one field with the content (high-level) and then create the file or not. I do not see much value in a boolean activation flag and the content option.

Hunter-Thompson:
> Wouldn't it makes sense to provide a higher level API than just an array of strings?

Could you provide an example? Il work on a Component based of the example.

> Additionally, the default value does not really make sense imho. So maybe it would be enough to have one field with the content (high-level) and then create the file or not. I do not see much value in a boolean activation flag and the content option.

Makes sense, will work on updating the PR.

eladb:
@Hunter-Thompson when I say higher-level API I mean something like this:

```ts
const owners = new CodeOwners(project);
owners.addOwners('*', '@global-owner1', '@global-owner2');
owners.addOwners('*.js', '@js-owner');
owners.addOwners('*.go', 'docs@example.com');
owners.addOwners('/build/logs/', '@doctocat');
owners.addOwners('docs/*', 'docs@example.com');
// ...
```

Additionally, we can offer an initializer API:

```ts
const owners = new CodeOwners(project, {
  owners: [
    { pattern: '*', owners: [ '@global-owner1', '@global-owner2' ] },
    { pattern: '*.js', owners: [ '@js-owner' ] },
  ],
});

owners.addOwners(...);
```

eladb (changes requested):
see comment

Hunter-Thompson:
> @Hunter-Thompson lmk if you'd like to follow up here or should we close for now

@eladb, will pick this up over the weekend.

Hunter-Thompson:
@eladb, made the required changes and updated the PR description. Let me know where you'd like changes.

src/github/codeowners.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+export class CodeOwners extends Component {
+  private owners: string[] = [];
+
+  constructor(github: GitHub, options: CodeOwnersProps) {
+    super(github.project);
+    options.owners?.forEach(this.addOwners.bind(this));
+  }
+
+  public addOwners(owner: Owner) {
```
eladb, line 22: Why not?

```ts
addOwners(pattern: string, ...owners: string[])
```

src/github/codeowners.ts
```diff
@@ -0,0 +1,31 @@
...
+import { GitHub } from './github';
+
+
+export interface Owner {
+  readonly patterns?: string;
+  readonly owners?: string[];
+}
+
+export interface CodeOwnersProps {
+  readonly owners?: Owner[];
```
eladb, line 12: Why not?

```ts
{
  owners: {
    pattern: [ owners ],
    ...
  }
  
}
```

eladb:
@Hunter-Thompson could you please update the APIs to the proposed option. I think the ergonomics are better.

Hunter-Thompson:
> @Hunter-Thompson could you please update the APIs to the proposed option. I think the ergonomics are better.

Sure thing, will make the changes when I get some time tomorrow.

eladb:
?

Hunter-Thompson:
> ?

Not getting time to work on this feature, please gimme a day or two.

Issue #811, opened 2021-05-19
bug: cyclic dependency error in Python distribution of projen


Chriscbr:
Since projen is built with jsii, it is also vended as a python library on pypi (see [here](https://pypi.org/project/projen/)).

However, it seems that using this module to synthesize even simple projects isn't possible due to cyclic dependencies. For example, if we install `projen` with pip and write a project definition in `.projenrc.py` like this:

```python
from projen import Project

project = Project()

project.synth()
```

Then running `python .projenrc.py` throws the following error:

```
Traceback (most recent call last):
  File "/Users/rybickic/Developer/projen-test/.projenrc.py", line 3, in <module>
    from projen import Project
  File "/Users/rybickic/Developer/projen-test/.env/lib/python3.9/site-packages/projen/__init__.py", line 323, in <module>
    from .deps import Dependencies as _Dependencies_53ce8fa0
  File "/Users/rybickic/Developer/projen-test/.env/lib/python3.9/site-packages/projen/deps/__init__.py", line 13, in <module>
    from .. import Component as _Component_2b0ad27f, Project as _Project_57d89203
ImportError: cannot import name 'Component' from partially initialized module 'projen' (most likely due to a circular import) (/Users/rybickic/Developer/projen-test/.env/lib/python3.9/site-packages/projen/__init__.py)
```

----

My best guess of the cause of this is that `Project` directly references `Component` in several of its methods, and likewise `Project` references `Component` in its constructor.

Is it possible we could resolve this cycle by changing one of these to an interface like `IProject` or `IComponent` instead?

I'm also curious how other JSII libraries like the AWS CDK go about avoiding (or resolving) cyclic dependencies like this.

----

Note: since this is just the first error that's thrown, it's not clear fixing this alone would be enough to get .projenrc.py working in full. When I ran `madge --circular --extensions ts src` it reports there are upwards of 45 potential circular dependencies as of posting (see below) -- though it's possible not all of these are relevant to synthesizing Python projects.

<details>

```
1) component.ts > project.ts > clobber.ts
2) project.ts > clobber.ts
3) component.ts > project.ts
4) file.ts > _resolve.ts
5) component.ts > project.ts > file.ts
6) project.ts > file.ts
7) component.ts > project.ts > gitpod.ts
8) project.ts > gitpod.ts
9) project.ts > gitpod.ts > yaml.ts > object-file.ts
10) project.ts > gitpod.ts > yaml.ts
11) yaml.ts
12) project.ts > ignore-file.ts
[...]

PR #812, opened 2021-05-19
feat(python): add experimental projenrc.py support

Chriscbr:
Actual synthesis with `.projenrc.py` is currently not functional because of #811, but this commit adds the CLI flag along with projenrc.py file generation so the feature can be experimentally tested going forward.

Since it's still not functional, I did not update the docs to mention `.projenrc.py` since it would probably cause more confusion than help.

---


eladb (approved):
Yey!

eladb:
The issue with cyclic dependencies is not going to be solved at the JSII level. We need to reorganize the submodules in PROJEN so that they don't have cycles between them... That's going to be a bit painful :-)

PR #821, opened 2021-05-24
fix: errors caused by ts-node v10 incompatibility with node v10

### LINKED ISSUE #820: projen new typescript --projenrc-ts fails on node v10
The cause appears to be that ts-node is unpinned, and it just updated from v9 to v10, and v10 does not support node v10 (which projen does support, and is running its tests using).

Related to an issue that is causing some new builds to fail, see #818

Chriscbr:
Fixes #820

This bug was also causing an integration test in the projen development repo to fail. However, just this fix alone doesn't (immediately) fix it because the integration test still tries to install ts-node v10. To mitigate this, I added `--no-post` to the test.

Relatedly, I also tried adding a unit/integ test for this happy path:
```ts
execProjenCLI(projectdir, ['new', 'typescript', '--projenrc-ts']);
const projenrc = directorySnapshot(projectdir)['.projenrc.ts'];
expect(projenrc).toMatchSnapshot();
```

... but running this failed because of a linking issue: `error Couldn't find any versions for "projen" that matches "^0.0.0"`. So this is the best I can do at the moment.

---


Chriscbr:
@eladb An alternative fix here could be to change the default minNodeVersion to v12. Node 10 isn't being actively maintained anymore (see <https://nodejs.org/en/about/releases/>) so I think it would be appropriate.

src/awscdk-app-ts.ts
```diff
@@ -178,7 +178,7 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
       this.tsconfig.exclude.push('cdk.out');
     }
 
-    this.addDevDeps('ts-node');
+    this.addDevDeps('ts-node@9');
```
eladb, line 181: Is that the same as `^9`? I suspect this means to pin to 9.0.0

Chriscbr: Whoops! That should be ^9

eladb (approved):

PR #824, opened 2021-05-26
feat(project): auto approve workflow

nija-at:
Set up an auto approve workflow that will approve pull requests with a
known label and from an allowed set of Github usernames.

Remove part of the auto merge functionality that allowed pull requests
with the 'auto-merge' label to be merged without an approval. `mergify`
does not support merging pull requests unapproved pull requests.
So this really never worked.

BREAKING CHANGE: The `mergifyAutoMerge` option in `NodeProjectOptions`
has been removed. To auto merge PRs, use the 'auto approve'
functionality via the `autoApproveOptions` option.

---


nija-at:
### Remaining work

- [ ] Test against a real project and check the auto approve actually works
- [ ] synth projen on projen
- [ ] error if the upgrade action identity is the same as approve action identity

src/github/auto-approve.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Component } from '../component';
+import { GithubWorkflow } from './workflows';
+import { Job, JobPermission } from './workflows-model';
+
+/**
+ * Options for 'AutoApprove'
+ */
+export interface AutoApproveOptions {
+  /**
+   * Only pull requests authored by these Github usernames will be auto-approved.
```
eladb, line 11: 1. This default might be surprising. For example, I am not sure it's a safe practice to auto-approve dependency upgrades.
2. Is there a way not to specify a user filter? E.g I want to auto approve all PRs with the defined label?

src/github/auto-approve.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * A GitHub secret name which contains a GitHub Access Token
+   * with write permissions for the `pull_request` scope.
+   *
+   * This token is used to approve pull requests.
+   *
+   * @default 'GITHUB_TOKEN' https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret
+   */
+  readonly secret?: string;
```
eladb, line 30: How can the default token approve PRs submitted by the GitHub user itself?

(Seems like the default behavior of auto-approving to only PRs from GitHub bot will just not work..)

src/github/auto-approve.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  readonly label?: string;
+
+  /**
+   * A GitHub secret name which contains a GitHub Access Token
+   * with write permissions for the `pull_request` scope.
+   *
+   * This token is used to approve pull requests.
+   *
+   * @default 'GITHUB_TOKEN' https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret
```
eladb, line 28: Do we really need to specify the token here? If the default token already has permissions to approve PRs?

src/project.ts
```diff
@@ -86,6 +86,22 @@ export interface ProjectOptions extends GitHubOptions {
    * @default {}
    */
   readonly logging?: LoggerOptions;
+
+  /**
+   * Sets up a Github workflow that automatically approves PRs that meet a certain
+   * criteria. See `autoApproveOptions` for the default criteria and change them.
+   *
+   * When enabled in combination with `mergify`, pull requests can be auto-approved
+   * and auto-merged. This is applied for the dependency upgrade PRs set up by projen.
```
eladb, line 95: I prefer that the default behavior won't be to auto approve dep upgrades as it poses a security risk

src/project.ts
```diff
@@ -86,6 +86,22 @@ export interface ProjectOptions extends GitHubOptions {
...
+   *
+   * @default false
+   */
+  readonly autoApproveEnabled?: boolean;
```
eladb, line 99:
```suggestion
  readonly autoApprove?: boolean;
```

src/github/auto-approve.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(project: Project, options: AutoApproveOptions = {}) {
+    super(project);
+
+    this.label = options.label ?? 'auto-approve';
+    const usernames = options.allowedUsernames ?? ['github-bot'];
+    const secret = options.secret ?? 'GITHUB_TOKEN';
+
+    if (!project.github) {
+      return;
```
eladb, line 47: I think this should probably fail and not just bail out.

Maybe you can simply require a GitHub object instead of Project (like GitHubWorkflow does)

eladb (changes requested):

src/github/auto-approve.ts
nija-at: > This default might be surprising. For example, I am not sure it's a safe practice to auto-approve dependency upgrades.

Good point. 

How about this - enable the auto approve workflow by default but do not add the 'auto-approve' label to any generated PRs by default. Does that work?

> Is there a way not to specify a user filter? E.g I want to auto approve all PRs with the defined label?

By providing an empty list. Is this good enough, or should there be clear option on the API?

src/github/auto-approve.ts
nija-at: It's here because of your previous comment because of the inability for default tokens to approve their own PRs. 
I'm still working on the best customer experience on configuring this.

src/project.ts
nija-at: Yes. see conversation https://github.com/projen/projen/pull/824#discussion_r641519102

src/project.ts
```diff
@@ -86,6 +86,12 @@ export interface ProjectOptions extends GitHubOptions {
    * @default {}
    */
   readonly logging?: LoggerOptions;
+
+  /**
+   * Configure the 'auto approve' workflow. See `autoApproveEnabled` to enable this workflow.
```
BenChaimberg, line 91: minor: autoApproveEnabled no longer exists

src/github/auto-approve.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  readonly label?: string;
+
+  /**
+   * A GitHub secret name which contains a GitHub Access Token
+   * with write permissions for the `pull_request` scope.
+   *
+   * This token is used to approve pull requests.
+   *
+   * Github forbids an identity to approve its own pull request.
```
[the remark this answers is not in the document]
nija-at: Yes, and that's what the message says.

What are you suggesting to do here? 

I weighed the benefits of making this optional vs mandatory. 
Leaving it optional would likely mean that the average experience being incorrectly configured tokens, lots of debugging and digging only to find that this property needed to be configured.
Making it mandatory, on the other hand, forces the user to read the documentation with a higher chance of correctly configuring it. They can still choose to set this to `GITHUB_TOKEN`.

Further, going from mandatory -> optional is backwards compatible while the opposite is not.

src/project.ts
nija-at:
```suggestion
   * Enable and configure the 'auto approve' workflow.
```

src/github/auto-approve.ts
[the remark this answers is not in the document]
eladb: Okay. Sounds good

src/github/auto-approve.ts
```diff
@@ -0,0 +1,80 @@
...
+   * Hence, this cannot be the default secret
```
eladb, line 29: It _can_ be the default token

nija-at:
```suggestion
   * If your project produces automated pull requests using the Github default token -
   * {@link https://docs.github.com/en/actions/reference/authentication-in-a-workflow `GITHUB_TOKEN` }
   * - that you would like auto approved, such as when using `depsUpgrade` property in `NodeProjectOptions`,
   * then you must use a different token here.
```

src/github/auto-approve.ts
```diff
@@ -0,0 +1,80 @@
...
+   * {@link https://docs.github.com/en/actions/reference/authentication-in-a-workflow 'GITHUB_TOKEN' }.
```
nija-at, line 30:
```suggestion
```

eladb (approved):

eladb:
@nija-at you'll need to rebuild locally and commit any changes (the self-mutation feature does not work for forks)

nija-at:
> @nija-at you'll need to rebuild locally and commit any changes (the self-mutation feature does not work for forks)

Done! @eladb

src/github/auto-approve.ts
```diff
@@ -0,0 +1,82 @@
...
+        uses: 'hmarr/auto-approve-action@v2.1.0',
+        with: {
+          'github-token': `\${{ secrets.${options.secret} }}`,
+        },
+      }],
+    };
+
+    const workflow = new GithubWorkflow(project.github, 'auto-approve');
+    workflow.on({
+      pullRequest: {
```
iliapolo, line 76: @nija-at This might be problematic. If the PR's are created with a custom token, than a user will likely want to pass the default github token here, which won't work because it has `read-only` permissions on this trigger. 

I think we should use `pull_request_target` to accommodate this use case as well.  See https://github.com/aws/aws-cdk/pull/14939

nija-at: changed.

src/__tests__/__snapshots__/integ.test.ts.snap
```diff
@@ -611,7 +611,7 @@ jobs:
             *Automatically created by projen via the \\"upgrade-projen\\" workflow*
           branch: github-actions/upgrade-projen
           title: \\"chore(deps): upgrade projen\\"
-          labels: auto-merge
+          labels: \\"\\"
```
eladb, line 614: This looks like a bug

src/github/auto-approve.ts
```diff
@@ -0,0 +1,82 @@
...
+        uses: 'hmarr/auto-approve-action@v2.1.0',
+        with: {
+          'github-token': `\${{ secrets.${options.secret} }}`,
+        },
+      }],
+    };
+
+    const workflow = new GithubWorkflow(project.github, 'auto-approve');
+    workflow.on({
+      pullRequestTarget: {
```
eladb, line 76: Can you please add some notes about why `pullRequestTarget` is needed here and emphasize the security considerations and why it is safe?

eladb (changes requested):

src/__tests__/__snapshots__/integ.test.ts.snap
nija-at: Are you referring to the 'auto-merge' being removed or that `labels` being an empty string?

eladb: The empty string

nija-at: Fixed

eladb (approved):

PR #825, opened 2021-05-27
feat(cli): --version flag

### LINKED ISSUE #497: CLI --version flag no longer available
Context: In #460 a field named `version` field was added for Java Maven project types, but since this allows `--version` to be used in the CLI to initialize the value of this field, the regular `--version` CLI flag (that would show the current projen version) had to be disabled (see [here](https://github.com/projen/projen/commit/9f940549863aa9076f63aceefe46953d8360ef0e#diff-2304ffbbb1cb4987906a47b08b7c3b590fc542f4247b733844049776fd8591f2R20)).

I think this is problematic because `--version` is a pretty standard flag for shell programs, and it's useful to be able to check what version of projen you are using for debugging.

Maybe there is some way to add `--version` as a flag/command that operates differently if you are not using any sub-command (such as `projen new`, `projen start`, etc.). Looking through the [yargs docs](http://yargs.js.org/docs/) could help here.

Similarly, I think it's misleading that the `--post` option is available for _all_ projen commands, even project-specific tasks (try running `npx projen test --help`), when it only makes sense in the context of the synthesis task. But this is lower priority since it doesn't have any negative impact.

Chriscbr:
Fixes #497

`yargs` provides a built-in utility that shows the script's version and immediately exits when the user passes the `--version` flag. Unfortunately, we are using `--version` as an field in new project types (so a user can run `projen new java --version 1.2.3` and the version will be pre-filled in their `projenrc` file), which causes a conflict. Currently, `yargs` doesn't provide ways to customize where this `--version` utility is valid (see https://github.com/yargs/yargs/issues/1918), so to work around this I manually implemented the `--version` feature so it only activates if you run `projen` with no subcommands.

Now the current behavior is:

```
$ projen --help
projen [command]

Commands:
  projen new [PROJECT-TYPE] [OPTIONS]  Creates a new projen project
  projen start                         Prints all project commands

Options:
      --post     Run post-synthesis steps such as installing dependencies. Use
                 --no-post to skip                           [boolean] [default: true]
  -w, --watch    Keep running and resynthesize when projenrc changes
                                                            [boolean] [default: false]
      --debug    Debug logs                                 [boolean] [default: false]
      --rc       path to .projenrc.js file
                                           [string] [default: "/path/to/.projenrc.js"]
      --help     Show help                                                   [boolean]
      --version  Show version number                                         [boolean]

$ projen new --help
(help information, with no --version option available)

$ projen new java --help
(help information, with java-specific --version information shown)

$ projen new java --help
(help information, with python-specific --version information shown)
```

---

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,6 +1,6 @@
 import { resolve } from 'path';
 import * as yargs from 'yargs';
-import { PROJEN_RC } from '../common';
+import { PROJEN_RC, PROJEN_VERSION } from '../common';
 import { TaskRuntime } from '../tasks';
 import { synth } from './synth';
 import { discoverTaskCommands } from './tasks';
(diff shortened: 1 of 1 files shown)


eladb (approved):
Nice!

PR #826, opened 2021-05-27
feat(cli): bash completion

adrianmace:
# Shell Completions!
This allows you to run `projen completion >> ~/.zshrc` or `projen completion >> ~/.bashrc` to enable rich auto-completion.

image

---


eladb (commented):
Nice!

Can we add this in the getting started section?

adrianmace:
> Nice!
> 
> Can we add this in the getting started section?

Absolutely. Will update the PR in the next 24hr. 

We also have the ability to define a custom command name in case you would prefer a different term to be used other than `completion`. Just let me know!

eladb:
> We also have the ability to define a custom command name in case you would prefer a different term to be used other than `completion`. Just let me know!

I think this is a good name. The only thing that we will need to do is disallow tasks to use this name (as well as "new", by the way). But we can add that in a future PR

eladb (approved):

PR #832, opened 2021-05-30
fix(license): apache license should not be modified

### LINKED ISSUE #830: Incorrect Apache-2.0 license file
The copyright piece in Apache-2.0 is an example source header. it should not be edited. To declare the copyright, we need to use a separated NOTICE file. Check this project as an example https://github.com/apache/kafka/blob/trunk/LICENSE
The original Apache-2.0 license: https://www.apache.org/licenses/LICENSE-2.0.txt

```diff
diff --git a/a b/b
index f49a4e1..3a48601 100644
--- a/a
+++ b/b
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2021 Elad Ben-Israel
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
```

eladb:
Do not modify the appendix in Apache 2.0 license. To allow other licenses to be modified (like MIT), change the substitution string to `$copyright_period` and `$copyright_owner`. If the text includes these tokens and the relevant options are not provided, fail.

Fixes #830

BREAKING CHANGE: The initializer of the `License` now takes `spdx` through the options struct instead of a positional argument. Additionally, an error is thrown if `MIT` license is selected but the `copyrightOwner` is not provided (because it is required in the license text).

---


wchaws (commented):
I don't think we should remove the APPENDIX part. IMO, The line`Copyright [yyyy] [name of copyright owner]` should be kept without any modification.

eladb:
> @eladb I don't think we should remove the APPENDIX part. IMO, The line`Copyright [yyyy] [name of copyright owner]` should be kept without any modification.

@wchaws can you please point to the guidance that indicates that this is what is expected by Apache?

wchaws:
> > @eladb I don't think we should remove the APPENDIX part. IMO, The line`Copyright [yyyy] [name of copyright owner]` should be kept without any modification.
> 
> @wchaws can you please point to the guidance that indicates that this is what is expected by Apache?

I've checked some projects under https://github.com/apache/. They both have `APPENDIX` in Apache-2.0 LICENSE.

eladb:
OK, reverting

src/license.ts
```diff
@@ -4,37 +4,23 @@ import { Project } from './project';
...
     const textFile = `${__dirname}/../license-text/${spdx}.txt`;
     if (!fs.existsSync(textFile)) {
       throw new Error(`unsupported license ${spdx}`);
     }
 
-    const years = options.copyrightPeriod ?? new Date().getFullYear().toString();
-    const owner = options.copyrightOwner ?? '';
-
-    this.text = fs.readFileSync(textFile, 'utf-8')
-      .replace('[yyyy]', years)
```
wchaws, line 36: I think we still need the substation in MIT-0

eladb: Fixed

PR #836, opened 2021-05-31
feat(release): support pre-release tags

eladb:
Introduce a `prerelease` option which can be used bumps versions with a pre-release tag (e.g. 1.0.0-beta.1, 1.0.0-beta.2, etc).

Same as for normal versions, the latest version is resolved by querying git tags and filtering based on the pre-release identifier. If a latest version with this pre-release tag cannot be located, the `initialVersion` option is used with the pre-release tag.

Since resolving versions is now slightly more complicated, extract this logic into a "builtin" called `release/resolve-version` which is executed as part of the `bump` task. To support this, we introduce the concept of `builtins` which are executables baked into projen itself and executed as part of a task sequence.

BREAKING CHANGE: `task.toShellCommand()` has been removed, as well as the option to render npm scripts as shell commands. Maintaining this task rendition is too expensive and limits the ability to expand the task system.
* **tasks:** The `Task` initializer no longer takes `Tasks` as the first argument (no longer needed).
* **node:** The `npmTaskExecution` option has been removed.

---


Chriscbr:
Some thoughts:

1. "builtins" are an interesting concept. Maybe it could be expanded some day to be less node.js-runtime specific? Maybe there's an opportunity for composing bash scripts using projen's `SourceCode` class or similar, and then executing those? Just random ideas.

2. Can <https://github.com/projen/projen/blob/main/scripts/readme-projects.js> be refactored into a "builtin"?

3. Do you think #337 still makes sense? w.r.t. implementation, I think ejecting is less trivial without "toShellCommand" available, though I might still be able to do it. But you mentioned concerns about "[limiting] the ability to expand the task system" so I don't want to impose more constraints if you have other ideas for task-related features.

PR #840, opened 2021-05-31
feat: non-text sample files

### LINKED ISSUE #768: React project type is missing logo
When React project type is created (`npx projen new react-ts`), there are some non-code files missing in the public folder. They are referenced in `/public/index.html`:

`<link rel="icon" href="%PUBLIC_URL%/favicon.ico" />`
and
`<link rel="apple-touch-icon" href = "%PUBLIC_URL%/logo192.png" />`

but they are not generated.

Is there any way to include non-code files? If not - it seems like a major limitation. I'm going to create custom project types (external modules), but I'm not sure how can I attach non-code files or store project content in separate files and use something like `fs.readFile` to read and generate them with projen (like you @eladb mentioned here: https://github.com/projen/projen/pull/229#pullrequestreview-516490059)

Chriscbr:
Closes #768.

Introduces new fields "sourcePath" and "sourceDir" for `SampleFile` and `SampleDir` respectively that allows you to specify the path of a file or directory respectively that should be copied from. Through this option, it is now possible to include non-text files in the sample code of a new project.

Also, updates the sample files that are provided in new react projects.

---


src/sample-file.ts
```diff
@@ -91,8 +127,55 @@ export class SampleDir extends Component {
...
 }
+
+/**
+ * A helper function that resolves the path to a raw file based on
+ * projen's current runtime environment.
+ *
+ * @param relativeSource - a relative path to a file in a module, starting with the module name
+ * @returns an absolute path
+ */
+function resolveSourcePath(relativeSource: string) {
```
Chriscbr, line 161: I spent a while trying to get this working... but there ought to be a cleaner way to do this. I don't believe resolving paths `require.resolve()` works for our intended case because it will always resolve to the directory of projen's runtime, even if the file we are looking for is in a third party module. (But I could be wrong).

Also, based on the way it's implemented right now, it's possible for a third-party module to copy files from other modules. For example, a third-party `VueProject` could create in its constructor:

```ts
// OK
new SampleFile(this, 'icon.png', { source: 'projen-vue/assets/icon.png' })
// seems kind of suspicious
new SampleFile(this, 'logo.svg', { source: 'projen/logo/projen.svg' });
```

This raises some security concerns for me. I think it would be safer if projects could only include files that are within their own directory - but I'm not sure how to add this kind of validation. (For example, in the `synthesize()` method of `SampleFile`, is there a way to figure out the path to the file where `new SampleFile` was written?)

Chriscbr:
This is appearing to be a difficult feature to test, at least with the API I've drafted here...

To see if this feature works with third-party project types "in the wild", it essentially requires locally setting up a custom projen module (like [taimos-projen](https://github.com/taimos/taimos-projen)) that _depends on_ a version of projen patched with this PR. Once that is setup, to see if the feature works you would have to try creating a new project `--from` this custom project type to see if the non-text sample files get created as expected.

Even if there's some a simple way to set this up (it seems to be pretty tricky in my experience), the fact that I can't easily emulate it in a integration test makes it seem impractical to maintain, so for now I'm going to close the PR.

If anyone has ideas for a different kind of API, or different way of implementing the feature, feel free to add a comment or create a new issue though. :-)

src/__tests__/sample-file.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  // THEN
+  expect(synthSnapshot(project)['welcome.txt']).toMatch('hello\nworld');
+});
+
+test('sample file from source', () => {
+  // GIVEN
+  const project = new TestProject();
+
+  // WHEN
+  new SampleFile(project, 'logo.svg', { source: 'projen/logo/projen.svg' });
```
eladb, line 22: I feel `source` can just be an absolute path instead of relative and it will dramatically simplify:

```ts
new SampleFile(project, 'logo.svg', { 
  source: path.join(__dirname, 'logo/projen.svg')
});
```

Chriscbr: facepalm

You're totally right. I've seen plenty of other APIs work that way...

src/sample-file.ts
```diff
@@ -25,17 +39,34 @@ export class SampleFile extends Component {
...
+    let contents;
+    if (this.options.contents) {
+      contents = this.options.contents;
+    } else if (this.options.source) {
+      if (fs.existsSync(this.options.source)) {
+        contents = fs.readFileSync(this.options.source);
+      } else {
+        this.project.logger.error(`Could not find sample file expected at path: ${this.options.source}. Skipping...`);
+        return;
+      }
```
Chriscbr, line 68: Included this "else" branch to handle the odd case that the source file can't be found -- in which case, I didn't think it made sense to still create the file with empty contents. I wasn't sure if this should explicitly throw an error (totally open to changing this).

src/web/react.ts
```diff
@@ -157,6 +159,7 @@ export class ReactTypeScriptProject extends TypeScriptAppProject {
         fileExt: 'tsx',
         srcdir: this.srcdir,
       });
+      new SampleDir(this, 'public', { source: path.join(__dirname, '..', '..', 'assets', 'react') });
```
eladb, line 162: Let's put this under `../../assets/web/react` (to match the same directory structure)

src/sample-file.ts
```diff
@@ -81,6 +127,10 @@ export class SampleDir extends Component {
    */
   constructor(project: Project, dir: string, options: SampleDirOptions) {
     super(project);
+    if (!options.files && !options.source) {
```
eladb, line 130: Why? Technically it should be possible to use both modes together as long as the file names are not conflicting...

Chriscbr: This branch throws an error if **neither** mode is selected. Nothing is barring the user form using both modes together.

I've added a unit test to make this more clear.

src/sample-file.ts
```diff
@@ -11,7 +12,20 @@ export interface SampleFileOptions {
...
+   * a text file).
+   *
+   * If your project is Typescript-based and has configured `testdir` to be a
+   * subdirectory of `src`, sample files should outside of the `src` directory,
+   * otherwise they may not be copied. For example:
+   * ```
+   * new SampleFile(this, 'assets/icon.png', { source: path.join(__dirname, '..', 'sample-assets', 'icon.png') });
+   * ```
+   */
+  readonly source?: string;
```
eladb, line 28:
```suggestion
  readonly sourcePath?: string;
```

src/sample-file.ts
eladb: I'd totally throw. I don't see any scenario where failing quietly is going to make sense.

I would not even check if the file exists and let readFile throw in that case.

src/sample-file.ts
```diff
@@ -61,9 +92,24 @@ export class SampleFile extends Component {
...
+   *
+   * If your project is typescript-based and has configured `testdir` to be a
+   * subdirectory of `src`, sample files should outside of the `src` directory
+   * otherwise they may not be copied. For example:
+   * ```
+   * new SampleDir(this, 'public', { source: path.join(__dirname, '..', 'sample-assets') });
+   * ```
    */
-  readonly files: { [fileName: string]: string };
+  readonly source?: string;
```
eladb, line 112:
```suggestion
  readonly sourceDir?: string;
```

eladb (commented):
Minor

eladb (approved):

PR #851, opened 2021-06-03
fix(jsii-project): option to disable release to npm

Hunter-Thompson:
`releaseToNpm: false` was not being honored by `jsii-project`, this resulted in an unwanted workflow for every project extended using `jsii-project`.

---


Hunter-Thompson:
This issue might be fixed by #850?

eladb (changes requested):
I am not sure that's the desired behavior. By definition, jsii projects are expected to be published at least to npm and this is why this is the default behavior.

Can we leave the default to always publish to npm but also allow disablement?

Hunter-Thompson:
@eladb, when I set `releaseToNpm` to true, the integration tests are failing. Unable to figure that out.

src/jsii-project.ts
```diff
@@ -203,12 +203,6 @@ export class JsiiProject extends TypeScriptProject {
     }
     this.addFields({ jsii });
 
-    this.publisher?.publishToNpm({
```
eladb, line 206: Revert this change and add an `if` statement here that will respect `options.releaseToNpm`

src/jsii-project.ts
```diff
@@ -145,9 +145,9 @@ export class JsiiProject extends TypeScriptProject {
       // this is needed temporarily because our release workflows use the 'gh'
       // cli and its not yet available in jsii/superchain
       releaseWorkflowSetupSteps: options.releaseWorkflowSetupSteps,
-      releaseToNpm: false, // we have a jsii release workflow
```
eladb, line 148: revert this change so that by default we pass `false` to super class

eladb (changes requested):

eladb (changes requested):
@Hunter-Thompson please add a unit test for this

PR #856, opened 2021-06-04
fix: gitattributes not recognizing projen files as generated

Chriscbr:
Files in `/.projen/` are not currently being recognized as generated in GitHub's UI. The reason is that according to the [gitattributes docs](https://git-scm.com/docs/gitattributes):

> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash `path/` syntax is pointless in an attributes file; use `path/**` instead)

---


Chriscbr:
Wow, what are the odds! It looks like in the workflow for this PR ([link](https://github.com/projen/projen/runs/2750471825)), the tests started around 23:59 UTC and continued past 00:00 UTC the next day, causing one of the bump tests fails at this line:

https://github.com/projen/projen/blob/1c32a60bfe3b239f7b95abcc9c3412540aa23c94/src/__tests__/release/bump.test.ts#L25

...because `today` is calculated earlier in the file, and by the time the comparison was made, the date had changed. We ought to fix this.

For now, I'll push a dummy commit to unblock myself...

eladb (approved):

PR #857, opened 2021-06-06
feat(deps): upgrade transitive deps of non upgraded packages

iliapolo:
The current dependencies upgrade task executes `npm-check-updates` to detect new versions of dependencies. It then runs the package manager specific **installation** command to install those new versions, along with any of its transitive dependencies.

The problem is that transitive dependencies of packages that dont have a new version don't get updated. 

You can see an example here: https://github.com/cdklabs/json2jsii/security/dependabot/yarn.lock/ws/open.

> The `ws` package is a transitive dependency of `jest`, but because `jest` itself hasn't released a new version, `ws` is also > not upgraded. Note that in this case we will never get a new version of `jest` because we are using an old major version > which is probably not being released anymore. 

The solution is to run the package manager specific **upgrade** command after the installation. 
Where this gets tricky is the need to respect the `include` and `exclude` configuration of the upgrade workflow. 

`yarn` and `pnpm` both have a mechanism for this (though I couldn't get `pnpm` to work), but `npm` does not support excluding packages when running `npm update`. 

Since projen is aware of all package dependencies, as well as what needs to be ignored, we can simply build an upgrade command that includes all the necessary packages. This approach, while seems funky, works the same for all package managers. 

---


.projen/tasks.json
```diff
@@ -151,6 +151,9 @@
         {
           "exec": "yarn install --check-files"
         },
+        {
+          "exec": "yarn upgrade @types/fs-extra @types/glob @types/ini @types/inquirer @types/jest @types/node @types/semver @types/yargs @typescript-eslint/eslint-plugin @typescript-eslint/parser all-contributors-cli eslint eslint-import-resolver-node eslint-import-resolver-typescript eslint-plugin-import jest jest-junit jsii jsii-diff jsii-docgen jsii-pacmak json-schema markmac npm-check-updates standard-version typescript @iarna/toml chalk decamelize fs-extra glob ini inquirer semver shx xmlbuilder2 yaml yargs"
```
eladb, line 155: Wouldn't it make more sense to run this *before* the `yarn install`?

src/node-package.ts
```diff
@@ -617,6 +617,41 @@ export class NodePackage extends Component {
     return this.renderInstallCommand(false);
   }
 
+  /**
+   * Render a package manager specific command to upgrade all requested dependencies.
+   */
+  public renderUpgradePackagesCommand(exclude: string[], include?: string[]): string {
+
+    const project = this.project;
```
eladb, line 625: Wondering if (for aesthetic reasons), it makes sense to call `xxx upgrade` with no arguments in case both `exclude` and `include` are empty (e.g. upgrade all can just use the standard command).

src/upgrade-dependencies.ts
```diff
@@ -137,6 +137,9 @@ export class UpgradeDependencies extends Component {
     // run "yarn/npm install" to update the lockfile and install any deps (such as projen)
     task.exec(this._project.package.installAndUpdateLockfileCommand);
 
+    // run upgrade command to upgrade transitive deps as well
+    task.exec(this._project.package.renderUpgradePackagesCommand(exclude, this.options.include));
```
eladb, line 141: Curious, why isn't `renderUpgradaePackagesCommand` implemented as a private method of the `UpgradeDependencies` class?

eladb (changes requested):

.projen/tasks.json
iliapolo: Would it? I'm not sure...`npm-check-updates` instructs you to run `yarn install` immediately after for example.
I was thinking its probably ok to skip `yarn install` altogether and let `yarn upgrade` do everything, but I didn't want to venture into testing this right now.

src/node-package.ts
iliapolo: Yeap i'll change 👍

src/upgrade-dependencies.ts
iliapolo: Wanted to keep all those package manager `switch` cases in one place. I agree it's creates some pollution though - want me to change?

eladb: Let's leave as-is for now. We should probably extract all of these into a separate `NodePackageManager` abstract class that hides these details from both components.

.projen/tasks.json
eladb: OK. Let's leave as is

src/node-package.ts
```diff
@@ -617,6 +617,50 @@ export class NodePackage extends Component {
...
+        // filter by exclude and include.
+        return `${command} ${project.deps.all
+          .map(d => d.name)
+          .filter(d => include ? include.includes(d) : true)
+          .filter(d => !exclude.includes(d))
+          .join(' ')}`;
+      };
+    }
+
+    let lazy = undefined;
```
eladb, line 643: What's going on? Seems like this is dead code

eladb: Strike that

eladb (approved):

src/upgrade-dependencies.ts
iliapolo: 👍

PR #858, opened 2021-06-06
feat(react): rewire webpack config

iliapolo:
Support basic react property overrides without needing to eject.

Based on https://www.npmjs.com/package/react-app-rewired.

You can see an example of this mechanism here: https://github.com/cdklabs/construct-hub-webapp/blob/main/config-overrides.js

Once this is merged, we can migrate construct-hub-webapp to use it. 

---


src/web/react.ts
```diff
@@ -150,29 +187,50 @@ export class ReactComponent extends Component {
...
+      const configOverrides = new SourceCode(this.project, 'config-overrides.js');
+      configOverrides.line(`// ${FileBase.PROJEN_MARKER}`);
+      configOverrides.line('/**');
+      configOverrides.line(' * Override CRA configuration without needing to eject.');
+      configOverrides.line(' *');
+      configOverrides.line(' * @see https://www.npmjs.com/package/react-app-rewired');
+      configOverrides.line(' */');
+      configOverrides.open('module.exports = function override(config, env) {');
+      for (const [key, value] of Object.entries(rewire)) {
+        configOverrides.line(`config.${key} = ${JSON.stringify(value)}`);
```
eladb, line 202:
```suggestion
        configOverrides.line(`config.${key} = ${JSON.stringify(value)};`);
```

src/web/react.ts
```diff
@@ -150,29 +187,50 @@ export class ReactComponent extends Component {
...
+      }
+      configOverrides.line('return config');
```
eladb, line 204:
```suggestion
      configOverrides.line('return config;');
```

src/web/react.ts
```diff
@@ -150,29 +187,50 @@ export class ReactComponent extends Component {
...
+      configOverrides.close('}');
```
eladb, line 205:
```suggestion
      configOverrides.close('};');
```

src/web/react.ts
```diff
@@ -150,29 +187,50 @@ export class ReactComponent extends Component {
       project.addDevDeps('@types/jest', '@types/node', '@types/react', '@types/react-dom');
     }
 
+    const rewire = options.rewire;
+
+    if (rewire) {
+      const configOverrides = new SourceCode(this.project, 'config-overrides.js');
```
eladb, line 193: Can this go under `.projen/react-config-overrides.js` to avoid polluting the root?

src/web/react.ts
```diff
@@ -150,29 +187,50 @@ export class ReactComponent extends Component {
...
+      configOverrides.line(' */');
+      configOverrides.open('module.exports = function override(config, env) {');
+      for (const [key, value] of Object.entries(rewire)) {
+        configOverrides.line(`config.${key} = ${JSON.stringify(value)}`);
+      }
+      configOverrides.line('return config');
+      configOverrides.close('}');
+    }
+
+    const reactScripts = rewire ? 'react-app-rewired' : 'react-scripts';
```
eladb, line 208: Don't we need to install this dependency?

eladb (changes requested):

src/web/react.ts
iliapolo: Looks like its [possible](https://github.com/timarney/react-app-rewired/commit/73596cc59e05ae7b0ec42b4332fca448bf30134b). I'll try it out.

src/web/react.ts
iliapolo: Whoops! Good catch...

src/web/react.ts
iliapolo: Moved 👍

eladb (approved):

PR #862, opened 2021-06-07
feat(deps): deduplicate dependencies

Chriscbr:
Some components are created multiple times (e.g. `UpgradeDependencies` in a default node project is created once to generate a `upgrade-dependencies` task and once more to generate an `upgrade-projen` task). If a component specifies a dependency, the dependency may get added to `deps.json` multiple times. This is not desirable because it's possible for versions/metadata to conflict, and the dependency will get listed multiple times when reading from `project.deps.all`.

This PR changes the default behavior of `project.deps.addDependency()` to override an existing dependency if a dependency has been defined with the same name AND same dependency type.

(I wasn't sure if there are use cases for having the same dep listed under multiple dependency types, but adding de-deplication so that a given dep can only be listed under a single dependency type would be an easy change.)

---


src/web/react.ts
```diff
@@ -184,7 +184,7 @@ export class ReactComponent extends Component {
     project.addDeps('react', 'react-dom', 'react-scripts@^4.0.0', 'web-vitals');
     project.addDevDeps('@testing-library/jest-dom', '@testing-library/react', '@testing-library/user-event');
     if (this.typescript) {
-      project.addDevDeps('@types/jest', '@types/node', '@types/react', '@types/react-dom');
```
Chriscbr, line 187: @types/node is already added in `TypeScriptProject` so including it here was unneeded.

src/__tests__/java/__snapshots__/java-project.test.ts.snap
```diff
@@ -414,15 +414,6 @@ exports[`dependencies 1`] = `
...
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.1</version>
-                <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
-                </configuration>
-            </plugin>
```
Chriscbr, line 425: This diff looks weird since I didn't change any corresponding tests, but turns out it's just an artifact of this line defining a dependency that is already included in the default `TestJavaProject` class:

https://github.com/projen/projen/blob/5955ada420872c82387ce8f6cc71d910307e9b83/src/__tests__/java/java-project.test.ts#L31-L36

src/__tests__/deps.test.ts
```diff
@@ -69,6 +69,19 @@ test('deps.all returns all the dependencies', () => {
   expect(p.deps.all).toMatchSnapshot();
 });
 
+test('can be overridden', () => {
+  // GIVEN
+  const p = new TestProject();
+
+  // WHEN
+  p.deps.addDependency('depy@^7', DependencyType.PEER);
+  p.deps.addDependency('depy@^9', DependencyType.PEER);
```
eladb, line 78: Mmm.. Not sure the desired behavior is to override. I feel like adding two deps with different version requirements should fail by default and (maybe) we can add an option to force.

eladb (changes requested):

src/__tests__/deps.test.ts
Chriscbr: OK, sounds reasonable.

We can def. add a force / allow-override flag, but for now just deleting and re-adding the dependency is a good enough workaround.

eladb (approved):

PR #867, opened 2021-06-08
fix(web): build tasks do not run projen

### LINKED ISSUE #754: web projects: "build" tasks do not run projen

Chriscbr:
Fixes #754

The default behavior for web projects has been updated to run `npx projen` (as well as any compile and test tasks, if those have been defined) before generating a production build via `react-scripts build` or `next build`.

In addition, `project.addTask` has been modified so that it's not possible to override an existing task with the same name. If you want to override an existing task, you can remove it first with `project.removeTask` and then add your task.

---


src/__tests__/web/__snapshots__/nextjs-ts-project.test.ts.snap
```diff
@@ -313,9 +313,18 @@ tsconfig.tsbuildinfo
...
         "name": "build",
         "steps": Array [
+          Object {
+            "exec": "npx projen",
+          },
+          Object {
+            "spawn": "test",
+          },
+          Object {
+            "spawn": "compile",
```
eladb, line 326: Is that what we want?

eladb (approved):

src/__tests__/web/__snapshots__/nextjs-ts-project.test.ts.snap
Chriscbr: I believe so? It's possible the compile step is redundant for web projects, but at least on fresh react/next.js projects it doesn't cause any conflicts. We can definitely add an option in `NodeProject` to remove the compile task (as you suggested) if needed

PR #868, opened 2021-06-08
fix(typescript): cannot collocate sources and compiled output

iliapolo:
Right now, if you want to collocate sources and compiled output, i.e use the same value for `srcdir` and `libdir`:

```typescript
new TypeScriptProject({
  name: 'test',
  defaultReleaseBranch: 'test',
  libdir: 'lib',
  srcdir: 'lib',
  });
```

If fails to build because the tsconfig both includes and excludes the `lib` directory, essentially making it so that there are no sources to compile. 

```json
  "include": [
    "lib/**/*.ts",
  ],
  "exclude": [
    "node_modules",
    "lib"
  ],
```

This PR enables providing the same value and collocating these files.
This makes it easier to onboard projen into existing projects that do have collocation and rely on it (regrettably) 

---


src/typescript.ts
```diff
@@ -257,21 +257,26 @@ export class TypeScriptProject extends NodeProject {
     if (!options.disableTsconfig) {
       const baseTsconfig: TypescriptConfigOptions = {
         include: [`${this.srcdir}/**/*.ts`],
-        exclude: [
-          'node_modules',
-          this.libdir,
-        ],
```
iliapolo, line 263: I believe this is OK to just remove because the `includes` keeps up scoped to the correct files anyway. Am I missing something here?

iliapolo:
@eladb Thoughts?

src/typescript.ts
eladb: Seems you are [correct](https://www.typescriptlang.org/tsconfig#exclude)

eladb (approved):
Nice.

PR #870, opened 2021-06-09
chore(deps): consolidate upgrade workflow if projen secret is available

iliapolo:
Currently we create two separate upgrade workflows, one for projen and one for everything else. 

The reason is that projen upgrades often require changes to github workflows, which requires special permissions from the token used to create the PR. 

However, if such a token is defined, and it usually is to make projen auto upgrade as well, we can just use that token and merge the two workflows into one. 

---


src/node-project.ts
```diff
@@ -633,13 +633,15 @@ export class NodeProject extends Project {
...
       ? DependenciesUpgradeMechanism.dependabot()
       : DependenciesUpgradeMechanism.githubWorkflow({
-        workflowOptions: options.workflowContainerImage ? {
+        // if projen secret is defined we can also upgrade projen here.
+        ignoreProjen: !options.projenUpgradeSecret,
+        workflowOptions: {
           // if projen secret is defined, use it (otherwise default to GITHUB_TOKEN).
           secret: options.projenUpgradeSecret,
-          container: {
+          container: options.workflowContainerImage ? {
```
iliapolo, line 641: I changed this because previously it only used the secret if `workflowContainerImage` is defined, which seems like a bug. It should always use the secret if it can.

eladb (approved):

PR #871, opened 2021-06-09
feat(node): auto merge dependency upgrades

iliapolo:
As things stand now, its too complicated to setup auto merges for dependency upgrade PR's. Here is what it takes:

```ts
  ...
  projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',

  // first we need to enable auto approval
  autoApproveOptions: {
    allowedUsernames: ['aws-cdk-automation'],
    secret: 'GITHUB_TOKEN',
    label: 'auto-approve'
  },

  // then we need to apply it on the deps upgrade workflow.
  // which means overriding the default values, and redefining them.
  depsUpgrade:  DependenciesUpgradeMechanism.githubWorkflow({
    workflowOptions: {
      labels: ['auto-approve'], // this must match the configured auto-approve label, which is why we explicitly set it above as well 
    },
    // since we are using a secret, we can include projen here and use a single workflow (this is done for you in the default values)
    ignoreProjen: false,
    secret: 'PROJEN_GITHUB_TOKEN',
  })
});
```

This PR proposes an alternative by introducing the `depsUpgradeAutoMerge` property, similar to the already existing `projenUpgradeAutoMerge` property. 

```ts
  ...
  projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',
  autoApproveOptions: {
    allowedUsernames: ['aws-cdk-automation'],
    secret: 'GITHUB_TOKEN',
  },
  depsUpgradeAutoMerge: true,
});
```

Even this simplification is somewhat cumbersome, and can be improved if we decide at some point to simply default to auto-merging dependencies if `autoApprove` is configured. 

I feel its ok to start off safe and make it explicit, we can always relax requirements later. 

BREAKING CHANGE: `DependabotOptions.autoMerge` was removed. Pass `project.autoApprove.label` into `DependabotOptions.labels` instead. In addition, `projenUpgradeAutoMerge` now defaults to `false`.

---


src/github/dependabot.ts
```diff
@@ -20,12 +20,6 @@ export interface DependabotOptions {
    */
   readonly versioningStrategy?: VersioningStrategy;
 
-  /**
-   * Automatically merge dependabot PRs if build CI build passes.
-   * @default true
-   */
-  readonly autoMerge?: boolean;
```
iliapolo, line 27: Decided to remove this and align the behavior with the upgrade workflow that simply accepts `labels` and moves the auto-merge responsibility to the caller for now.

src/github/dependabot.ts
```diff
@@ -165,24 +165,6 @@ export class Dependabot extends Component {
...
-        actions: {
-          merge: {
-            method: 'squash',
-            commit_message: 'title+body',
-            strict: 'smart',
-            strict_method: 'merge',
-          },
-        },
-      });
-    }
```
iliapolo, line 184: This shouldn't be here anyway since it doesn't comply with branch protection rules (missing approval)

src/node-project.ts
```diff
@@ -171,10 +171,22 @@ export interface NodeProjectOptions extends ProjectOptions, NodePackageOptions,
...
 
+  /**
+   * Automatically merge deps upgrade PRs when build passes.
+   * Auto-approves the pull request which mergify will then merge.
+   *
+   * Only applies if `autoApproveOptions` are defined.
+   *
+   * @default - true
+   */
+  readonly depsUpgradeAutoMerge?: boolean;
```
iliapolo, line 188: Should be probably called `depsUpgradeAutoApprove` since mergify might be disabled...but this does convey better intent and is aligned with the `projenUpgradeAutoMerge` property. 

I can be persuaded otherwise though.

src/github/dependabot.ts
```diff
@@ -156,6 +155,7 @@ export class Dependabot extends Component {
             interval: options.scheduleInterval ?? DependabotScheduleInterval.DAILY,
           },
           'ignore': () => this.ignore.length > 0 ? this.ignore : undefined,
+          'labels': options.labels ?? [],
```
eladb, line 158:
```suggestion
          'labels': options.labels ? options.labels : undefined,
```

src/node-project.ts
```diff
@@ -171,10 +171,22 @@ export interface NodeProjectOptions extends ProjectOptions, NodePackageOptions,
    * Automatically merge projen upgrade PRs when build passes.
    * Auto-approves the pull request which mergify will then merge.
    *
-   * @default - "true" if mergify auto-merge is enabled (default)
+   * Only applies if `autoApproveOptions` are defined.
+   *
+   * @default - true
```
eladb, line 176:
```suggestion
   * @default true
```

src/node-project.ts
```diff
@@ -171,10 +171,22 @@ export interface NodeProjectOptions extends ProjectOptions, NodePackageOptions,
...
    */
   readonly projenUpgradeAutoMerge?: boolean;
 
+  /**
+   * Automatically merge deps upgrade PRs when build passes.
+   * Auto-approves the pull request which mergify will then merge.
+   *
+   * Only applies if `autoApproveOptions` are defined.
```
eladb, line 184: Should this fail if autoApprove is not configured?

src/node-project.ts
eladb: I prefer to change both to be called `autoApproveUpgrades` and `autoApproveProjenUpgrades`.

eladb (changes requested):
I think the semantics should be "autoApprove" all the way.

Auto-merge of approved PRs is a different concern.

I think we should do something about the autoApprove config. Still feels clunky, but maybe we can leave this for a subsequent PR.

src/node-project.ts
iliapolo: @elad What if move these down to the `autoApprove` configuration (and break the existing config)?

```ts
autoApprove: {
  projenUpgrade: true,
  depsUpgrade: true
}
```

This also means auto approve will have to be configured avoiding the need to validate it.

eladb: But autoApprove doesn't know about deps upgrade. It's a separate component.

src/node-project.ts
iliapolo: Yes. Changed.

src/github/dependabot.ts
Chriscbr: "BREAKING CHANGE" boilerplate should probably be added to the PR description if we're removing an option from the API.

src/node-project.ts
iliapolo: Yeah I know, ok changed the names and added a deprecation

eladb (approved):

PR #877, opened 2021-06-10
fix: "yarn link projen" is broken

### LINKED ISSUE #151: Cannot link local projen fork to new package
I was interested in adding a new project type to projen, but I've seemed to hit a roadblock. My plan was to use [yarn link](https://classic.yarnpkg.com/en/docs/cli/link/) to create a symlink between a local version of projen and a test package in which I'm going to use it. However, when I try this and run `yarn projen` to actually synthesize things (in the test package), projen automatically deletes the symlink and runs `yarn install` to replace it with the latest version of projen, overriding the symlink. It seems to be caused by https://github.com/eladb/projen/blob/8aa5e311e298acc11a5df833fcc838bd70401a01/src/node-project.ts#L999

I was able to work around this temporarily commenting out some of the symlinking lines, but is there a better way to test things? I imagine there's some kind of escape hatch in place, but I couldn't seem to figure it out.

eladb:
By default, do not pin projen to the CLI version and let the dependency manager maintain the version. This also has the affect of enabling normal use of `yarn link`.
Update CONTRIBUTING to describe the local test workflow.

Fixes #151

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,24 +69,26 @@ When your local version of projen builds successfully, you can test it to create
 a new project by going into another directory and invoking the binary directly:
 
 ```console
-$ pwd
-/path/to/projen
-$ cd ..
-$ mkdir testing
-$ cd testing
-$ ../projen/bin/projen new <project-type>
+$ cd /path/to/local/projen
+$ yarn link
+$ cd/
+
+# existing project
+$ cd /my/other/project
+$ yarn link projen
+
+# new project
+$ mkdir /my/new/project
+$ cd /my/new/project
+$ /path/to/local/projen/bin/projen new TYPE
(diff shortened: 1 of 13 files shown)


Chriscbr:
Awesome! 🙌  The updated instructions seem to work for me

Chriscbr:
I should qualify my earlier comment by saying if I've made local changes to the way `projen` sets up new projects, it still seems like I still need to directly invoke my projen binary sometimes - for example to test my latest PR, this wasn't good enough:

```
$ yarn link projen
$ npx projen new react-ts  # ignores the linked projen, and instead uses latest public projen
```
and I had to instead use:
```
$ yarn link projen
$ ../projen/bin/projen new react-ts
```

eladb:
Yes, I think `npx projen` won't work for a linked version. I'll update the guide.

PR #889, opened 2021-06-13
feat: remove "start" command

eladb:
The `start` interactive command doesn't seem like it's adding any value to users.

BREAKING CHANGE: `projen start` is no longer supported.

---


hoegertn:
Does this solve maintenance issues? I used the `pj start` almost daily but with the new bash completion I can work around it

Issue #648, opened 2021-03-24
fix(external-projects): projen as a peerDependency


gradybarrett:
### Use Case

Given that I have an external project that is built using projen
When people consume that external project
Then they should be able to bring their desired projen version, and it should not clash with the version of projen used to build the external project.

### Context

I've been writing an external project, built using the projen JsiiProject. I've noticed that in order for project creation `--from` an external project to work correctly, I have to declare projen as a full dependency in my external project. But it seems to me that external projects conform to the description of a "plugin" given in the [peerDependencies docs](https://nodejs.org/en/blog/npm/peer-dependencies/). That makes sense to me because I want people to be able to bring the latest and greatest projen to my external project without it clashing with a projen version declared in my external project. If, however, I make projen a peer dependency in my external project it won't yet be available during project synthesis, which means project creation fails b/c peerDependencies of the external project aren't yet satisfied.

https://github.com/projen/projen/blob/b9152631d7f79d83df175cda497fe3f01cb3b46f/src/cli/cmds/new.ts#L437-L442

I can hack this by installing projen alongside the new project, but that feels strange.

```typescript
  }

  logging.info(`installing external module ${spec}...`);
  // new.ts:440
  exec(`yarn add --modules-folder=${baseDir}/node_modules --silent --no-lockfile --dev ${spec}`, { cwd: baseDir });
  exec(`yarn add --modules-folder=${baseDir}/node_modules --silent --no-lockfile --dev projen`, { cwd: baseDir });

  // if package.json did not exist before calling yarn add, we should remove it
```

It would be nice to just use the version of projen used by npx until we get to the point where projen installs all dependencies for me, but I'm not sure if that's easy or possible to do in a consistent way across platforms.

Have I just come at this from the wrong angle? Or is this what it seems, which is a bit of a chicken before the egg situation?

**Edit:** Just a comment on the current experience. If I put projen as a dep and peerDep, projen ends up adding it as a devDep as well, which results in an invalid package.json. And if I try to use `projenDevDependency: false` to remove it, it gets added back b/c of the `pinnedDevDependency` behavior. I can set `peerDependencyOptions.pinnedDevDependency` to false, but that may not be desirable if there are other peer deps present. Lastly, my projen dep and peer dep don't match - both added as bare module name `'projen'`. The peer dep is rendered as `^0.17.17`, but the dep is rendered in locked form `0.17.17`.

So the final config I have to use to produce a valid package.json satisfy jsii is:

```javascript
[...]

PR #893, opened 2021-06-15
fix(cli): new from external module with multiple types

### LINKED ISSUE #887: PROJECT-TYPE cli argument not working for external modules with multiple types
`ProjectType` is being used for two different concepts:

1. As a definition for a projen project type: https://github.com/projen/projen/blob/main/src/inventory.ts#L26
2. As an enum used for jsii purposes: https://github.com/projen/projen/blob/main/src/project.ts#L523

Because all projects inherit from the `Project` class, the `ProjectOptions` includes the `projectType` property: https://github.com/projen/projen/blob/main/src/project.ts#L84

The `new` cli command exposes `PROJECT-TYPE` as an argument, which gets camelcase-d to `projectType`: https://github.com/projen/projen/blob/main/src/cli/cmds/new.ts#L18

This argument is intended to be used to determine the projen project class to use. But it also sets the `ProjectOption.projectType` property. The config for this property is:

```
{
    "path": ["projectType"],
    "parent": "ProjectOptions",
    "name": "projectType",
    "fqn": "projen.ProjectType",
    "docs": "Which type of project this is (library/app).",
    "type": "ProjectType",
    "kind": "enum",
    "switch": "project-type",
    "default": "ProjectType.UNKNOWN",
    "optional": true,
    "featured": true
}
```

The above clashes with the idea of `projectType` representing the [projen project type](https://github.com/projen/projen/blob/main/src/inventory.ts#L26).

As a result, I am seeing the following error when bootstrapping a new projen project using an external module with multiple project types defined:

```
npx projen new --from "file:../my-projen-types" foo
```

results in

```
evalmachine.<anonymous>:6
  projectType: ProjectType.FOO,
               ^

ReferenceError: ProjectType is not defined
    at evalmachine.<anonymous>:6:16
```

External modules work for a single type, because `PROJECT-TYPE` is not required and thus does not set the `projectType` option, which then results in the default (`ProjectType.UNKNOWN`).

Chriscbr:
Fixes #887

---

### DIFF
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -227,8 +227,14 @@ function commandLineToProps(type: inventory.ProjectType, argv: Record<string, un
  * @param args Command line arguments (incl. project type)
  */
 async function newProjectFromModule(baseDir: string, spec: string, args: any) {
-  // install projen if it isn't already installed or linked
-  exec(`yarn list --depth=0 --pattern projen || yarn add --modules-folder=${baseDir}/node_modules --silent --no-lockfile --dev projen`, { cwd: baseDir });
+  const installCommand = `yarn add --modules-folder=${baseDir}/node_modules --silent --no-lockfile --dev`;
+  if (args.projenVersion) {
+    exec(`${installCommand} projen@${args.projenVersion}`, { cwd: baseDir });
+  } else {
+    // do not overwrite existing installation
+    exec(`yarn list --depth=0 --pattern projen || ${installCommand} projen`, { cwd: baseDir });
+  }
+
   const specDependencyInfo = yarnAdd(baseDir, spec);
 
   // collect projects by looking up all .jsii modules in `node_modules`.
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -11,11 +11,11 @@ import { exec } from '../../util';
 import { tryProcessMacro } from '../macros';
 
 class Command implements yargs.CommandModule {
-  public readonly command = 'new [PROJECT-TYPE] [OPTIONS]';
+  public readonly command = 'new [PROJECT-TYPE-NAME] [OPTIONS]';
   public readonly describe = 'Creates a new projen project';
 
   public builder(args: yargs.Argv) {
-    args.positional('PROJECT-TYPE', { describe: 'optional only when --from is used and there is a single project type in the external module', type: 'string' });
+    args.positional('PROJECT-TYPE-NAME', { describe: 'optional only when --from is used and there is a single project type in the external module', type: 'string' });
     args.option('synth', { type: 'boolean', default: true, desc: 'Synthesize after creating .projenrc.js' });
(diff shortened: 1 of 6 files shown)


Chriscbr:
To reproduce the original bug, I tried using @taimos/projen (thanks @hoegertn) -- though doing it the naive way fails for me:

```
npx projen new --from @taimos/projen taimos-ts-lib
```

with the error: `Error: cannot find module 'projen'`.

If I try `npm i projen && npx projen new ...` instead, I get the error: `Error: No projects found after installing @taimos/projen. The module must export at least one class which extends projen.Project`.

I'm still not sure what causes this. But once that's fixed, I think it would be good to use this as an integration test of the `npx projen new --from <path> <pjid>` flow.

But I was able to work around this by downloading @taimos/projen locally, building it to generate the .jsii, and running:

```
npx projen new --from "file:../taimos-projen" taimos-ts-lib
```

This yields the error message from #887, so I was able to test that this PR fixes the bug.

eladb:
> with the error: Error: cannot find module 'projen'.

I think this is a bug we have at the moment with bootstrapping projects from external modules. It's a result of the fact that the CLI does not add a `projen` symlink to `node_modules` now. I think the solution is to also add `projen` when bootstrapping from an external module.

src/cli/cmds/new.ts
```diff
@@ -11,11 +11,11 @@ import { exec } from '../../util';
 import { tryProcessMacro } from '../macros';
 
 class Command implements yargs.CommandModule {
-  public readonly command = 'new [PROJECT-TYPE] [OPTIONS]';
+  public readonly command = 'new [PROJECT-TYPE-NAME] [OPTIONS]';
```
Chriscbr, line 14: `PROJECT-TYPE-ID` or `PROJEN-ID` or some other variation could probably work just as well (just thought of this now...)

Chriscbr:
@eladb thanks, that fixed the issue. Added an integ test.

gradybarrett:
Installing projen should fix most of #648. Not sure if the peer dep discrepancies have been solved by any recent PRs, but if so then you could close #648.

Chriscbr:
From #648:

> When people consume that external project
> Then they should be able to bring their desired projen version, and it should not clash with the version of projen used to build the external project.

This PR partially satisfies that - though we might have to figure out what counts as bringing a desired projen version. With these changes, if you run `yarn add --dev projen@x.y.z` before `npx projen new --from <module> <pjid>`, it will respect the version of projen you specified.

But it will not automatically pick up if you have a specific version of projen installed globally via `yarn global add projen`.

gradybarrett:
I'm not sure we have to satisfy that at bootstrap-time though. I think it's sufficient to use latest projen to bootstrap, at which point folks can choose the version they want and regenerate.

The only caveat is if someone passes a projen version in as a CLI option, then we could use that version during bootstrapping couldn't we?

Chriscbr:
Cool, I think that makes sense. It should be possible to use the verison doing bootstrapping as well, I'll take a look.

eladb (approved):

Issue #298, opened 2020-11-11
Strongly-typed GithubWorkflow


eladb:
The `GithubWorkflow` class is very leaky. Let's model the GH workflow APIs.

pgollucci:
yaml2json and json-to-ts-cli are useful

pgollucci:
isn't this going to result in us being the Chef of ".dotfiles" for projects? we'll always be behind.  This is why JSII generates from CFT weekly right ?

eladb:
I've been mostly thinking about the workflow schema, not actions...

I think the basic schema for GitHub workflows is not going to dramatically change, so it's probably not too much of a maintenance effort to add new features as they emerge. We should also include an "escape hatch" for users to be able to break the abstraction (of course).

gradybarrett:
@eladb are you suggesting modeling this down to the granularity of Events, ActivityTypes, etc? e.g. [webhook-events](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#webhook-events)

eladb:
All the way down. The more we model the more users would love this

gradybarrett:
Cool. I love it. I'm going to give it a shot if you don't mind.

I'm not intimately familiar with Workflows yet, so this is a slightly selfish chance to gain that knowledge. :D It'll take me a while to get through the docs, but I'll try to post a PR once I get the top-level classes done to get some feedback.

jackdreinhardt:
@pgollucci I like the idea of generating these types, I've explored this a bit in a project on my own. https://www.schemastore.org/json/ has Github Workflow and Github Actions schemas (and many more). Generating types from these using something like [json-schema-to-typescript](https://www.npmjs.com/package/json-schema-to-typescript) on a weekly cadence would offload the maintenance effort

gradybarrett:
@jackdreinhardt, schema-driven definitions would be awesome. I started down that road with modeling workflows, but quickly found that the half-dozen or so jsonschema to TS conversion libraries I tried all behaved radically differently. They also all lacked core schema feature support to produce useful output for this complex of a schema.

In experimenting with the conversion libs, how did you get around them not supporting certain schema keywords?

My impression so far is that we want more than just types, so that's how I'm proceeding right now. I share concerns about modeling a schema that may change, but also share Elad's perspective that the schema is not likely to change radically. Provided appropriate escape hatches are in place, folks should be able to adopt new features before they're officially modeled.

[...]

Issue #137, opened 2020-10-01
RFC: gitlab CI support


dariusjs:
I've created this as an RFC, as I am not sure if these types of extensions are considered. It would be nice if Github as the default workflow was configurable so you could have different providers such as Gitlab.

Currently the NodeBuildWorkflow has tight coupling with the GithubWorkflow class as it extends it on node-project. If the relationship was different gitlab CI support could be added.

Gitlab CI places a file called .gitlab-ci.yml in the root of a project, a little different to Githubs ./workflows structure. Sample gitlab CI files are here.
https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/templates/Bash.gitlab-ci.yml

I've tried working on this myself, but wasn't sure how to proceed because of the class extension.

robertd:
Gitlab's package registry support would be great too. They’ve recently released package registry support for everyone to use in their core offerings (https://about.gitlab.com/releases/2020/08/22/gitlab-13-3-released/#package-registry-now-available-in-core)

jackdreinhardt:
Agreed. I see a lot of potential abstracting projects from CI providers. Ideally, the developer could choose any provider (GitHub Workflows, Gitlab CI, TravisCI, CircleCI, etc). The developer could also choose to generate a 'headless' project if projen doesn't yet support their project type, where only CI files are generated.

eladb:
I like this direction. I've been working in that direction here: https://github.com/projen/projen/pull/387

ThomasSteinbach:
Cool. Projen is very interesting, however better gitlab support would be nice.

delprofundo:
Yeah I feel the same as Thomas. Trying to understand what we need to do with GitLab is bewildering and from the silence at the end of issues asking Q's about it its clear its not worth GL users time.

mrgrain:
Yes @delprofundo there is unfortunately no out-of-the-box support for any of the higher lever projects at the moment.
What we have is some helper classes to create gitlab configuration: https://projen.io/gitlab.html

So the general process would be:
- Use the project you want
- Manually rebuild the GitHub Actions workflow in GitLab
- Remove or ignore the unneeded GH Actions workflows

[...]

PR #895, opened 2021-06-15
feat(github): task-based github workflows

eduardomourar:
There is no real logic change within this PR. The only thing is that a new task-based GitHub workflow is being publicly exposed, as well as in use now by `NodeProject` and `Release` classes.

Related to #74 and #298.
 
---


src/__tests__/release/__snapshots__/release.test.ts.snap
```diff
@@ -88,10 +106,16 @@ jobs:
...
           fetch-depth: 0
+      - name: Anti-tamper check
+        run: git diff --ignore-space-at-eol --exit-code
       - name: Set git identity
         run: |-
           git config user.name \\"Automation\\"
           git config user.email \\"github-actions@github.com\\"
+      - name: Bump to next version
+        run: projen bump
+        env:
```
eladb, line 117: All those snapshot changes seem like bugs 🐜

eladb (changes requested):

src/github/generic-workflows.ts
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Permissions for the build job.
+   */
+  readonly permissions: JobPermissions;
+}
+
+/**
+ * A GitHub generic workflow.
+ */
+export class GenericWorkflow extends GithubWorkflow {
```
Chriscbr, line 102: Will it be possible to expand this beyond GitHub-specific workflows? Based on the name `GenericWorkflow`, it reads to me like a universal workflow, i.e. one that is generic across CI providers, so it would be great if we could later upgrade this to support #137.

eduardomourar: I believe it is best to rename this to `GenericGitHubWorkflow` to avoid confusion, because the implementation is quite specific to GitHub.

src/__tests__/release/__snapshots__/release.test.ts.snap
eduardomourar: done

eduardomourar:
@eladb , I fixed the snapshots. Please, could you have a look into this PR and let me know what do you think?

src/github/generic-workflows.ts
```diff
@@ -0,0 +1,211 @@
...
+   * The triggers for the workflow.
+   *
+   * @default - by default workflows can only be triggered by manually.
+   */
+  readonly trigger?: Triggers;
+
+  /**
+   * Initial steps to run before the source code checkout.
+   */
+  readonly initialSteps?: JobStep[];
```
eladb, line 49: Rename to `preCheckoutSteps`?

src/github/generic-workflows.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * Steps to run after the main build step.
+   */
+  readonly postBuildSteps?: JobStep[];
+
+  /**
+   * Disables anti-tamper checks in the workflow.
+   */
+  readonly antitamperDisabled?: boolean;
```
eladb, line 80: Reverse logic:

```suggestion
  readonly antitamper?: boolean;
```

src/github/generic-workflows.ts
```diff
@@ -0,0 +1,211 @@
...
+  readonly task: Task;
+
+  /**
+   * Main build step used in the workflow.
+   * @default {run: `projen ${task.name}`}
+   */
+  readonly buildStep?: JobStep;
+
+  /**
+   * Steps to run after the main build step.
```
eladb, line 73: Please add `@default` to all options

eladb (changes requested):
This looks really good, but I am not sure about the name... `GenericGithubWorkflow` is not really conveying the idea behind this workflow.

Trying to think what's the abstraction here and I think it's about tasks. It's basically a workflow backed by a projen task.

Perhaps we can call it `TaskGithubWorkflow`? Not sure...

src/node-project.ts
```diff
@@ -951,108 +969,23 @@ export class NodeProject extends Project {
...
-    const postSteps = options.postSteps ?? [];
-
-    const antitamperSteps = (options.antitamperDisabled || !this.antitamper) ? [] : [{
-      name: 'Anti-tamper check',
-      run: 'git diff --ignore-space-at-eol --exit-code',
-    }];
-
-    const job: Mutable<workflows.Job> = {
-      runsOn: 'ubuntu-latest',
+    const project = github.project as NodeProject;
```
eladb, line 976:
```suggestion
    const project = this;
```

src/node-project.ts
```diff
@@ -951,108 +969,23 @@ export class NodeProject extends Project {
     );
   }
 
-  private createBuildWorkflow(name: string, options: NodeBuildWorkflowOptions): GithubWorkflow {
-    const buildJobId = options.jobId;
-
+  private createBuildWorkflow(options: NodeWorkflowOptions): TaskGithubWorkflow {
```
eladb, line 972: Feels like this method can be removed and the code moved back to the call site.

src/release/release.ts
```diff
@@ -136,23 +135,24 @@ export interface ReleaseOptions extends ReleaseProjectOptions {
  * By default, no branches are released. To add branches, call `addBranch()`.
  */
 export class Release extends Component {
+  public static readonly BUILD_JOBID = 'release';
```
eladb, line 138: Why do we need this to be public?

eladb (changes requested):
Can you make sure to rebase this branch with `main`? There are a few changes in the `Release` class that I don't see here.

eduardomourar:
> Can you make sure to rebase this branch with `main`? There are a few changes in the `Release` class that I don't see here.

Yes, working on it. I am trying to figure out how to resolve the conflicts, but I will get there.

src/release/release.ts
eduardomourar: I usually like the idea of making those default values accessible to the user, but I will leave private to align with what it was originally.

eladb: As a general rule I rather expose the minimum amount of API possible. It's always possible to extend an API but impossible to narrow without breaking.

src/node-project.ts
```diff
@@ -597,15 +618,19 @@ export class NodeProject extends Project {
...
-        antitamperDisabled: mutableBuilds, // <-- disable anti-tamper if build workflow is mutable
-        image: options.workflowContainerImage,
-        codeCov: options.codeCov ?? false,
-        codeCovTokenSecret: options.codeCovTokenSecret,
-      });
+        task: this.buildTask,
+        buildStep: {
+          name: 'Build',
+          run: this.runTaskCommand(this.buildTask),
+        },
```
eladb, line 627: This feels wrong (why do we need to pass buildTask twice). Isn't the whole point of the workflow to run the build task?

src/release/release.ts
```diff
@@ -136,23 +135,24 @@ export interface ReleaseOptions extends ReleaseProjectOptions {
  * By default, no branches are released. To add branches, call `addBranch()`.
  */
 export class Release extends Component {
+  private static readonly BUILD_JOBID = 'release';
```
eladb, line 138: Just leave this as a file-local constant.

As a general practice it's better to keep commits that include refactoring to the bear minimum and not "pollute" them with stylistic changes so it's easier to have confidence

src/release/release.ts
```diff
@@ -291,49 +293,21 @@ export class Release extends Component {
...
-    steps.push({
-      name: 'Set git identity',
-      run: [
-        'git config user.name "Automation"',
-        'git config user.email "github-actions@github.com"',
-      ].join('\n'),
-    });
-
-    steps.push(...this.preBuildSteps ?? []);
+    const postBuildSteps = new Array<JobStep>(...this.postBuildSteps);
```
eladb, line 296: Just not just pass `this.postBuileSteps` as is?

src/release/release.ts
```diff
@@ -346,41 +320,45 @@ export class Release extends Component {
...
+        // otherwise tags are not checked out
+        'fetch-depth': 0,
+      },
+      antitamper: false,
+      preBuildSteps,
+      task: this.task,
+      buildStep: {
+        name: 'Release',
+        run: this.project.runTaskCommand(releaseTask),
+      },
```
eladb, line 358: ??

eladb (changes requested):

src/node-project.ts
eduardomourar: we are overriding the default value for the name within the build step

src/release/release.ts
eduardomourar: the same as above

src/node-project.ts
eduardomourar: i will make a commit removing `buildStep` property and you will see they would look like in the snapshots

src/release/release.ts
eduardomourar: done and moved the other constants in the `task-workflow.ts` as well to align with your comment

src/release/release.ts
eduardomourar: this is needed (improved the unit test to cover it), because we are running the `createWorkflow` method while looping over every branch.

src/node-project.ts
eladb: We can just use the name of the task in the step inside the workflow - no worries if the step display name will change.

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Permissions for the build job.
+   */
+  readonly permissions: JobPermissions;
+}
+
+/**
+ * A GitHub workflow for common build tasks within a project.
+ */
+export class TaskGithubWorkflow extends GithubWorkflow {
```
eladb, line 114: We should rename `GithubWorkflow` to `Workflow` at some point (because now it's under the `github` submodule), but this one can already be named appropriately:

```suggestion
export class TaskWorkflow extends GithubWorkflow {
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+  readonly jobId: string;
+
+  constructor(github: GitHub, options: TaskGithubWorkflowOptions) {
+    super(github, options.name);
+    this.jobId = options.jobId ?? DEFAULT_JOB_ID;
+    this.github = github;
+    this.createWorkflow(options);
+  }
+
+  protected getMainStep(options: TaskGithubWorkflowOptions) {
```
eladb, line 125: Why is this needed?

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+    return options.buildStep ?? {
+      name: options.task.name,
+      run: this.github.project.runTaskCommand(options.task),
+    };
+  }
+
+  protected createWorkflow(options: TaskGithubWorkflowOptions) {
```
eladb, line 132: Why does this have to be protected? If we don't have a concrete use case for this to be protected (which is effectively public API), always prefer `private`.

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+    const antitamperSteps = antitamper ? [{
+      name: 'Anti-tamper check',
+      run: 'git diff --ignore-space-at-eol --exit-code',
+    }] : [];
+
+    if (options.artifactsDirectory) {
+      finalSteps.push({
+        name: 'Upload artifact',
+        uses: 'actions/upload-artifact@v2.1.1',
+        if: 'always()',
```
eladb, line 162: why do we need `always()` here?

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+        with: {
+          name: options.artifactsDirectory,
+          path: options.artifactsDirectory,
+        },
+      });
+    }
+
+    const job: Mutable<Job> = {
```
eladb, line 170: Does it have to be mutable?

```suggestion
    const job: Job = {
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+        // this will identify any non-committed files generated during build (e.g. test snapshots)
+        ...antitamperSteps,
+
+        ...finalSteps,
+      ],
+    };
+
+    this.addJobs({ [this.jobId]: job });
+
+    return this;
```
eladb, line 216: Not needed:

```suggestion

```

src/github/task-workflow.ts
eladb, line 132: This method feels redundant

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+        {
+          name: 'Checkout',
+          uses: 'actions/checkout@v2',
+          ...checkoutWith,
+        },
+
+        // perform an anti-tamper check immediately after we run projen.
+        ...antitamperSteps,
+
+
```
eladb, line 189:
```suggestion
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+      ],
+    };
+
+    this.addJobs({ [this.jobId]: job });
+
+    return this;
+  }
+}
+
+type Mutable<T> = { -readonly [P in keyof T]: T[P] };
```
eladb, line 220: Not needed

src/node-project.ts
```diff
@@ -1178,4 +1032,4 @@ export class DependenciesUpgradeMechanism {
   }
 }
 
-type Mutable<T> = { -readonly [P in keyof T]: T[P] };
+export type NodeWorkflowOptions = Omit<TaskGithubWorkflowOptions, 'buildStep'>;
```
eladb, line 1035: Why is this needed?

src/github/task-workflow.ts
```diff
@@ -0,0 +1,220 @@
...
+   * @default - not set
+   */
+  readonly postBuildSteps?: JobStep[];
+
+  /**
+   * Enables anti-tamper checks in the workflow.
+   *
+   * @default true
+   */
+  readonly antitamper?: boolean;
```
eladb, line 90: I don't think anti-tamper is needed at the TaskWorkflow level. It's only used for the build workflow (the release workflow includes it as part of the task), so I'd just leave it there.

This will also mean that you no longer need the `finalSteps` option because basically `postBuildSteps` can cover all the post build activity.

eladb (changes requested):

src/github/task-workflow.ts
eduardomourar: you would want to upload artifacts even if the previous steps have failed (e.g. coverage report, internal logs, etc)

eladb: Is that what that means? Ha! TIL.

(can you add a comment?)

src/node-project.ts
eduardomourar: the interface NodeWorkflowOptions was in that file already, but i will remove it

src/github/task-workflow.ts
eduardomourar: no reason. removed

src/github/task-workflow.ts
eduardomourar: done

src/github/task-workflow.ts
eduardomourar: it does not

src/github/task-workflow.ts
eduardomourar: will do

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+   * The main task to be executed.
+   */
+  readonly task: Task;
+
+  /**
+   * Main build step used in the workflow.
+   *
+   * @default - by default we will run `projen ${task.name}`
+   */
+  readonly buildStep?: JobStep;
```
eladb, line 76: Remove this please

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * @default - not set
+   */
+  readonly postBuildSteps?: JobStep[];
+
+  /**
+   * Workflow environment variables.
+   * @default {}
+   */
+  readonly env?: Record<string, string>;
```
eladb, line 89:
```suggestion
  readonly env?: Record<string, string | undefined>;
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+   * Permissions for the build job.
+   */
+  readonly permissions: JobPermissions;
+}
+
+/**
+ * A GitHub workflow for common build tasks within a project.
+ */
+export class TaskWorkflow extends GithubWorkflow {
+  readonly github: GitHub;
```
eladb, line 101:
```suggestion
  private readonly github: GitHub;
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+  readonly jobId: string;
```
eladb, line 102:
```suggestion
  public readonly jobId: string;
```

src/node-project.ts
```diff
@@ -523,19 +523,44 @@ export class NodeProject extends Project {
...
 
-      const updateRepo = new Array<any>();
+      const postBuildSteps = new Array<any>();
       const gitDiffStepId = 'git_diff';
       const hasChangesCondName = 'has_changes';
       const hasChanges = `steps.${gitDiffStepId}.outputs.${hasChangesCondName}`;
       const repoFullName = 'github.event.pull_request.head.repo.full_name';
 
+      // disable anti-tamper if build workflow is mutable
+      const antitamperSteps = (!mutableBuilds ?? this.antitamper) ? [{
```
eladb, line 539: Remove

src/node-project.ts
```diff
@@ -597,15 +628,17 @@ export class NodeProject extends Project {
           repository: repo,
         } : undefined,
 
-        postSteps: updateRepo,
+        preBuildSteps: [
+          ...antitamperSteps,
```
eladb, line 632: Remove

eladb (changes requested):

src/node-project.ts
eduardomourar: i think i don't understand. we will still have an options called `antitamper` for a NodeProject, so we have to add the checks steps there, right?

eladb: Oh sorry! Didn't realize this is in NodeProject. Taking it back.

src/github/task-workflow.ts
eduardomourar: i get the following JSII error
> Compilation errors prevented the JSII assembly from being created
src/github/task-workflow.ts:82:12 - error JSII3999: Encountered optional value in location where a plain type reference is expected
82   readonly env?: Record<string, string | undefined>;
              ~~~
src/github/task-workflow.ts(137,7): error TS2322: Type 'Record<string, string | undefined> | undefined' is not assignable to type 'Record<string, string> | undefined'.
  Type 'Record<string, string | undefined>' is not assignable to type 'Record<string, string>'.
    Index signatures are incompatible.
      Type 'string | undefined' is not assignable to type 'string'.
        Type 'undefined' is not assignable to type 'string'.

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+  readonly container?: ContainerOptions;
+
+  /**
+   * Adds an 'if' condition to the workflow.
+   */
+  readonly condition?: string;
+
+  /**
+   * A directory name which contains artifacts to be uploaded (e.g. `dist`).
+   *
```
eladb, line 33:
```suggestion
   * A directory name which contains artifacts to be uploaded (e.g. `dist`). If this is set, the 
   * contents of this directory will be uploaded as an artifact at the end of the workflow run.
   *
```

src/github/task-workflow.ts
```diff
@@ -0,0 +1,180 @@
...
+   * @default - not set
+   */
+  readonly artifactsDirectory?: string;
+
+  /**
+   * The triggers for the workflow.
+   *
+   * @default - by default workflows can only be triggered by manually.
+   */
+  readonly trigger?: Triggers;
```
eladb, line 44:
```suggestion
  readonly triggers?: Triggers;
```

src/github/task-workflow.ts
eladb: Ping

src/github/task-workflow.ts
eladb: Changed my suggestion - I believe this should work

eladb (commented):
Getting there... A few minor comments remaining :-D Thanks for bearing with me

src/github/task-workflow.ts
eduardomourar: done

eladb (approved):

PR #897, opened 2021-06-16
feat(cdk8s): option to specify module versions

### LINKED ISSUE #896: bug: cdk8s-plus-17 version conflicting
Since we use `cdk8sVersion` to set the version for cdk8s-plus-17, and cdk8s-plus-17 has a peer dependency of a version lower than the current cdk8sVersion used. We have a conflict.

```
Error: Error: Declared dependency on version 1.0.0-beta.21 of cdk8s, but version 1.0.0-beta.24 was found
```

Having a cdk8sPlusVersion option would help in this case.

Hunter-Thompson:
Option to set a version for cdk8s-plus-17, and cdk8s-cli, pinned or unpinned. Defaults to `this.cdk8sVersion` for backward compatibility.

Fixes #896 

---


eladb (approved):

eladb (changes requested):
I guess we should also offer a way to control the CDK8s CLI version. All of these modules now have separate versions.

Hunter-Thompson:
@eladb, made the changes.

eladb (approved):

PR #899, opened 2021-06-18
fix(self mutation): GitHub auto-merge may commit the previous commit

### LINKED ISSUE #898: Self-mutation does not work well with GitHub's auto-merge
The self-mutation procedure on Pull Request validation does not integrate well with GitHub's "request auto-merge" feature. 

When he code on the PR requires a mutation, the validation process will add a new commit on the pull request (self-mutation), but still mark the original commit as "successful". Unfortunately, this means that if auto-merge was enabled on the PR, GitHub might eagerly merge the original commit as soon as all it's checks pass; instead of merging the self-mutation commit.

This problem can be fixed by having the PR validation check fail if it inserts a self-mutation commit on the PR, as that new commit will have all checks successful. This will ensure GitHub auto-merges the correct commit into the base branch.

RomainMuller:
Adds an explicit check failure in case self mutation was
added on a PR, so that GitHub's built-in "resuest auto-merge"
does not risk merging the wrong commit into the base branch.

Fixes #898

---


eladb (approved):

eladb:
Tested this and it works great! Thanks!

PR #903, opened 2021-06-19
docs: add ARCHITECTURE.md

### LINKED ISSUE #647: Add an ARCHITECTURE.md
Inspired from reading https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html

I think it could be useful just to document some high level concepts of how the library is written, such as how the library makes use of [constructs](https://github.com/aws/constructs), and how some of the core projen functions work (like the synthesis / cleanup process, and how project options are discovered using jsii).

Some of projen's architecture is still up in the air, e.g.:
- At the last projen meetup, Elad mentioned the possibility of shifting from an inheritence based project model (where there is a strict hierarchy from Project -> NodeProject -> TypeScriptProject -> JsiiProject -> ConstructLibrary -> AwsCdkConstructLibrary) to a more modular project model
- How subprojects are modeled and implemented (there is a system in place, and we have some docs, but I think there's a bit of confusing behavior, and I know monorepos are a priority on the roadmap)

So I might omit writing about those for now.

- [x] I can try to implement this

Chriscbr:
Closes #647.

Not a comprehensive doc - just focused on file synthesis for v1. I've tried to keep this at a high level while including details that I thought would be most useful, and hopefully won't be outdated 6 months from now. As always, open to feedback. :-)

Even though the sub-headers are questions, the doc is spiritually meant to be just for architecture things, not a FAQ - open to restructuring things / renaming headers to make this more clear if needed.

I've also tried to avoid linking to source code for the most part in case things get refactored / to avoid stale links.

[View rendered version](https://github.com/Chriscbr/projen/blob/docs/architecture/ARCHITECTURE.md)

---


eladb (approved):
Link from CONTRIBUTING.md

Issue #904, opened 2021-06-22
Unable to overwrite task definition without removing all dependent tasks


john-tipper:
[This change ](https://github.com/projen/projen/commit/a621e396aa65e1b6334b47b8e552e0e0ec531baa#diff-67c4aaef15040e0527e1b1955306c8cff66858840340a94b9a5d13752d4d99ab) removed the ability to add a task definition without removing it first. However, you cannot remove a task unless all dependent tasks are removed also.  This means that it's not possible to overwrite a task without first removing all dependent tasks, then removing the task you wish to overwrite, re-adding the amended task, then replacing all of the dependent tasks.

The fix is to simply remove this if statement: https://github.com/projen/projen/blob/a621e396aa65e1b6334b47b8e552e0e0ec531baa/src/tasks/tasks.ts#L54-56

PR #905, opened 2021-06-22
fix: #904 allow overwrite of tasks

john-tipper:
This PR allows for tasks to be overwritten: without this, a task may not be changed without removing the whole of the DAG of dependent tasks and the task itself, then adding the change, then replacing the whole of the DAG.

---

### DIFF
--- a/src/tasks/tasks.ts
+++ b/src/tasks/tasks.ts
@@ -51,9 +51,6 @@ export class Tasks extends Component {
    */
   public addTask(name: string, options: TaskOptions = {}) {
     const task = new Task(name, options);
-    if (this._tasks[name]) {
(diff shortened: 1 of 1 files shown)


src/tasks/tasks.ts
```diff
@@ -51,9 +51,6 @@ export class Tasks extends Component {
    */
   public addTask(name: string, options: TaskOptions = {}) {
     const task = new Task(name, options);
-    if (this._tasks[name]) {
```
eladb, line 54: Maybe we should make this explicit? Something like `replaceTask()`?

eladb:
@john-tipper I am closing this for now. Feel free to reopen when ready to continue

Issue #782, opened 2021-05-12
Allow release on demand


pgarbe:
Since the change to the tag-based release projen makes it impossible to use other release mechanism. 

My use case is to run the "release" task on a developer machine to increase the version, create the changelog, set a tag, and commit everything. In my CodePipeline it publish the package if the commit contains a version tag.

Even with some ~hacks~ escape hatches it's not possible to completely restore the old behavior. Projen will still overwrite the package.json version with "0.0.0" when it runs. The workaround is also very messy as it's not easy to change `versionrc.json`.

eladb:
@pgarbe yes, this use case is not very well supported. I need to think what's the best way to address this.

pgollucci:
What's wrong with what you did in `aws/constructs`?  removing push triggers via `releaseEveryCommit: false`?

eladb:
Keep open please

Issue #726, opened 2021-04-23
Manual releases are broken in new tag-based bumps


smacintyre:
The readme states:

> Bump & release scripts with CHANGELOG generation based on Conventional Commits

This should be updated since the CHANGELOG is no longer generated and release is no longer supported. As per  #634 

Also please document how to use the new bump versioning. It doesn't work out of the box as the old release pattern did -- at least for me. How can we use it?

eladb:
@smacintyre See https://github.com/projen/projen/issues/724 for details on bootstrapping

smacintyre:
@eladb I'm not sure if that applies. I'm talking about an existing project where we have been using the previous `release` task. So there are existing tags. 

Here is my output:
```bash
$ yarn bump
yarn run v1.22.10
$ npx projen bump
🤖 bump | condition: ! git log --oneline -1 | grep -q "chore(release):"
🤖 bump | git describe --tags --match="v*" --first-parent --abbrev=0 > .version.tmp.json
🤖 bump | standard-version
✔ bumping version in package.json from 0.0.0 to 0.9.6
✔ created .changelog.tmp.md
✔ outputting changes to .changelog.tmp.md
```

The only change is to `package.json` according to git:
```bash
$ git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   package.json

no changes added to commit (use "git add" and/or "git commit -a")
```

This gets wiped out if I run `yarn projen`. No new tags are created; there is nothing to push.

```bash
$ git push --follow-tags origin main
Everything up-to-date
```

I'm sorry, I'm I completely missing something? How is this new versioning process supposed to work?

eladb:
Are you performing manual bumps/releases. In the new model Bumps only happen during release builds (in a GitHub workflow).

The checked in package.json file always says 0.0.0

Of course we need to document!

smacintyre:
@eladb Yes, doing manual releases

eladb:
Okay so I think the new scheme doesn't support manual releases (and it should). 

Let's repurpose this issue for that

smacintyre:
We generally do a bump at the end of a sprint. We don't use Github's release. We use CDK Pipelines for releases
.

gradybarrett:
@eladb what do you think about enabling standard-version tagging (and possibly the changelog) if the Github release workflow is disabled?

[...]

PR #917, opened 2021-07-04
feat(release): introduce "release" task

eladb:
Extract some steps from the GitHub release workflow into a projen task so it can be executed manually via `projen release`.

The `release` task **prepares a release** by performing the following: `bump` + `build` + `unbump` + anti-tamper. 

The task will emit two files under the `dist/` directory which can be used to publish a release: `version.txt` and `changelog.md` (both are .gitignored). The GitHub workflow will pick up those files (and the rest of the contents under `dist`) and publish them as GitHub releases and to the various package managers.

Next steps: this is an initial step to address #782 and #726. The next step is to add tasks that can be used to publish a release. These tasks can be composed of `publish:github`, `publish:npm`, `publish:nuget`, which we can then use in the GitHub workflow. Additionally, we will likely want to allow manual creation of git tags and committing of a CHANGELOG file when projects are not hosted on GitHub.

Misc: add more information to task inspection and execution outputs (namely, include environment variables).

---


package.json
```diff
@@ -63,6 +64,7 @@
     "standard-version": "^9",
     "typescript": "^3.9.10"
   },
+  "peerDependencies": {},
```
eladb, line 67:
```suggestion
```

eladb:
@Chriscbr can you take a quick look to see I didn't mess this up?

.github/workflows/release.yml
```diff
@@ -13,7 +13,6 @@ jobs:
       contents: write
     env:
       CI: "true"
-      RELEASE: "true"
```
Chriscbr, line 16: "release" is the only task that uses this yes?

src/tasks/runtime.ts
```diff
@@ -158,18 +170,25 @@ class RunTask {
...
+  private resolveEnvironment(parents: string[]) {
+    let env = this.runtime.manifest.env ?? {};
+
+    // add env from all parent tasks one by one
+    for (const parent of parents) {
+      env = {
+        ...env,
+        ...this.runtime.tryFindTask(parent)?.env ?? {},
+      };
+    }
```
Chriscbr, line 183: A unit test for this would be nice to have

Chriscbr (approved):
+1 on moving towards modeling github workflows as much as possible within projen tasks.

Terminology question - what's the difference between "releasing" and "publishing"? It feels to me as though the codebase uses them interchangeably, but it's not really clear since we have have public APIs using these names (`Release` and `Publisher`).

Related, it looks like we are dividing releasing/publishing into a two step process (preparing the files for distribution, and then posting them on npm/maven/etc.). If the "release" task doesn't actually release, maybe another name would be better (e.g. "prepare-release", "publish:prepare")?

~Can you add some info in `docs/` about projen's manual release features/options?~ 😄 edit: This part hasn't been finished yet, so nvm

Looks great!

.github/workflows/release.yml
eladb: Yap.

src/tasks/runtime.ts
eladb: Done.

PR #918, opened 2021-07-04
chore: remove task categories

Chriscbr:
Since the start menu was removed in v0.24.6, task categories are no longer used, and can be removed.

I'm open to leaving them in if we think they can be used in other ways in the future (as metadata or otherwise).

BREAKING CHANGE: Task categories have been removed.

---


eladb:
Lol I was just thinking we should get rid of these!

src/node-package.ts
```diff
@@ -686,7 +686,7 @@ export class NodePackage extends Component {
   }
 
   private determineVersion(currVersion?: string) {
-    if (!this.isRelaseBuild) {
```
eladb, line 689: ☹️

src/tasks/model.ts
```diff
@@ -24,6 +24,7 @@ export interface TaskCommonOptions {
    * Category for start menu.
    *
    * @default TaskCategory.MISC
+   * @deprecated - start menu has been removed (use `projen --help`)
```
eladb, line 27: Just remove this I think. I doubt it will break anyone

eladb (changes requested):

src/node-package.ts
Chriscbr: do you want this omitted?

eladb (approved):

PR #926, opened 2021-07-06
feat: upgrade default @types/node to v12

Chriscbr:
Node v10 is [no longer being supported](https://nodejs.org/en/about/releases/), so this PR updates the default version of @types/node vended in TypeScript projects to v12, as well as the minNodeVersion used by projen itself to v12.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2.2.0
         with:
-          node-version: 10.17.0
+          node-version: 12.19.16
       - name: Install dependencies
         run: yarn install --check-files --frozen-lockfile
       - name: Set git identity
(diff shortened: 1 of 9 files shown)


eladb (changes requested):
This change might be problematic because the `jsii/superchain` image which is used in many JSII projects uses node v10.

@RomainMuller what do you think?

Chriscbr:
@eladb I see there's also a version of jsii/superchain with node v14 [(link](https://github.com/aws/jsii/tree/main/superchain)). One alternative could be to try and update all github actions to use this image and just use node v14 in all projen-dependent projects -- but I expect this would cause a lot of breakage / incompatibilities since aws-cdk is still supporting node v10. So for now I will close this.

eladb:
@Chriscbr let's resurrect this pull request. I think we need to move up to node 12 or node 14

PR #929, opened 2021-07-06
chore: do not use “projen publish:xxx” in publish jobs

eladb:
Since the publish job does not checkout the sources, we cannot really use projen tasks there. Instead, just emit the actual command and environment into the workflow. Not a huge deal.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,8 +61,10 @@ jobs:
           name: dist
           path: dist
       - name: Release
-        run: /bin/bash ./projen.bash publish:npm
+        run: npx -p jsii-release@latest jsii-release-npm
         env:
+          NPM_DIST_TAG: latest
+          NPM_REGISTRY: registry.npmjs.org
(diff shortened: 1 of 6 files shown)


Chriscbr:
@eladb Is checking out the sources not possible? Or would it just incur some kind of extra costs? Just curious since it seems like rendering the full command is a bit of an abstraction leak, especially if users wants to extend `publish:xxx` to have multiple steps.

eladb:
> @eladb Is checking out the sources not possible? Or would it just incur some kind of extra costs? Just curious since it seems like rendering the full command is a bit of an abstraction leak, especially if users wants to extend `publish:xxx` to have multiple steps.

Yes, I wanted the job environment to be minimal. The main reason I want to avoid checking out the sources, installing projen and invoking it is that it will add constraints to the job environment (eg the docker image) and I want to minimize those constraints so it will be possible to optimize the publish job.

But you are bringing a good point about the fact that these tasks should not be extended/modified and we can at least add some protection against that so it won't trip people up

Issue #894, opened 2021-06-15
jsii lookup issues


joostvdwsd:
We have nested projen module dependencies. It follows this structure:

core module (implements base concepts extended from projen)
feature modules (Implements feature projects extended from the core module)

This used to work perfectly, however with this feature this was broken:
https://github.com/projen/projen/pull/744

During projen new --from <feature module> it will not see it as a valid module

This has to do because previously inventory.ts would lookup all jsii modules. Now it does only the current module (feature) + projen. It will not find the core module.

Not enough time before my holyday to create a pr but it would be nice if this would work again

Issue #605, opened 2021-03-10
Only select project from selected project during new --from


joostvdwsd:
With the creation of a new project all projects are listed inside the node_modules folder. Projen native projects are excluded:

https://github.com/projen/projen/blob/51259cf8556e9fcc6e5ed07c21e6dc528a66738a/src/cli/cmds/new.ts#L325

It would be nice to only list the projects of the actually specified project in the --from.

We have a structure like this:
package : @company-projen/core
package: @company-projen/business-unit

If i want to create a package from the business-unit project it will also list all project from the core project which is undesired (if that was the idea I would have specified the core package)

Especially if we want to make small packages with one specific project type we still have to specify the project name eventhough the package itself only exports 1 project type

PR #937, opened 2021-07-12
fix(inventory): Recursive JSII module Lookup 

joostvdwsd:
When developing external modules we encountered some issues with the projen new command:
1) https://github.com/projen/projen/issues/894 Prevented us to upgrade
2) https://github.com/projen/projen/issues/605 Was a small anoyance that prevented us to have single project packages to work smoothly

Both problems occur with the following package structure:

FeaturePackage => GenericOrganisationProjenPackage => projen

This would end up in the following node_modules structure:
```
  node_modules
  - FeaturePackage
    - node_modules
      -  GenericOrganisationProjenPackage
  - projen
```

Before this PR the logic of the inventory was to scan the full folder of node modules for .jsii modules. It would list all project types and had a exclusive logic to filter out projen native projects. Projen would only see the top level packages with this logic, eventhough the jsii references were correct

This PR changes the call of the discovery function for newProjectFromModule. It will now only pass the requested module. Recusively it will scan for jsii types based on the jsii dependencies. This way it will discover all 3 packages. This will solve issue 1 (and potentially scan a bit faster because it will not search through all package folders now)

As a side optimalisation, it will only list the projects of the requested module to see during the new phase. This means it will allow single project feature packages without project specification.

fixes: 894
fixes: 605

---


joostvdwsd:
Also changes the resolveProjectType (called from Project constructor) to use the same lookup logic. In te original code this logic was kinda duplicated between discover and resolveProjectType

joostvdwsd:
@eladb any issues with this one? Kinda need at least this logic. We can’t upgrade past 17.x.x.  later releases are fixing some windows related builds

eladb (approved):

joostvdwsd:
Ok, that was stupid. Had for some reason false positive tests. Now corrected and reverted some redundant changes to the datamodels in inventory

eladb:
Please re-run build in your local clone and push the updated files.

joostvdwsd:
Finally time, was a regenerate of projen with modified workflow files. Now runs successfully in my repo

eladb:
@Chriscbr do you mine doing a quick review of this?

src/cli/cmds/new.ts
```diff
@@ -237,12 +237,20 @@ async function newProjectFromModule(baseDir: string, spec: string, args: any) {
 
   const specDependencyInfo = yarnAdd(baseDir, spec);
 
-  // collect projects by looking up all .jsii modules in `node_modules`.
-  const modulesDir = path.join(baseDir, 'node_modules');
-  const modules = fs.readdirSync(modulesDir).map(file => path.join(modulesDir, file));
+  // Remove optional semver information from spec to retrieve the module name
+  const moduleName = spec.replace(/\@([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?$/, '');
```
Chriscbr, line 241: Wow, semver tags are more complicated than I realized! But I guess regex does that to everything...

src/inventory.ts
```diff
@@ -87,15 +112,29 @@ export function discover(...moduleDirs: string[]) {
 
     const manifest = fs.readJsonSync(jsiiFile);
     for (const [fqn, type] of Object.entries(manifest.types as JsiiTypes)) {
-      jsii[fqn] = type;
+      jsii[fqn] = {
+        ...type,
+      };
```
Chriscbr, line 117: Not totally sure I understand the need for this - but I suppose making a shallow copy might make things safer?

src/inventory.ts
```diff
@@ -104,34 +143,25 @@ export function discover(...moduleDirs: string[]) {
...
-  const projenManifest = readJsiiManifest('projen');
-  for (const [fqn, type] of Object.entries(projenManifest.types as JsiiTypes)) {
-    jsii[fqn] = type;
-  }
+  // try picking the manifest. We only need the base folder but this is directly a nice check if we request from a valid jsii package
+  const jsiiManifestFile = require.resolve(`${moduleName}/.jsii`, {
+    paths: [
+      process.cwd(),
+    ],
+  });
```
Chriscbr, line 160: The original control flow in this function called `require.resolve()` without a `paths` argument (which I'm guessing would also have done a lookup in the current directory), so I'm assuming this is just making things more explicit?

Chriscbr (commented):
@eladb No glaring issues I can see.

It would be nice to have these components further unit tested eventually (I like that this PR is helping break functions into smaller parts). Setting up some live dummy projects on npm that demonstrate some of these edge cases could also help (e.g. a package that vends multiple project types, or a package that vends a project type that is _not_ a direct subclass from a core Project project type). Or some way to otherwise mock these. But not sure if it's worth the effort atm.

src/cli/cmds/new.ts
joostvdwsd: Ha, nice right..... I only needed the package name here. But a simple regex would always fail on scoped packages with the semver behind it. So therefor the regex for semver itself. Ideally I would have a generic method in the whole package for splitting-up a spec. Which would then also be used in things like the dependency handling

src/inventory.ts
joostvdwsd: hmm, good catch. This was due to an earlier iteration where I had an extra property, I can easily change it back to make it more compact. Don't think the copy is needed here

Issue #342, opened 2020-11-30
Where do I put Lambda code for a construct?


pszabop:
CDK Constructs sometimes need Lambda code.  For example see this [non-projen construct](https://github.com/eladb/cdk-dynamo-table-viewer):

The `dynamo-table-viewer` puts the code in `lib/lambda`.  That won't work as projen wipes the lib directory since the files in there are created files.

Putting the code in `src/lambda` won't work either, as the relative path after deploying the construct as a module is `lib/lambda`.

What is the best location to place lambda code for a Construct?

My assumption is the correct way to deploy the code is `Lambda.Code.fromAsset()`.  I just don't know where to put the asset.

Chriscbr:
> The dynamo-table-viewer puts the code in lib/lambda. That won't work as projen wipes the lib directory since the files in there are created files.

It should be possible to change this directory by passing a different name specifying a `libdir` option in the project constructor - would that fix the issue?

eladb:
Why not under `src/lambda` and reference it relative?

lukewiwa:
yeah this is kinda broken as the lambda code in the src directory gets compiled when I'd like to do the work elsewhere and just add a js file

Issue #783, opened 2021-05-12
Feature request: support lambdas written in TypeScript


haimlit:
### Background

projen generates CDK projects written in TypeScript. Lambda handlers don't support TypeScript natively, and require compiling their code from TypeScript to Javascript/node first. In order to write a construct containing a lambda handler written in TypeScript, a two phased compilation is required - one which is ad-hoc for lambda handlers, compiling them from TS to JS, and another which is the standard TS project compilation.

### Current state

Currently, to perform the ad-hoc lambda handler compilation, additional code is required on projen to add a new task to compile the lambda `*.ts` files. Here's a snippet which does this for a specific lambda handler:

```typescript
const bundleLambdaHandlerTask = project.addTask('bundle:handler/myHandler', {
  description: 'bundle handler/myHandler',
  exec: [
    'esbuild',
    '--bundle',
    `${project.srcdir}/handler/myHandler.ts`,
    '--target="node14"',
    '--platform="node"',
    `--outfile="${project.srcdir}/handler/myHandler.js"`,
    '--external:aws-sdk',
  ].join(' '),
});
project.compileTask.spawn(bundleLambdaHandlerTask);
```

It could be improved by scanning for all files with a pattern of `handler/*.lambda.ts` for example. [Link to `contruct-hub` which does this type of scan](https://github.com/cdklabs/construct-hub/blob/main/.projenrc.js#L138), where a different approach was taken by generating full lambda construct code behind the scenes.

### Feature request

This boilerplate could be saved if projen would have that step support OOTB.
See API proposal below to extend `AwsCdkConstructLibraryOptions`.
All values are also proposed to be the default ones.

```typescript
bundleLambda: {
  // AWS lambdas have AWS SDK available at runtime, which some lambdas use.
  // This flag saves the developer from explicitly declaring AWS SDK under devDeps section.
  addAwsSdkDep: true

  // Suffix to be used to scan for lambdas.
  suffix: 'lambda.ts'

  // Let developer choose whether the compiled lambda handler should be committed or not.
  addToGitIgnore: true
}
```

### Pitfall

On one hand, `aws-sdk` shouldn't be a dependency and get bundled, as it's a runtime dependency only. However, `eslint` doesn't play well with `aws-sdk` being a devDependency, as when a lambda imports `aws-sdk`, `eslint` requires disabling `import/no-extraneous-dependencies` rule. This should probably be part of the solution.

eladb:
Keep

PR #940, opened 2021-07-17
feat(awcdk-construct): lambda bundling

haimlit:
Solves #783 by introducing lambda bundling as part of build.
Includes construct generation, inspired by [construct-hub .projenrc](https://github.com/cdklabs/construct-hub/blob/main/.projenrc.js#L190).

Worth discussing:

1. By convention, files with suffix 'lambda' are assumed to be lambda handlers and are targeted for bundling and construct generation.
2. There are flags to turn off partial functionality (construct generation, customizing suffix for lambda) - is this desired? Or should we start with just on/off switch?
3. Initially implemented a flag whether to include `aws-sdk` dep, removed as it's too granular.
4. Initially included lambda handler stub file generation on the `src` folder, to enable running tests which synthesize the lambda construct. However, I've since noticed that the practice across most CDK constructs is to run the tests against the `lib` folder and not `src`, so if this practice is a working assumption then the stub isn't needed, which is why it's removed from this PR.

Fixes https://github.com/projen/projen/issues/342

---


eladb (changes requested):
At first glance, can you please add some docs to https://github.com/projen/projen/blob/main/docs/awscdk-construct.md - that would be a good way to describe the feature as this is been code reviewed?

src/__tests__/awscdk-construct.test.ts
```diff
@@ -58,6 +60,113 @@ describe('constructs dependency selection', () => {
...
+    // GIVEN
+    const project = new TestProject({
+      cdkVersion,
+    });
+
+    // WHEN
+    const snapshot = synthSnapshot(project);
+
+    // THEN construct created
+    expect(snapshot[`src/__tests__/integration/lambda/mylambda-${handlerExt}.ts`]).toMatchSnapshot();
```
eladb, line 77: Something odd is happening here. Where are the .lambda. files coming from. The TestProject is supposed to be located in a temporary directory...

src/awscdk-construct.ts
```diff
@@ -68,6 +72,22 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    * @featured
    */
   readonly cdkTestDependencies?: string[];
+
+  readonly bundleLambda?: boolean;
```
eladb, line 76: Docstrings please

src/__tests__/integration/lambda/mylambda-js.lambda.js
```diff
@@ -0,0 +1,3 @@
+export function handler() {
```
eladb, line 1: Can we add a test that checks that dependencies are bundled?

src/awscdk-construct.ts
```diff
@@ -159,6 +179,7 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
     this.cdkVersion = options.cdkVersionPinning ? options.cdkVersion : `^${options.cdkVersion}`;
     this.cdkDependenciesAsDeps = options.cdkDependenciesAsDeps ?? true;
 
+    this.logger.info('hello!!!!');
```
eladb, line 182:
```suggestion
```

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+   * @param bundleLambdaOptions Bundle lambda options
+   * @private
+   */
+  private bundleLambdas(bundleLambdaOptions?: BundleLambdaOptions) {
+    this.logger.debug('running bundleLambdas');
+    const lambdaSuffix = (bundleLambdaOptions?.suffix ?? '.lambda');
+    const generateConstruct = bundleLambdaOptions?.generateConstruct ?? true;
+    const handlerFilePattern = `${this.srcdir}/**/*${lambdaSuffix}\\.@(ts|js)`;
+
+    this.addDevDeps('aws-sdk'); // common lambda runtime dependency
```
eladb, line 266: I don't think that's needed. If users want they can add it easily (and control the version etc)

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+    this.addDevDeps('esbuild');
```
eladb, line 267: Add a major version `@^N`

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+
+    this.eslint?.addOverride({
```
eladb, line 269: Add a comment explaining why this is needed

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+      },
+      files: [handlerFilePattern],
+    });
+
+    // create bundle compile task for each lambda handler, generate convenience construct
+    glob.sync(handlerFilePattern).forEach(file => {
+      this.logger.debug(`found lambda handler ${file}`);
+      const parsedPath = path.parse(file);
+      const bundleDirName = `${parsedPath.name}.bundle`;
+      const bundleDir = path.join(parsedPath.dir, `${parsedPath.name}.bundle`);
```
eladb, line 281:
```suggestion
      const bundleDir = path.join(parsedPath.dir, bundleDirName);
```

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+      rules: {
+        'import/no-extraneous-dependencies': 'off',
+      },
+      files: [handlerFilePattern],
+    });
+
+    // create bundle compile task for each lambda handler, generate convenience construct
+    glob.sync(handlerFilePattern).forEach(file => {
+      this.logger.debug(`found lambda handler ${file}`);
+      const parsedPath = path.parse(file);
```
eladb, line 279: Please extract this to a public method called `addLambdaBundle(...)` so users can disable auto-detect and explicitly add bundles if they want.

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+      const bundleDirName = `${parsedPath.name}.bundle`;
+      const bundleDir = path.join(parsedPath.dir, `${parsedPath.name}.bundle`);
+      const compiledHandlerPath = bundleDir.replace(this.srcdir, this.libdir);
+      const constructFilePath = path.join(parsedPath.dir, parsedPath.name.replace(lambdaSuffix, '') + '.ts');
+
+      if (generateConstruct) {
+        new NodeLambdaConstructSourceCode(this, constructFilePath, bundleDirName);
+      }
+      const bundleLambdaTask = this.addTask(`bundleLambda:${file}`, {
```
eladb, line 288: I think in this context, "bundle:" is sufficient as a prefix.

```suggestion
      const bundleLambdaTask = this.addTask(`bundle:${file}`, {
```

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+        description: `Bundle lambda: ${file}`,
+        exec: [
+          'esbuild',
+          '--bundle',
+          file,
+          '--target="node14"',
```
eladb, line 294: Extract to an option. This also needs to impact the generated construct.

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+          '--platform="node"',
+          `--outfile="${path.join(compiledHandlerPath, 'index.js')}"`,
+          '--external:aws-sdk',
```
eladb, line 297: This needs to be an option (especially if people want to control the version of the SDK)

src/node-lambda-construct-source-code.ts
```diff
@@ -0,0 +1,47 @@
+import * as path from 'path';
```
eladb, line 1: Move to `src/cdk` (I am slowly trying to organize the src tree)

src/awscdk-construct.ts
```diff
@@ -225,10 +250,63 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
     return `${module}@${this.cdkVersion}`;
   }
+
+  /**
+   * Enables CDK apps to consume construct library lambdas.
+   *
+   * @param bundleLambdaOptions Bundle lambda options
+   * @private
+   */
+  private bundleLambdas(bundleLambdaOptions?: BundleLambdaOptions) {
```
eladb, line 260: I am wondering if it makes sense to extract all of this logic into a separate `Component` and just add it here. This way it will be possible e.g. to add that to other types of projects, such as CDK apps...

(and add it to CDK apps as well)

src/node-lambda-construct-source-code.ts
```diff
@@ -0,0 +1,47 @@
...
+    super.line();
+    super.open(`export interface ${propsName} extends lambda.FunctionOptions {`);
+    super.close('}');
+    super.line();
+    super.open(`export class ${constructName} extends lambda.Function {`);
+    super.open(`constructor(scope: Construct, id: string, props: ${propsName} = {}) {`);
+    super.open('super(scope, id, {');
+    super.line('runtime: lambda.Runtime.NODEJS_14_X,');
+    super.line('handler: \'index.handler\',');
+    super.line(`code: lambda.Code.fromAsset(path.join(__dirname, '/${handlerDirPath}')),`);
```
eladb, line 41:
```suggestion
    super.line(`code: lambda.Code.fromAsset(path.join(__dirname, handlerDirPath)),`);
```

src/node-lambda-construct-source-code.ts
```diff
@@ -0,0 +1,47 @@
...
+    super.line('import * as lambda from \'@aws-cdk/aws-lambda\';');
+    super.line('import { Construct } from \'constructs\';');
+    super.line();
+    super.open(`export interface ${propsName} extends lambda.FunctionOptions {`);
+    super.close('}');
+    super.line();
+    super.open(`export class ${constructName} extends lambda.Function {`);
+    super.open(`constructor(scope: Construct, id: string, props: ${propsName} = {}) {`);
+    super.open('super(scope, id, {');
+    super.line('runtime: lambda.Runtime.NODEJS_14_X,');
```
eladb, line 39: This needs to be configurable

src/awscdk-construct.ts
eladb, line 76: I am thinking that enable/disable should just be for automatic bundling, and the options at this level should reflect the defaults applied to the auto-detected handlers. Additionally users should be able to call something like `addBundleLambda(handlerPath[, options])` to explicitly specify a handler for bundling with its own options.

eladb (changes requested):
Please also take a look at the latest version in construct-hub to make sure we didn't miss anything.

src/node-lambda-construct-source-code.ts
haimlit: 👍 I actually did exactly did initially, but reverted to avoid a bigger refactor, or having inconsistent paths.

src/node-lambda-construct-source-code.ts
haimlit: Note sure I understood why would that work. The generated code has no `handlerDirPath` variable.

src/awscdk-construct.ts
haimlit: What's the motivation for adding this to CDK apps? I thought CDK apps are the end consumer of constructs, and won't benefit from the bundling process during build time, and the CDK bundling works well for that use case. Given the bundling part is redundant, is the lambda construct generation useful enough to be used on CDK apps as well?

If you think the answer is yes, I'll extract to `Component`

src/awscdk-construct.ts
haimlit: 👍 , Changing to `autoBundleLambda` to make the distinction clear.

src/node-lambda-construct-source-code.ts
eladb: lol you are right. So just remove the / from the string.

src/awscdk-construct.ts
eladb: I think there's definitely benefits to this approach for CDK apps as well and we can give people choice

src/node-lambda-construct-source-code.ts
haimlit: Adding this as a configuration will force us to add a peer dependency on CDK (unless you see another way). Is this desired?

eladb: Why? I would just add explicit support for a bundle of nodejs versions that are supported both by esbuild and lambda

src/awscdk-construct.ts
haimlit: `aws-sdk` can easily add to `devDep`, but can't be easily added to `esbuild ... --external:aws-sdk`, as I didn't make it configurable. I tried to walk the fine line of providing some basic defaults and the fully blown bundler on CDK's Lambda NodeJS.

Do you think I should make externals configurable, adding each to both `devDep` and as `esbuild` arg? Or just keep the hardcoded `aws-sdk` one?

src/node-lambda-construct-source-code.ts
haimlit: Isn't it kind of duplicating `lambda.Runtime` class? Perhaps duplicating and taking supported esbuild node versions, then it's not that bad.

src/__tests__/awscdk-construct.test.ts
haimlit: I thought this is intended behavior for testing projects to contain the files in `__tests__` when running TestProject, but given your response I understand this is unexpected behavior. After digging a bit, I found out the root cause is that `process.cwd()` determines where things like `glob` are going to act on. So basically all the project source files can potentially be discovered, and the files I've defined on the integration folder aren't special at all.

So now let's talk about alternatives - using plain `new SampleCode(project, 'mylambda.lambda.ts')` doesn't work - Lambda bundling doesn't pick that up, as lambda handler scan happens on project constructor, while file is created only during synthesis.

I considered moving everything to integ tests, but I think integ tests should be used much more sparingly, especially as the integration is reading a file and doing specific assertions. I think we have a new type of beast here.

After a lot of experimenting, the following works:

```typescript
    // GIVEN
    const prepProject = new TestProject({
      cdkVersion,
    });
    new SampleFile(prepProject, 'src/mylambda.lambda.ts', {
      contents: '// empty lambda handler',
    });
    synthSnapshot(prepProject, false); // added flag to skip output folder removal in `utill.ts`
    process.chdir(prepProject.outdir); // reuse folder for next project

    const project = new TestProject({
      cdkVersion,
    });

    // THEN
    expect(snapshot['src/mylambda.ts']).toBeDefined();
```

I like this far better than writing an integ test, despite this being a bit verbose and ridden with implementation details. I wanted to extract to a util, but the magic happens between constructor calls. The only way I could extract it to a util is by wrapping all project calls in lazy init functions, which makes the code much less readable.

So I'm going to change to the above, let me know if you have a better idea.

src/__tests__/integration/lambda/mylambda-js.lambda.js
haimlit: I think it'll take much more work beyond the scope of this PR. We can test that a bundle task is created, but I'm not aware of a way to test its execution, as it's pretty much running `npx projen build`. I'll do a manual test, please let me know if you can think of a reasonable way to test this.

src/__tests__/awscdk-construct.test.ts
eladb: This just uncovered a bug in the lambda discovery code. 

We need to pass the project `outdir` to your `glob()` call.

And in the test, just use ```
writeFileSync(join(project.outdir, 'src', 'handler.lambda.ts'))
```

src/__tests__/integration/lambda/mylambda-js.lambda.js
eladb: Okay fair enough

src/awscdk-construct.ts
eladb: We are slowly realizing that the pre-bundling approach works much better then the "synth-time" bundling. Eventually I can envision that most customers will use this over `NodejsFunction` and friends (as much as I like them). To that end, I don't see a reason why certain features should be supported there and not here.

haimlit: 👍 , I guess I was trying to start simple a bit too much, I'll update with more bundling options then.

PR #944, opened 2021-07-20
fix(release): use posix path for changelog and bump file

MarkMcCulloh:
When using projen on a windows machine on a project with the version component, the workflows and tasks contained `\` in their path, thus resulting in a difference between windows/not-windows synths. It also breaks the release script when running in github actions.

This PR forces the use of posix slashes for those files.

---

### DIFF
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,4 +1,4 @@
-import { join } from 'path';
+import { posix } from 'path';
 import { Component } from './component';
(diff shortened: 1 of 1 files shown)


eladb (approved):

MarkMcCulloh:
@eladb Thank you for the quick approval! Could you approve the workflow run as well?

PR #950, opened 2021-07-25
fix(jsii): unable to publish to Maven GitHub Packages

### LINKED ISSUE #947: Publishing Maven/Java to GitHub Packages
Hey,

thanks for creating this cool project! I've started building an AWS CDK construct library for internal use, which I would like to publish to GitHub Packages. In #355 some work was already done to make this possible, but either a change broke it in the mean time or I'm missing something.

I've configured my .projenrc.js like shown below, but the workflow fails with an authorization error in the Release to Maven job. I've also configured the MAVEN_USERNAME secret to match my username.

```javascript
const project = new AwsCdkConstructLibrary({
  publishToMaven: {
    javaPackage: 'com.github.mkerix.something',
    mavenGroupId: 'com.github.mkerix',
    mavenArtifactId: 'something',
    mavenServerId: 'github',
    mavenRepositoryUrl: 'https://maven.pkg.github.com/mkerix/something',
    mavenPassword: 'GITHUB_TOKEN',
  },
} 
```

I believe the issue is that [publishToMaven](https://github.com/projen/projen/blob/88379ed5588e40338fcbc1a3e5b0f37033c64ac8/src/release/publisher.ts#L111) is missing the additional permissions to write to packages. For comparison, [publishToNpm](https://github.com/projen/projen/blob/88379ed5588e40338fcbc1a3e5b0f37033c64ac8/src/release/publisher.ts#L71) sets this based on a conditional check.

Furthermore, the release process to GitHub could be a made a bit nicer by auto-configuring the username to match the repository owner. This would bring it closer to the Maven release process that people are used to from GitHub Actions.

mKeRix:
This pull request adapts the GitHub Actions job generation for Maven in a way that publishing to GitHub Packages (a) is possible again and (b) roughly matches the steps required for publishing npm to Packages. To do this, I ended up changing two things:

- the permissions of the job are set correctly to `packages: write` when needed (same logic as for npm)
- the `MAVEN_USERNAME` is set to be the GitHub Actor automatically if no override was given, like shown in the GitHub docs

To do the latter I introduced a concept of `expressions` to the `PublishJobOptions`, which allow more freedom than the secrets option, but use the same logic. I also converted the secrets implementation to use these expressions under the hood, so they are essentially a shortcut for an expression now.

Jobs that publish to Maven Central or other repositories will be generated in the same way still, however the `MAVEN_USERNAME` env variable moved spots (hence the change in release.yml).

I tested it in a private project and can confirm that it works now, see below for an excerpt from my test config and the result as a screenshot.

```typescript
const project = new AwsCdkConstructLibrary({
  publishToMaven: {
    javaPackage: 'com.github.mkerix.projenmaventest',
    mavenArtifactId: 'projen-maven-test',
    mavenGroupId: 'com.github.mkerix',
    mavenServerId: 'github',
    mavenRepositoryUrl: 'https://maven.pkg.github.com/mKeRix/projen-maven-test',
    mavenPassword: 'GITHUB_TOKEN',
  },
});
```

Bildschirmfoto 2021-07-25 um 15 23 12

Closes #947.

---


src/release/publisher.ts
```diff
@@ -242,7 +259,8 @@ interface PublishTaskOptions {
 interface PublishJobOptions extends PublishTaskOptions {
   readonly image?: string;
   readonly name: string;
-  readonly secrets?: { [name: string]: string };
```
eladb, line 245: I would just replace `secrets` with `workflowEnv` and just pass `${ secrets.Xxx }` for values (or implement a helper function `secret(x)` which returns that.

src/release/publisher.ts
```diff
@@ -109,6 +110,9 @@ export class Publisher extends Component {
    * @param options Options
    */
   public publishToMaven(options: JsiiReleaseMaven = {}) {
```
eladb, line 112: Please add information about this behavior to the docstrings in `JsiiReleaseMaven`

Bonus points for adding this also to https://github.com/projen/projen/blob/main/docs/publisher.md

src/release/publisher.ts
```diff
@@ -118,12 +122,19 @@ export class Publisher extends Component {
         MAVEN_SERVER_ID: options.mavenServerId,
```
eladb, line 122: Can we default to `github` here if `isGitHubPackages` is enabled and if it is actually set, should we verify that it is always set to `github` to avoid errors?

eladb (changes requested):

src/release/publisher.ts
mKeRix: Implemented! Wasn't sure where the best place for the helper function is, so I ended up putting it close to its use. Happy to move it if you think it fits better somewhere else.

eladb (approved):
looks great!

mKeRix:
@eladb I regenerated the project and pushed it again - that should hopefully take care of the broken build.

Issue #596, opened 2021-03-05
feature request: eject out of projen


Chriscbr:
Discussed briefly in #337 comments

As a user, it would be nice to be able to "eject" myself out of projen if I wanted to. This should change permissions all projen-controlled files so that they are not readonly (and removing the "Generated by projen" markers, perhaps), as well as exporting the tasks in some kind of uniform way. That is, for NPM projects, all tasks would be rendered in the `scripts` section of `package.json` (currently achievable with `npmTaskExecution: NpmTaskExecution.Shell`), while for other project types (Python, Java, etc.), the tasks could be automatically exported in some universal format like a Makefile or shell script.

Inspired by the similar feature in other CLI tools, such as [create-react-app eject](https://create-react-app.dev/docs/available-scripts/#npm-run-eject).

eladb:
Keep

Chriscbr:
Keep

PR #951, opened 2021-07-26
feat(tasks): render tasks into Makefile

Chriscbr:
Adds the capability to synthesize tasks into a Makefile format. I believe this is a prerequisite for implementing a solution of #596.

Just experimenting at the moment, this PR isn't close to being ready yet.

- [ ] environment variables not working yet
- [ ] ensure output is deterministic (no direct paths to node etc.)
- [ ] research better ways to handle recursive tasks in make?
- [ ] refactor task code to minimize duplicate code
- [ ] have the default `make` task (when no target is specified) be projen's "default" task if possible
- [ ] unit tests

---


src/tasks/tasks.ts
```diff
@@ -21,13 +32,19 @@ export class Tasks extends Component {
...
+  private readonly makefile?: Makefile;
+
+  constructor(project: Project, options: TasksOptions = {}) {
     super(project);
 
     const manifestFile = Tasks.MANIFEST_FILE;
     this._tasks = {};
     this._env = {};
 
+    if (options.makefile) {
```
eladb, line 44:
```suggestion
    if (options.makefile ?? false) {
```

src/tasks/tasks.ts
```diff
@@ -115,4 +132,85 @@ export class Tasks extends Component {
...
+   */
+  public getFullEnvironment(task: Task) {
+    let env = this._env ?? {};
+
+    env = {
+      ...env,
+      ...task._renderSpec().env ?? {},
+    };
+
+    return env;
```
eladb, line 148:
```suggestion
    return {
      ...this._env,
      ...task._renderSpec().env,
    };
```

Makefile
```diff
@@ -0,0 +1,145 @@
+.PHONY: all
```
eladb, line 1: Projen marker

eladb (commented):
I like this direction but I don't like that we need to maintain two runtimes. I am wondering if we should only use make files to run tasks. Could be a better long term approach...

Chriscbr:
Time permitting I'll try to return to the PR this weekend

Chriscbr:
I'm planning to revisit this soon, so keep your eyes peeled

PR #963, opened 2021-08-03
fix(jsii): error if docgen is explicitly enabled

eladb:
Disable `docgen` when calling `super()` since `JsiiProject` uses `jsii-docgen` instead of
the normal TypeScript docgen tooling.

Fixes #960

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #965, opened 2021-08-03
fix: integration test errors on macOS

Chriscbr:
Resolves a bug running integration tests on macOS introduced in #963.

---

### DIFF
--- a/src/__tests__/util.ts
+++ b/src/__tests__/util.ts
@@ -81,7 +81,7 @@ export function mkdtemp() {
  */
 export function synthSnapshot(project: Project): any {
   // defensive: verify that "outdir" is actually in a temporary directory
-  if (!path.resolve(project.outdir).startsWith(os.tmpdir())) {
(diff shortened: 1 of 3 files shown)


src/__tests__/util.ts
```diff
@@ -81,7 +81,7 @@ export function mkdtemp() {
  */
 export function synthSnapshot(project: Project): any {
   // defensive: verify that "outdir" is actually in a temporary directory
-  if (!path.resolve(project.outdir).startsWith(os.tmpdir())) {
```
Chriscbr, line 84: When I tried doing some print debugging, I found `project.outdir` had a value like:
```
/private/var/folders/XXXXX/T/projen-test-XXXXX/project-dir
```
while `os.tmpdir()` had a value like:
```
/var/folders/XXXXX/T
```
Not really sure if there is some kind of quirky behavior with how `os.tmpdir()` is working, or if I'm just missing something obvious. 🤷‍♂️ This is the best duct-tape fix I could come up with...

eladb (approved):

PR #977, opened 2021-08-09
feat(upgrade): rename "upgrade-dependencies" to "upgrade" 

eladb:
"upgrade" is sufficient and aligns better with the current one-word naming of standard workflows.

---


Chriscbr:
@eladb It's worth mentioning that this briefly surprised me since I use `yarn <task>` pretty often as an alternative to `npx projen <task>`, and to date, it hasn't been an issue since all of our task names haven't intersected with built-in yarn operation names. But in this situation, `yarn upgrade` won't run the projen task because of the name reuse.

That said, I don't think reverting the change is all that worthwhile - there are a lot of `yarn` operations and avoiding using all of the names is probably a bit restrictive.

eladb:
@Chriscbr 👍

PR #980, opened 2021-08-11
fix: rename license files to correct SPDX ids

hyandell:
Hi Projen folk, I'm a part of the AWS OSPO and I noticed that the license filenames were not using SPDX codes. My understanding is that these are licenses that someone generating project configuration might use as the license for their output. 

Some of them also seemed like odd choices, either due to being unusual or being old dated versions of licenses that folk probably do not use anymore.

I don't know what the effect of doing this is - I'm not a clueful projen user. 

---


Chriscbr (approved):
Thanks for the PR! This looks great to me - I didn't realize there were incorrect identifiers, and I think it likely makes sense to focus on just including common licenses.

re: implications, there is some code in https://github.com/projen/projen/blob/main/src/license.ts that will take a user-provided identifier and automatically add the right license file to a user's project - but I don't think your changes should really break anyone since none of the licenses changed are particularly common.

hyandell:
Awesome - thank you Christopher :)

PR #992, opened 2021-08-18
fix: tests under lib/ are included in npm tarballs

### LINKED ISSUE #955: When tests are under `lib/` they are included in npm tarballs
When projects are configured with `testdir: "src/__tests__"` then test code will end up under `lib/__tests__`.

The `.npmignore` file has a `!/lib` at the end, which overwrites the `/lib/__tests__` exclusion that comes before it.

Chriscbr:
Fixes #955 

I don't think it's possible to fix #955 without disabling the automatic sorting behavior in , so I tried changing the implementation to see if it makes sense. The main difference is that the underlying data structure is an array, and not a pair of sets (current implementation) or an object (implementation from 8+ months ago).

One feature I added is that `IgnoreFile` will eliminate existing patterns that are "overridden" by the patterns you add. So if you add `/foo/`, then not only will `!/foo/` be removed, but also any other patterns starting with `!/foo/` such as `!/foo/bar/baz` or `!/foo/**/*.ts`. Likewise, adding `!/foo/` will remove any patterns starting with `/foo/`. This kind of logic only applies to directories (we don't want `/foo` to overwrite `!/foo.txt` because `foo` might be the name of a file with no extension).

I also deprecated the `npmIgnore` and `gitIgnore` fields in `NodeProject`. Since any patterns added via these fields would be applied before patterns added in subclasses such as `TypeScriptProject`, this will lead to a [confusing developer experience](https://github.com/projen/projen/issues/453) under the new order-dependent implementation. Instead, users should use `project.addPackageIgnore()` and `project.addGitIgnore()`.

One downside of disabling sorting is that the generated output is now dependent on the order in which patterns are added, so refactorings could cause npmignore files to get randomly reshuffled.

---


joostvdwsd:
This PR also removed the lib/__tests__/util.ts in the test folder. Probably not correct since its from a test folder but we were using this to grab a snapshot from our own generated projects.
Maybe an idea to include this file in the main package?

PR #999, opened 2021-08-20
feat: `.projenrc.json`

### LINKED ISSUE #990: Representation of projenrc as json
We would like to be able to declaratively talk about projen rcs.

### LINKED ISSUE #994: All arguments passed as strings when initializing external projen module
`npx projen@latest new --from @taimos/projen taimos-ts-lib --stale=false` generates the following in `.projenrc.js`:

```ts
const project = new TaimosTypescriptLibrary({
  defaultReleaseBranch: 'main',
  devDeps: ['@taimos/projen'],
  name: 'projen-test',
  stale: 'false', // error! should not be in quotes
});
```

The reason for this is that if you try initializing a built-in projen project type, the project type is loaded and processed by `yargs` so that it knows any "stale" flag should have a boolean value. But for initializing external project types, the project type is not loaded, so all types are (by default) provided as strings.

A temporary solution for this might be to add logic that coerces truthy/falsey strings into true and false. But ideally we should install the module, retrieve the jsii information, and then pass the information through a [yargs parser](https://yargs.js.org/docs/#api-reference-parseargs-context-parsecallback) configured based on the jsii information (like we do for built-in project types) so that the arguments are parsed into the correct types, and to improve the error messaging if arguments are provided incorrectly.

Edit: actually, a simpler fix might just be to add the appropriate parsing logic here: https://github.com/projen/projen/blob/be7c9fc7dab79c284107c217a12ae865fb7a46f1/src/cli/cmds/new.ts#L273-L289

Chriscbr:
Adds a `--projenrc-json` option for all project types that will generate a `.projenrc.json` file for managing the project instead of `.projenrc.js`. The `.projenrc.json` can contain any fields that can be passed to the constructor of the appropriate project type. The JSON file must also contain a field named "projectTypeFqn" that contains the fully-qualified name of the project type (e.g. `projen.web.React`, `my-jsii-module.MyProjectType`, ...).

Since projenrc.json only allows passing properties that could be passed to the project's constructor it is less flexible than other projenrc types and does not provide access to the other API.

Closes #990 
Fixes #994

Example CLI usage: `npx projen new typescript --projenrc-json`.
Example `.projenrc.json`:

```json
{
  "projectTypeFqn": "projen.TypeScriptProject",
  "defaultReleaseBranch": "main",
  "name": "projen-test",
  "projenrcJson": true
}
```

Comments:

- It would be nice to eventually define schemas for projenrc.json so that users editing JSON manually can get autocomplete etc.
- It could be beneficial to unify projenrc options/flags later (instead of having `--projenrc-json`, `--projenrc-js`, `--projenrc-java` etc., we could use `--projenrc=js` etc.)

---


src/json/run-projenrc.task.ts
```diff
@@ -0,0 +1,61 @@
...
+const newProjectCode = `const project = new ${projectType.typename}(${renderedOptions});`;
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const mod = require(modPath);
+const ctx = vm.createContext(mod);
+
+vm.runInContext([
+  newProjectCode,
+  'project.synth();',
+].join('\n'), ctx);
```
Chriscbr, line 61: Some of this logic is copied from `new.ts`, we can probably extract it out...

src/__tests__/util.ts
```diff
@@ -50,6 +50,11 @@ export interface SynthOutput {
   [filePath: string]: any;
 }
 
+beforeAll((done) => {
+  process.env.PROJEN_TEST = 'true';
```
eladb, line 54: Add a note why this is needed

src/cli/cmds/new.ts
```diff
@@ -228,7 +229,11 @@ function commandLineToProps(cwd: string, type: inventory.ProjectType, argv: Reco
  */
 async function newProjectFromModule(baseDir: string, spec: string, args: any) {
   const installCommand = `yarn add --modules-folder=${baseDir}/node_modules --silent --no-lockfile --dev`;
-  if (args.projenVersion) {
+  if (process.env.PROJEN_TEST) { // for integration tests
+    exec(`mkdir -p node_modules && ln -s ${process.execPath} node_modules/projen`, { cwd: baseDir });
+  } else if (process.env.PROJEN_LINK_LOCAL) {
```
eladb, line 234: Is this documented somewhere?

src/json/projenrc.ts
```diff
@@ -0,0 +1,55 @@
...
+      return; // already exists
+    }
+
+    const bootstrap = this.project.newProject;
+    if (!bootstrap) {
+      return;
+    }
+
+    const json = {
+      projectTypeFqn: bootstrap.fqn,
```
eladb, line 48: Let's rename this to something a bit less intimidating. Maybe just `type`? or `base`?

src/json/run-projenrc.task.ts
eladb: Yes, that would be nice

eladb (changes requested):
Looks great @Chriscbr - I love the creative reuse of the javascript rendering.

Can we add an issue to also support YAML? It should be straightforward given YAML is JSON.

Added a few minor comments.

alexforsyth:
This looks awesome! Thanks so much. @Chriscbr let's say i'm using an external project. Does that change anything? How would I know what version I need to use?

eladb (approved):

eladb (approved):

PR #1009, opened 2021-08-28
feat(release): Create an issue on failed releases

iliapolo:
Add ability to create GitHub issues for each failing publishing task. 
This allows for better visibility into those scenarios, that usually require manual intervention. 

Examples:

- https://github.com/iliapolo/cdk8s-core/issues/35
- https://github.com/iliapolo/cdk8s-core/issues/34

---


src/release/publisher.ts
```diff
@@ -43,6 +43,21 @@ export interface PublisherOptions {
    * @default "latest"
    */
   readonly jsiiReleaseVersion?: string;
+
+  /**
+   * Create an issue when a publish task fails.
+   *
+   * @default false
+   */
+  readonly issueOnFail?: boolean;
```
eladb, line 52: Names are inconsistent between Publisher and Release

src/release/publisher.ts
```diff
@@ -252,31 +273,60 @@ export class Publisher extends Component {
...
+          id: 'extract-version',
+          run: 'echo "::set-output name=VERSION::$(cat dist/version.txt)"',
+        },
+        {
+          name: 'Create Issue',
+          if: '${{ failure() }}',
+          uses: 'imjohnbo/issue-bot@v3',
+          with: {
+            labels: this.issueOnFailLabel,
+            title: `Publishing v\${{ steps.extract-version.outputs.VERSION }} to ${opts.registryName} failed`,
```
eladb, line 309:
```suggestion
            title: `Publishing v$(cat dist/version.txt) to ${opts.registryName} failed`,
```

src/release/publisher.ts
```diff
@@ -252,31 +273,60 @@ export class Publisher extends Component {
...
+    const perms = opts.permissions ?? { contents: JobPermission.READ };
+
+    if (this.issueOnFail) {
+      steps.push(...[
+        {
+          name: 'Extract Version',
+          if: '${{ failure() }}',
+          id: 'extract-version',
+          run: 'echo "::set-output name=VERSION::$(cat dist/version.txt)"',
+        },
```
eladb, line 302: seems redundant:

```suggestion
        {
          name: 'Extract Version',
          if: '${{ failure() }}',
          id: 'extract-version',
          run: 'echo "::set-output name=VERSION::$(cat dist/version.txt)"',
        },
```

eladb (commented):
Interesting approach.. not sure this is going to be super effective but I am curious.

Technically you could have opened tickets by querying the workflow status (similar to how the GitHub badge works).

.projenrc.js
```diff
@@ -64,6 +64,7 @@ const project = new JsiiProject({
     distName: 'projen',
     module: 'projen',
   },
+  failureIssue: true,
```
Chriscbr, line 67: Not in love with this API name. Maybe `notifyOnFailure` or `issueOnFailure`?

Chriscbr (changes requested):
Could you generalize this just a little bit to add an API (maybe a method call on `project.release`) so users can add their own custom steps to run on release failures? Or if this is already doable, just add some documentation for it in perhaps.

use case: If I want to run https://github.com/marketplace/actions/send-email when release fails, there should be a natural way to add this, based on the failure() mechanism we are using here.

Let's add a section in docs/publisher.md about this as well.

src/release/publisher.ts
iliapolo: This doesn't work - the title is being passed as is to the action, it doesn't undergo bash interpolation befoehand.

src/release/publisher.ts
iliapolo: See https://github.com/projen/projen/pull/1009#discussion_r698753772

.projenrc.js
iliapolo: Yeah there's no good name really, i'll go with `issueOnFailure` since it was my initial name as well.

src/release/publisher.ts
iliapolo: Fixed

.projenrc.js
iliapolo: Changed

iliapolo:
@eladb 

> Interesting approach.. not sure this is going to be super effective but I am curious.
> Technically you could have opened tickets by querying the workflow status (similar to how the GitHub badge works).

Yeah this was my initial plan. I ended going down this route because:

1. I think its a nice projen feature, it gives better visibility as to which step exactly failed as well as which version.
2. It provides a nice way for maintainers to put such failures on project boards and prioritize this work.
3. It avoids the problem of handling workflows that are still in-progress. That is, what if you query the workflow status and it returns in-progress? do you query the last 2? last 5? We could have come up with a good enough heuristic, but this seemed simpler.

iliapolo:
@Chriscbr 

> Could you generalize this just a little bit to add an API (maybe a method call on project.release) so users can add their own custom steps to run on release failures?

That would be cool, but I don't currently have the capacity for it - as it requires some non trivial API design. 

I think we can start off with this and eventually evolve it to a nice `onFailure` notification system.

Chriscbr (approved):

API.md
```diff
@@ -416,6 +416,8 @@ new AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
   * **stability** (<code>string</code>)  Package's Stability. __*Optional*__
   * **antitamper** (<code>boolean</code>)  Checks that after build there are no modified files on git. __*Default*__: true
   * **artifactsDirectory** (<code>string</code>)  A directory which will contain artifacts to be published to npm. __*Default*__: "dist"
+  * **failureIssueLabel** (<code>string</code>)  The label to apply to issues indicating publish failures. __*Default*__: "failed-release"
+  * **issueOnFailure** (<code>boolean</code>)  Create a github issue on every failed publishing task. __*Default*__: false
```
eladb, line 420: These two options should have the same prefix so it's clear they are related.

I think `failureIssue` and `failureIssueLabel` is my preference.

iliapolo: Switched it back 👍

iliapolo:
@eladb You good with merging this?

eladb (approved):

PR #1010, opened 2021-08-28
feat(javascript): JavascriptFile component

MarkMcCulloh:
Adds a component that allows you to generate a javascript file based on actual runtime code.

Use cases I can immediately think of for this:
1. Node tooling config files that must be a js files rather than json for certain scenarios (bundlers like webpack)
2. JS files meant to be used as project scripts
3. General JS codegen

While this could be accomplished with a plain TextFile where the js is simply written to it or loaded from the filesystem, you will likely lose real flexibility with that approach. The goal of this component is to make this code more easily composable. So for example you would be able to make a base webpack config component with this but allow for consumers to override/augment the content with their own code.

This PR generally uses a pretty naive approach, by just using .toString() on an existing function. Also, the API is possibly a little overcomplicated. This is something I started implementing for my own use (for composable [Vite](https://vitejs.dev/) configs, and I would love some advice to see if this is worth having in projen itself.

---


pgollucci:
comemnt

MarkMcCulloh:
Since this was reopened I take it there's interest in this feature. I'm still looking for feedback on the API, I'm not quite happy with it and I'm still looking for advice on that if possible.
After rebasing I'm getting weird test errors so I'll have to look into that as well. Doesn't really seem related to this feature.

test/javascript/javascript-file.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    setToValue: 'greeting',
+  });
+
+  file.appendFunction({
+    fn: (greeting) => {
+      console.log(greeting);
+    },
+    prefix: '// do a log\n',
+    postfix: '\n// did a log',
+  });
```
Chriscbr, line 39: When I read the API, I first thought the only output would be functions, but it looks like the API is actually a lot more expressive than that. How can we model the API closer to what the intent of the author is? example:

```js
file.addCode({
  code: () => {
    console.log("hello world!");
    let cats = parseInt(process.env.CAT_COUNT);
  },
  type: SnippetType.INLINE,
});

file.addSnippet({
  code: () => {
    console.log("hello world!");
    let cats = parseInt(process.env.CAT_COUNT);
  },
});

// named after https://developer.mozilla.org/en-US/docs/Glossary/IIFE
file.addIIFE({
  code: (cats) => {
    console.log(`you have ${cats} cats!`);
  },
  callWith: ["cats"],
});

file.addGlobalFunction({
  code: (arg1, arg2) => {
    return arg1 + arg2;
  },
  name: "globalFunction",
  exported: true,
});
```

which renders:

```js
console.log("hello world!");
let cats = parseInt(process.env.CAT_COUNT);

(function (cats) {
  console.log(`you have ${cats} cats!`);
})(cats);

function globalFunction(arg1, arg2) {
  return arg1 + arg2;
}
```

Chriscbr (commented):
I opened this because I think @pgollucci might have had interest in using this for postcss or tailwindcss config files - but I've gone ahead and added some comments as well. :)

One suggestion might be to add a test case that creates a functional config file (like the ones previously mentioned, or e.g. `vite.config.js`) to show how it's used for an example use case.

src/javascript/javascript-file.ts
```diff
@@ -0,0 +1,122 @@
+import { js } from 'js-beautify';
+import { TextFileOptions, TextFile, Project } from '..';
```
Chriscbr, line 2: I think this should fix the build errors:
```suggestion
import { Project } from '../project';
import { TextFile, TextFileOptions } from '../textfile';
```

test/javascript/javascript-file.test.ts
MarkMcCulloh: Oh I love these examples! I'll add these as well as some other more intent-focused methods when I have some time for this PR

MarkMcCulloh:
I recently stumbled on [pulumi's function serialization](https://github.com/pulumi/pulumi/tree/master/sdk/nodejs/runtime/closure) implementation and I think it would be great to use here, replacing the naive approach. It would automatically serialize the closure as needed (variables, imports, etc.), which would simplify the API quite a bit while making it much more powerful. Probably also a bit slower, but that seems like a worthy price to pay. 

I was probably going to extract their code into a new package to make it more generic since I've had a need for this sort of thing in other places too.

Major problem though: I don't think it would be possible to do it synchronously. @Chriscbr do you think it would be a bad idea to do this anyways and just let the promise go un-awaited in synthesis? Since node shouldn't exit if it isn't done, I'm not sure it would be noticed (as long as it goes well). I'm not a huge fan, but it may be worth it.

Chriscbr:
@MarkMcCulloh I'm happy to accept any contributions that are appropriately licensed. 🙂 Bundling any complex or projen-independent logic into a separate library sounds like a good idea.

re: async -- Maybe we can dispatch it through a separate node process / keep it synchronous using `execSync`? It's not elegant but I think it'd be an OK hack for now. (Changing the general synth process to be async is sort of a "one-way door" in terms of the API design.)

PR #1013, opened 2021-08-30
RFC 1013: External GitHub Repository Settings

eladb:
Requesting comments and feedback on managing external GitHub repository settings such as branch protection rules and secrets via projen. See [rendered version](https://github.com/projen/projen/blob/benisrae/rfc-project-settings/rfcs/github-project-settings.md).

RFC for https://github.com/projen/projen/pull/1013

---


rfcs/github-project-settings.md
```diff
@@ -0,0 +1,76 @@
...
+Done
+```
+
+_not sure that’s feasible:_ You will also notice that your `.github/workflows/release.yml` workflow was updated and now includes
+a check that verifeis that your repository has the desired secrets. This is sort of an "anti-tamper check" for external settings.
+If your repository does not have the needed secerts, an error will be displayed with instructions on how to store them.
+
+A similar approach can be taken to configure other non-source-code repository settings such as branch protection rules:
+
+```
```
eladb, line 56:
```suggestion
```js
```

mmcculloh-dms:
I love the idea of projen managing project related things outside of just files. I've wanted to do this recently to set up SonarCloud projects.

I would be concerned about projen being able to make changes that are basically "invisible" though. I've always liked that projen makes changes that can be seen and committed explicitly. I feel like there should be some sort of state file or declarative file that describes the intent.

mmcculloh-dms:
I suppose it's true that the projenrc describes intent, so that's covered. Perhaps the state is more important, which I don't think the projenrc has. That would be like saying the the typescript files for a CDK project contain state. 

For everything in projen, state is covered by the content of the files themselves. But for this, it wouldn't be.

intent => state
add gitignore pattern => change in .gitignore
add branch protection => ???

eladb:
@mmcculloh-dms I understand your point, but I'd argue that the files projen generates can be considered an "implementation detail" of the abstractions presented by projen. In most cases, if users trust the abstraction, they don't really need to care about these state files which are actually "leaks" in the abstraction provided by projen. We need them for practical reasons but not from a software engineering perspective. In fact, in the AWS CDK, we recommend that users don't commit the synthesized CloudFormation templates to their repo. As far as AWS CDK users are concerned, these templates are just intermediate build artifacts. You can take an analogy from the compilation world - developers do not normally commit the output executable into their repository. They trust that the "abstraction" provided by the compiler (i.e. their programming language) will produce the correct output every time the compiler is executed.

kcwinner:
For branch protection, etc would it be more clear to use the [settings.yml](https://github.com/apps/settings) method? This way there are no side-effects prior to a PR being merged in. If that was the plan it isn't necessarily clear in the RFC.

eladb:
> For branch protection, etc would it be more clear to use the [settings.yml](https://github.com/apps/settings) method? This way there are no side-effects prior to a PR being merged in. If that was the plan it isn't necessarily clear in the RFC.

TIL about [settings.yml](https://github.com/apps/settings). Definitely seems like a nice way to enable many of these capabilities. I'll add to the RFC.

ahammond:
Regarding secrets, I would like to see a way to handle secrets that are shared between multiple repos. I think a reasonable implementation would be an ordered list of SecretManager secrets. Each SM secret dict gets pulled and applied, updating a dict which is then persisted to GitHub.

eladb:
> Regarding secrets, I would like to see a way to handle secrets that are shared between multiple repos. I think a reasonable implementation would be an ordered list of SecretManager secrets. Each SM secret dict gets pulled and applied, updating a dict which is then persisted to GitHub.

This is sort of what I am proposing in this RFC. Lmk if it makes sense.

eladb:
> TIL about settings.yml. Definitely seems like a nice way to enable many of these capabilities. I'll add to the RFC.

On second thought, there's a bit of a security challenge relying on external apps for core projen features. Using external apps requires additional security vetting in some companies (including Amazon), and this type of app is less likely to be approved given the administrative nature of this app.

rfcs/github-project-settings.md
```diff
@@ -0,0 +1,76 @@
...
+  },
+});
+```
+
+> NOTE: not sure I like the union-like API, but we can debate that later.
+
+Now, as usual, execute `npx projen`.
+
+Under the hood, projen uses `gh secret list` to check if your repo has the desired secrets. 
+If not, it will retrieve them from AWS Secrets Manager and store them in your repository via the GitHub API.
```
Chriscbr, line 40: If we are cool with making API calls to AWS to retrieve secrets, why even go through the hassle of managing multiple secrets in GitHub? If projen-managed GitHub workflows are the only processes using a given secret, is it possible to shortcut this design and have the workflow script to retrieve the secret at workflow-runtime? (this way, only a single secret is stored per GitHub repo).

Chriscbr (commented):
Love the idea! I can see a lot of value to not having to configure secrets by hand.

I think the risks I see are less around the feature's particular benefits, and more around 1) how we are choosing to let users to opt in (or opt out) of these features, and 2) all of the possible ways this abstraction could unintentionally feel leaky, as with other projen abstractions.  

For 1), I am curious if would be opt-in or opt-out, or if it would automatically be bundled with existing `GitHub` component/settings, etc. For managing source-controlled files, we have at least a couple of escape hatches for users, but I imagine for something API call based, new escape hatches might be needed. (Unless it's leveraging the existing task system). Not sure.  

For 2), this feature is unique in that the state we're managing isn't in files but remotely on GitHub. And so it raises the question, what happens when users tamper with these settings? If these settings are always mutable through the web UI, it still leads to a clunky experience (even if there is an anti-tamper check added to GitHub workflows). Moreover, it's extra knowledge that projen users have to absorb ("if I need to install a package, I can't use the language-idiomatic install command, I need to edit projenrc... if I need to change the default branch, I can't use GitHub's web UI, I need to edit projenrc..."). The discussion probably is beyond the scope of this RFC, but it has to be at least mentioned.

rfcs/github-project-settings.md
eladb: If we want to retrieve secrets from workflows, it means that we will need to store the AWS credentials in the GitHub settings (going back to square one). It means the role for these **must** be scoped down dramatically to retrieve only the specific secret that contains the tokens (require a bit of an elaborate AWS setup) and _also_ means that the workflow jobs will be able to access tokens that they don't really need (unless we create a different set of AWS credentials for each job).

All in all, doesn't seem like this really helps :-). I'll add to "alternative considered"

Chriscbr: Ah, so the solution proposed by the RFC actually doesn't involve storing AWS credentials on GitHub at all - that point wasn't clear to me. (But it makes sense from a security standpoint). So any secret-updating has to be done manually. Got it. 👍

PR #1019, opened 2021-08-31
feat: testing utils

joostvdwsd:
This will make test utils part of the public api (or at least resolvable for node). (src/__ tests __/utils.ts => src/test-utils.ts)
Didn't include the test-utils.ts in the index.ts on purpose. This is because it disables logging when included.

No inclusion in index.ts mean not available for default export. Therefor this is only available through:
```typescript
import { TestProject } from 'projen/lib/test-utils'
```

Another result is that it wont be available for other languages through jsii. Maybe in the future this could be all refactored when there is need in non-node projects.

---


eladb (changes requested):
1. Missing an export of this new file in the root `index.ts` - without it, these APIs won't be accessible externally.
2.  Some of the APIs exported are not JSII compatible -- JSII cannot translate free-standing functions.
3. The current `test-utils.ts` file is exporting too much. Let's start with the bare minimum and add later as needed. For example, I am not sure `TestProject` itself is too useful outside of the projen test suite. I think we need to do some refactoring so that only the relevant APIs will be exported.

My understanding is that you are primarily looking for a way to synthesize a project to capture a snapshot of all the files as a dictionary, correct? If that's the case, I think we should create a wrapper class with a static method:

```ts
import { synthSnapshot } from './_test-utils.ts';

export class Testing {
  public static synth(project: Project): Record<string, any> {
    return synthSnapshot(project);
  }

  private constructor() {} // utility
}
```

Then, usage will be:

```ts
import { Testing } from 'projen';
const dict = Testing.synth(myProject)
```

The rest of the functions should likely go under a non-exported private file (usually we prefix them with an underscore to indicate they are internal).

So:

1. Rename this file to `_test-utils.ts`
2. Create a new file called `testing.ts` which exports the `Testing` class above.
3. Add `export * from './testing'` to `index.ts`.

joostvdwsd:
Actually the reason I didn't include it in the public api (for now) was this line in the test-utils.ts:
```typescript

logging.disable(); // no logging during tests

```

This will be executed during import inside index.ts.

eladb:
Let's move this line to the constructor of TestProject for now

eladb:
@joostvdwsd lmk when this is ready for another round

mbonig:
Why was this closed? I was just about to submit a similar PR. I could really use this for simple snapshot testing of my projen constructs.

mbonig:
See my PR, if there is a way to merge my new one into this PR, otherwise do we continue over there?

PR #1021, opened 2021-08-31
feat(release): publishing npm packages to AWS CodeArtifact

### LINKED ISSUE #986: Publishing npm to AWS CodeArtifact
Hi,

I would like to publish npm packages to AWS CodeArtifact.
The special point at this npm registry is the authorization token which is only valid for 12 hours.
A new token can be requested using AWS CLI command `aws codeartifact login` (see [documentation](https://docs.aws.amazon.com/codeartifact/latest/ug/tokens-authentication.html)).

Would you accept a PR that implements this feature?

**Proposed solution:**
Get a new authorization token for AWS CodeArtifact in the release workflow before publishing the artifact to the npm registry.

I did a first test with this code snipped in release.yml:
```yaml
- name: Release
  run: |
    export NPM_TOKEN=`aws codeartifact get-authorization-token --domain my_domain --domain-owner 111122223333 --query authorizationToken --output text`
    npx -p jsii-release@latest jsii-release-npm
  env:
    NPM_DIST_TAG: latest
    NPM_REGISTRY: my_domain-111122223333.d.codeartifact.region.amazonaws.com/npm/my_repo/
    AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
    AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
    AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
```

This command has to be inserted as an additional command in section `run` if publishing to AWS CodeArtifact is enabled:
```bash
export NPM_TOKEN=`aws codeartifact get-authorization-token --domain my_domain --domain-owner 111122223333 --query authorizationToken --output text`.
```

AWS CLI uses the following environment variables:
- AWS_ACCESS_KEY_ID -> new configuration in `.projenrc.js`
- AWS_SECRET_ACCESS_KEY -> use GitHub secrets (instead of NPM_TOKEN)
- AWS_DEFAULT_REGION -> new configuration in `.projenrc.js`

In addition, command `aws codeartifact get-authorization-token` requires values for domain and domain owner. I would try to extract them from the URL in NPM_REGISTRY.

Trigger for publishing to AWS CodeArtifact: npm registry URL that contains regex `.codeartifact.*.amazonaws.com`.

I would be happy to have your feedback.

jumic:
This PR introduces publishing npm packages to AWS CodeArtifact.

The authentication will be done using AWS CLI. This depends on PR https://github.com/cdklabs/jsii-release/pull/150 which has to be merged first.

`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` have to be defined as GitHub Secrets to follow the [IAM best practices](https://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html#iam-user-access-keys
). The names of those variables can be overridden. The default values will be applied if an AWS CodeArtifact URL is detected.

Closes #986.

---


src/node-package.ts
```diff
@@ -290,6 +291,22 @@ export interface NodePackageOptions {
...
    */
   readonly npmTokenSecret?: string;
+
+  /**
+   * GitHub secret which contains the AWS access key ID to use when publishing packages to AWS CodeArtifact.
+   * This property must be specified only when publishing to AWS CodeArtifact (`npmRegistryUrl` contains AWS CodeArtifact URL).
+   *
+   * @default "AWS_ACCESS_KEY_ID"
+   */
+  readonly awsAccessKeyIdSecret?: string;
```
eladb, line 301: I am wondering if this should be collapsed under `codeArtifactOptions` (and a `CodeArtifactOptions` struct) which can later be reused for other publishing targets and will also make it easier to understand that this is only relevant for CodeArtifact:

```ts
export interface CodeArtifactOptions {
  readonly accessKeyIdSecret?: string;
  readonly secretAccessKeySecret?: string;
}

export interface NodePackageOptions {
  readonly codeArtifactOptions?: CodeArtifactOptions;
}
```

eladb (changes requested):

src/node-package.ts
jumic: Yes, I think it's a good idea to change this. I updated the code as proposed.

src/node-package.ts
```diff
@@ -362,7 +388,17 @@ export class NodePackage extends Component {
   /**
    * GitHub secret which contains the NPM token to use when publishing packages.
    */
-  public readonly npmTokenSecret: string;
+  public readonly npmTokenSecret?: string;
+
+  /**
+   * GitHub secret which contains the AWS access key ID to use when publishing packages to AWS CodeArtifact.
+   */
+  public readonly awsAccessKeyIdSecret?: string;
```
eladb, line 396: Why not use the `codeArtifactOptions` here as well?

eladb (changes requested):

src/node-package.ts
jumic: I changed it. Now, variables `awsAccessKeyIdSecret` and `awsSecretAccessKeySecret` don't exist in the code anymore.

src/node-package.ts
```diff
@@ -728,12 +764,33 @@ export class NodePackage extends Component {
       throw new Error(`"npmAccess" cannot be RESTRICTED for non-scoped npm package "${this.packageName}"`);
     }
 
+    const isAwsCodeBuildRegistry = npmRegistryUrl && AWS_CODEARTIFACT_REGISTRY_REGEX.test(npmRegistryUrl);
```
eladb, line 767: Can we move this logic to the `Publisher` class? Feels like it belongs in there.

eladb (changes requested):

src/node-package.ts
jumic: I moved the logic to `Publisher` class. Class `NodePackage` uses this new method, now.

hoegertn:
@eladb Can we keep this open or get this merged?

jumic:
@Chriscbr Thanks for reopening this PR. I just merged the current main branch to this branch.

Now, the PR has status "Changes requested" in the list of my pull requests. From my point of view, all reviews are marked as resolved. Am I missing something? The status of this PR should be "ready for review". Thanks.

Chriscbr (approved):

Chriscbr:
Everything looks good to me / I think you re-requested a review correctly, the PR must have just slipped through the cracks - thanks for the reminder.

hoegertn:
@jumic What do you think about adding support for Github OIDC instead of access key and secret key? Would love this.

jumic:
@hoegertn I just had a first look at Github OIDC. Definitely a good idea for this feature. I will submit an additional PR.

PR #1022, opened 2021-08-31
fix: collocated TypeScript compiler files not ignored

### LINKED ISSUE #1020: (bug): collocated Typescript compiler files not ignored after projen@0.27.35
It looks like something changed in the way .gitignore files are generated in #992.

The net impact is that for projects with collocated source + build files (e.g., srcdir === libdir), the `.gitignore` directive for ignoring the `.d.ts` and `.js` files was removed.

I can see that the check is still present in `typescript.ts`:

https://github.com/projen/projen/blob/e66a80f287f643865b0b4ea86081101e96b7f6d7/src/typescript.ts#L268-L275

However, something else must be causing these lines to be removed, as they aren't present in the project once upgraded to ^0.27.35. This even shows up in the Typescript snapshot tests from the above PR:

https://github.com/projen/projen/pull/992/files#diff-11f4dca84bdff503e734bc67efc73c7f54ca3f3ad387a399e61a64316acc5264L20-L21

Example project hit by this:
https://github.com/cdklabs/aws-delivlib/pull/1028/

Chriscbr:
Fixes #1020

---


src/__tests__/__snapshots__/typescript.test.ts.snap
```diff
@@ -45,10 +45,38 @@ junit.xml
 !/test/
 !/tsconfig.json
 !/lib/
+/lib/**/*.js
+/lib/**/*.d.ts
```
Chriscbr, line 49: This is the evidence the bug was fixed.

eladb (approved):

PR #1027, opened 2021-09-02
feat: split Project and GitHubProject

### LINKED ISSUE #1011: Add a more abstract Project base class
The existing Project class has quite a few things in it. If one is making a new project type, it would be nice to be able to limit what's inherited from the base to be able to pick and choose components more freely. 

Of the options provided to the base class, these seem to be the "core" options:
- name
- parent
- outdir
- projectType
- logging
- projenrcJson

eladb:
Extract some opinionated (mostly GitHub-specific) behavior from the base `Project` class to a new subclass called `GitHubProject` so it will be possible to create project types that don’t carry those opinions.

Going forward, we would want to get rid of this project type, but this requires a bit more refactoring and decoupling from GitHub (e.g. abstract workflows). To that end, the new type is marked “deprecated”.

Misc: Deprecate `projectType` and remove from `@featured`.

Resolves #1011

---


src/project.ts
```diff
@@ -652,3 +495,201 @@ export interface NewProject {
...
+   */
+  readonly clobber?: boolean;
+
+  /**
+   * The README setup.
+   *
+   * @default - { filename: 'README.md', contents: '# replace this' }
+   * @example "{ filename: 'readme.md', contents: '# title' }"
+   */
+  readonly readme?: SampleReadmeProps;
```
Chriscbr, line 562: I feel like LICENSE/README at least concepts are more software-universal and could probably stay in the base ProjectType, but just 2 cents. Omitting sample files from the base Project type could also be desirable.

Chriscbr (approved):

PR #1028, opened 2021-09-03
fix: Add ability to override default python version for Poetry projects

lmarsden:
* The Python version for poetry projects is currently hard-coded to ^3.6
* This hard-coded Python version has a knock-on effect to the versions of other packages that are retrieved by poetry. Some projects require different versions of Python.
* Remove the auto-addition of the Python dependency in favour of specifying it explicitly in the 'deps' section of each project alongside other project dependencies e.g. `"python@^3.6,<3.9"`. This allows individual projects to specify the version(s) of python they use.

---


Chriscbr (commented):
If a python version is still required, maybe it should be exposed as a specific field to make it clear it should be configured? For example, we could add a field like `minPythonVersion` (and `maxPythonVersion` if appropriate) to `PythonPackagingOptions`, similar to what we have for Node projects, and have the dependency automatically added if poetry is used.

lmarsden:
> If a python version is still required, maybe it should be exposed as a specific field to make it clear it should be configured? For example, we could add a field like `minPythonVersion` (and `maxPythonVersion` if appropriate) to `PythonPackagingOptions`, similar to what we have for Node projects, and have the dependency automatically added if poetry is used.

I have reinstated the `python@^3.6` default python version for poetry projects for backwards compatibility. Poetry users will be used to declaring `python` as a package dependency, so I don't feel it merits `min/maxPythonVersion` fields on `PythonProject`. I considered adding `min/maxPythonVersion` to `PoetryPyprojectOptionsWithoutDependencies` but they don't naturally fit here either, as they are not written as standalone sections into `pyproject.toml` like the other fields. So went with simply defaulting to the existing version in cases where it is not specified, making this a non-breaking change but also fixing the issue that it was previously not possible to override the hard-code version.

src/python/poetry.ts
```diff
@@ -78,11 +75,19 @@ export class Poetry extends Component implements IPythonDeps, IPythonEnv, IPytho
 
   private synthDependencies() {
     const dependencies: { [key: string]: any } = {};
+    var pythonDefined: boolean = false;
```
Chriscbr, line 78:
```suggestion
    let pythonDefined: boolean = false;
```

Chriscbr (changes requested):
OK - yeah let's try this out and see if it makes for a simpler experience. Thanks for updating the docs as well! Just one small nitpick below.

src/python/poetry.ts
lmarsden: Thanks, suggested change above now applied.

Chriscbr (changes requested):
I should have called this out in the last review, but we should also add at tests here as well.

lmarsden:
Test now added

Chriscbr (approved):

Chriscbr:
Same issue as #1030 (not due to your code changes). This should fix it:
```
git checkout main
git pull
git checkout <pr-branch>
git merge main
yarn projen && yarn build  # in case build steps changed
git add .
git commit
git push
```

lmarsden:
Build now fixed. PR was automatically merged once this was done.

PR #1033, opened 2021-09-04
fix: release workflow jobs not idempotent

### LINKED ISSUE #983: Release workflow idempotence
Context: For JSII projects that are configured using projen to publish to multiple package registries (using options like `publishToMaven`, `publishToPyPi`, etc.), the default GitHub workflow generated by projen currently works like the following:

-  a job named `release` bumps the version
- jobs named `release_github`, `release_npm`, `release_pypi` etc. publishes artifacts to the appropriate registries

The command `npx projen release`, which is run as part of the `release` job above, is idempotent by itself (if I run it multiple times locally, it bumps to the same version every time). But the entire workflow is not idempotent altogether.

For example, if the `release` job succeeds and the `release_github` job succeeds, but `release_npm` fails, then re-running the entire workflow manually (because [individual jobs can't be re-run](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234)) causes the version to be bumped _again_ causing the latest commit to be tagged with two versions. I validated this behavior in a test repo.

This is a problem because if one of the release jobs fails (due to a recently expired token, etc.), since there is no way to re-run the job to re-try releasing a package to a registry.

### LINKED ISSUE #1032: publish:release task cannot be run locally
The task generated by `project.publisher.publishToGitHubReleases()` includes a GitHub workflow substitution that isn't valid to run in bash:

https://github.com/projen/projen/blob/672fba2cf46a320027987523793fa06b754896ce/src/release/publisher.ts#L101

We could add some notion of tasks that can only be run in certain environments, but I think for now the easiest fix is just to switch this from the [context syntax](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) to the [environment variable syntax](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables), i.e. by using `$GITHUB_REPOSITORY` instead.

Chriscbr:
Fixes #983
Fixes #1032

Addresses #983 by making both the "bump" and "publish:github" tasks idempotent (running them more than once on the same commit should have no new effect and does not error).

I tested this by pushing the change directly to my fork (Chriscbr/projen) and validating that clicking the "re-run all jobs" error on the build workflow did not cause new versions to be pushed.. I did not validate whether the tasks for publishing to other registries (publish:npm, publish:pypi, etc.) are idempotent.

---


src/release/publisher.ts
```diff
@@ -87,6 +87,24 @@ export class Publisher extends Component {
...
+    // release script that does not error when re-releasing a given version
+    const idempotentRelease = [
+      'errout=$(mktemp);',
+      `${ghRelease} 2> $errout && true;`,
+      'exitcode=$?;',
+      'if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then',
+      'cat $errout;',
+      'exit $exitcode;',
+      'fi',
+    ].join(' ');
```
Chriscbr, line 106: The publish:github task/job fails if you run it a second time because it will say "Release.tag_name already exists", so I wrote this workaround so it will ignore that particular error (for idempotence). I would write this as a multi-line script but then it would not work as an individual task. This feels a little hacky and I'm less experienced writing bash scripting, so very open to suggestions or cleaner solutions.

src/release/bump-version.ts
```diff
@@ -157,3 +139,53 @@ async function tryReadVersionFile(versionFile: string) {
...
   return JSON.parse(await readFile(versionFile, 'utf8'));
 }
+
+/**
+ * Determines the latest release tag.
+ * @param major (optional) A major version line to select from
+ * @param prerelease (optional) A pre-release suffix.
+ * @returns the latest tag, and whether it is the first release or not
+ */
+function determineLatestTag(cwd: string, major: number | undefined, prerelease: string | undefined): { latest: string; firstRelease: boolean } {
```
Chriscbr, line 149: Extracted this to a function since to make the script a bit easier to read / break into pieces

src/release/bump-version.ts
```diff
@@ -131,13 +93,33 @@ export async function bump(cwd: string, options: BumpOptions) {
...
+
+  if (currentTags.includes(`v${latest}`)) {
+    logging.info('Skipping version bump - using existing tagged version.');
+
+    // delete the existing tag (locally)
+    // if we don't do this, CHANGELOG is not generated by standard-version
+    exec(`git tag --delete v${latest}`, { cwd });
+
+    cmd.push('--release-as');
+    cmd.push(latest);
```
Chriscbr, line 108: To make the bump task idempotent, I initially wanted to skip running `standard-version`, but this caused errors because then no changelog is generated, and the release artifacts are expected by the subsequent publishing jobs.

src/release/publisher.ts
```diff
@@ -87,6 +87,24 @@ export class Publisher extends Component {
...
+      `gh release create ${getVersion}`,
+      '-R $GITHUB_REPOSITORY',
+      `-F ${changelogFile}`,
+      `-t ${getVersion}`,
+    ].join(' ');
+
+    // release script that does not error when re-releasing a given version
+    const idempotentRelease = [
+      'errout=$(mktemp);',
+      `${ghRelease} 2> $errout && true;`,
```
eladb, line 100: Why do you need `&& true` here?

src/release/publisher.ts
eladb: I think this is reasonable. Although a builtin might be nicer

eladb (approved):

src/release/publisher.ts
Chriscbr: Spent some time trying to remember since I wrote this a week ago... looks like I found [this StackOverflow post](https://stackoverflow.com/questions/6112540/return-an-exit-code-without-closing-shell) and thought the solution applied to my situation. But looks like it's actually very unrelated, so it's almost definitely cruft / just a no-op here.

PR #1047, opened 2021-09-08
feat(release): release triggers

### LINKED ISSUE #726: Manual releases are broken in new tag-based bumps

gradybarrett:
Manual publishing to git. This includes prepending the
changelog artifact onto a project-level changelog,
commiting that change as a release commit, and tagging that
release commit.

Closes #726

---


src/release/release.ts
```diff
@@ -204,6 +217,12 @@ export class Release extends Component {
       });
     }
 
+    this.publisher.publishToGit({
+      changelogFile: join(this.artifactsDirectory, this.version.changelogFileName),
```
gradybarrett, line 221: Joining artifact directory from version rather than hard coding artifact directory in the task. Any concerns with updating the other publish tasks as well and removing the ARTIFACT_DIR constants?

gradybarrett:
I didn't push the commit or tag, but can if that's desirable.

src/release/publisher.ts
```diff
@@ -572,9 +586,24 @@ export interface GitHubReleasesPublishOptions {
...
 
   /**
-   * The location of an .md file that includes the changelog for the release.
-   *
-   * @example changelog.md
-   */
+    * The location of an .md file (relative to `dist/`) that includes the changelog for the release.
+    *
+    * @example changelog.md
+    */
```
eladb, line 592:
```suggestion
   * The location of an .md file (relative to `dist/`) that includes the changelog for the release.
   *
   * @example changelog.md
   */
```

src/release/release.ts
eladb: Sounds good to me

src/util.ts
```diff
@@ -280,4 +280,18 @@ export function kebabCaseKeys<T = unknown>(obj: T, recursive = true): T {
...
   return result as any;
+}
+
+export function normalizeVersion(version: string) {
+  const ver = version.trim();
+
+  return ver.startsWith('v') ? ver.substr(1) : ver ;
+}
+
+export async function tryReadFile(file: string) {
```
eladb, line 291: This function seems like it's only used by tests. Please move to `test/util.ts`

src/util.ts
```diff
@@ -280,4 +280,18 @@ export function kebabCaseKeys<T = unknown>(obj: T, recursive = true): T {
     result[decamelize(k).replace(/_/mg, '-')] = v;
   }
   return result as any;
+}
+
+export function normalizeVersion(version: string) {
```
eladb, line 285: I can't see where this is used.

src/version.ts
```diff
@@ -24,6 +25,12 @@ export interface VersionOptions {
 export class Version extends Component {
   public static readonly STANDARD_VERSION = 'standard-version@^9';
 
+  public static async tryReadVersion(versionFile: string) {
```
eladb, line 28: Doesn't feel like this should be a public API as it is only used by builtin tasks. Move this to an internal file next to the builtins.

src/release/update-changelog.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!version) {
+    throw new Error(
+      `Unable to determine version from ${versionFile}. Cannot proceed with changelog update. Did you run 'bump'?`,
+    );
+  }
+
+  const inputChangelogContent = await readFile(inputChangelog, 'utf-8');
+
+  if (!inputChangelogContent.includes(version)) {
```
eladb, line 60: Should we use the same search pattern as we do with the output? e.g. `[${version}]`?

src/release/release.ts
```diff
@@ -204,6 +217,12 @@ export class Release extends Component {
       });
     }
 
+    this.publisher.publishToGit({
```
eladb, line 220: Does it make sense to always include this task? Feels like it can be slightly misleading because users who don't use manual publishing may be confused why this task even exists. And vice versa - if you want manual publishing, then all the workflows related to automated publishing don't make sense.

So it feels like maybe we need this to be an opt-in thing, mutually exclusive with automated release workflow.

@Chriscbr any suggestions on now to model this?

eladb (changes requested):

src/version.ts
gradybarrett: I was thinking from the perspective of encapsulating the standard for how to read and process a version file. But maybe this is a case of yagni.

src/release/update-changelog.ts
gradybarrett: We should! Fixed.

src/release/release.ts
gradybarrett: Pushed up 6a5e4e8d. Pretty minor change, but marked it as breaking since it might change the output artifacts directory for existing projects.

src/release/release.ts
Chriscbr: How about we only call this if `releaseEveryCommit` is false?

Are manual git releases compatible with scheduled releases (e.g. once a week - configured by `this.releaseSchedule`), or are those also incompatible?

Maybe it's possible to model these options as some a enum-like value like `ReleaseStrategy` which a user could provide as:

- `ReleaseStrategy.schedule({ cronExpression: ..., changelog: true })`
- `ReleaseStrategy.continuous({ changelog: false })`
- `ReleaseStrategy.manual({ changelog: true })`

I think that would be a pretty cool-looking API, at least if my mental model is on the same page as everyone else's.

Chriscbr (commented):
Didn't look at all of the implementation, but this looks cool. Let's be sure to update the docs / add some guidance about how to use this. 🙏 

We should probably make it clear in the docs if this is the only release method that creates/manages a CHANGELOG file for you.

src/release/release.ts
gradybarrett: Technically it's compatible because `publish:git` is idempotent, but I'm not sure it makes sense as part of those flows.
I was thinking along the same lines regarding an enum-like approach. The first time I interacted with `project.release` I half expected it to have a type of `ReleaseType` (`ReleaseStrategy` is a better name 😄).

gradybarrett: @Chriscbr @eladb I'm most of the way through adding in ReleaseStrategy, but I noticed that technically releaseEveryCommit and releaseSchedule can coexist - there's no check for mutual exclusion. Was that intended?

Chriscbr: Probably not intended, no.

src/release/release-strategy.ts
```diff
@@ -0,0 +1,139 @@
...
+   *
+   * Only defined if this is a scheduled release.
+   *
+   * @example '0 17 * * *' - every day at 5 pm
+   *
+   * @default undefined
+   */
+  readonly schedule?: string;
+}
+
```
gradybarrett, line 58: @Chriscbr let me know what you think of this.

src/release/release-strategy.ts
```diff
@@ -0,0 +1,139 @@
...
+  readonly changelog?: boolean;
+}
+
+interface ReleaseOptions {
+  /**
+   * Whether or not a project-level changelog should be maintained.
+   *
+   * @default false
+   */
+  readonly changelog?: boolean;
```
gradybarrett, line 38: This could also be the project changelog file path. 🤔 

``js
ReleaseStrategy.manual({ changelog: 'foo/changelog.md' });
``

gradybarrett: The first option didn't work when I thought about integrating changelogs with automated releases in the future.

src/release/release.ts
gradybarrett: Refactored and added ReleaseStrategy. Let me know what you guys think.

gradybarrett:
> Didn't look at all of the implementation, but this looks cool. Let's be sure to update the docs / add some guidance about how to use this. 🙏

We should probably make it clear in the docs if this is the only release method that creates/manages a CHANGELOG file for you.

Should have this covered now between class, method, and release.md docs. Let me know if there are any holes anywhere.

src/version.ts
eladb: I try to avoid exposing APIs that I am not sure are actually needed :-D

docs/releases.md
```diff
@@ -40,3 +40,46 @@ releaseBranches: {
...
+`publish:git` task for you to use. The OOB manual release workflow would then look like:
+
+```terminal
+$ yarn build
+$ yarn release
+$ yarn publish:git
+$ git push --follow-tags origin main
+```
+
+You can optimize the release experience using the `publish` and `pushArtifacts` manual release options
```
eladb, line 70: Kind of feels like the default for `yarn release` in "manual" mode should be to perform publishing. The piecemeal tasks are an implementation detail.

src/release/release.ts
```diff
@@ -216,11 +252,17 @@ export class Release extends Component {
...
     if (githubRelease) {
       this.publisher.publishToGitHubReleases({
-        changelogFile: this.version.changelogFileName,
-        versionFile: this.version.versionFileName,
+        changelogFile: join(this.artifactsDirectory, this.version.changelogFileName),
+        versionFile: join(this.artifactsDirectory, this.version.versionFileName),
       });
     }
 
+    this.releaseStrategy.isManual && this.publisher.publishToGit({
```
eladb, line 260: please use an "if" statement

docs/releases.md
```diff
@@ -40,3 +40,46 @@ releaseBranches: {
...
 ```
+
+## Release Strategies
+
+If the project type supports it, then you can specify a `releaseStrategy`. You can use this to control whether
+or not your releases are automated as well as any unique artifacts associated with releases such as project-level
+changelogs.
+
+```js
+releaseStrategy: ReleaseStrategy.scheduled({ schedule: '0 17 * * *' }),
```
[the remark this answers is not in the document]
eladb: @Chriscbr how will this be usable if projenrc is in JSON? We either need to explicitly support "union-like classes" or we need to prevent them in the "options" API...

eladb (changes requested):
@gradybarrett I love it that we are modeling release triggers more formally.

See some initial notes. I hope I'll have time for a more thorough review later this week.
This is a very sensitive area of the code base 👍

src/version.ts
gradybarrett: Fair enough. :) I'll move it for now.

docs/releases.md
gradybarrett: Yep, that's fair. Do you think we should even expose the ability to shut off publishing in manual mode?

```js
releaseStrategy: ReleaseStrategy.manual({
  changelog: true,
  changelogPath: 'my/changelog.md'
}),
```

docs/releases.md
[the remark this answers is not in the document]
gradybarrett: I don't think the current implementation will serialize without establishing a new standard and expectation for how to serialize/de-serialize union-like classes. Supporting means that the projenrc.json/yaml representations will have special/different rules.

For example:

```json
{
  "releaseTrigger":
    "type": "continuous"
    "options": {
      "changelog": "true"
    }
}
```

which when parsed executes the `ReleaseTrigger.<type>` factory and passes it provided options. Potentially not great for dev ux continuity.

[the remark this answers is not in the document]
Chriscbr: Created an issue: #1062

Not sure what you mean by the "options" API. We support other union-like classes currently in project constructors, like `depsUpgrade: DependenciesUpgradeMechanism` - but as you mentioned, they can't be provided in the CLI during `projen new` or in projenrc.json.

docs/releases.md
gradybarrett: Simplified a bit.

src/release/publisher.ts
```diff
@@ -98,13 +98,45 @@ export class Publisher extends Component {
...
+   *
+   * @param options Options
+   */
+  public publishToGit(options: GitPublishOptions) {
+    const releaseTagFile = options.releaseTagFile;
+    const changelog = options.changelogFile;
+    const projectChangelogFile = options.projectChangelogFile;
+    const gitBranch = options.gitBranch ?? 'main';
+
+    const taskName = (gitBranch == 'main' || gitBranch == 'master') ? 'publish:git' : `publish:git:${gitBranch}` ;
```
eladb, line 114:
```suggestion
    const taskName = (gitBranch === 'main' || gitBranch === 'master') ? 'publish:git' : `publish:git:${gitBranch}` ;
```

src/release/release-trigger.ts
```diff
@@ -0,0 +1,141 @@
...
+interface ReleaseTriggerOptions {
+  /**
+   * Project-level changelog file path.
+   *
+   * Ignored if `changelog` is false
+   */
+  readonly changelogPath?: string;
+
+  /**
+   * Whether or not this is a continuous release.
```
eladb, line 39: Add a sentence explaining what continuous release means (i.e. "releases every commit")

src/release/release-trigger.ts
```diff
@@ -0,0 +1,141 @@
...
+   * Creates a manual release trigger.
+   *
+   * Use this option if you want totally manual releases.
+   *
+   * This will give you the same release task + a git publish command
+   * that will handle project-level changelog management and release
+   * tagging.
+   *
+   * To release you can run `yarn release` followed by `yarn publish:git`,
+   * then push the generated artifacts.
```
eladb, line 70: Doesn't `publish:git` already pushes?

src/release/release.ts
```diff
@@ -16,24 +18,46 @@ const LATEST_COMMIT_OUTPUT = 'latest_commit';
 export interface ReleaseProjectOptions {
   /**
    * Automatically release new versions every commit to one of branches in `releaseBranches`.
+   *
    * @default true
+   *
+   * @deprecated Use `releaseTrigger` instead
```
eladb, line 24:
```suggestion
   * @deprecated Use `releaseTrigger: ReleaseTrigger.continuous()` instead
```

src/release/release.ts
```diff
@@ -16,24 +18,46 @@ const LATEST_COMMIT_OUTPUT = 'latest_commit';
...
   /**
-    * CRON schedule to trigger new releases.
-    *
-    * @default - no scheduled releases
-    */
+   * CRON schedule to trigger new releases.
+   *
+   * @default - no scheduled releases
+   *
+   * @deprecated Use `releaseTrigger` instead
```
eladb, line 33:
```suggestion
   * @deprecated Use `releaseTrigger: ReleaseTrigger.scheduled()` instead
```

src/release/release.ts
```diff
@@ -16,24 +18,46 @@ const LATEST_COMMIT_OUTPUT = 'latest_commit';
...
+   */
   readonly releaseSchedule?: string;
 
+  /**
+   * The release trigger to use.
+   *
+   * @default - Continuous releases
```
eladb, line 40:
```suggestion
   * @default - Continuous releases (`ReleaseTrigger.continuous()`)
```

src/release/release.ts
```diff
@@ -16,24 +18,46 @@ const LATEST_COMMIT_OUTPUT = 'latest_commit';
...
 
+  /**
+   * Project-level changelog file path.
+   *
+   * `publish:git` will write a project-level changelog to this file as part of
+   * git publication.
+   *
+   * @default "CHANGELOG.md"
+   */
+  readonly projectChangelogFile?: string;
```
eladb, line 59: This only applies with manual releases, right? Shouldn't this just be part of `ReleaserTrigger.manual()`?

eladb (changes requested):
Great job. A few additional minor comments and I think we are ready!

src/release/release-trigger.ts
gradybarrett: Whoops, forgot to update that. Thanks.

src/release/release-trigger.ts
gradybarrett: Done

src/release/release.ts
gradybarrett: Yes. I'll remove it from the others.

I originally intended to add support for it everywhere, but I think it's better to wait and add that incrementally.

gradybarrett: Oh yes, this is no longer used. Should have been removed in the release trigger refactor.

gradybarrett: Removed.

docs/releases.md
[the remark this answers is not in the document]
gradybarrett: Resolving this now that we have #1062.

eladb (approved):

gradybarrett:
Thanks @eladb! 🎉

eladb:
Thank you for such high quality contribution 🙏

hoegertn:
https://github.com/projen/projen/blame/180393e9795f4d0f4d7720268c0a87328a38469c/src/release/update-changelog.task.ts#L19

Is this name mismatch intended? I get errors after upgrading and trying to use the release job. to sure if this is the issue but seems weird to me.

gradybarrett:
@hoegertn definitely not intended. Just missed a rename when switching to use the new release tag file.

#1116 to fix.

gradybarrett:
There still possibly something weird about release tag prefixes and the dist/changelog.md. I'll look more at it tomorrow. Might just be a local setup thing. In any case, when there's no prefix defined we need to strip off the `v` prefix from the release tag. I'll fix that tomorrow unless someone beats me to it.

hoegertn:
Can confirm this: https://github.com/projen/projen/blame/main/src/release/update-changelog.ts#L64

Fails due to the v being on the tag but not in the changelog file

PR #1048, opened 2021-09-08
feat(eslint): import aliasing

mtimbs:
allow specification of configuration to remove lint errors for projects that utilise import aliasing.

e.g `import { bar } from '@src/foo'`

## Notes
I'm not entirely happy with the API here. I'd have preferred to take an option for `alias` as an object. However, when trying to compile I got jsii errors saying map types were unsupported

```typescript
readonly alias?: {
  map: [string, string][];
  extensions: string[];
};
```

If there is a way to support this kind of API I believe it would be more intuitive for the consumer than having different options for the map and the extensions as submitted in this PR.

```typescript
  /**
   * Enable import alias for module paths
   * @default undefined
   */
  readonly aliasMap?: { [key: string]: string; };

  /**
   * Enable import alias for module paths
   * @default undefined
   */
  readonly aliasExtensions?: string[];
```

### Possible future task

All TypeScript projects need to configure import aliasing in 3 places
- tsconfig
- eslint config
- jest config

tsconfig and jest already exposed ways to set this but eslint did not. Given that projen generates all three of these files, there is opportunity to define the mapping in one place and inject it into all three config files. I'd have to do some digging into project structure to figure out how this might work but happy to help with this if you think it's worth having

---


src/eslint.ts
```diff
@@ -47,6 +48,18 @@ export interface EslintOptions {
    * @default false
    */
   readonly prettier?: boolean;
+
+  /**
+   * Enable import alias for module paths
+   * @default undefined
+   */
+  readonly aliasMap?: [string, string][];
```
Chriscbr, line 56: Because we are using [jsii](https://aws.github.io/jsii/) under the hood to compile projen and generate APIs in other languages, I don't believe we can use certain fancier types like [string, string] in our exported interfaces. Instead we should probably change this to a record/dictionary type:
```suggestion
  readonly aliasMap?: { [key: string]: string };
```

...and add some code in the constructor that converts this back into an array of tuples.

Chriscbr (changes requested):
Neat, it would be great to support this!

src/eslint.ts
mtimbs: Ok, I'll make the necessary changes. I also considered automatically configuring this based on the `.tsconfig` that we have access to already within Eslint - but opted for this approach to be more explicit. If we need to deviate from the normal eslint structure it might be ok to have this implicit configuration based on aliases declared in the `.tsconfig`?

Chriscbr: Do you need to define import aliases in both places for source code to compile and lint without errors?

If the user provides import aliases/paths within the tsconfig options and we could pass them to the eslint config as well (so they don't have to write them twice), I think that could be convenient! I'm also okay with leaving them separate if needed.

mtimbs: Yeah unfortunately need to define them in 3 places. Jest config, tsconfig and eslint. Let me try and work some magic to infer them from tsconfig as I think that would be a nicer developer experience (so long as it is documented what is happening)

mtimbs: I ended up making the changes to the types only. The reference to tsconfig available was just the path, not the actual config options

src/__tests__/eslint.test.ts
```diff
@@ -60,3 +60,54 @@ describe('prettier', () => {
...
+
+    // WHEN
+    new Eslint(project, {
+      dirs: ['src'],
+
+    });
+
+    // THEN
+    expect(synthSnapshot(project)['.eslintrc.json']).toMatchSnapshot();
+  });
```
Chriscbr, line 83: I think it's OK to remove this test - doesn't seem to validate anything specific to the aliases feature.

src/eslint.ts
```diff
@@ -305,6 +322,12 @@ export class Eslint extends Component {
           '@typescript-eslint/parser': ['.ts', '.tsx'],
         },
         'import/resolver': {
+          ...( options.aliasMap && options.aliasExtensions && {
```
Chriscbr, line 325: Looking at https://www.npmjs.com/package/eslint-import-resolver-alias, I think aliasExtensions can be omitted, so it should probably be:
```suggestion
          ...( options.aliasMap && {
```

Chriscbr (changes requested):
Just a few small changes and this should be good to ship.

src/__tests__/eslint.test.ts
mtimbs: Done

src/__tests__/eslint.test.ts
```diff
@@ -31,10 +31,11 @@ describe('prettier', () => {
       defaultReleaseBranch: 'master',
     });
 
+
     // WHEN
     new Eslint(project, {
-      dirs: ['mysrc'],
-      prettier: true,
+      dirs: ['src'],
```
Chriscbr, line 37: Please revert the changes here

Chriscbr (changes requested):

src/__tests__/eslint.test.ts
mtimbs: Sorry! All done

Chriscbr (approved):

Chriscbr:
Looks like the snapshots need to be updated: https://github.com/projen/projen/runs/3627825547#step:9:9

Issue #1050, opened 2021-09-09
“Node v10.24.1 has reached end-of-life” warning message in workflow 


wchaws:
I found this warning in https://github.com/cdklabs/cdk-ecr-deployment/pull/120/checks?check_run_id=3551950707 workflow

```
Ran all test suites.
🤖 build » test » eslint | env: PATH=/usr/lib/node_module...
🤖 build » test » eslint | eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern src test build-tools .projenrc.js
🤖 build » compile | env: PATH=/usr/lib/node_module...
🤖 build » compile | jsii --silence-warnings=reserved-word --no-fix-peer-dependencies
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!                                                                                                           !!
!!  Node v10.24.1 has reached end-of-life and will no longer be supported in new releases after 2021-09-01.  !!
!!  Please upgrade to a supported node version as soon as possible.                                          !!
!!                                                                                                           !!
!!  As of the current release, supported versions of node are:                                               !!
!!  - ^12.7.0                                                                                                !!
!!  - ^14.5.0                                                                                                !!
!!  - ^16.3.0                                                                                                !!
!!                                                                                                           !!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
🤖 build » compile » docgen | env: PATH=/usr/lib/node_module...
🤖 build » compile » docgen | jsii-docgen
🤖 build » package | env: PATH=/usr/lib/node_module...
🤖 build » package | jsii-pacmak
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!                                                                                                           !!
!!  Node v10.24.1 has reached end-of-life and will no longer be supported in new releases after 2021-09-01.  !!
!!  Please upgrade to a supported node version as soon as possible.                                          !!
[...]

Issue #1034, opened 2021-09-04
update jsii docker image in GitHub workflows to fix failing builds


dontirun:
the `jsii/superchain` docker image should ultimately be updated to the the `jsii/superchain:1-buster-slim-node14` image, but use `jsii/superchain:node14` in the interim

1. Node 10 is deprecated and builds are starting to fail as packages start to drop support, [as seen here](https://github.com/cdklabs/cdk-nag/runs/3510017434?check_suite_focus=true), thus switching to a Node.js 14 based image would be appropriate
2. There have been updates to the `jsii/superchain` image that have deprecated many tags (`latest`, `node14` , etc) and the recommendation is to switch to the `:node(10|14)(-buster-slim|-nightly)` tags and use the new debian based image. However these images cause problems with the current GitHub workflows as [seen here](https://github.com/cdklabs/cdk-nag/runs/3510520231?check_suite_focus=true), due to the commands (like checking out and making directories) needing `sudo` permissions. While ultimately projen should use those images, the failing builds need to be fixed

Proposed remediation is to immediately switch to  `jsii/superchain:node14` to fix the failing builds then switch to `jsii/superchain:1-buster-slim-node14` after issues with permissions with GitHub actions have been resolved

PR #1052, opened 2021-09-09
fix: node 10.x is end-of-life so default to 14.x

dontirun:
Related to #1034 , #1050

BREAKING CHANGE: The default minimum Node.js version for node projects is now 14.x since Node 10.x is now end-of-life. Therefore we also change the default build container for JSII projects is now `jsii/superchain:node14`.

---


eladb (approved):

.projenrc.js
```diff
@@ -41,7 +41,7 @@ const project = new JsiiProject({
 
   projenDevDependency: false, // because I am projen
   releaseToNpm: true,
-  minNodeVersion: '10.17.0',
+  minNodeVersion: '14.17.0',
```
iliapolo, line 44: @eladb With this, users are forced to run their workflows with node 14, since otherwise projen won't be installed.

For projects choosing to only support node > 14, thats fine. But for projects who still like to support node 12, and specify:

```js
minNodeVersion: '12.0.0',
workflowNodeVersion: '14.0.0'
``` 

It means their workflows aren't testing the declared supported node version. Was this the intent? 
I wonder if we should stick to node 12 as long as its not EOL (7 months away)

> See https://endoflife.date/nodejs

eladb: That's true only for jsii workflows. Normal node/typescript projects don't use the jsii superchain image so they will run with the correct node version configured via setup-node.

I think we just need to stop using superchain for jsii projects and setup the environment using GitHub actions instead. Would also speed up things dramatically. Would be an interesting project to add some mechanism/modeling around build dependencies that will reflect into the workflow.

iliapolo: Not sure I understand how is this related to jsii. 

How would a non jsii project configure both workflow and engine node version to 12 if projen requires 14?

hoegertn:
@eladb I totally agree on Node14 being the default, but do we need `14.17.x` order would an earlier minor version be ok too?

Issue #1051, opened 2021-09-09
feat: support for a Go project


Hunter-Thompson:
Just like Python, TS, and Node, it would be awesome to be able to support a Go project.

```
const project = new GoProject({})
```

Would probably require a RFC for this since its a pretty big task.

eladb:
Yes! Contributions are welcome

PR #1061, opened 2021-09-12
RFC 1051: Go Project

Hunter-Thompson:
Requesting comments and feedback on support for a Go Project. [Rendered version.](https://github.com/smallcase/projen/blob/rfc/go-project/rfc/go-project.md)

I will update the doc with API specifications once I get some more feedback.

RFC for #1051 

---

### DIFF
--- a/rfc/go-project.md
+++ b/rfc/go-project.md
@@ -6,12 +6,11 @@ Just like TS, Node and Python, having support for a Go project would be great.
 
 Get dependencies if vendoring is disabled, run `go fmt`, run the linter, run go test, and finally run build.
 
-## Tests
-
-
 ## Releases
 
 Setup semantic versioning and release using semantic-release.
+
+## Tests
 
 Expose an API that interacts with `go test`, allows arguments to be passed.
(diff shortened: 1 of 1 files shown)


polothy:
Yah Go!

Hunter-Thompson:
I will work on this when time permits. Also looking for more comments from fellow go programmers on what they'd like to see.

polothy:
What you outlined is what I do for all my Go projects.  My only comment is that instead of `go fmt` I just use `golangci-lint run --fix` which can fix any formatting issues depending on which linters I enable.  But, `go fmt` or `goimports` would be good if someone wanted to opt out of golangci-lint.

Hunter-Thompson:
Thought it would be a lot easier and nicer to use go for this, so I wrote a little something. It's experimental.

Took inspiration from projen, hope that's alright : )

https://github.com/Hunter-Thompson/gojen

eladb:
@Hunter-Thompson Thanks for sharing. Would be great to add this to projen

PR #1066, opened 2021-09-13
feat(tsconfig): expose tsconfig: emitDecoratorMetadata prop

hassanazharkhan:
If the project was created using `--projenrc-ts`, then there is no way out to set the tsconfig `emitDecoratorMetadata` porp,
after this getting merged we could use this `.projent.ts `

---


eladb (approved):

eladb:
Diff issue: https://github.com/projen/projen/pull/1066/checks?check_run_id=3589824893#step:9:14

Chriscbr:
Looks like some trailing spaces on lines 130 and 133 are causing the diff to fail:
https://github.com/projen/projen/runs/3592259167#step:9:14

PR #1067, opened 2021-09-13
feat: addOverride can override array elements

Chriscbr:
Previously, it was not possible to override specific array elements through the `addOverride` escape hatch, even though this is possible in some other CDK libraries (like the AWS CDK). This adds this feature back.

---


src/__tests__/util.test.ts
```diff
@@ -100,7 +100,7 @@ describe('deepMerge (destructive: false)', () => {
 
   test('overwrites non-objects', () => {
     // GIVEN
-    const original = { a: [] };
+    const original = { a: 'foo' };
```
Chriscbr, line 103: These two tests were invalidated via the changes, so I adjusted them slightly. I think this is necessary for the behavior we want to have, which is that overriding [1, 2, 3] with { '0': 5, '2': 8 } should produce [5, 2, 8].

In another world, we could keep this test passing by adding logic that checks if the keys of the object you're overriding with are integers or not, and changes the overriding behavior accordingly. But I don't think it's worth adding this complexity unless there is a use case that needs it.

eladb (approved):

PR #1069, opened 2021-09-14
fix: pullRequestTemplateContents was using the wrong type

hassanazharkhan:
`pullRequestTemplateContents` should be a `string[]`

---


eladb (approved):
Indeed

PR #1073, opened 2021-09-15
feat(tsconfig): expose "baseUrl" and "paths" in compilerOptions

hassanazharkhan:
Added missing props in `TypeScriptCompilerOptions` which are needed to re-map the imports.

---

### DIFF
--- a/src/typescript-config.ts
+++ b/src/typescript-config.ts
@@ -345,6 +345,20 @@ export interface TypeScriptCompilerOptions {
    * Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
    */
   readonly allowSyntheticDefaultImports?: boolean;
+
+  /**
+   * Lets you set a base directory to resolve non-absolute module names.
+   *
+   * You can define a root folder where you can do absolute file resolution.
+   */
+  readonly baseUrl?: string;
+
+  /**
+   * A series of entries which re-map imports to lookup locations relative to the baseUrl, there is a larger coverage of paths in the handbook.
+   *
+   * paths lets you declare how TypeScript should resolve an import in your require/imports.
+  */
+  readonly paths?: { [key: string]: string[] };
 }
 
 export class TypescriptConfig {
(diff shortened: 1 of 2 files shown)


.all-contributorsrc
```diff
@@ -477,6 +477,69 @@
       "contributions": [
         "code"
       ]
+    },
```
Chriscbr, line 480: Can you revert the changes to this file and the README? We should keep this PR focused on one change and update the contributors list in a separate PR.

src/typescript-config.ts
```diff
@@ -345,6 +345,22 @@ export interface TypeScriptCompilerOptions {
...
+  readonly baseUrl?: string;
+
+  /**
+   * A series of entries which re-map imports to lookup locations relative to the baseUrl, there is a larger coverage of paths in the handbook.
+   * paths lets you declare how TypeScript should resolve an import in your require/imports.
+   *
+   * @see https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping
+   * @default undefined
+  */
+  readonly paths?: { [key: string]: string[] };
```
Chriscbr, line 363: There should also be a change to API.md - it should be automatically updated when you run the compile task.

Chriscbr (changes requested):

.all-contributorsrc
hassanazharkhan: Done

src/typescript-config.ts
hassanazharkhan: Done

Chriscbr (approved):

src/__tests__/python/__snapshots__/poetry.test.ts.snap
```diff
@@ -147,6 +147,11 @@ cython_debug/
         "type": "runtime",
         "version": "0.0.2",
       },
+      Object {
```
Chriscbr, line 150: I think if you just omit this file from the PR, the build should pass.

hassanazharkhan: Done

PR #1077, opened 2021-09-16
fix(new): Use NPM for `projen new` instead of Yarn

### LINKED ISSUE #885: Use NPM to add projen and external module instead of Yarn
When using an external project type, projen installs the targeted package with yarn. 
While I understand the desire to use Yarn as much as possible, I think it would make more sense to use the default package manager. This brings a few benefits:

1. Since the user must have node, then they will always be able to use projen regardless of whether or not they have yarn
2. If the project being loaded is configured to use NPM or PNPM, then there really is no reason for yarn to be a part of the process at all and at worse can cause confusion
3. (This could be accomplished without switching to NPM, but still) Peer deps of the external project are installed by default through NPM, which may or may not be necessary for the project to synth
4. `yarn add` has unfortunate behavior when being run in a subfolder of another node project. It attempts to add it to the package.json of the parent folder even though it adds it to the current dir's node_modules. This can make it very slow as it tries to resolve the dependency tree in an unecessary way. NPM would not have this behavior with --prefix, although this could probably be solved in yarn with --cwd

MarkMcCulloh:
Fixes #885 

Updated cdk-appsync-project version based on https://github.com/kcwinner/cdk-appsync-project/pull/186 (Thanks @kcwinner !) to make sure the tests were working.

See the linked issue for the reasoning behind the change.

---


src/cli/cmds/new.ts
```diff
@@ -308,7 +290,7 @@ function yarnAdd(baseDir: string, spec: string): string {
  * @returns The string that includes the install command ("yarn add ...")
  */
 function renderInstallCommand(dir: string, module: string): string {
-  return `yarn add --modules-folder=${dir}/node_modules --silent --no-lockfile --ignore-engines --dev ${module}`;
+  return `npm i --silent -D -f --no-package-lock --prefix=${dir} ${module}`;
```
MarkMcCulloh, line 293: ℹ️  Note the `-f`. If this is not used, then there are almost always going to be peer dependency issues failures. Most external projects have projen as a peer dependency, but since projen has a 0 major version number and moves quickly, it's pretty impossible to satisfy so NPM install will fail. Yarn didn't install peer deps, so it avoided this. In case the external projects has peer deps necessary for synthesis, we should make sure to install them.

src/cli/cmds/new.ts
```diff
@@ -257,32 +256,15 @@ async function newProject(baseDir: string, type: inventory.ProjectType, args: an
...
-
-    const packageDir = fs.mkdtempSync(path.join(os.tmpdir(), 'projen-'));
-    const copy = path.join(packageDir, 'pkg.tgz');
-    fs.copyFileSync(spec, copy);
-
-    spec = copy;
-
-    dependencyInfo = `${moduleName}@${spec}`;
-  }
-
```
MarkMcCulloh, line 285: ℹ️  No longer needed because NPM doesn't have that issue. See the linked issue for the reasoning behind the change.

src/cli/cmds/new.ts
Chriscbr: Hmm, I wonder if the silent flag really helps us at all. Maybe take it out

Also since we don't have to type this command ourselves, might as well spell out the arguments to improve readability. 🙂 
```suggestion
  return `npm install --save-dev -f --no-package-lock --prefix=${dir} ${module}`;
```

src/cli/cmds/new.ts
Chriscbr: Awesome!

Chriscbr (changes requested):
Looks good to me, just one small comment to address.

Chriscbr (approved):

eladb (commented):
Thanks!

PR #1083, opened 2021-09-17
chore: remove files from projen's npm distribution

Chriscbr:
---


.projenrc.js
```diff
@@ -148,9 +150,20 @@ const setup = project.addTask('devenv:setup');
...
+project.npmignore.exclude('/.all-contributorsrc');
+
+project.npmignore.exclude('/scripts/');
+project.npmignore.exclude('/ARCHITECTURE.md');
+project.npmignore.exclude('/CODE_OF_CONDUCT.md');
+project.npmignore.exclude('/CONTRIBUTING.md');
+project.npmignore.exclude('/VISION.md');
+project.npmignore.exclude('/SECURITY.md');
+project.npmignore.exclude('/.gitattributes');
+project.npmignore.exclude('/.gitpod.yml');
```
Chriscbr, line 167: It probably makes sense to exclude .gitpod.yml in all project types, but the Gitpod component doesn't have any knowledge of the npmignore component (since it's not exclusive to node projects) - so I'm not sure where the logic should really belong...

PR #1100, opened 2021-09-27
feat: signoff dependency upgrade pr's

iliapolo:
Ok, this PR started off with me just wanting to add support for signing-off dependency upgrade PR's. This is done by passing the `signoff` argument to the [action](https://github.com/peter-evans/create-pull-request#action-inputs). 

This is necessary for project that want to enable the [DCO](https://github.com/apps/dco) check, i.e every CNCF project.

> For example: https://github.com/cdk8s-team/cdk8s-core/pull/68

The problem is that with the current API, enabling this sign-off would require something like:

```ts
depsUpgrade: DependencyUpgradeMechanism.githubWorkflow({
  signoff: true,
})
```

Which seems fairly ok. Problem is that once we provide a value to `depsUpgrade`, all its default values are no longer being applied, and those defaults are important. So in reality, this would actually require:

```ts
depsUpgrade: DependenciesUpgradeMechanism.githubWorkflow({
  signoff: true,
  ignoreProjen: false,
  workflowOptions: {
    labels: ['auto-approve'],
    secret: 'PROJEN_GITHUB_TOKEN',
    container: {
      image: 'jsii/superchain',
    },
  },
}),
```

Which is really painful. 

One option would be to keep the `DependenciesUpgradeMechanism` API and make it so user defined values are merged with default values, instead of overwriting them. This has some unnatural implications on the implementation, and is starting to become a code smell. 

We've briefly discussed our discontent from the `DependenciesUpgradeMechanism` API in the past and I figured now is a good time to re-evaluate it. 

The simplest and straight-forward API I came up with at the moment looks like so:

```ts
depsUpgrade: true, // not required - just here for demonstration 
depsUpgradeOptions: {
  signoff: true,
},
``` 

Basically the change is to get rid of this custom `DependenciesUpgradeMechanism` class and allow passing its options directly. This is easier to configure since it doesn't require an additional import in `.projenrc`, and is consistent with the API we currently provide for dependabot (i.e [`dependabot`](https://github.com/projen/projen/blob/main/src/node-project.ts#L129) and [`dependabotOptions`](https://github.com/projen/projen/blob/main/src/node-project.ts#L137))

I know we are doing our best to maintain backwards compatibility, but doing so in this case doesn't feel right, it will conflate the configuration too much and introduce awkward and bug prone implementation code. So I opted to break the API and throw an error linking to migration instructions for those who were explicitly configuring `depsUpgrade`. 

In the future we should think of a better way to collapse `dependabot` and `depsUpgrade` to a more ergonomic API.

--------------------

BREAKING CHANGE: `depsUpgrade` is now a `boolean` indicating whether or not to enable dependency upgrades (defaults to `true`). Options that were previously passed to `DependenciesUpgradeMechanism.githubWorkflow()` can now be directly passed to `depsUpgradeOptions`, and options passed to `DependenciesUpgradeMechanism.dependabot()` should be now passed to `dependabotOptions`, which is no longer deprecated. 

---


src/upgrade-dependencies.ts
```diff
@@ -285,8 +299,7 @@ export class UpgradeDependencies extends Component {
           'title': title,
           'labels': this.options.workflowOptions?.labels?.join(',') || undefined,
           'body': description,
-          'author': `${COMMIT_AUTHOR_NAME} <${COMMIT_AUTHOR_EMAIL}>`,
-          'committer': `${COMMIT_AUTHOR_NAME} <${COMMIT_AUTHOR_EMAIL}>`,
```
iliapolo, line 289: Replaced by adding the `Set git identity` step, which is also necessary for the sign-off.

src/github/constants.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Represents the github-actions user.
+ *
+ * Use this when you need to perform a commit as part of your workflow.
+ */
+export const GITHUB_ACTIONS_USER = {
+  name: 'github-actions',
+  email: 'github-actions@github.com',
+};
```
iliapolo, line 9: These details used to be scattered over the code, so I consolidated to a github-actions user that should be used from now on. Also changed `Automation` to `github-actions` to be consistent with the github username that appears in the UI on occasion.

src/__tests__/node-project.test.ts
```diff
@@ -202,7 +201,18 @@ describe('deps upgrade', () => {
     });
 
     const snapshot = yaml.parse(synthSnapshot(project)['.github/workflows/upgrade.yml']);
-    expect(snapshot.jobs.pr.steps[3].with.labels).toStrictEqual(project.autoApprove?.label);
+    expect(snapshot.jobs.pr.steps[4].with.labels).toStrictEqual(project.autoApprove?.label);
```
iliapolo, line 204: Step added to the workflow so index changed from `3` to `4`.

src/__tests__/node-project.test.ts
```diff
@@ -258,29 +268,36 @@ describe('deps upgrade', () => {
 
   test('github actions - with projen secret', () => {
     const project = new TestNodeProject({
-      depsUpgrade: DependenciesUpgradeMechanism.githubWorkflow(),
       projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',
     });
     const snapshot = synthSnapshot(project);
     expect(snapshot['.github/workflows/upgrade.yml']).toBeDefined();
-    expect(snapshot['.github/workflows/upgrade-projen.yml']).toBeDefined();
+    expect(snapshot['.github/workflows/upgrade-projen.yml']).toBeUndefined();
```
iliapolo, line 275: This was actually a bug in the test that went undetected due to a bug in the code...

src/__tests__/upgrade-dependencies.test.ts
```diff
@@ -34,12 +34,12 @@ test('upgrades command doesnt include ignored packages', () => {
...
     deps: ['dep1', 'dep2'],
-    depsUpgrade: DependenciesUpgradeMechanism.githubWorkflow({
+    depsUpgradeOptions: {
       exclude: ['dep2'],
-    }),
+    },
   });
 
-  const deps = 'jest jest-junit npm-check-updates standard-version dep1';
+  const deps = 'jest jest-junit npm-check-updates projen standard-version dep1';
```
iliapolo, line 42: This was actually a bug in the test that went undetected due to a bug in the code...

Chriscbr (approved):
Interesting - yeah the tradeoff here feels reasonable to me. I think providing a unified API (where "dependabot" or "GitHub workflow" are two supported implementations of a common interface) could be useful down the line, but for now the `DependenciesUpgradeMechanism` class isn't doing much for us.

src/node-project.ts
```diff
@@ -714,9 +724,17 @@ export class NodeProject extends GitHubProject {
...
+        labels: autoApproveLabel(depsAutoApprove),
+        ...(options.dependabotOptions ?? {}),
+      });
+      ignoresProjen = dependabotConf?.ignoresProjen;
+    }
+
+    if (depsUpgrade) {
+      const upgradeDependencies = new UpgradeDependencies(this, {
         // if projen secret is defined we can also upgrade projen here.
         ignoreProjen: !options.projenUpgradeSecret,
```
Chriscbr, line 739: This is outside of the immediate scope of this PR, but there's currently a small bug where the projen repository itself should be auto-approving PRs, but it is not because the upgrade.yml workflow is using the `GITHUB_TOKEN` workflow instead of `PROJEN_GITHUB_TOKEN`. Adding `projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN'` to the constructor should fix this - but if we do so, then this "ignoreProjen" flag gets flipped and projen will try upgrading itself.

Could you try adding this line and see if that fixes it? :)
```suggestion
      const upgradeDependencies = new UpgradeDependencies(this, {
        // if projen secret is defined we can also upgrade projen here.
        ignoreProjen: !options.projenUpgradeSecret || this.package.packageName === 'projen',
```

iliapolo:
@eladb You good with this change?

src/node-project.ts
iliapolo: I dont think this is necessary actually. The upgrade task won't try to upgrade `projen` because its not a dependency of `projen`. The only thing the explicitly upgrades projen is:

https://github.com/projen/projen/blob/epolon/upgrade-sign-off/src/node-project.ts#L754

Which already special cases `projen` itself. We should change `.projenrc` to configure:

```ts
depsUpgrade: {
  workflowOptions: {
    secret: 'PROJEN_UPGRADE_TOKEN`
  }
}
```

And that should do it. 

I'll take of this in a subsequent PR

src/github/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Workflow flow to configure git with the github actions identity.
+ */
+export const SET_GIT_IDENTITY_WORKFLOW_STEP = {
```
eladb, line 14: add type

src/upgrade-dependencies.ts
```diff
@@ -79,6 +78,13 @@ export interface UpgradeDependenciesOptions {
    */
   readonly ignoreProjen?: boolean;
 
+  /**
+   * Add Signed-off-by line by the committer at the end of the commit log message.
+   *
+   * @default false
```
eladb, line 84: Why not `true` by default?

src/util.ts
```diff
@@ -193,8 +193,11 @@ export function isObject(x: any): x is Obj<any> {
  */
 export function deepMerge(objects: Array<Obj<any> | undefined>, destructive: boolean = false) {
   function mergeOne(target: Obj<any>, source: Obj<any>) {
+    console.log('Target is:');
+    console.dir(target);
     for (const key of Object.keys(source)) {
       const value = source[key];
+      console.log(`Source: Key=${key}, value=${value}`);
```
eladb, line 200: revert

eladb (changes requested):

src/util.ts
iliapolo: Whoops, good catch.

src/upgrade-dependencies.ts
iliapolo: Hmm, no reason I guess...will change 👍

src/github/constants.ts
iliapolo: 👍

iliapolo:
@eladb Ready for another pass

src/upgrade-dependencies.ts
```diff
@@ -79,6 +78,13 @@ export interface UpgradeDependenciesOptions {
    */
   readonly ignoreProjen?: boolean;
 
+  /**
+   * Add Signed-off-by line by the committer at the end of the commit log message.
+   *
+   * @default true.
```
eladb, line 84:
```suggestion
   * @default true
```

eladb (approved):

PR #1104, opened 2021-09-28
fix: speed up discovery of external types

### LINKED ISSUE #1101: projen new command running into really long loops while checking .jsii files
I am facing an issue with `projen new`  command for external constructs library. It seems that the command takes a lot of time to run through when there are lot of dependencies in the external library. One such example would be a constructs library which our team uses.

The issue  is caused because of checking of .jsii files by projen recursively, introduced ~0.17.68. The earlier versions of projen like 0.17.65 work fine.
The issue seems to be here: https://github.com/projen/projen/blob/main/src/inventory.ts#L106
Having deps like @aws-cdk/@aws-cloudfront which have lot of dependencies in turn causes the loops to be really long with large IO times for reading all jsii files.

The command to reproduce it is:
`npx projen new --from @rioclaid/cdk-constructs rio-fargate-web-app`
This is one of the libraries we use where we can see the large loops in action. It can take upto ~45 mins for the command to run.

pgarbe:
by avoiding re-probing same dependencies multiple times

No new tests added as the functionality will still be the same.
It's just a performance boost when the extnernal module
has multiple dependencies (e.g. aws cdk v1).

Fixes #1101

---

### DIFF
--- a/src/inventory.ts
+++ b/src/inventory.ts
@@ -105,12 +105,19 @@ export function discover(...moduleDirs: string[]) {
  */
 function discoverJsiiTypes(...moduleDirs: string[]) {
   const jsii: JsiiTypes = {};
+  const discoveredManifests: Array<string> = [];
 
   const discoverJsii = (dir: string) => {
     const jsiiFile = path.join(dir, '.jsii');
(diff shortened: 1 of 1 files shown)


Chriscbr (approved):
🙏

PR #1112, opened 2021-10-03
feat(upgrade): multi branch upgrades

iliapolo:
Some projects have multiple branches being maintained simultaneously. For example:

- [`constructs`](https://github.com/aws/constructs) manages two major versions in two different branches. (`main` and `10.x`)
- [`cdk8s-plus`](https://github.com/cdk8s-team/cdk8s-plus) manages different packages, each in its own branch (`k8s-22/main`, `k8s-21/main`, `k8s-20/main`)

Currently, the `upgrade` workflow will only upgrade dependencies on the default repository branch. This is because scheduled workflows only trigger on the default branch. 

To fix this, we allow passing a `branches` property, which will create a dedicated `upgrade-${branch}.yml` workflow for each branch. For example: https://github.com/iliapolo/cdk8s-plus/tree/k8s-22/main/.github/workflows 

---


src/upgrade-dependencies.ts
```diff
@@ -120,8 +120,16 @@ export class UpgradeDependencies extends Component {
 
     const task = this.createTask();
 
+    // this will translate into the default branch of the repo
+    // just like not specifying anything.
+    const defaultBranch = undefined;
+
+    const branches = options.workflowOptions?.branches ?? [defaultBranch];
```
eladb, line 127: I think that the default should be that all release branches will get upgrades, not just the default branch.

iliapolo: Good idea. I'll change

iliapolo: @eladb Done. I opted to keep the branch suffix on the workflow filename even for default branches such as `master` and `main`, which created a lot of snapshot and test changes. Let me know if you'd rather special case them.

eladb (approved):

src/upgrade-dependencies.ts
eladb: That's ok

PR #1117, opened 2021-10-05
fix(release): prefix env variable clash

gradybarrett:
I think PREFIX is clashing with an existing environment
variable used by nodejs or maybe nvm. When I run `yarn bump`
locally using manual releases and no prefix, PREFIX gets set
to `/usr/local`. Given the significance of that path I assume
that PREFIX is being used by some other tool on my system.

Making PREFIX naming more specific to avoid this problem.

---


gradybarrett:
@Chriscbr ran into this problem while testing manual releases. I'll dig a bit deeper into my own system configs tomorrow but I think we're going to have to name env variables more specifically to account for less predictable host system setups. Let me know what you think, and curious if you or anyone else runs into this as well.

eladb (approved):

PR #1123, opened 2021-10-07
fix(awscdk-app-ts): error message when cdkVersion isn't specified

Chriscbr:
I've seen multiple users run into this error message when `cdkVersion` isn't specified for certain cdk project types:

```
.../node_modules/semver/classes/comparator.js:37
      throw new TypeError(`Invalid comparator: ${comp}`)
            ^
TypeError: Invalid comparator: ^undefined
```

cdkVersion is a required field, and it's automatically provided in .projenrc.js if you are generating from the CLI, but if you remove it (or you're writing a .projenrc.js from scratch), it's not clear why the project fails to synthesize unless your IDE happens to catch it.

To fix this, we add clauses that check for this value at runtime. A solution that automatically performs these checks at runtime would be more desirable, but this requires some discussion since the information about field optionality would have to be retrieved from somewhere, like e.g. the .jsii file.

---


eladb (approved):
We've been talking about adding runtime type checks in jsii (injecting code during compilation). 

Maybe that's worth bringing up again with @RomainMuller and @MrArnoldPalmer

PR #1125, opened 2021-10-07
feat: create projects programmatically

### LINKED ISSUE #1081: Programmatic project bootstrapping
In a projenrc.js file, creating a Project and then calling `project.synth()` will create the project in the appropriate directory, but normally it does not generate the projenrc file (since it is expected that the user is managing/creating this file).

We would like to add behavior so that running the project synthesis _does_ create a projenrc.js (or projenrc.java, etc.) file in the same directory as the project.

Currently it's possible to achieve this using some special flags that projen uses under the hood:

```ts
  __new__: {
    fqn: 'projen.TypeScriptProject',
    args: {
      outdir: '/tmp/outdir',
      defaultReleaseBranch: 'main',
      name: 'MyProject',
    },
    comments: false,
  },
  outdir: '/tmp/outdir',
  defaultReleaseBranch: 'main',
  name: 'MyProject',
});
project.synth();
```

But perhaps we could standardize this API like so:

```ts
  bootstrap: true,
  
  // (optional) flags provided in "projen new"
  bootstrapOptions: {
    rcFile: '.projenrc.js',
    synth: true,
    post: true,
    comments: false,
    git: false
  },

  outdir: '/tmp/outdir',
  defaultReleaseBranch: 'main',
  name: 'MyProject',
});
project.synth();
```

Chriscbr:
Closes #1081

---


src/cli/util.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as fs from 'fs-extra';
+import * as logging from '../logging';
+import { exec } from '../util';
+
+/**
+ * Installs the npm module (through `npm install`) to node_modules under `projectDir`.
+ * @param spec The npm package spec (e.g. `foo@^1.2` or `foo@/var/folders/8k/qcw0ls5pv_ph0000gn/T/projen-RYurCw/pkg.tgz`)
+ * @returns The installed package name (e.g. `@foo/bar`)
+ */
+export function installPackage(baseDir: string, spec: string): string {
```
Chriscbr, line 11: Moved some functions to a utility file so that I could access them for integration tests.

src/javascript/render-options.ts
```diff
@@ -33,6 +33,12 @@ export interface RenderProjectOptions {
    * @default false
    */
   readonly bootstrap?: boolean;
+
+  /**
+   * A list of fields to omit from the initial projenrc file.
+   * @default - none
+   */
+  readonly omitFromBootstrap?: string[];
```
Chriscbr, line 41: When we run `Projects.createProject()`, this option is used so that the "outdir" field can be omitted from the generated projenrc, since it makes the snapshot tests inconsistent. If users do want to include the "outdir" in the generate projenrc, we can revisit this later.

src/projects.ts
```diff
@@ -0,0 +1,105 @@
...
+   * Directory that the project will be generated in.
+   */
+  readonly dir: string;
+
+  /**
+   * Fully-qualified name of the project type (usually formatted
+   * as `module.ProjectType`).
+   * @example `projen.TypescriptProject`
+   */
+  readonly projectFqn: string;
```
Chriscbr, line 18: Replaced the "type" option with "projectFqn" in order to make the API more ergonomic.

src/projects.ts
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Should we call `project.synth()` or instantiate the project (could still
+   * have side-effects) and render the .projenrc file.
+   */
+  readonly synth: boolean;
+
+  /**
+   * Should we execute post synthesis hooks? (usually package manager install).
+   */
+  readonly post: boolean;
```
Chriscbr, line 40: Should we make these fields optional? (if so, what would the defaults be?) I'm open to changing this.

src/projects.ts
```diff
@@ -0,0 +1,105 @@
...
+   * Fully-qualified name of the project type (usually formatted
+   * as `module.ProjectType`).
+   * @example `projen.TypescriptProject`
+   */
+  readonly projectFqn: string;
+
+  /**
+   * Option values.
+   */
+  readonly params: Record<string, any>;
```
eladb, line 23:
```suggestion
  readonly projectOptions: Record<string, any>;
```

src/projects.ts
eladb: Yes, let's them optional and true by default.

src/projects.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Should we render commented-out default options in the projenrc file?
+   * Does not apply to projenrc.json files.
+   */
+  readonly comments: NewProjectOptionHints;
```
eladb, line 29: This should also be optional and default to FEATURED

eladb, line 29:
```suggestion
  readonly optionHints: NewProjectOptionHints;
```

src/projects.ts
```diff
@@ -0,0 +1,105 @@
...
+   * This function creates the project type in-process (with in VM) and calls
+   * `.synth()` on it (if `options.synth` is not `false`).
+   *
+   * At the moment, it also generates a `.projenrc.js` file with the same code
+   * that was just executed. In the future, this will also be done by the project
+   * type, so we can easily support multiple languages of projenrc.
+   */
+  public static createProject(options: CreateProjectOptions) {
+    createProject(options);
+  }
```
eladb, line 59: Add a private ctor

Chriscbr:
Something about how I'm doing tests is causing everything to break, still a little more work needed

eladb (approved):

PR #1139, opened 2021-10-15
fix: make release workflow path separators platform-invariant

### LINKED ISSUE #1138: Projen generates inconsistent release workflow path separators in Windows vs Github actions
When building an `AwsCdkConstructLibrary` with `releaseToNpm: true` on a Windows desktop, Projen generates a `.github/workflows/release.yml` with backwards slashes rather than forward slashes. 

Because of this, the first release will fail, as Projen changes the backward slashes to forward slashes during Projen's tamper check.

image

Instead of using Node's `path.join`, we could synthesize these workflow paths with a POSIX path separator.

misterjoshua:
Fixes #1138

---


misterjoshua:
@Chriscbr I've made your requested changes. Let me know if you need anything else. Thanks.

PR #1175, opened 2021-10-26
Used ts-eager instead of ts-node

AminFazlMondo:
https://github.com/bustle/ts-eager#ts-eager

---


eladb:
I believe we want type-checking so I am going to punt on this change for now.

PR #1184, opened 2021-10-29
fix: :bug: update the default CDK version to 1.129.0 

nikhil-zadoo:
The current cdk version has dependency issues with jest which fails with error "npm ERR!   peer jest@"^27.0.0" from ts-jest@27.0.7".
The later versions have fixed this dependency issue. So updating the default cdk version to the latest.

---


src/awscdk-app-ts.ts
```diff
@@ -23,7 +23,7 @@ export interface AwsCdkTypeScriptAppOptions extends TypeScriptProjectOptions {
   /**
    * AWS CDK version to use.
    *
-   * @default "1.95.2"
+   * @default "v1.129.0"
```
Chriscbr, line 26:
```suggestion
   * @default "1.129.0"
```

src/awscdk-construct.ts
```diff
@@ -8,7 +8,7 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
   /**
    * Minimum target version this library is tested against.
    *
-   * @default "1.95.2"
+   * @default "v1.129.0"
```
Chriscbr, line 11:
```suggestion
   * @default "1.129.0"
```

Chriscbr (changes requested):

src/awscdk-construct.ts
Chriscbr: @nikhil-zadoo It looks like this still needs to be updated

Chriscbr (approved):

nikhil-zadoo:
Can someone help with the failing build?
I am not sure what permissions are missing.

neilkuan:
hi @nikhil-zadoo 

you need to update the `test/__snapshots__/inventory.test.ts.snap` and commit again.
see build fail log...
```bash
Run git diff --exit-code || echo "::set-output name=has_changes::true"
diff --git a/test/__snapshots__/inventory.test.ts.snap b/test/__snapshots__/inventory.test.ts.snap
index 977fec6..ba3974e 100644
--- a/test/__snapshots__/inventory.test.ts.snap
+++ b/test/__snapshots__/inventory.test.ts.snap
@@ -3581,7 +3581,7 @@ Array [
         "switch": "cdk-test-dependencies",
       },
       Object {
-        "default": "\\"v1.129.0\\"",
+        "default": "\\"1.129.0\\"",
         "docs": "Minimum target version this library is tested against.",
         "featured": true,
         "fullType": Object {
```

PR #1188, opened 2021-10-30
feat: enforce conventional commit with mergify

### LINKED ISSUE #1186: Enforce conventional commit titles in PR titles
This is a kind of automated check that we have in the AWS CDK repo, and it's helpful for generating correctly formatted changelogs with `standard-version`. It can be implemented as a mergify check.

Naturally it should be possible to opt out of this.

Chriscbr:
Closes #1186

Adds a `conventionalCommit` boolean option to the `Mergify` component that will add a check that pull request titles follow [Conventional Commits](https://www.conventionalcommits.org). Enabled as true by default (but I'm open to changing this). Implementation is based on the example here: https://docs.mergify.com/actions/post_check/

This can be disabled by disabling mergify, or by disabling the feature specifically. For example:

```ts
  // ...
  githubOptions: {
    mergify: true, // default
    mergifyOptions: {
      conventionalCommits: false,
    },
  },
});
```

---


eladb (changes requested):
I am wondering if we should instead just install a workflow that enforces conventional commits? (e.g. https://github.com/marketplace/actions/conventional-changelog-action).

Generally, I'd like us to start migrating off of Mergify and move towards GitHub workflow/actions.

PR #1192, opened 2021-11-01
feat: support alwaysTryTypes config for typescript eslint import resolver

njlynch:
Without this setting, the eslint-import-resolver-typescript will not try to
resolve imports on the `@types/*` packages, leading to errors like the following
at build:

```
error  Unable to resolve path to module 'aws-lambda'  import/no-unresolved
```

By adding `{ alwaysTypeTypes: true }` to the typescript import resolver config,
the error is resolved.

---

### DIFF
--- a/src/eslint.ts
+++ b/src/eslint.ts
@@ -60,6 +60,11 @@ export interface EslintOptions {
    * @default undefined
    */
   readonly aliasExtensions?: string[];
+
+  /**
+   * Always try to resolve types under `<root>@types` directory even it doesn't contain any source code
+   */
+  readonly typescriptAlwaysTryTypes?: boolean;
 }
 
 /**
@@ -331,6 +336,7 @@ export class Eslint extends Component {
           node: {},
           typescript: {
             project: tsconfig,
+            ...( options.typescriptAlwaysTryTypes && { alwaysTryTypes: true } ),
           },
         },
       },
--- a/src/eslint.ts
+++ b/src/eslint.ts
@@ -62,9 +62,11 @@ export interface EslintOptions {
   readonly aliasExtensions?: string[];
 
   /**
-   * Always try to resolve types under `<root>@types` directory even it doesn't contain any source code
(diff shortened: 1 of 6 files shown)


src/eslint.ts
```diff
@@ -60,6 +60,11 @@ export interface EslintOptions {
    * @default undefined
    */
   readonly aliasExtensions?: string[];
+
+  /**
+   * Always try to resolve types under `<root>@types` directory even it doesn't contain any source code
+   */
+  readonly typescriptAlwaysTryTypes?: boolean;
```
eladb, line 67: Any reason this shouldn't be enabled by default?

(Add a `@default` docstring)

eladb (changes requested):

src/eslint.ts
```diff
@@ -60,6 +60,11 @@ export interface EslintOptions {
    * @default undefined
    */
   readonly aliasExtensions?: string[];
+
+  /**
+   * Always try to resolve types under `<root>@types` directory even it doesn't contain any source code
```
iliapolo, line 65: Can you add a sentence explaining why this is needed? what happens if this is disabled? 

Also mention the default value.

src/eslint.ts
iliapolo, line 67: I think we use `ts` throughout.

```suggestion
  readonly tsAlwaysTryTypes?: boolean;
```

iliapolo (changes requested):

src/eslint.ts
njlynch: None that I know of; I was just being conservative. Happy to default to `true`.

src/eslint.ts
njlynch: Done.

src/eslint.ts
njlynch: Done.

iliapolo (approved):

PR #1200, opened 2021-11-03
fix(node): bundled assets cannot be located during tests

eladb:
When test code is not executed from `lib/`, bundled assets cannot be located because they are under `lib/`. To address this we now place assets under a root `assets/` directory which has the same relative path whether code runs from `lib/` or `src/`.

BREAKING CHANGE: minor tweaks of the `LambdaFunction` and `Bundler` APIs.

---


docs/bundling.md
```diff
@@ -0,0 +1,30 @@
+# Bundling
+
+The `Bundler` component (for Node.js projects) can be used to produce JavaScript
+bundles from source files.
```
Chriscbr, line 4: I'm curious if is linked to Node.js just for ergonomics reasons / being opinionated, or if there is a technical limit?

Also (dumb question), is this only needed if I'm writing TypeScript code? Or does esbuild also bundle javascript files as an entrypoint too? The word "source files" is a bit ambiguous here

eladb: I guess it might be possible to use this for other project types but this specific bundler is a JavaScript bundler, so I figured it makes sense to couple this with Node.js projects.

src/awscdk/lambda-function.ts
```diff
@@ -21,13 +20,14 @@ export interface LambdaFunctionCommonOptions {
...
-   * @default - by default, the "aws-sdk" module will be excluded from the
-   * bundle. Note that if you use this option you will need to add "aws-sdk"
-   * explicitly.
+   * If not specified the default bundling options specified for the project
+   * `Bundler` instance will be used.
+   *
+   * @default - defaults
    */
-  readonly externals?: string[];
+  readonly bundlingOptions?: BundlingOptions;
```
iliapolo, line 30: Mark as breaking change

src/awscdk/lambda-function.ts
```diff
@@ -83,17 +78,11 @@ export interface LambdaFunctionOptions extends LambdaFunctionCommonOptions {
...
  *   srcdir: myProject.srcdir,
- *   libdir: myProject.libdir,
+ *   entrypoint: 'src/foo.lambda.ts',
  * });
  */
 export class LambdaFunction extends Component {
-  /**
-   * The bundle task for this function.
-   */
-  public readonly bundleTask: Task;
```
iliapolo, line 95: Breaking change

src/javascript/bundler.ts
```diff
@@ -164,7 +189,7 @@ export interface BundleOptions {
   /**
    * Include a source map in the bundle.
    *
-   * @default true
+   * @default false
```
iliapolo, line 192: Breaking change

src/awscdk/lambda-function.ts
```diff
@@ -110,37 +99,46 @@ export class LambdaFunction extends Component {
...
-      throw new Error(`${options.entrypoint} must be under ${options.srcdir}`);
-    }
-
     // allow Lambda handler code to import dev-deps since they are only needed
     // during bundling
     const eslint = Eslint.of(project);
     eslint?.allowDevDeps(options.entrypoint);
 
-    const entrypoint = relative(options.srcdir, options.entrypoint);
+    const entrypoint = options.entrypoint;
```
iliapolo, line 107: Feels like a good place to validate file existence

src/javascript/bundler.ts
```diff
@@ -77,72 +88,86 @@ export class Bundler extends Component {
    * `bundle:$name`).
    * @param options Bundling options
    */
-  public addBundle(name: string, options: BundleOptions): Task {
+  public addBundle(entrypoint: string, options: AddBundleOptions): Bundle {
```
iliapolo, line 91: Breaking change (return type)

iliapolo (approved):
Just a bunch of breaking changes that popped up. I'll leave you to decide which ones to explicitly mention.

Chriscbr (approved):

PR #1211, opened 2021-11-06
feat: allow customization of versionrc file

### LINKED ISSUE #1204: Allow configuring custom URL formats to versionrc to generate changelog to others than GitHub repos
Currently only standard kind of `versionrc.json` is generated (at [here](https://github.com/projen/projen/blob/856a33d45e7b3ab759d5522b46adf1747bb2abf6/src/release/bump-version.ts#L104-L123)). It would be nice to be able to configure URL formats for e.g. Atlassian stack (BitBucket + Jira). This would require possibility to configure following values:
- commitUrlFormat
- compareUrlFormat
- issueUrlFormat
- userUrlFormat

Example configuration [here](https://github.com/conventional-changelog/standard-version/issues/227#issuecomment-538532266).

More generic option would just allow adding e.g. `versionrc` object to projenrc, which then gets merged with the standard versionrc (overwriting values if given same than exist in standard one).

Hi-Fi:
allows to use changelog easily with repos not in public GitHub

fixes #1204

---

### DIFF
--- a/src/release/release.ts
+++ b/src/release/release.ts
@@ -145,6 +145,14 @@ export interface ReleaseProjectOptions {
    * @default - no prefix
    */
   readonly releaseTagPrefix?: string;
+
+  /**
+   * Custom configuration used when creating changelog with standard-version package.
+   * Given values either append to default configuration or overwrite values in it.
+   *
+   * @default - standard configuration applicable for GitHub repositories
+   */
+  readonly versionrcOptions?: Record<string, any>;
 }
 
 /**
@@ -242,6 +250,7 @@ export class Release extends Component {
     this.version = new Version(project, {
       versionInputFile: this.versionFile,
       artifactsDirectory: this.artifactsDirectory,
+      versionrcOptions: options.versionrcOptions,
     });
 
     this.publisher = new Publisher(project, {
--- a/src/version.ts
+++ b/src/version.ts
@@ -19,6 +19,11 @@ export interface VersionOptions {
    * are emitted.
    */
   readonly artifactsDirectory: string;
+
+  /**
+   * Custom configuration for versionrc file used by standard-release
+   */
+  readonly versionrcOptions?: Record<string, any>;
 }
 
 export class Version extends Component {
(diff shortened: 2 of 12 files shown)


Hi-Fi:
@eladb @Chriscbr Does this look valid?

src/release/release.ts
```diff
@@ -145,6 +145,14 @@ export interface ReleaseProjectOptions {
...
    */
   readonly releaseTagPrefix?: string;
+
+  /**
+   * Custom configuration used when creating changelog with standard-version package.
+   * Given values either append to default configuration or overwrite values in it.
+   *
+   * @default - standard configuration applicable for GitHub repositories
+   */
+  readonly versionrc?: object;
```
Chriscbr, line 155: I think based on the naming scheme we use in other projen APIs, `versionrcOptions` would be a more fitting name here (also leaves open the possibility of `versionrc` in the future to be a boolean enable/disable property, or a string property to rename the file, etc.). Also I think we can make the type slightly more specific so the API is friendlier in other JSII languages:
```suggestion
  readonly versionrcOptions?: Record<string, any>;
```

src/version.ts
```diff
@@ -19,6 +19,11 @@ export interface VersionOptions {
    * are emitted.
    */
   readonly artifactsDirectory: string;
+
+  /**
+   * Custom configuration for versionrc file used by standard-release
+   */
+  readonly versionrc?: object;
```
Chriscbr, line 26: Same changes here

Chriscbr (changes requested):
Thanks @Hi-Fi for the pull request! Exposing these options makes sense to me, I just have a small nitpick regarding the API naming / types. Otherwise looks good to me. 👍

src/release/release.ts
Hi-Fi: Made change through the PR to use `versionrcOptions` (or one in capital letters)

Hi-Fi:
> Thanks @Hi-Fi for the pull request! Exposing these options makes sense to me, I just have a small nitpick regarding the API naming / types. Otherwise looks good to me. 👍

I think those are not nitpicking, as naming is quite often the hard part and changing of arguments is a hassle to those who have started to use those.

Chriscbr (approved):

Chriscbr (approved):

PR #1224, opened 2021-11-13
feat: enforce conventional commits with PR linter

### LINKED ISSUE #1071: pr-linter support
I see that `aws-cdk` has a super cool custom linter to enforce http://conventionalcommits.org/ in the naming of PRs. I also see that https://github.com/marketplace/actions/pr-linter appears to be a less cool, but entirely reasonable general purpose solution for the same problem. I would love to have an easy way to include a pr-linter action for my cdk projects, and anything else which is applying conventional commits.

Chriscbr:
Introduces `pullRequestLint` and `pullRequestLintOptions` fields to `GitHub`, which provides a validation workflow that runs on all pull requests. By default, it has one linter `semanticTitle` that checks if PR titles follow [Conventional Commits](https://www.conventionalcommits.org/). This component is enabled by default, and can be disabled with `githubOptions: { pullRequestLint: false }`.

Supersedes #1188
Closes #1071 

---


src/github/semantic-pr.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Requires pull requests to follow Conventional Commits.
+ * @see https://www.conventionalcommits.org/
+ */
+export class SemanticPullRequest extends Component {
+  constructor(github: GitHub, options: SemanticPullRequestOptions = {}) {
+    super(github.project);
+
+    const types = options.types ?? [
```
eladb, line 31: Since the default is that only `fix` and `feat` are included in the change log, I rather we only allow `chore` as an additional type which basically indicates (not included in the change log).

So I think the default should be `[ 'feat', 'fix', 'chore' ]`.

src/github/semantic-pr.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @default false
+   */
+  readonly requireScope?: boolean;
+}
+
+/**
+ * Requires pull requests to follow Conventional Commits.
+ * @see https://www.conventionalcommits.org/
+ */
+export class SemanticPullRequest extends Component {
```
eladb, line 27: I am wondering if we should make this component/workflow a bit more general purpose so it can include additional PR linting jobs (for example, like @iliapolo's linter which validates that FEAT PRs always include README updates and FIX PRs always include a change to a test). Other linting could be to validate DCO, commit message, BREAKING CHANGES, etc.

So to that extend, maybe wan just call this `PullRequestLint` and the workflow `pull-request-lint.yml`).

And then initially we just include support for semantic titles, but we can add more stuff later.

eladb (changes requested):

src/github/semantic-pr.ts
Chriscbr: Yeah, I was unsure on the right level of abstraction to use here... the way you describe it makes a lot more sense. 😄 Right now the semantic title checking is performed by a GitHub action I found, but that should just be an implementation detail.

src/github/semantic-pr.ts
Chriscbr: Updated.

Chriscbr:
Updated the PR. I considered abbreviating the options as "prLint" or "prLintOptions" in some places, but leaned against it since I could easily imagine times where someone might type in "pull" and see no autocomplete results just because of this shorthand.

eladb (approved):

PR #1227, opened 2021-11-15
feat(publishing): allow assuming role for AWS CodeArtifact

AminFazlMondo:
Added ability to assume role prior to get authorization token from AWS CodeArtifact

---


src/node-package.ts
```diff
@@ -307,6 +307,14 @@ export interface CodeArtifactOptions {
...
     */
   readonly secretAccessKeySecret?: string;
+
+  /**
+    * ARN of AWS role to be assumed prior to get authorization token from AWS CodeArtifact
+    * This property must be specified only when publishing to AWS CodeArtifact (`registry` contains AWS CodeArtifact URL).
+    *
+    * @default "AWS_SECRET_ACCESS_KEY"
+    */
+  readonly roleToAssume?: string;
```
eladb, line 317: Should be under `CodeArtifactOptions`

eladb (changes requested):
See comment

src/node-package.ts
AminFazlMondo: Does it not need to be here in order to use when defining projects?

eladb: Users would define it via `codeArtifactOptions`

AminFazlMondo: Do you think the following code change is not necessary?
https://github.com/projen/projen/blob/9135e799b7451fd1ea48cf7c33398a86d38a6e9d/src/node-project.ts#L630

eladb (approved):
LGTM

PR #1228, opened 2021-11-15
feat(awscdk-construct): awscdk-construct used aws-cdk v2 by default

corymhall:
This updates the awscdk-construct to use cdk v2 by default instead of
v1.

Some of the things that are updated:

- `constructs@^10.0.5` & `aws-cdk-lib` are added as a devDependency in addition to
  peerDependency
- the `cdkAssert` property is deprecated and replaced with
  `cdkAssertions`

BREAKING CHANGE: cdkVersion default value updated to `2.0.0`
- `cdkAssert` deprecated and default changed to `false`
- new property `cdkAssertions` added and defaulted to `true`

---


src/awscdk-construct.ts
```diff
@@ -119,32 +131,49 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
+      const constructsMajorVersion = semver.minVersion(options.constructsVersion)?.major;
+      if (constructsMajorVersion !== undefined &&
+        constructsMajorVersion < 10 && cdkMajorVersion === 2) {
+        throw new Error('CDK 2.x requires constructs version >= 10');
+      }
       this.addPeerDeps(`constructs@^${options.constructsVersion}`);
       this.addDevDeps(`constructs@${options.constructsVersion}`);
     } else if (cdkMajorVersion === 1) {
+      if (options.cdkAssert) {
+        this.addDevDeps(this.formatModuleSpec('@aws-cdk/assert'));
```
eladb, line 149: addPeerDeps automatically adds a dev dependency, so I don't think this is needed

eladb (commented):
Nice!

src/awscdk-construct.ts
corymhall: It looks like we have been explicitly setting that to `false`

https://github.com/projen/projen/blob/dcfcfde305885778b804abe624f0f000b177a3de/src/awscdk-construct.ts#L127

eladb: Oh right because it is explicitly taken cared of in addCdkDependency(). 

I am wondering what's the utility of add addCdkDependency() for v2... maybe we should throw an error there?

corymhall: I think it is still used for the `-alpha` modules.

eladb: Yes you are right.

I do think that we can enable the pinned dev dependency option when using v2

corymhall: Ok i'll update that.

corymhall: Ok I reworked this a little. I kept the v1 functionality the same, but for v2 `pinDevDependency` will not be set so will default to `true`. Also, `cdkDependencies` will not apply for v2. You will just add any `alpha` modules as `peerDeps` and they will automatically get set as `devDeps` as well due to `pinDevDependency`.

src/awscdk-construct.ts
```diff
@@ -54,16 +54,30 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
...
+   *
    * @default true
    */
   readonly cdkDependenciesAsDeps?: boolean;
 
   /**
    * Install the @aws-cdk/assert library?
-   * @default true
+   *
+   * @depricated - use cdkAssertions instead
```
bilalq, line 66: Typo here. Won't surface the deprecation warning in tools looking for it unless the typo is corrected, I believe.

corymhall: Good catch!

corymhall:
@eladb I think this is ready for a (hopefully) final review.

src/awscdk/awscdk-construct.ts
```diff
@@ -110,40 +124,54 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
   constructor(options: AwsCdkConstructLibraryOptions) {
+    const cdkVersion = options.cdkVersionPinning ? options.cdkVersion : `^${options.cdkVersion}`;
+    const cdkMajorVersion = semver.minVersion(cdkVersion)?.major ?? 2;
     super({
       ...options,
-      peerDependencyOptions: {
+      peerDependencyOptions: cdkMajorVersion === 1 ? {
         pinnedDevDependency: false,
-      },
+      } : undefined,
```
eladb, line 133: Setting this explicitly to undefined means users can't configure this behavior via `options`. Maybe switch the order so options are applied after this default

src/awscdk/awscdk-construct.ts
```diff
@@ -110,40 +124,54 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
     if (!options.cdkVersion) {
       throw new Error('Required field cdkVersion is not specified.');
     }
+    this.cdkVersion = cdkVersion;
 
-    this.cdkVersion = options.cdkVersionPinning ? options.cdkVersion : `^${options.cdkVersion}`;
     this.cdkDependenciesAsDeps = options.cdkDependenciesAsDeps ?? true;
 
-    const cdkMajorVersion = semver.minVersion(this.cdkVersion)?.major ?? 1;
     if (options.constructsVersion) {
```
eladb, line 128: Can we turn this into a switch statement so top level conditions are all evaluated against the same value (cdkMajor)

src/awscdk/awscdk-construct.ts
```diff
@@ -42,7 +41,8 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
   /**
    * Which AWS CDK modules (those that start with "@aws-cdk/") does this library
    * require when consumed?
-   * @featured
+   *
+   * For CDK 2.x use 'peerDeps' instead
```
eladb, line 45: Is this enforced? (I'd expect an error if this is used with v2)

src/awscdk/awscdk-construct.ts
```diff
@@ -55,16 +55,30 @@ export interface AwsCdkConstructLibraryOptions extends ConstructLibraryOptions {
    * If this is disabled, `cdkDependencies` will be added to `devDependencies` to ensure
    * they are present during development.
    *
+   * Not used for CDK 2.x
```
eladb, line 58: Is this enforced?

eladb (changes requested):
What about setting the default to v2 for awscdk-app-xxx? Do we want to include it here or in a subsequent commit?

src/awscdk/awscdk-construct.ts
corymhall: Added error messages and deprecated the property.

src/awscdk/awscdk-construct.ts
corymhall: Added error messages and deprecated the property.

corymhall:
> What about setting the default to v2 for awscdk-app-xxx? Do we want to include it here or in a subsequent commit?

I'll create a separate PR for the app side of things.

eladb (approved):
Looks 👍

Chriscbr:
Awesome!! 🙌

PR #1237, opened 2021-11-18
fix: convert generated lambda paths to posix

### LINKED ISSUE #1234: bug: lambda autodiscover produces invalid paths on windows
Projen emits `lambda.Function` with Windows style path separators on Windows, but doesn't take into account that backslash tries to escape the first letter of each path part.

image

misterjoshua:
Fixes #1234
---


misterjoshua:
By building on Linux and linking on Windows, I was able to get this correct result:

image

I'm not sure how I can add tests to Projen to make sure this doesn't regress - perhaps in the future, we could look at adding a matrix build or some sort of matrix integration test.

src/awscdk/lambda-function.ts
```diff
@@ -205,4 +205,11 @@ export class LambdaRuntime {
...
     public readonly esbuildTarget: string) {
   }
+}
+
+/**
+ * Converts the given path string to posix if it wasn't already.
+ */
+function convertToPosixPath(p: string) {
+  return p.split(sep).join(posix.sep);
 }
```
Chriscbr, line 215: @misterjoshua Is it possible we could instead update the variable definitions earlier in the file to use `posix.join(...)` and `posix.relative(...)` etc.? Could you test if that works?

Chriscbr (changes requested):

src/awscdk/lambda-function.ts
misterjoshua: I'll see what I can do.

misterjoshua: @Chriscbr I've updated the variables earlier in the file.

This way of doing it seems harrier to me. The POSIX functions need more care to ensure they receive only POSIX paths, so I need to rely more heavily on `convertToPosixPath`. That said, my opinions here aren't strong as long as it works on Windows.

Here's a screenshot from Windows:

image

Chriscbr: @misterjoshua  Ah, I see... I think I was assuming that all our inputs (project.outdir, options.constructFile, options.entrypoint, and possibly even bundle.outfile) could be provided as POSIX paths, and we could just decide at render time whether to print paths as POSIX or Windows. Based on your code I'm guessing we can't assume that about the inputs here?

For now I'm OK with going back to the fix you provided when you first submitted this PR since it only touched a couple lines.

misterjoshua: > @misterjoshua Ah, I see... I think I was assuming that all our inputs (project.outdir, options.constructFile, options.entrypoint, and possibly even bundle.outfile) could be provided as POSIX paths, and we could just decide at render time whether to print paths as POSIX or Windows. Based on your code I'm guessing we can't assume that about the inputs here? For now I'm OK with going back to the fix you provided when you first submitted this PR since it only touched a couple lines.

@Chriscbr Project's outdir is definitely a native path and many components seem to depend on that, so I think changing that one from native to POSIX would be a pretty big breaking change for Windows users. AutoDiscover and Bundler also provide native paths to LambdaFunction, but I don't think changes there would be that impactful, so I think consistency is the only issue in those cases. If we're to be consistent with Project's `outdir`, I think it's best to assume input paths are native paths.

I agree that we should decide at render time which format is appropriate. LambdaFunction knows best here, and converting native paths to POSIX works fine in this case. So, I'll revert the latest commits and get us back to ec364c4.

Chriscbr (approved):

PR #1247, opened 2021-11-22
fix(awscdk-app-ts): unsetting context when cdk version > 2

### LINKED ISSUE #1205: AwsCdkTypeScriptApp should not populate project.cdkConfig.json.obj.context when cdkVersion >= 2
After upgrading to projen v0.32.5, I try running cdk and see 

```
Unsupported feature flag '@aws-cdk/core:enableStackNameDuplicates'. This flag existed on CDKv1 but has been removed in CDKv2. CDK will now behave as the same as when the flag is enabled.
```

We are using CDKv2. The offending line is in `cdk.json`. Workaround is

```
// Workaround projen breaking CDKv2 by adding feature flags.
// https://docs.aws.amazon.com/cdk/latest/guide/work-with-cdk-v2.html
project.cdkConfig.json.obj.context = {};
```

mbonig:
fixes #1205 

Setting context to undefined when CDK major version is >=2 

Only question: should it be undefined or an empty object?
---

### DIFF
--- a/src/awscdk-app-ts.ts
+++ b/src/awscdk-app-ts.ts
@@ -1,7 +1,7 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import * as semver from 'semver';
-import { CdkTasks, AutoDiscover, LambdaFunctionCommonOptions } from './awscdk';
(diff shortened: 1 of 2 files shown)


Chriscbr (approved):
Thanks! 🙏

Issue #1103, opened 2021-09-28
Escape hatches do not work for release.yml


Chriscbr:
Escape hatches work as usual for most GitHub workflow files:

```js
const workflow = project.tryFindObjectFile('.github/workflows/build.yml');
workflow.addOverride('env.SERVER', 'production');
```

But `tryFindObjectFile` returns undefined for `.github/workflows/release.yml` because it gets created during preSynthesis of the `Release` component. We should find a way to generate the workflow using token values / lazy values so that the functionality works normally.

We should also think of ways to prevent this kind of error in the future, perhaps by limiting the kinds of actions that can be performed during pre/post synthesis (?), or by using some kind of linting rule, etc.

PR #1250, opened 2021-11-23
feat(upgrade): allowing to prepand commands to upgrade via upgrade options

AminFazlMondo:
---

### DIFF
--- a/API.md
+++ b/API.md
@@ -167,6 +167,7 @@ Name|Description
 [SampleFileOptions](#projen-samplefileoptions)|Options for the SampleFile object.
 [SampleReadmeProps](#projen-samplereadmeprops)|SampleReadme Properties.
 [SourceCodeOptions](#projen-sourcecodeoptions)|Options for `SourceCodeFile`.
+[TaskCommandStep](#projen-taskcommandstep)|Task command step definition.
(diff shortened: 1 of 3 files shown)


eladb (changes requested):
why is this needed?

Can't you just use .prependExec() on the upgrade task?

AminFazlMondo:
Thanks @eladb ,
I wanted to use this in one of my [projects](https://github.com/AminFazlMondo/Lerna-Projen), exactly same as what you describe above.
But the hrere is what I get
```
const project = new NodeProject(....)
let upgradeTask = this.tasks.tryFind('upgrade')
// This return undefined
project.synth()
upgradeTask = this.tasks.tryFind('upgrade')
// This returns the task
upgradeTask?.prependExec(....)
```
But after synth the changes to the task does not get reflected. Would you be able to help me if I am making any mistake?

Chriscbr:
@AminFazlMondo You cannot modify tasks after `project.synth()` is called.

> Can't you just use .prependExec() on the upgrade task?

@eladb It looks like the issue is that `upgrade-dependencies` creates its task during `preSynthesize`... definitely a code smell. Same issue as #1103

AminFazlMondo:
Thanks @Chriscbr

PR #1255, opened 2021-11-25
fix: release.yml cannot be modified with escape hatches

### LINKED ISSUE #1103: Escape hatches do not work for release.yml

Chriscbr:
Fixes #1103

Resolves a bug where the workflows generated by the `Release` component cannot be accessed or modified using escape hatches because they are generated during the `preSynthesize()` stage of the component lifecycle.

This requires making some changes to `Publisher`, which `Release` relies upon, so that it can provide a link to the jobs it will generate even if `Publisher`'s jobs are not all available at the time construction finishes.

In addition, workflows that are not created via the `github.addWorkflow` method were not getting added to `GitHub` so they could not be discovered with `tryFindWorkflow`. This is fixed by adding a register hook within the `GithubWorkflow` constructor.

---


src/github/workflows.ts
```diff
@@ -46,16 +46,20 @@ export class GithubWorkflow extends Component {
...
 
   private events: workflows.Triggers = { };
   private jobs: Record<string, workflows.Job> = { };
+  private _providers = new Array<IJobProvider>();
 
   constructor(github: GitHub, name: string, options: GithubWorkflowOptions = {}) {
     super(github.project);
 
+    // register to catch duplicate workflow name errors early
+    github._registerWorkflow(name, this);
```
Chriscbr, line 61: An alternative to tracking workflows in `GitHub` is would be to do a search for all components that are `instanceof GithubWorkflow` whenever someone accesses `github.workflows` or calls `github.tryFindWorkflow()`. I don't have a strong opinion on which way to do this / I'm open to trying that instead if this `_registerWorkflow` hook is undesirable

src/release/publisher.ts
```diff
@@ -67,7 +67,7 @@ export interface PublisherOptions {
  *
  * Under the hood, it uses https://github.com/aws/jsii-release
  */
-export class Publisher extends Component {
+export class Publisher extends Component implements IJobProvider {
```
Chriscbr, line 70: Instead of having `Publisher` implement `IJobProvider`, I could create a separate component that is generated by `Publisher` to isolate the behavior

Chriscbr:
I'm not sure I love this solution, though it does work. Open to other ideas of course. :)

It's worth mentioning this bug also exists in `UpgradeDependencies` -- that component generates resources at pre-synth time because it requires a list of all dependenciess and a list of all `Release` branches. So neither the workflow nor task generated by it are can be escape hatched around.

src/github/workflows.ts
eladb: Might be more idiomatic to look up than to register. I think this is how findFile works.

src/github/workflows.ts
```diff
@@ -100,12 +104,28 @@ export class GithubWorkflow extends Component {
     };
   }
 
+  /** @internal */
+  public _addJobsFromProvider(provider: IJobProvider) {
```
eladb, line 108: Add docs

src/github/workflows.ts
```diff
@@ -208,3 +228,10 @@ function arrayOrScalar<T>(arr: T[] | undefined): T | T[] | undefined {
   }
   return arr;
 }
+
+export interface IJobProvider {
+  /**
+   * A collection of jobs that may be dynamically generated.
+   */
+  readonly jobs: Record<string, workflows.Job>;
```
eladb, line 236: I'd use a method instead of a property here to signal that this can be late bound.

Something like `renderJobs()`

src/github/workflows.ts
eladb, line 108: I'd make this public can all it `addJobsLater()`

eladb (changes requested):
Generally like the design. A few semantic comments

src/github/workflows.ts
Chriscbr: Done

Chriscbr: Good suggestion. I like how colloquial the name sounds

src/github/github.ts
```diff
@@ -107,6 +101,12 @@ export class GitHub extends Component {
...
    */
   public tryFindWorkflow(name: string): undefined | GithubWorkflow {
-    return this._workflows[name];
+    for (const workflow of this.workflows) {
+      if (workflow.name === name) {
+        return workflow;
+      }
+    }
+
+    return undefined;
```
eladb, line 110:
```suggestion
    return this.workflows.find(w => w.name === name);
```

eladb (approved):

PR #1270, opened 2021-11-30
feat(awscdk): integration tests

eladb:
Add support for CDK integration tests.

To use, create a file with an `xyz.integ.ts` suffix under the test tree and run `projen integ:xyz:deploy` to deploy it to your development account and capture a snapshot of the synthesized output. During build, as part of the `test` phase, the snapshot will be verified.

BREAKING CHANGE: `AutoDiscover` now requires `testdir`.

---


Chriscbr (approved):
Looks great! Definitely should be possible to try this out on construct hub

PR #1276, opened 2021-12-01
fix: error in IntegrationTest is not thrown correctly

Chriscbr:
`getDependency` throws an error if the dependency isn't found, so the "ts-node dependency is required for integration tests" error message never appears to the user.

Also updated the docs since it wasn't apparent that `AutoDiscover` was also auto-discovering integration tests.

---

### DIFF
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -69,9 +69,7 @@ export class IntegrationTest extends Component {
 
     const app = `ts-node ${entry}`;
 
-    try {
-      project.deps.getDependency('ts-node');
-    } catch (e) {
+    if (!project.deps.tryGetDependency('ts-node')) {
       throw new Error('ts-node dependency is required for integration tests.');
     }
 
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -69,7 +69,9 @@ export class IntegrationTest extends Component {
 
     const app = `ts-node ${entry}`;
 
-    if (!project.deps.getDependency('ts-node')) {
+    try {
+      project.deps.getDependency('ts-node');
+    } catch (e) {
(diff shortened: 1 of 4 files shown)


src/awscdk/integration-test.ts
```diff
@@ -69,7 +69,9 @@ export class IntegrationTest extends Component {
 
     const app = `ts-node ${entry}`;
 
-    if (!project.deps.getDependency('ts-node')) {
```
eladb, line 72: Can we add `tryGetDependency()`?

eladb (changes requested):

src/awscdk/integration-test.ts
Chriscbr: Done!

src/deps/dependencies.ts
```diff
@@ -95,6 +95,27 @@ export class Dependencies extends Component {
...
+   * Returns undefined if the dependency does not exist with that name OR
+   * there is more than one dependency type for this dependency.
+   *
+   * @param name The name of the dependency
+   * @param type The dependency type. If this dependency is defined only for a
+   * single type, this argument can be omitted.
+   *
+   * @returns a copy (cannot be modified)
+   */
+  public tryGetDependency(name: string, type?: DependencyType): Dependency | undefined {
```
eladb, line 110: Try to avoid catching exceptions in happy flows since they could be thrown for many different reasons. 

Implement `getDependency()` by means of `tryGetDepencency()` instead of the other way.

eladb (changes requested):

src/deps/dependencies.ts
Chriscbr: Ahh yeah this makes sense. Updated.

Chriscbr:
Superseded by #1302

PR #1282, opened 2021-12-02
feat: allow customising or disabling git-push used by manual release

### LINKED ISSUE #1267: Make publish:git configurable?
My project is using `AwsCdkConstructLibrary` with `releaseTrigger: ReleaseTrigger.manual()` set. When running `yarn build` Projen automatically creates a commit, tags it and pushes it to my remote. I'd like more control over this. 

I'm wanting to run a pipeline in Gitlab which runs `yarn release` whenever a commit is pushed to my **main** branch. As `yarn release` pushes automatically I need to be able to tell Gitlab to NOT start another pipeline when this push occurs. Gitlab supports this with the option `-o ci.skip`

Currently publish:git does this:

```
git push --follow-tags origin main
```

I'd like it to do this:

```
git push --follow-tags -o ci.skip origin main
```

A way to override the git command would probably suffice?

willdady:
This adds the ability for a user to customise (or disable) the git-push command which is run as part of a manual release.

closes #1267 

---

### DIFF
--- a/src/release/publisher.ts
+++ b/src/release/publisher.ts
@@ -128,7 +128,7 @@ export class Publisher extends Component implements IJobProvider {
       publishTask.builtin('release/update-changelog');
     }
     publishTask.builtin('release/tag-version');
-    
+
     if (options.gitPushCommand !== '') {
       const gitPushCommand = options.gitPushCommand || `git push --follow-tags origin ${gitBranch}`;
       publishTask.exec(gitPushCommand);
@@ -757,10 +757,10 @@ export interface GitPublishOptions extends VersionArtifactOptions {
    * @default "main"
    */
   readonly gitBranch?: string;
-  
+
   /**
-   * Override git-push command. 
-   * 
+   * Override git-push command.
+   *
    * Set to an empty string to disable pushing.
    */
   readonly gitPushCommand?: string;
--- a/src/release/publisher.ts
+++ b/src/release/publisher.ts
@@ -129,7 +129,7 @@ export class Publisher extends Component implements IJobProvider {
     }
     publishTask.builtin('release/tag-version');
     
-    if (options.gitPushCommand !== null) {
+    if (options.gitPushCommand !== '') {
       const gitPushCommand = options.gitPushCommand || `git push --follow-tags origin ${gitBranch}`;
       publishTask.exec(gitPushCommand);
     }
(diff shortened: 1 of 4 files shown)


src/release/publisher.ts
```diff
@@ -753,4 +757,11 @@ export interface GitPublishOptions extends VersionArtifactOptions {
    * @default "main"
    */
   readonly gitBranch?: string;
+  
+  /**
+   * Override git-push command. 
+   * 
+   * Set to null to disable pushing.
+   */
+  readonly gitPushCommand?: string | null;
```
Chriscbr, line 766:
```suggestion
  readonly gitPushCommand?: string;
```

Projen is compiled with JSII to other languages where `undefined` and `null` will be treated as equivalent, so distinguishing between them may lead to errors.

How about we instead add a line in the docstring that says e.g. `To disable pushing, use a no-op command such as "true"`?

Chriscbr (changes requested):
Thanks for submitting this PR! Just one requested API change below.

src/release/publisher.ts
willdady: Thanks @Chriscbr. How about, `To disable pushing, set to an empty string`? 

I feel `gitPushCommand: true` reads a little weirdly for something which is disabling something.

Chriscbr: Fair point. I suppose the alternative here is to add another field. This solution sounds good to me. 🙂

Chriscbr (approved):

Chriscbr:
From the build error it looks like PR code just needs to be linted and the docs updated (`yarn eslint` and `yarn docgen` probably)

willdady:
Build is failing as it's trying to do a git-push to _my_ fork?

Chriscbr:
The actual error is in the build step right before (this is a bit confusing). It should be possible to fix this by merging the latest changes from `main` and re-building

willdady:
> The actual error is in the build step right before (this is a bit confusing). It should be possible to fix this by merging the latest changes from `main` and re-building

@Chriscbr Thanks. I've just done this now.

Chriscbr:
Sorry for the confusion - when I said it will be fixed by "re-building" I meant running a build on your local machine, and then committing any of the changes that it makes (it may update test snapshots or docs) to your branch. If you look at the diff, you can see it incorrectly includes changes to "cdk8sPlusVersion" in API.md that are unrelated to your PR.

PR #1293, opened 2021-12-06
fix: passing optional gitPushCommand to publishToGit method

willdady:
Custom git push command was not actually being passed through to `publishToGit` method. This should have been part of #1282.

---


eladb (changes requested):
Can you please add a test?

Chriscbr (approved):

PR #1296, opened 2021-12-06
feat: support specifiying alternative github runner

### LINKED ISSUE #1295: Cannot use Github self-hosted runners
Generated github workflows hardcode the runner selection (`runs-on`) to `ubuntu-latest` with no option to select an alternative.

Consequently, it prevents the use of self-hosted runners which are mandated in certain organizations.

msessa:
Fixes #1295

Introduce the ability to override the default github runner selection in various parts of projen that will provision workflows

---


src/release/release.ts
```diff
@@ -151,7 +151,14 @@ export interface ReleaseProjectOptions {
    *
    * @default - standard configuration applicable for GitHub repositories
    */
+
```
Chriscbr, line 154: Please undo

src/release/release.ts
```diff
@@ -151,7 +151,14 @@ export interface ReleaseProjectOptions {
...
   readonly versionrcOptions?: Record<string, any>;
+
+  /**
+   * Github Runner selection label(s)
+   * '@default "ubuntu-latest"
```
Chriscbr, line 159:
```suggestion
   * @default "ubuntu-latest"
```

src/github/auto-approve.ts
```diff
@@ -31,6 +31,12 @@ export interface AutoApproveOptions {
    * `NodeProjectOptions`, then you must use a different token here.
    */
   readonly secret: string;
+
+  /**
+   * Github Runner selection label(s)
+   * @default "ubuntu-latest"
+   */
+  readonly runsOn?: string | string[];
```
Chriscbr, line 39: Since we are using JSII to compile projen to other languages, any TypeScript type unions will end up converted into "object" or similar equivalents in other languages (see https://aws.github.io/jsii/specification/2-type-system/#type-unions) - so we'd like to avoid this if possible.

Could we instead just use `string[]` and coerce the value to a string if the array is length 1 (and add some error handling if the array is empty)? I'd also be okay with typing it as `string` and allowing a comma-delimited list.

Chriscbr (changes requested):
Looks great @msessa ! Thanks for adding the unit tests as well. Just a few comments.

msessa:
@Chriscbr Thanks for the review!

I've addressed your comments. 
In regards to the type union:
Since the runner selection is mostly a new option for the user, I've opted to stick to string-list as it most closely reflects the expected type in the workflow yaml and github's strict ordering expectations ( `self-hosted` **must** be the first entry followed by additional selectors eg: `ubuntu-18.04` )

in order to avoid large amounts of changes in the test snapshots, at rendering time the value is coerced to a scalar when only one item is passed in the list.
Additionally an error might be thrown if an empty list is provided.

Chriscbr (approved):

PR #1303, opened 2021-12-07
fix: remove gitattributes spacing alignment

### LINKED ISSUE #1085: Add minimum spacing to .gitattributes file
For many projects, the `.gitattributes` file generated by projen may list dozens of glob patterns. When a change is made to add a new pattern, if the pattern is longer than all previous patterns, then the generated .gitattributes file will adjust the indentation of *all* the patterns on the page, making it hard to tell when reviewing the diff what patterns were actually added or changed.

One way to improve this experience would be to apply a minimum amount of spacing (perhaps 80 or 100 characters) so that, as long as glob patterns are below that length, the same spacing would be applied.

Chriscbr:
Fixes #1085

---


eladb (approved):
Thank you thank you thank you

PR #1306, opened 2021-12-07
fix(cleanup): file pruning is too heavy handed

### LINKED ISSUE #1265: Feature Request: Cache list of projen-managed files
When projen prunes it has to go through files to find the marker. If generated files were instead written into some sort of files.json in the .projen file, pruning could instead read from that list (which I guess would also contain itself). This provides the obvious efficiency bonus as well as the added benefit of a simple way to inspect what files are being managed.

In the case that some files get orphaned somehow, a cli/project arg can be added to do a full cleanup like the current behavior. This would have to be done manually, but I think it would make sense to optimize the more common case that files are being properly tracked between projen invocations.

MarkMcCulloh:
Added a files.json in the projen folder to store a list of the files generated by projen on the last run. Using this file, projen can selectively clean up files on subsequent runs.

With this implementation, generated files are cleaned up regardless of whether or not they have the `PROJEN_MARKER` in them. A file is considered a candidate for deletion if it is marked as readonly and is in the files array of a project.
If files.json is not present or otherwise invalid, the previous method of cleanup is used.

fixes #1265

---


src/cleanup.ts
```diff
@@ -1,16 +1,27 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
+import { existsSync, readFileSync } from 'fs-extra';
 import * as glob from 'glob';
 import { PROJEN_MARKER } from './common';
 import * as logging from './logging';
 
+export const FILE_MANIFEST = '.projen/files.json';
```
eladb, line 8: There's a const for `.projen` somewhere :-)

src/cleanup.ts
```diff
@@ -1,16 +1,27 @@
...
+
 export function cleanup(dir: string, exclude: string[]) {
-  try {
-    for (const f of findGeneratedFiles(dir, exclude)) {
-      fs.removeSync(f);
+  const fileManifestPath = path.resolve(dir, FILE_MANIFEST);
+  if (existsSync(fileManifestPath)) {
+    const fileFile = JSON.parse(readFileSync(fileManifestPath, 'utf-8'));
+    fileFile.files.forEach((file: string) => {
```
eladb, line 14: style: please use a normal `for` loop instead of `.forEach()`

src/cleanup.ts
```diff
@@ -1,16 +1,27 @@
...
 
+export const FILE_MANIFEST = '.projen/files.json';
+
 export function cleanup(dir: string, exclude: string[]) {
-  try {
-    for (const f of findGeneratedFiles(dir, exclude)) {
-      fs.removeSync(f);
+  const fileManifestPath = path.resolve(dir, FILE_MANIFEST);
+  if (existsSync(fileManifestPath)) {
+    const fileFile = JSON.parse(readFileSync(fileManifestPath, 'utf-8'));
```
eladb, line 13: Handle the case when this file is corrupted (fall back to previous behavior)

src/project.ts
```diff
@@ -206,6 +206,13 @@ export class Project {
     if (projenrcJson) {
       new Projenrc(this, options.projenrcJsonOptions);
     }
+
+    new JsonFile(this, FILE_MANIFEST, {
+      omitEmpty: true,
+      obj: () => ({
+        files: this.files.filter(f => f.readonly).map(f => f.path.replace(/\\/g, '/')),
```
eladb, line 213: why do we need the `replace()` here?

src/cleanup.ts
```diff
@@ -1,16 +1,27 @@
...
+    fileFile.files.forEach((file: string) => {
+      fs.removeSync(path.resolve(dir, file));
+    });
+  } else {
+    try {
+      for (const f of findGeneratedFiles(dir, exclude)) {
```
eladb, line 19: Can you please add a unit test?

eladb (changes requested):

eladb:
@MarkMcCulloh looks great. I've updated the title to reflect the problem this fix is solving. 
Also, please add details to your PR description about the "how" this is is implemented.

src/project.ts
MarkMcCulloh: Forcing posix paths for windows users to prevent any issues there

MarkMcCulloh: If this is not done it would be possible for discrepancies between window/non-windows users in this file. Seems to be worth avoiding that.

src/util/synth.ts
```diff
@@ -49,6 +49,14 @@ export interface DirectorySnapshotOptions {
...
    */
   readonly excludeGlobs?: string[];
+
+  /**
+   * Only read the names of files and not the contents.
+   * Value of `[path]` will be `true` if it exists.
+   *
+   * @default false include file content
+   */
+  readonly excludeContent?: boolean;
```
Chriscbr, line 59: optional: The type signature is different from "excludeGlobs" so it might make sense to name it differently
```suggestion
  /**
   * Only snapshot the names of files and not their contents.
   * The value for a path will be `true` if it exists.
   *
   * @default false include file content
   */
  readonly omitContents?: boolean;
```

src/cleanup.ts
```diff
@@ -1,10 +1,30 @@
...
+        for (const file of fileManifest.files) {
+          fs.removeSync(path.resolve(dir, file));
+        }
+
+        return;
+      }
+    }
+  } catch (e) {
+    logging.warn(`warning: failed to clean up generated files using file manifest: ${e.stack}`);
+  }
```
Chriscbr, line 25: It might simplify the control flow / improve readability to break when conditions fail instead of nesting the happy cases together. e.g.

```ts
const fileManifestPath = path.resolve(dir, FILE_MANIFEST);
if (!existsSync(fileManifestPath)) {
  return legacyCleanup(...);
}

const fileManifest = JSON.parse(readFileSync(fileManifestPath, 'utf-8'));
if (!fileManifest.files) {
  return legacyCleanup(...);
}

// actual work
```

test/cleanup.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const postDirSnapshot = directorySnapshot(p.outdir, { excludeContent: true });
+  const postFiles = Object.keys(postDirSnapshot);
+
+  const deletedFiles = preFiles.filter(f => !postFiles.includes(f));
+
+  // THEN
+  expect(deletedFiles).toEqual(fileList);
+  expect(deletedFiles).toMatchSnapshot();
+});
```
Chriscbr, line 31: Perhaps we should explicitly validate here that foo/bar.txt was NOT deleted even though it does not have the projen marker in it - since I believe this is new behavior?

MarkMcCulloh: There is a new behavior, but it's the opposite of that: `foo/bar.txt` IS deleted even though it does not have the projen marker in it. I updated the test with some more assertions as well as the PR description to be more clear.

src/cleanup.ts
MarkMcCulloh: I made some changes and I think the flow is more clear although it is different than you suggested. Let me know if it could still be better, I'm happy to oblige 😁

src/project.ts
eladb: k. add a comment explaining that in the code plz (isn't there a better way to do it? `path.posix` or something?)

src/util/synth.ts
```diff
@@ -49,6 +49,14 @@ export interface DirectorySnapshotOptions {
...
    */
   readonly excludeGlobs?: string[];
+
+  /**
+   * Only snapshot the names of files and not their contents.
+   * The value for a path will be `true` if it exists.
+   *
+   * @default false include file content
+   */
+  readonly omitContents?: boolean;
```
eladb, line 59: Nit: always use "positive" names (avoid "omit", "disable", etc). So I'd call this something like `includeContents` (with `true` as default) or `onlyFileNames` (with `false as a default).

eladb (changes requested):
Please resolve all remaining conversation

src/project.ts
MarkMcCulloh: These paths may or may not be fully user-controlled, so unfortunately afaik this is the only place this logic could be to catch this.
I do wish `path` had a helper for this.

eladb (approved):

PR #1314, opened 2021-12-09
feat(release): allow specifying npm dist-tag per branch

eladb:
Adds `npmDistTag` to release branch definition which will be configured in the release workflow triggered for this branch.

Implementation notes: to support this we needed to refactor the publisher so that when its jobs are rendered, they have the branch context. This also means that now we can ensire that when a publish command is executed manually, it will be done only on the correct release branch.

The root `npmDistTag` setting applies to the default branch.

BREAKING CHANGE: `npmDistTag` only effects publishing (and supported for each release branch) setting and so it is no longer set in `package.json`.
* `workflow.addJobsLater()` is no longer supported. Use `postSynthesis` hooks and call `.addJobs()` as needed.
* `publish:xxxx` tasks for automated releases are not defined by default since normally they should only be executed from within workflows. Set `publishingTasks: true` to create them.

---


eladb:
Copy: @njlynch -- this is for `constructs`.

test/release/release.test.ts
```diff
@@ -500,3 +522,117 @@ test('publisher can use custom github runner', () => {
...
+describe('npmDistTag', () => {
+  test('determines npm dist-tag used in the workflow', () => {
+    // GIVEN
+    const project = new TestProject();
+
+    // WHEN
+    const release = new Release(project, {
+      releaseBranches: {
+        'main-3': { majorVersion: 3, npmDistTag: 'latest-3' },
+      },
```
Chriscbr, line 535: I assume the outputs are the same if the branch is specified through `addBranch()`? I think this is probably caught by other test cases but just double checking

Chriscbr (approved):
Overall strategy looks good to me. I wasn't a big fan of `addJobsLater()` either 🙃

src/release/publisher.ts
```diff
@@ -315,91 +352,104 @@ export class Publisher extends Component implements IJobProvider {
...
-    });
+    }));
   }
 
-  private addPublishJob(opts: PublishJobOptions) {
-    const jobname = `release_${opts.name}`;
-    if (jobname in this._jobs) {
-      throw new Error(`Duplicate job with name "${jobname}"`);
-    }
+  private addPublishJob(factory: (branch: string, branchOptions: Partial<BranchOptions>) => PublishJobOptions) {
```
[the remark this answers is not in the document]
eladb: Meh... As long as it's internal

test/release/release.test.ts
eladb: Everything going through `addBranch()` eventually so it's covered.

eladb: I would be nice if the main branch and the "other" branches are all defined under `releaseBranches` but I got a bit tangled with the defaults... I'll revisit sometime.

PR #1320, opened 2021-12-10
fix: codecov should run on main branch to generate baseline

ahammond:
Per https://docs.codecov.com/docs/error-reference
"In order to produce an accurate coverage comparison,
Codecov must have the coverage report for the pull request base."

Users who are using squash commits can't re-use the last uploaded
coverage from the previously merged feature-branch as the base.

---

### DIFF
--- a/src/javascript/node-project.ts
+++ b/src/javascript/node-project.ts
@@ -594,12 +594,28 @@ export class NodeProject extends GitHubProject {
     if (release) {
       this.addDevDeps(Version.STANDARD_VERSION);
 
+      // run codecov if enabled or a secret token name is passed in
+      // AND jest must be configured
+      const postBuildSteps = options.postBuildSteps ?? [];
+      if ((options.codeCov || options.codeCovTokenSecret) && this.jest?.config) {
+        postBuildSteps.unshift({
+          name: 'Upload coverage to Codecov',
+          uses: 'codecov/codecov-action@v1',
+          with: options.codeCovTokenSecret ? {
+            token: `\${{ secrets.${options.codeCovTokenSecret} }}`,
+            directory: this.jest.config.coverageDirectory,
+          } : {
+            directory: this.jest.config.coverageDirectory,
+          },
+        });
+      }
+
       this.release = new Release(this, {
         versionFile: 'package.json', // this is where "version" is set after bump
         task: this.buildTask,
         branch: options.defaultReleaseBranch ?? 'main',
         ...options,
-
+        postBuildSteps,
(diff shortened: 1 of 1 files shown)


Chriscbr (changes requested):
Please add a unit test

src/javascript/node-project.ts
```diff
@@ -594,12 +594,28 @@ export class NodeProject extends GitHubProject {
     if (release) {
       this.addDevDeps(Version.STANDARD_VERSION);
 
+      // run codecov if enabled or a secret token name is passed in
```
eladb, line 597: Why do we need to add this to the release workflow? Can we just create a separate workflow to keep the code coverage baseline?


Wondering if this might be a good opportunity to extract a `CodeCov` component which will take care of adding itself to the workflows.

eladb (changes requested):

src/javascript/node-project.ts
ahammond: Ooooooh, that'd be cool. The only reason I figured to pack it in here is that tests have already run and there's coverage data free for the taking.

ahammond:
Should I add a UT for this or make a separate workflow?

src/javascript/node-project.ts
eladb: Let's create a `CodeCov` component and bake this knowledge in there (and also move the stuff from the build workflow in)

eladb:
## Please read this to avoid merge hell

Hola! apologies for hassle. I've enabled Prettify as part of #1453. This means that most of the source files have changed and you will likely see quite a lot of merge conflicts when you try to merge from `main`.

*But there's an easy way to handle this.*

First, merge your branch from this tag:

```sh
$ git fetch origin
$ git merge 1020d0ad3f95bb4e1201f8640d440ffb43ccf230
```

Edit the `.projenrc.js` file in your branch and add this to **line 57** (right
after `codeCov: true`):

```js
  prettier: true,
```

Now, run:

```sh
$ git clean -fdx .
$ yarn
$ bash projen.bash
$ yarn eslint
$ git add .
$ git commit -m 'format code'
```

This will reformat all the code in your branch to match `main`.

Now, merge from main:

```sh
$ git merge origin/main
```

Sorry again for the inconvenience...

src/javascript/node-project.ts
ahammond: @eladb is there any guidance on how to do this (create a CodeCov component)? I'm willing to do some work on this, but I'm out of my depth. Is there an example component I could pattern this after?

PR #1323, opened 2021-12-12
feat(awscdk-app-ts): support cdk watch

eladb:
`awscdk-app-ts` projects now support a new `watch` task which will deploy your app and start monitoring source files for changes.

When changes are identified, bundles will be recreated and a "hotswap deployment" will be performed.

---


src/awscdk/cdk-tasks.ts
```diff
@@ -26,12 +26,17 @@ export class CdkTasks extends Component {
...
+   */
+  public readonly watch: Task;
+
   constructor(project: Project) {
     super(project);
 
     this.synth = project.addTask('synth', {
       description: 'Synthesizes your cdk app into cdk.out (part of "yarn build")',
-      exec: 'cdk synth',
+      exec: 'cdk synth > /dev/null', // redirect to /dev/null to hide template
```
Chriscbr, line 39: `cdk synth` by default outputs contents to both stdout and cdk.out? Never occurred to me until now how weird that is

test/tasks/tasks.test.ts
```diff
@@ -356,6 +356,21 @@ test('lock() can be used to disallow modifications', () => {
   expect(() => t.prependSay('hoho')).toThrow(error);
 });
 
+test('it is possible to edit the descriptio', () => {
```
Chriscbr, line 359:
```suggestion
test('it is possible to edit the description', () => {
```

Chriscbr (approved):

src/awscdk/cdk-tasks.ts
eladb: It's a bit of an odd behavior. Only if the app has a single stack then it will also emit it to stdout

PR #1327, opened 2021-12-13
fix(awscdk): some issues with CDK v2

### LINKED ISSUE #1173: 1-buster-slim breaks workflows using cdk v2.0.0-rc.25
My PR's build is failing with
```
error @aws-cdk/assert@2.0.0-rc.25: The engine "node" is incompatible with this module. Expected version ">= 14.15.0". Got "12.22.7"
```
while trying to upgrade All The Things in my corp `cdk-library`. 
```
❯ npx projen --version
0.27.51
```
One of the things I see is that the upgrade swapped:
```js
      image: jsii/superchain:node14
      image: jsii/superchain:1-buster-slim
```
in the workflows. I don't know what `1-buster-slim` is, but... it looks like it's breaking stuff.

Workaround in `.projenrc.js`:
```js
  workflowContainerImage: 'jsii/superchain:node14', // Workaround buster-slim issue
```

### LINKED ISSUE #1288: AWS CDK v2 Support
CDK v2 has been released!

https://aws.amazon.com/about-aws/whats-new/2021/12/aws-cloud-development-kit-cdk-generally-available/

Would it be possible to initialize a new CDK v2 project using projen?

### LINKED ISSUE #1326: AwsCdkTypeScriptApp still uses deprecated assert library
The AwsCdkTypeScriptApp still has a [hard coded](https://github.com/projen/projen/blob/main/src/awscdk/awscdk-app-ts.ts#L106) do the deprecated assert library. 

To keep it kind of compatible, I'd move that line to the CDK v1 block, and add the assertion library in the v2 block. 

I can provide a pull request, but would like to discuss how it should look like as it might be a breaking change.

eladb:
Extract the logic that handles CDK dependencies (and supports v1/v2 differences) into a new component called `AwsCdkDeps` and reuse across  `AwsCdkTypeScriptApp` and `AwsCdkConstructLibrary`.

* Modify `LambdaFunction` to determine how to render import statements in the generated function construct source code.
* Fix the sample test code in `awscdk-app-ts` to work with the new assertions library.
* Change the default CDK version for `awscdk-app-ts` to `^2`.
* Change `cdkVersion` and `constructsVersion` to reflect version requirement instead of a specific version. This means that existing projects will need to be migrated. Added a NOTICE message during synth to inform.
* Determines workflow container image based on CDK version

Fixes #1288 
Fixes #1326
Fixes #1173

BREAKING CHANGE: The `cdkVersion` and `constructsVersion` options are now *semver ranges* and not just specific versions. To enable the previous behavior, use a caret prefix `^`.
* **awscdk:** `LambdaFunction` and `AutoDiscover` now requires `cdkDeps` so they can interact with CDK version manager.

---


corymhall (commented):
The only thing I keep going back and forth on is if we should have any special handling of the `-alpha` modules. Right now we are just requiring users to add them as `peerDeps` or `deps`, but maybe we should keep using `cdkDependencies`? By removing `cdkDependencies` we are removing some of the abstraction when maybe we shouldn't.

Otherwise LGTM.

corymhall (approved):

eladb:
> The only thing I keep going back and forth on is if we should have any special handling of the `-alpha` modules.

I feel like we might want to add a separate abstraction for this in the future

src/awscdk/awscdk-app-ts.ts
```diff
@@ -251,22 +211,22 @@ app.synth();`;
...
 import { MyStack } from '../src/${appEntrypointName}';
 
 test('Snapshot', () => {
   const app = new App();
   const stack = new MyStack(app, 'test');
 
-  expect(stack).not.toHaveResource('AWS::S3::Bucket');
-  expect(app.synth().getStackArtifact(stack.artifactId).template).toMatchSnapshot();
+  const template = Template.fromStack(stack);
+  expect(template.toJSON()).toMatchSnapshot();
```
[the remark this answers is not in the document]
eladb: We can tweak it later. I kind of like the fact that this is actually capturing the app.

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,261 @@
...
+   */
+  readonly cdkVersion: string;
+
+  /**
+    * Semantic version requirement for the `constructs` library.
+    *
+    * @default - for CDK 1.x the default is "^3.2.27", for CDK 2.x the default
+    * is "^10.0.5".
+    */
+  readonly constructsVersion?: string;
```
Chriscbr, line 23: The indentation of *'s is a inconsistent throughout this file (cleanup)

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    if (this.cdkDependenciesAsDeps) {
+      this.addDeps(...deps.map(m => this.formatModuleSpec(m)));
+    } else {
+      this.addDevDeps(...deps.map(m => this.formatModuleSpec(m)));
+    }
+  }
+
+  /**
+   * Adds CDK modules as test dependencies.
```
Chriscbr, line 181: nit: explain what this means (I'm not sure if "test dependencies" is universally understood)

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,261 @@
...
+   * @param deps names of cdk modules (e.g. `@aws-cdk/aws-lambda`).
+   * @deprecated not supported for CDK v2, use `deps` (for apps) or `peerDeps` (for libraries) instead
+   */
+  public addCdkDependencies(...deps: string[]) {
+    if (deps.length > 0 && this.cdkMajorVersion !== 1) {
+      throw new Error('addCdkDependencies() is not supported for CDK 2.x and above, use addDeps() or addPeerDeps() instead');
+    }
+
+    // this ugliness will go away in cdk v2.0
+    this.addRuntimeDeps(...deps.map(m => this.formatModuleSpec(m)));
```
Chriscbr, line 171: What do you mean? can we not guard this in a "if cdkMajorVersion !== 2" branch? Do we need "runtime" type dependencies even in library type projects?

test/__snapshots__/new.test.ts.snap
```diff
@@ -78,16 +78,15 @@ exports[`projen new awscdk-app-ts 1`] = `
 Object {
   ".projenrc.js": "const { awscdk } = require(\\"projen\\");
 const project = new awscdk.AwsCdkTypeScriptApp({
-  cdkVersion: \\"1.129.0\\",
+  cdkVersion: \\"^2\\",
```
Chriscbr, line 81: Should we call this out as a breaking change?

test/__snapshots__/new.test.ts.snap
```diff
@@ -78,16 +78,15 @@ exports[`projen new awscdk-app-ts 1`] = `
...
-  // deps: [],                    /* Runtime dependencies of this module. */
-  // description: undefined,      /* The description is just a string that helps people understand the purpose of the package. */
-  // devDeps: [],                 /* Build dependencies for this module. */
-  // packageName: undefined,      /* The \\"name\\" in package.json. */
-  // release: undefined,          /* Add release management to this project. */
+  // deps: [],                /* Runtime dependencies of this module. */
+  // description: undefined,  /* The description is just a string that helps people understand the purpose of the package. */
+  // devDeps: [],             /* Build dependencies for this module. */
+  // packageName: undefined,  /* The \\"name\\" in package.json. */
+  // release: undefined,      /* Add release management to this project. */
```
Chriscbr, line 89: I wonder if we should highlight / feature any options relating to the new features you've added (feel free to leave as is)

test/awscdk/awscdk-app.test.ts
```diff
@@ -1,27 +1,25 @@
 import { mkdirSync, writeFileSync } from 'fs';
```
Chriscbr, line 1: Ahh just realizing now I moved a lot of the source files as part of The Big Refactoring, but didn't move/rename the tests to match up. I should clean up the rest when I have a chance

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,261 @@
...
+    *
+    * @deprecated For CDK 2.x use 'devDeps' instead
+    */
+  readonly cdkTestDependencies?: string[];
+}
+
+export interface AwsCdkDepsOptions extends AwsCdkDepsCommonOptions {
+  /**
+   * The type of dependency to use for runtime AWS CDK and `constructs` modules.
+   */
```
Chriscbr, line 86: Why is this required? Perhaps include guidance on which kind of dependency type to use?

Chriscbr (changes requested):
Nice encapsulation!

Please check if any of the docs need to be updated or moved relating to dependencies - e.g. https://github.com/projen/projen/blob/main/docs/awscdk-construct.md#dependencies

src/awscdk/awscdk-deps.ts
eladb: Fixed

src/awscdk/awscdk-deps.ts
eladb: added

src/awscdk/awscdk-deps.ts
eladb: The name of this private method is misleading. It will add a dependency based on `dependencyType`. I'll rename.

eladb: Renamed to `addDepsWithRequestedType()` and simplified this method.

src/awscdk/awscdk-deps.ts
eladb: Renamed to `addCdkDevDependencies()`.

src/awscdk/awscdk-construct.ts
```diff
@@ -213,45 +81,56 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
-   * and "peerDependencies". This is because currently npm would not
-   * automatically install peer dependencies that are not declared as concerete
-   * dependencies by the consumer, so this is a little npm "hack" so that
-   * consumers will not need to depend on them directly if they don't interact
-   * with them.
-   * See `cdkDependenciesAsDeps` for changing the default behavior.
+   * Since this is a library project, dependencies will be added as peer dependencies.
    *
    * @param deps names of cdk modules (e.g. `@aws-cdk/aws-lambda`).
+   * @deprecated Not supported in v2. For v1, use `project.cdkDeps.addCdkDependencies()`
```
Chriscbr, line 89:
```suggestion
   * @deprecated Not supported in v2. For v1, use `project.cdkDeps.addV1Dependencies()`
```

src/awscdk/awscdk-construct.ts
```diff
@@ -213,45 +81,56 @@ export class AwsCdkConstructLibrary extends ConstructLibrary {
...
-    }
+    return this.cdkDeps.addV1Dependencies(...deps);
   }
 
   /**
-   * Adds CDK modules as test dependencies.
+   * Adds AWS CDK modules as dev dependencies.
    *
    * @param deps names of cdk modules (e.g. `@aws-cdk/aws-lambda`).
+   * @deprecated Not supported in v2. For v1, use `project.cdkDeps.addCdkDevDependencies()`
```
Chriscbr, line 99:
```suggestion
   * @deprecated Not supported in v2. For v1, use `project.cdkDeps.addV1DevDependencies()`
```

Chriscbr (changes requested):

test/__snapshots__/new.test.ts.snap
eladb: Leaving as is for now

test/__snapshots__/new.test.ts.snap
eladb: This is actually not a breaking change because `cdkVersion` is a required property. It will just change the initial value for new projects, but won't break any existing users.

PR #1330, opened 2021-12-13
chore: use major-pinned version of @types/node

RomainMuller:
`@types/node` versions numbers match the node runtime versions'
`major.minor`, however new releases are only created when API changes
are included in a node release... We might for example have dependencies
that require `node >= 12.22`, but as 12.21 and 12.22 did not include API
changes, `@types/node@12.20.x` is the "correct" version to use. As it is
not possible to easily determine the correct version to use, we pick up
the latest version.

BREAKING CHANGE: the default version of `@types/node` installed if no
`minNodeVersion` is specified on Node projects is now `^12` instead of
`^14`.

---

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -23,7 +23,7 @@
     },
     {
       "name": "@types/node",
-      "version": "^12.7.0",
+      "version": "^12",
       "type": "build"
     },
     {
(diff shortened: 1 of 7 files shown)


eladb (approved):

eladb:
Why `chore`?

PR #1331, opened 2021-12-13
fix: do not use aws-cdk/assert in v2

pgarbe:
Applied similar logic like in AwsCdkTypeScriptApp. 

But it keeps the aws-cdk/assert library as I didn't want to add another option.

---


eladb:
@pgarbe Can you take a look at #1327 -- I believe it fixes this issue as well...

pgarbe:
@eladb Yes it does. I'll close this one.

PR #1338, opened 2021-12-14
fix(awscdk): issues with CDK v2 libraries and apps (take 2)

### LINKED ISSUE #1173: 1-buster-slim breaks workflows using cdk v2.0.0-rc.25

### LINKED ISSUE #1288: AWS CDK v2 Support

### LINKED ISSUE #1326: AwsCdkTypeScriptApp still uses deprecated assert library

eladb:
Extract the logic that handles CDK dependencies (and supports v1/v2 differences) into a new component called `AwsCdkDeps` and reuse across  `AwsCdkTypeScriptApp` and `AwsCdkConstructLibrary`.

This is a follow up on #1327 which we had to revert due to breaking upgrade flows. The difference is that this change does NOT introduce breaking behavioral change to `cdkVersion` and dev/peer dependencies of constructs.

* Modify `LambdaFunction` to determine how to render import statements in the generated function construct source code.
* Fix the sample test code in `awscdk-app-ts` to work with the new assertions library.
* Change the default CDK version for `awscdk-app-ts` to `2.1.0`.
* Determines workflow container image based on CDK version

Fixes #1288
Fixes #1326
Fixes #1173

BREAKING CHANGE: `LambdaFunction` and `AutoDiscover` now requires `cdkDeps` so they can interact with CDK version manager.

---


src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,293 @@
...
+  public readonly cdkMinimumVersion: string;
+
+  /**
+   * Whether CDK dependencies are added as normal dependencies (and peer dependencies).
+   * @deprecated Not used for CDK 2.x
+   */
+  public readonly cdkDependenciesAsDeps: boolean;
+
+  /**
+   * The major version of the AWX CDK (e.g. 1, 2, ...)
```
Chriscbr, line 112:
```suggestion
   * The major version of the AWS CDK (e.g. 1, 2, ...)
```

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,293 @@
...
+        break;
+    }
+
+    this.project.deps.addDependency(`constructs@${versionRequirement}`, this.dependencyType);
+
+    return versionRequirement;
+  }
+
+  /**
+   * Addsa depdenencies on the AWS CDK framework (e.g. `@aws-cdk/core` for V1 or `aws-cdk-lib` for V1).
```
Chriscbr, line 217:
```suggestion
   * Adds a dependency on the AWS CDK framework (e.g. `@aws-cdk/core` for V1 or `aws-cdk-lib` for V1).
```

src/awscdk/awscdk-deps.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    const testDeps = new Array<string>();
+
+    if (options.cdkAssert ?? true) {
+      testDeps.push('@aws-cdk/assert');
+    }
+
+    // @aws-cdk/assertions is only available starting v1.111.0
+    if (semver.gte(this.cdkMinimumVersion, '1.111.0') && (options.cdkAssertions ?? true)) {
```
Chriscbr, line 264: Good check 👍

src/javascript/node-package.ts
```diff
@@ -838,6 +838,12 @@ export class NodePackage extends Component {
     if (pinned) {
       for (const dep of this.project.deps.all.filter(d => d.type === DependencyType.PEER)) {
         let req = dep.name;
+
+        // skip if we already have a runtime dependency on this peer
+        if (this.project.deps.tryGetDependency(dep.name, DependencyType.RUNTIME)) {
+          continue;
+        }
```
Chriscbr, line 845: Feels like this should be a unit test

Chriscbr (approved):

src/javascript/node-package.ts
eladb: I added one

PR #1341, opened 2021-12-14
chore(docs):  typescript project doc

kaizen3031593:
There is no TypeScript doc in the repository so I thought I would get the ball rolling on starting one.

---


kaizen3031593:
This is just a start -- hoping to get some guidance on what else I should include.

Also -- I'm wondering if it would make sense if the main project README linked to some of these docs rather than the API reference page. I feel like it would improve visibility of the `docs` directory, as well as act as a "softer landing" for new projen users.

docs/typescript.md
```diff
@@ -0,0 +1,82 @@
...
+The `projen new` command will also generate a `.projenrc.js` file which includes
+the definition of your project with any options you specified in the command
+line:
+
+```js
+const { typescript } = require('projen');
+  const project = new typescript.TypeScriptProject({
+    defaultReleaseBranch: 'main',
+    name: 'my-project',
+  });
```
eladb, line 35:
```suggestion
const project = new typescript.TypeScriptProject({
  defaultReleaseBranch: 'main',
  name: 'my-project',
});
```

docs/typescript.md
```diff
@@ -0,0 +1,82 @@
...
+project.synth();
+```
+
+> At this point, projenrc is in JavaScript, but in the future we plan to allow
+> specifying your project definitions in TypeScript.
```
eladb, line 40: This is already supported. You'll need to run `projen new` with `--projenrc-ts` I think. 

Let's describe that as an alternative when running `new`

eladb (commented):
This is awesome!

@Chriscbr take a look and approve if it lgty

Chriscbr (approved):
Looks great. Thanks for the docs! 🙏

Issue #942, opened 2021-07-18
CI/CD abstraction


joostvdwsd:
Hi all,

i was wondering if there were already some attempts to make the pipeline/workflow generation more abstract.
Projen now generates github workflows. We also generate azure devops pipelines in our company packages and i’ve seen people talk about gitlab and travis ci.

In our packages i see a lot of similarities with the github flow and it should be easy to make this more plugin/interface based. 

This is probably a major change. I would not mind taking some of the burdon for this since in the end we would also be helped by a more standardized approach within our company (projen is becomming more and more a goto tool after we standardized  company project types with all security an dcompliancy rules in code) 

especially with the tasks of projen the workflows are getting very simple so a generic approach should be doable. As long as you have your escape hatches in place per implementation it should work

eladb:
We haven't made an attempt yet, but definitely interested in this space. 

My general thinking (as you implied) is to enhance "tasks" so they will be able to model all the standard workflows and render into platform-specific workflows.

Since this is a pretty major change, I would prefer it if you took some time to write an RFC with your proposal. You can use the CDK team's [RFC template](https://github.com/aws/aws-cdk-rfcs/blob/master/0000-template.md) as a baseline and submit a PR with a document under a new directory called `rfcs/`.

joostvdwsd:
Ok, gonna give it a try. Got some general idea on how to structure it. 

Btw, maybe a nice idea to "slowly" work towards it. for instance, in projen you have the release jobs where you have command lines for npm, maven and python packages. This kind of logic could already be nicely implemented as a projen task through:
release:npm
release:maven
release:pypi

as task this also has a nice remark. Would these tasks also build, would they run on the existing dist folder or would it build when there is no dist folder

PR #1357, opened 2021-12-20
feat: GitLab CI model and base classes

dontirun:
Related to #137, #942

This PR adds the model and some opinionated base classes for GitLab CI configurations (which should be reviewed). This currently does not tie the classes into a project configuration. I have not added any tests, but would be happy to add them as this PR progresses.

Base schema from [here](https://gitlab.com/gitlab-org/gitlab/-/raw/master/app/assets/javascripts/editor/schema/ci.json)
Initial TypeScript Interfaces (before opinionated modifications) generated from [here](https://app.quicktype.io/#)

---


src/gitlab/configuration-model.ts
```diff
@@ -0,0 +1,770 @@
...
+  /** Upload artifacts regardless of job status. */
+  ALWAYS = 'always',
+  /** Upload artifacts only when the job fails. */
+  ON_FAILURE = 'on_failure',
+  /** Upload artifacts only when the job succeeds (this is the default). */
+  ON_SUCCESS = 'on_success',
+}
+
+/**
+ * Default settings for the CI Configuration.
```
Chriscbr, line 28: Please clarify what these defaults mean/do (do they apply just to the jobs within this file?)

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+       */
+  public configureDefault(defaultConfig: Default) {
+    if (this.default === undefined) {
+      this.default = defaultConfig;
+    } else {
+      throw new Error(
+        `${this.name}: GitLab CI default has been previously set.`,
+      );
+    }
+  }
```
Chriscbr, line 83: I think it might be preferable to remove this API and simply let users provide a value the constructor and make `default` property public and mutable through getters/setters.

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+      }
+      this.include.push(include);
+    }
+  }
+
+  /**
+       * Set the GitLab pages job. Throws error if page job has been previously set.
+       * @param job The job to use.
+       */
+  public configurePagesJob(job: Job) {
```
Chriscbr, line 102: See comment on configureDefault

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  /**
+       * Add additional services
+       * @param services The services to add.
+       */
+  public addServices(...services: Service[]) {
+    for (const service of services) {
+      if (this.services.includes(service)) {
```
Chriscbr, line 118: I'm not sure if this will work as intended since Service is an interface and the objects might be equal-by-value but different in identity. Please add a unit test for this

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+       * Groups jobs into stages. All jobs in one stage must complete before next stage is
+       * executed. Defaults to ['build', 'test', 'deploy'].
+       */
+  public readonly stages: string[] = [];
+  public readonly variables: Record<string, number | VariableConfig | string> =
+    {};
+  private workflow?: Workflow;
+  public readonly jobs: Record<string, Job> = {};
+
+  constructor(project: Project, name: string) {
```
Chriscbr, line 58: I think it should be possible to configure many of the options at construction time, no?
```suggestion
  constructor(project: Project, name: string, options: CiConfigurationOptions) {
```

Same goes for the GitlabConfiguration and NestedConfiguration classes.

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+      this.services.push(service);
+    }
+  }
+
+  /**
+       * Add a globally defined variable to the CI configuration.
+       * @param variables The variables to add.
+       */
+  public addGlobalVariables(
+    variables: Record<string, number | VariableConfig | string>,
```
Chriscbr, line 132: We can't represent type unions in the languages JSII transliterates to, so as this is a public API, I'd prefer changing the type signature to `Record<string, any>` here.

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+      }
+      this.variables[key] = value;
+    }
+  }
+
+  /**
+       * Set the CI workflow configuration. Throws error if timeout settings have been previously set.
+       * @param workflow The workflow setting.
+       */
+  public configureWorkflow(workflow: Workflow) {
```
Chriscbr, line 148: See comment on configureDefault

src/gitlab/gitlab-configuration.ts
```diff
@@ -0,0 +1,31 @@
...
+  public readonly nestedTemplates: Record<string, NestedConfiguration> = {};
+  constructor(project: Project) {
+    super(project, 'gitlab-ci');
+  }
+
+  /**
+       * Creates and adds nested templates to the includes of the main CI. Additionally adds their stages to the main CI if they are not already present.
+       * @param names The template names.
+       */
+  public createNestedTemplates(...names: string[]) {
```
Chriscbr, line 18: Explain in the docstring how to customize the nested templates since it seems like in this method we can only provide the names of the new templates.

Would it make sense if instead the took as a parameter `Record<string, NestedConfigurationOptions>` or `Record<string, CiConfigurationOptions>` instead?

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,224 @@
...
+  public readonly jobs: Record<string, Job> = {};
+
+  constructor(project: Project, name: string) {
+    super(project);
+    this.project = project;
+    this.name = name;
+    this.path =
+      this.name === 'gitlab-ci'
+        ? '.gitlab-ci.yml'
+        : `.gitlab/ci-templates/${name.toLocaleLowerCase()}.yml`;
```
Chriscbr, line 65: Perhaps document this behavior in the class docstring and/or in `docs/gitlab.md`

Chriscbr (changes requested):
This looks like a great start! I think it would be awesome to have GitLab components in projen.

I've added some initial comments below. Be sure to add tests, and if you could also write up a short high level summary in `docs/gitlab.md` it would be very helpful for understanding the API as well.

src/gitlab/configuration-model.ts
```diff
@@ -0,0 +1,770 @@
...
+// Schema from: https://gitlab.com/gitlab-org/gitlab/-/raw/master/app/assets/javascripts/editor/schema/ci.json
+// Initial TypeScript Interfaces from: https://app.quicktype.io/#
+
+/**
+ * Cache Definition.
+ */
+export interface Cache {
+  /**
+       * Defines when to save the cache, based on the status of the job (Default: Job Success).
+       */
```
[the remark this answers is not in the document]
dontirun: Running `npx projen eslint` is reformatting the multi-line 'nested' docstrings to look like this. I agree it's ugly. I'll change these to be single line to make it look better.

src/gitlab/gitlab-configuration.ts
dontirun: I like the idea of `Record<string, CiConfigurationOptions>` . I'll add some details on how to make further configuration changes in the docstring

src/gitlab/configuration-model.ts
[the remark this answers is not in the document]
dontirun: FIxed in https://github.com/projen/projen/pull/1357/commits/d0f34e044b1ca39ad40233c42eb4002c3e91ed21

src/gitlab/configuration-model.ts
dontirun: FIxed in https://github.com/projen/projen/pull/1357/commits/d0f34e044b1ca39ad40233c42eb4002c3e91ed21

src/gitlab/configuration.ts
dontirun: Updated in https://github.com/projen/projen/pull/1357/commits/cee5171ae03af5ebdc4da06eb2f5331d3c5a94fe

src/gitlab/configuration.ts
dontirun: Updated in https://github.com/projen/projen/pull/1357/commits/cee5171ae03af5ebdc4da06eb2f5331d3c5a94fe

src/gitlab/configuration.ts
dontirun: fixed  and added tests in https://github.com/projen/projen/pull/1357/commits/a9c83f899dce32885b209531fdf9c1fca2ae40bd

src/gitlab/configuration.ts
dontirun: Resolved in https://github.com/projen/projen/pull/1357/commits/6d42ed0b3bd482a03fb6924aff5f964a01fd5f90

src/gitlab/configuration.ts
dontirun: Resolved in https://github.com/projen/projen/pull/1357/commits/d0f34e044b1ca39ad40233c42eb4002c3e91ed21

src/gitlab/configuration.ts
dontirun: Updated in https://github.com/projen/projen/pull/1357/commits/cee5171ae03af5ebdc4da06eb2f5331d3c5a94fe

src/gitlab/gitlab-configuration.ts
dontirun: Resolved in https://github.com/projen/projen/pull/1357/commits/26866b2b9f27a0450049df1f47c01b03b28b82f1

src/gitlab/configuration.ts
dontirun: I added some details to the docstring in https://github.com/projen/projen/pull/1357/commits/26866b2b9f27a0450049df1f47c01b03b28b82f1, but I think a `docs/gitlab.md` file would be helpful to have as well

dontirun: Added `docs/gitlab.md` documentation in https://github.com/projen/projen/pull/1357/commits/e2ec61092f40ccc8a8ab32aea75735b9587ea4da

docs/gitlab.md
```diff
@@ -0,0 +1,67 @@
+# GitLab
+
+[GitLab CI/CD](https://docs.gitlab.com/ee/ci/yaml/gitlab_ci_yaml.html) uses a `.gitlab-ci.yml` configuration located in the root of your repository. Optionally `.gitlab-ci.yml` files can be split into multiple files to increase readability, or reduce duplication of the same configuration in multiple places by using [include](https://docs.gitlab.com/ee/ci/yaml/#include).
```
Chriscbr, line 3: Let me know if this sounds any better
```suggestion
[GitLab CI/CD](https://docs.gitlab.com/ee/ci/yaml/gitlab_ci_yaml.html) uses a `.gitlab-ci.yml` file located at the root of your repository to specify continuous integration and continuous deployment processes. GitLab CI configuration files can be split into multiple files using the [include](https://docs.gitlab.com/ee/ci/yaml/#include) fields in order to increase readability or reduce duplication of the same configuration in multiple places.
```

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,221 @@
...
+  Default,
+  Include,
+  Job,
+  Service,
+  VariableConfig,
+  Workflow,
+} from './configuration-model';
+
+export interface CiConfigurationOptions {
+  readonly default?: Default;
```
Chriscbr, line 16: Please add docstrings for the other options in this interface (default, workflow, jobs)

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+      this.include.push(include);
+    }
+  }
+
+  /**
+   * Add additional services.
+   * @param services The services to add.
+   */
+  public addServices(...services: (string | Service)[]) {
```
Chriscbr, line 111: Does specifying a service with a string have the same meaning as using `{ name: 'the-string' }`? if so, I feel like this is an implementation detail that probably shouldn't be covered in our API -- plus introducing type unions makes the API more obtuse when this gets compiled to other languages using JSII.

Can we simplify the API like this maybe:
```suggestion
  public addService(options: Service) {
```
(Users can call this in a for loop to add multiple services if they want)

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,221 @@
...
+export class CiConfiguration extends Component {
+  public readonly project: Project;
+  public readonly name: string;
+  /** Path to CI file */
+  public readonly path: string;
+  /**
+   * The workflow YAML file.
+   */
+  public readonly file: YamlFile;
+  public default?: Default;
```
Chriscbr, line 46: These fields should be kept readonly since they may refer to objects/arrays, and leaving them mutable by anyone can lead to object reference issues. We can add (safe) getters/setters/methods to mutate these later if needed
```suggestion
  public readonly default?: Default;
```

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,221 @@
...
+   * Can be `Include` or `Include[]`. Each `Include` will be a string, or an
+   * object with properties for the method if including external YAML file. The external
+   * content will be fetched, included and evaluated along the `.gitlab-ci.yml`.
+   */
+  public readonly include: Array<Include | string> = [];
+  /**
+   * A special job used to upload static sites to Gitlab pages. Requires a `public/` directory
+   * with `artifacts.path` pointing to it.
+   */
+  public pages?: Job;
```
Chriscbr, line 57:
```suggestion
  public readonly pages?: Job;
```

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,221 @@
...
+  public readonly services: Array<Service | string> = [];
+  /**
+   * Groups jobs into stages. All jobs in one stage must complete before next stage is
+   * executed. Defaults to ['build', 'test', 'deploy'].
+   */
+  public readonly stages: string[] = [];
+  public readonly variables: Record<string, number | VariableConfig | string> = {};
+  public workflow?: Workflow;
```
Chriscbr, line 65:
```suggestion
  public readonly workflow?: Workflow;
```

Chriscbr (changes requested):

dontirun:
I previously missed that the Global `services` keyword [is deprecated](https://docs.gitlab.com/ee/ci/yaml/#globally-defined-image-services-cache-before_script-after_script).

In general it might make more sense to make a property in the `CiConfiguration` Class for each of the properties in the `Default` interface (ex. `defaultAfterScript`) and reassemble everything in the `renderCI` method.

This would also give more flexibility in adding more methods to help the user configure these properties 

```typescript
  private renderCI() {
    return {
      default: { afterScript: this.defaultAfterScript, ... },
      include:
        this.include.length > 0 ? snakeCaseKeys(this.include) : undefined,
      pages: snakeCaseKeys(this.pages),
      services:
        this.services.length > 0 ? snakeCaseKeys(this.services) : undefined,
      variables:
        Object.entries(this.variables).length > 0 ? this.variables : undefined,
      workflow: snakeCaseKeys(this.workflow),
      stages: this.stages.length > 0 ? this.stages : undefined,
      ...snakeCaseKeys(this.jobs),
    };
  }
}

```

Chriscbr (changes requested):
Just a few more small changes and I think this PR is about ready to ship 👍 

On further thought, I'm not totally sure what the best practices are here regarding leaving class properties public or private, since this class is both like an L1 (not a lot of opinionated defaults) but also like an L2 (it some provides convenience methods). I want to say it probably makes sense to make most of the properties private by default so that they can be changed as needed in the future and make the class more L2-like, but I don't have a strong opinion on this.

@eladb do you have any guidance here?

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,348 @@
...
+  /**
+   * The project the configuration belongs to.
+   */
+  public readonly project: Project;
+  /**
+   * The name of the configuration.
+   */
+  public readonly name: string;
+  /**
+   * Path to CI file egenrated by the configuration.
```
dontirun, line 67:
```suggestion
   * Path to CI file generated by the configuration.
```

src/gitlab/configuration.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  /**
+   * Add additional yml/yaml files to the CI includes
+   * @param includes The includes to add.
+   */
+  public addIncludes(...includes: Include[]) {
+    for (const additional of includes) {
+      if (!this.isValidInclude(additional)) {
+        throw new Error(`${this.name}: GitLab CI ${additional} is not a valid include configuration.`);
+      }
```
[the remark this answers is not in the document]
dontirun: Something like this?

```typescript
/**
 * Throw an error if the provided Include is invalid.
 * @see https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/config/external/mapper.rb
 * @param include the Include to validate.
 */
private assertIsValidInclude(include: Include) {
  const combos = [include.local, (include.file && include.project), include.remote, include.template];
  const len = combos.filter(x => Boolean(x)).length
  if (len !== 1) {
    throw new Error(
      `${this.name}: GitLab CI include ${include} contains ${len} property combination(s).
      A valid include configuration specifies *one* of the following property combinations.
      * local
      * file, project
      * remote
      * template  
      `
    );
  }
}
```

dontirun:
> Just a few more small changes and I think this PR is about ready to ship 👍
> 
> On further thought, I'm not totally sure what the best practices are here regarding leaving class properties public or private, since this class is both like an L1 (not a lot of opinionated defaults) but also like an L2 (it some provides convenience methods). I want to say it probably makes sense to make most of the properties private by default so that they can be changed as needed in the future and make the class more L2-like, but I don't have a strong opinion on this.
> 
> @eladb do you have any guidance here?

My thoughts here were to make this closer to a L1, but with some private properties (with convenience methods) to prevent the creation of CI files with invalid field configurations.

In the future I would like something closer to the `GithubProject` (or what comes out of this [RFC](https://github.com/projen/projen/issues/942)) where standard jobs are created for the different workflows that projen uses for GitHub (Release, build, etc). That future `GithubProject` would probably have more private properties as it would be closer to a L3

src/gitlab/configuration.ts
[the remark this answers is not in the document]
Chriscbr: Yes that looks good.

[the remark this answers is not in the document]
dontirun: Updated in https://github.com/projen/projen/pull/1357/commits/b85f20e419841c9702317913f937f6718d319c72

Chriscbr (approved):

eladb:
## Please read this to avoid merge hell

Hola! apologies for hassle. I've enabled Prettify as part of #1453. This means that most of the source files have changed and you will likely see quite a lot of merge conflicts when you try to merge from `main`.

*But there's an easy way to handle this.*

First, merge your branch from this tag:

```sh
$ git fetch origin
$ git merge 1020d0ad3f95bb4e1201f8640d440ffb43ccf230
```

Edit the `.projenrc.js` file in your branch and add this to **line 57** (right
after `codeCov: true`):

```js
  prettier: true,
```

Now, run:

```sh
$ git clean -fdx .
$ yarn
$ bash projen.bash
$ yarn eslint
$ git add .
$ git commit -m 'format code'
```

This will reformat all the code in your branch to match `main`.

Now, merge from main:

```sh
$ git merge origin/main
```

Sorry again for the inconvenience...

Issue #613, opened 2021-03-14
projen new: interactively ask for required options


joostvdwsd:
Sometimes there are properties mandatory for a project type without the posibility for auto lookup.
In that case it would be great to have projen ask for the configuration option through the CLI. 

I know i've been asking a couple of idea's for the "new" concept. This is one of the few sections of projen without the posibility of customization. Having a more customizable bootstrap would also be great. 
Also I don't mind sending some pr's but before I do stuff I think its nice to know if the idea's are welcome and will fit within the future idea of projen.

joostvdwsd:
The idea I had was something like this:

@default #Specify your team name

This would generate a cli input line with the specified question

mrgrain:
Prior Art:

https://github.com/projen/projen/pull/2926
https://github.com/projen/projen/pull/2975

PR #1365, opened 2021-12-24
chore: RFC 1365: Project creation prompts

MarkMcCulloh:
There are no open issues on this currently, but a relevant one is #613.

Requesting feedback on the proposal for an interactive experience in `projen new`.

[Rendered Version](https://github.com/MarkMcCulloh/projen/blob/prompts-rfc/rfcs/project-creation-prompt.md)

Not sure what to do about the PR title starting with `chore:`, too bad there's not an `rfc` type that I'm aware of.

---


rfcs/project-creation-prompt.md
```diff
@@ -0,0 +1,66 @@
...
+### Prompt format
+
+By default, values sent to the `prompts` library will be derived by the information in the JSII. For example, enums should have their choices populated automatically. 
+
+In case the default logic is insufficient,  `@prompt...` doc tags can be added to override the default. See below for reference.
+
+#### Docstring tag reference
+
+- `@featured` - Functions as it did before, but now also signifies that this option is prompt-able
+- `@promptType` - See https://github.com/terkelg/prompts#-types
```
eladb, line 46: Isn't the jsii type sufficient here?

rfcs/project-creation-prompt.md
```diff
@@ -0,0 +1,66 @@
...
+- `@promptMessage` - See https://github.com/terkelg/prompts#message
```
eladb, line 47: Can you maybe provide some
Examples on user cases  for where these will be used

eladb (changes requested):
Looks good to me. @Chriscbr thoughts?

rfcs/project-creation-prompt.md
MarkMcCulloh: In most cases yes, although I can think of at least one advanced use case for overriding it. Updated with this example:

- An array would have a default prompt type of [list](https://github.com/terkelg/prompts#listmessage-initial).  
  `@promptType multiselect` would allow you to provide a set of distinct options instead.

rfcs/project-creation-prompt.md
MarkMcCulloh: A couple of examples have been added 👍

rfcs/project-creation-prompt.md
```diff
@@ -0,0 +1,74 @@
...
+### Project option prompts
+
+After a project type has been selected, the user can be shown a list of featured options to set interactively. For a project option to be available as a prompt during `projen new`:
+
+- `@featured` tag in property docstring
+- Must be able to be set via CLI arg (is a primitive or enum)
+- Must not have already been provided as a CLI arg
+
+By default, `projen new` will not show prompts unless the `--prompt` flag is provided.
+There is one exception: If a project option meets the above criteria but is also:
```
Chriscbr, line 34: Is this saying no prompts at all will be shown by default, or that the only prompt shown by default will be the one to choose the project type?

Chriscbr (commented):
Looks good to me!

What happens if someone uses `projen new` (without any flags) outside of a TTY, such as in a bash script? error perhaps?

rfcs/project-creation-prompt.md
MarkMcCulloh: I corrected this to say `projen new <project>` instead of `projen new`. I think the overall idea is that prompts are not shown by default unless synthesis is not possible without them. So in the case of `projen new`, without additional info projen can't do anything. But in the case of `projen new typescript` for example, there is no current need for a prompt unless the user wants to opt-in with --prompt.

I added a section with some example situations to illustrate as well.

MarkMcCulloh:
> What happens if someone uses `projen new` (without any flags) outside of a TTY, such as in a bash script? error perhaps?

Good point, added a note on that in the doc: 

> We should check if in a TTY before a prompt is presented and error otherwise.

Chriscbr (approved):

PR #1367, opened 2021-12-25
feat: extend prettier support

### LINKED ISSUE #1318: Improve prettier support
There are two major problems with the way prettier works in projen today:

### Problem 1: Setting `prettier` to true in eslintOptions causes unnecessary red lines in IDEs

Let's say you have something like this in your `.projenrc.ts`
```typescript
eslintOptions: {
  dirs: ['src'],
  devdirs: ['test'],
  prettier: true,
},
```

The eslint setup runs prettier rather than just shutting off its own formatting rules right now. This results in red squiggly lines while editing that would go away once I hit save. That's really annoying. Prettier's own docs go into why this is generally not recommended: https://prettier.io/docs/en/integrating-with-linters.html

Having a projen option to choose whether to run with the prettier config or plugin for eslint would be much nicer. Let formatting run via prettier and linting via eslint.

### Problem 2: Prettier configuration options aren't supported

I know [the prettier team wishes they hadn't added support for options](https://prettier.io/docs/en/option-philosophy.html), but the reality is that they're here to stay and there are people like me that would opt for a different tool if it didn't support some (I can't stand semi-colons). As it stands, I have a manually managed/curated `.prettierrc.yml` file outside of projen that *mostly* works. None of the generated code respects it, which I guess isn't a huge deal. And of course, it's not managed by projen. Prettier supports configuration being defined in package.json, so if options are accepted in projen, they can be set there.

henrysachs:
allow to create a prettierrc with all config values aswell as the corresponding ignore file
it also supports overrides

I'm quite sure you already know prettier as a tool, it formats code. The initial support was already implemented but a lot of teams use prettier options for their repos. These can now be set and will be written into `prettierrc.json`. Also it allows you to ignore certain files with the prettierignore. I build a sample repo to verify this but also wrote some tests to verify that certain configs work and the corresponding files get created. 

fixes #1318

Co-authored-by: @Crapatrick <patrick@domnick.dev>

---


henrysachs:
I dont know if we should update the docs but the tests only run if 

`export EDITOR="vim"`

but some people for example me have configured their editor to be vscode for GIT or SOPS

henrysachs:
onfortunately the bump test doesn't run locally for me which blocks the build. I will probably turn it off locally to build projen and tinker with the config. It seems to run in CI.

henrysachs:
i also uploaded my sample repo to see the new feature in action: https://github.com/henrysachs/prettierprojen/blob/main/.projenrc.js

I will now try to implement a basic test and would like some feedback on the implementation. Maybe @bilalq wants to take a look if that matches what his issue described :)

henrysachs:
so as all checks went green i mark you @eladb and @Chriscbr and kindly ask you for a review. I don't see any other buttons to do this in the github ui and hope this is the intended way.

src/cdk/jsii-project.ts
```diff
@@ -136,6 +136,7 @@ export interface JsiiGoTarget extends GoPublishOptions {
  */
 export class JsiiProject extends TypeScriptProject {
   public readonly eslint?: Eslint;
+  public readonly prettier?: Prettier;
```
eladb, line 139: Can we add this at a lower level? `TypeScriptProject` or even `NodeProject`?

Also, doesn't seem like this is being assigned anywhere.

src/javascript/node-project.ts
```diff
@@ -824,6 +840,10 @@ export class NodeProject extends GitHubProject {
     this.npmignore?.addPatterns(pattern);
   }
 
+  public addPrettierIgnore(pattern: string) {
```
eladb, line 843: Docstrings

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
...
+  AUTO = 'auto',
+  CR = 'cr',
+  CRLF = 'crlf',
+  LF = 'lf',
+}
+
+/**
+ * How to handle whitespaces in HTML.
+ *
+ * Respect the default value of CSS display property.
```
eladb, line 171: Please move the descriptions of the specific options to docstrings of individual members instead of the enum type itself.

(throughout this file)

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
...
+export type Files = string[] | string;
+
+/**
+ * The options to apply for this override.
+ */
+export interface PrettierOverrideOptions {
+  /**
+   * Include parentheses around a sole arrow function parameter.
+   */
+  readonly arrowParens?: ArrowParens;
```
eladb, line 212: There seem to be duplication between `PrettierOverrideOptions` and `PrettierConfig`. Can we reuse somehow?

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
+import { Component } from '../component';
+import { NodeProject } from '../javascript';
+import { JsonFile } from '../json';
+import { Project } from '../project';
+
+export interface PrettierConfig {
```
eladb, line 6: Rename to `PrettierOptions` to match convention

eladb, line 6: All public APIs should have docstrings:

```suggestion
/**
 * Options for `Prettier`.
 */
export interface PrettierConfig {
```

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
...
+   * Print (to stderr) where a cursor at the given position would move to after formatting.
+   * This option cannot be used with --range-start and --range-end.
+   */
+  readonly cursorOffset?: number;
+  /**
+   * Control how Prettier formats quoted code embedded in the file.
+   */
+  readonly embeddedLanguageFormatting?: EmbeddedLanguageFormatting;
+  /**
+   * Which end of line characters to apply.
```
eladb, line 29: All optional (`?`) props should have a `@default` docstring tag which describes the default either explicitly as a value or via a description (see examples in the code).

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
...
+    return project.components.find(isPrettier);
+  }
+  /**
+   * Direct access to the prettier configuration (escape hatch)
+   */
+  public readonly config: PrettierConfig;
+  /**
+   * Direct access to the prettier overrides (escape hatch)
+   */
+  public readonly overrides: PrettierOverride[];
```
eladb, line 382: We prefer not to expose mutable arrays. You can use this pattern:

```ts
public readonly _overrides: PrettierOverride[];

public get overrides() { return [...this._overrides]; }
public addOverride(...) { this._overrides.push(...); }
```

src/javascript/prettier.ts
```diff
@@ -0,0 +1,401 @@
...
+ */
+export class Prettier extends Component {
+  public static of(project: Project): Prettier | undefined {
+    const isPrettier = (c: Component): c is Prettier => c instanceof Prettier;
+    return project.components.find(isPrettier);
+  }
+  /**
+   * Direct access to the prettier configuration (escape hatch)
+   */
+  public readonly config: PrettierConfig;
```
eladb, line 378: These are not really escape hatches. Since you are using `JsonFile`, users can already escape hatch into this file directly. Escape hatches should not part of the API of the component.

Seems like this is just exposing the underlying configuration for mutation, which seems fine.

src/typescript/typescript.ts
```diff
@@ -337,6 +351,10 @@ export class TypeScriptProject extends NodeProject {
       this.tsconfigEslint = this.tsconfigDev;
     }
 
+    if (options.prettier ?? false) {
```
eladb, line 354: Move to `NodeProject`.

src/typescript/typescript.ts
```diff
@@ -48,6 +48,19 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
    */
   readonly eslintOptions?: EslintOptions;
 
+  /**
+   * Setup prettier.
+   *
+   * @default false
```
eladb, line 54: Why not `true`?

eladb (changes requested):

src/cdk/jsii-project.ts
henrysachs: i just took inspiration from eslint there as i find those two kinda complementary. Thats also why i haven't moved it to `NodeProject`. But I can do this as I prefer it too.

src/typescript/typescript.ts
henrysachs: I was a bit hesitant as Prettier is opionated about formatting when turned on. Also than i need to toggle the eslint flag too. Would that be fine for you? if so the next topic would be if prettier in enabled in `NodeProject` eslint wouldn't exist but that shouldn't be a problem as the getter would just return undefined.

src/javascript/prettier.ts
henrysachs: okay i just those comments from eslint tbh. I will change the comment.

henrysachs:
@eladb I have implemented all remarks but I'm still unsure about activating prettier by default because it is pretty opinionated about formatting and i'm not sure if any user wants to have this strict rules about this. Also I would need to set it in eslint where its currently off by default. I'm fine with both just want to clarify the impact. 😅

henrysachs:
had to update my fork because the build was failing

henrysachs:
@eladb not to stress you but had you had a chance looking at this? I think it resolved all points you had except the default value which I would like to discuss. ☺️

src/javascript/node-project.ts
```diff
@@ -833,6 +867,16 @@ export class NodeProject extends GitHubProject {
     this.npmignore?.addPatterns(pattern);
   }
 
+  /**
+  * Defines Prettier ignore Patterns
+  * these patterns will be added to the file .prettierignore
+  *
+  * @param {string} pattern filepatterns so exclude from prettier formatting
```
eladb, line 874: type is already annotated in the code:

```suggestion
  * @param pattern filepatterns so exclude from prettier formatting
```

src/javascript/prettier.ts
```diff
@@ -0,0 +1,338 @@
+import { Component } from '../component';
+import { NodeProject } from '../javascript';
+import { JsonFile } from '../json';
+import { Project } from '../project';
+/**
+ * Options for Prettier
+ *
+ * @see https://prettier.io/docs/en/options.html
+ * @export
+ * @interface PrettierOptions
```
eladb, line 10: redundant:

```suggestion
```

src/javascript/prettier.ts
```diff
@@ -0,0 +1,338 @@
...
+   *
+   * @default []
+   * @see https://prettier.io/docs/en/configuration.html#configuration-overrides
+   */
+  readonly overrides?: PrettierOverride[];
+}
+/**
+ * Options to set in Prettier directly or through overrides
+ *
+ * @export
```
eladb, line 24: Remove

src/javascript/prettier.ts
```diff
@@ -0,0 +1,338 @@
...
+  /**
+   * Direct access to the prettier configuration
+   */
+  public readonly config: PrettierOptions;
+  /**
+   * Direct access to the prettier overrides
+   *
+   * @internal
+   */
+  public readonly _overrides: PrettierOverride[];
```
eladb, line 311: Why does this need to be public?

eladb (changes requested):
A few additional minor changes. I am okay leaving the default disabled for now.

src/javascript/prettier.ts
henrysachs: I thought jsii had issues with privates. Changed to private as its available through the getter and setter.

henrysachs:
oh i was at the stage if resolving conflicts too :D

eladb (approved):

eladb:
@henrysachs Great job!

PR #1383, opened 2021-12-27
fix(workflows): status check update is not working

### LINKED ISSUE #1362: Build: Gracefully handle self-mutation in forks
When a PR is submitted from a fork, we cannot push any updates to the forked branch (for security reasons). This means that today when self-mutation happens, the build fails with a cryptic error (i.e. cannot push to the branch).

Ideally, the build workflow should behave differently when it runs from a fork. It should clearly report the diff and tell the user they need to update their branch manually (maybe suggest a command).

eladb:
Use a Personal Access Token to push changes and create pull requests from the `build` and `upgrade` workflows. This implies that GitHub will trigger workflows as a result, and removes the need to manually create a "fake" status check update. It also simplifies the upgrade workflow because now it doesn't need to re-implement "build", which just got fancier in JSII projects (packaging tasks are executed in parallel).

Since this cannot be done with a normal GITHUB_TOKEN, and combined with the limitations of this token in upgrade workflows (unable to push workflow updates), we decided that from now, projen _requires_ a dedicated personal access token (PAT) in the repo.

By default, the name of this secret is `PROJEN_GITHUB_TOKEN` (which is the name most projects use today). If this secret does not exist in the repository, at the moment it will result in a poor experience. We will consider improving this experience in the future.

Also fixes #1362 by adding a separate job that only executes when build is triggered by forks, and prints a nice error with the diff.

BREAKING CHANGE: A secret called `PROJEN_GITHUB_TOKEN` with `workflows`, `packages` and `repo` scopes is now **required** for all projects. As much as this is a bit of a burden, most projects already have such a token because otherwise projen upgrade workflows cannot really function.

---


Chriscbr:
I wonder if the default behavior for the `upgrade` workflow should be to always use a PAT (and to make this required in projenrc)?

.github/workflows/build.yml
```diff
@@ -36,50 +31,85 @@ jobs:
...
+        name: Create Patch
         run: |-
-          if ! git diff --exit-code; then
-            git add .
-            git commit -m "chore: self mutation"
-            git push origin HEAD:${{ github.event.pull_request.head.ref }}
-            echo "::set-output name=self_mutation_commit::$(git rev-parse HEAD)"
-          fi
+          git add .
+          git diff --patch --staged > .upgrade.tmp.patch
```
iliapolo, line 42: For clarity:

```suggestion
          git diff --patch --staged > .self-mutation.tmp.patch
```

.github/workflows/build.yml
```diff
@@ -36,50 +31,85 @@ jobs:
...
+      - name: Push changes
+        run: |-2
+            git add .
+            git commit -m "chore: self mutation"
+            git push origin HEAD:${{ github.event.pull_request.head.ref }}
   package-js:
     needs: build
     runs-on: ubuntu-latest
     permissions: {}
+    if: ${{ ! needs.build.outputs.self_mutation_happened }}
```
iliapolo, line 112: Hm, this is a nice touch that would go away if you apply my suggestion from above, but maybe its worth it given the simplification it brings? 

Maybe a middle ground would be to have the `Apply patch` job set the output, and make this job require `Apply patch`.

.github/workflows/upgrade-main.yml
```diff
@@ -11,18 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-    outputs:
```
iliapolo, line 14: My god, what a beautiful diff for this file :)

src/build/build-workflow.ts
```diff
@@ -1,15 +1,18 @@
...
 
 const BRANCH_REF = '${{ github.event.pull_request.head.ref }}';
 const REPO_REF = '${{ github.event.pull_request.head.repo.full_name }}';
-const PRIMARY_JOB_ID = 'build';
+const BUILD_JOBID = 'build';
 const SELF_MUTATION_STEP = 'self_mutation';
-const SELF_MUTATION_COMMIT = 'self_mutation_commit';
+const SELF_MUTATION_HAPPENED_OUTPUT = 'self_mutation_happened';
+const IS_FORK = 'github.event.pull_request.head.repo.full_name != github.repository';
+const NOT_FORK = 'github.event.pull_request.head.repo.full_name == github.repository';
```
iliapolo, line 15: I think that should work...

```suggestion
const NOT_FORK = `!(${IS_FORK})`;
```

src/build/build-workflow.ts
```diff
@@ -243,58 +175,129 @@ export class BuildWorkflow extends Component {
...
+          uses: 'actions/checkout@v2',
+          with: {
+            repository: REPO_REF,
+            ref: BRANCH_REF,
+          },
+        },
+        ...WorkflowActions.downloadApplyGitPatch(),
+        {
+          name: 'Found diff after build (update your branch)',
+          run: 'git diff --exit-code',
```
iliapolo, line 251: Feels like we no longer need to run `git diff`, but rather inspect the patch created in the build.

src/github/auto-approve.ts
```diff
@@ -29,8 +29,10 @@ export interface AutoApproveOptions {
    * {@link https://docs.github.com/en/actions/reference/authentication-in-a-workflow `GITHUB_TOKEN` }
    * - that you would like auto approved, such as when using the `depsUpgrade` property in
    * `NodeProjectOptions`, then you must use a different token here.
+   *
+   * @default "GITHUB_TOKEN"
    */
-  readonly secret: string;
+  readonly secret?: string;
```
iliapolo, line 35: I wonder if we even need this now. Maybe we should always approve PR's with the default github token, this will enforce that automatic PR's are only created with a PAT, which is a better practice as we have been proven. 

Something to consider, we can follow up on this later as well.

src/github/workflow-actions.ts
```diff
@@ -0,0 +1,63 @@
...
+import { JobStep } from './workflows-model';
+
+const GIT_PATCH_FILE = '.upgrade.tmp.patch';
+const RUNNER_TEMP = '${{ runner.temp }}';
+
+/**
+ * A set of utility functions for creating GitHub actions in workflows.
+ */
+export class WorkflowActions {
+  public static createUploadGitPatch(options: WorkflowActionsCommonOptions = {}): JobStep[] {
```
iliapolo, line 11: docstring

src/github/workflow-actions.ts
```diff
@@ -0,0 +1,63 @@
...
+      {
+        if: options.if,
+        name: 'Upload patch',
+        uses: 'actions/upload-artifact@v2',
+        with: { name: GIT_PATCH_FILE, path: GIT_PATCH_FILE },
+      },
+    ];
+  }
+
+  public static downloadApplyGitPatch(options: WorkflowActionsCommonOptions = {}): JobStep[] {
```
iliapolo, line 30: docstring

src/github/workflow-actions.ts
eladb: Done

src/github/workflow-actions.ts
eladb: Done

.github/workflows/build.yml
eladb: Changed to .repo.patch so its more general

.github/workflows/build.yml
```diff
@@ -36,50 +31,85 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           directory: coverage
-      - name: Self mutation
+      - name: Check self-mutation
```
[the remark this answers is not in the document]
eladb: As you noted below, I need to know if self mutation happened in order to conditionally executes steps and jobs later. Your proposal to move this to apply patch won't change the fact that we still need this logic, and I am not sure it is really going to make a difference.

Additionally, this is a cheap operation.

Additionally, refactoring this and retesting is such a hassle, so I rather just leave it as is if that's okay with you.

src/github/auto-approve.ts
eladb: Yes, I agree, but let's follow up later with this.

src/build/build-workflow.ts
eladb: What is the benefit of inspecting the patch over `git diff`?

src/build/build-workflow.ts
eladb: OK

.github/workflows/build.yml
eladb: As mentioned above, I think the "middle ground" is essentially identical to what I've done.

.github/workflows/build.yml
```diff
@@ -36,50 +31,90 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions: {}
+    if: ${{ needs.build.outputs.self_mutation_happened &&
+      github.event.pull_request.head.repo.full_name != github.repository }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
```
eladb, line 66: double checkout

src/build/build-workflow.ts
iliapolo: The benefit would be not maintaining two `git diff` commands, that should be identical. 
For example, I just realized the command the creates the patch runs the diff with `--staged` so that it detects new files. This flag belongs here as well.

Inspecting the patch would prevent such mismatches.

iliapolo (approved):

iliapolo:
Looks awesome. Just this one: https://github.com/projen/projen/pull/1383#discussion_r776334094

src/release/release.ts
```diff
@@ -465,9 +457,7 @@ export class Release extends Component {
 
     // anti-tamper check (fails if there were changes to committed files)
     // this will identify any non-committed files generated during build (e.g. test snapshots)
-    if (this.antitamper) {
-      releaseTask.exec('git diff --ignore-space-at-eol --exit-code');
-    }
+    releaseTask.exec('git diff --ignore-space-at-eol --exit-code');
```
iliapolo, line 460: Also relates to https://github.com/projen/projen/pull/1383#discussion_r776334094.

Though in this case of course we don't have a patch to inspect.

Chriscbr:
> A secret called PROJEN_GITHUB_TOKEN with workflows, packages and repo scopes is now required for all projects.

@eladb Is this referring to just projects that want to use GitHub workflows? Or just projects that want to use the upgrade workflow specifically? I feel like it should still be possible to use a lot of projen's `GitHub` functionality without this token, no?

eladb:
> > A secret called PROJEN_GITHUB_TOKEN with workflows, packages and repo scopes is now required for all projects.
> 
> 
> 
> @eladb Is this referring to just projects that want to use GitHub workflows? Or just projects that want to use the upgrade workflow specifically? I feel like it should still be possible to use a lot of projen's `GitHub` functionality without this token, no?

This is needed by the build and upgrade workflows

Issue #1381, opened 2021-12-27
Broken build for `jsii` projects with `peerDependencies`


iliapolo:
Ever since https://github.com/projen/projen/pull/1359, packaging occurs in a different job than the build job. 

The package jobs consume the `js` artifact created by the build job and run `jsii-pacmak` on the assembly inside the artifact. 
Problem is that for projects that have `peerDependencies`, those dependencies are never installed. 

These `peerDependencies` are required for `jsii-pacmak` because they might appear in the assembly. 

This causes:

```console
Run jsii_version=$(node -p "JSON.parse(fs.readFileSync('.jsii')).jsiiVersion.split(' ')[0]")
npx: installed 114 in 13.291s
[jsii-pacmak] [INFO] Found 1 modules to package
[jsii-pacmak] [INFO] Packaging NPM bundles
[jsii-pacmak] [INFO] Loading jsii assemblies and translations
Error: Cannot find module '@aws-cdk/assets'
Require stack:
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/node_modules/jsii-reflect/lib/util.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/node_modules/jsii-reflect/lib/class.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/node_modules/jsii-reflect/lib/assembly.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/node_modules/jsii-reflect/lib/index.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/lib/index.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/bin/jsii-pacmak.js
- /home/runner/.npm/_npx/1740/lib/node_modules/jsii-pacmak/bin/jsii-pacmak
```

PR #1384, opened 2021-12-27
feat: pwd artifact

iliapolo:
Instead of only using the `artifactsDirectory` as the build artifact, use the entire current working directory. 

This is much more flexible since it allows downstream jobs access to the entire repo. It effectively makes the post build jobs run on the same environment as they would have if the were executed from the same job.

### Caveat 

For node projects, `node_modules` is not included in the artifact because it takes forever to compress and upload. 
If a post build job needs `node_modules`, the job is responsible for installing it, using the lock files available in the articact.

Fixes https://github.com/projen/projen/issues/1381

---


.projen/tasks.json
```diff
@@ -197,7 +194,7 @@
           "exec": "jsii_version=$(node -p \"JSON.parse(fs.readFileSync('.jsii')).jsiiVersion.split(' ')[0]\")"
         },
         {
-          "exec": "npx jsii-pacmak@$jsii_version -v --outdir $PWD/dist --target go"
+          "exec": "npx jsii-pacmak@$jsii_version -v --target go"
```
iliapolo, line 197: I think the `--outdir` wasn't necessary given:

https://github.com/projen/projen/blob/dae6ab47cc021efbf1fbc997f9b939495d64a5b6/src/cdk/jsii-project.ts#L192-L198

.github/workflows/build.yml
```diff
@@ -88,17 +91,10 @@ jobs:
...
-          rm -fr dist/js
+      - name: Install Dependencies
+        run: yarn install --check-files --frozen-lockfile
       - name: Create js artifact
-        run: >-
-          jsii_version=$(node -p
-          "JSON.parse(fs.readFileSync('.jsii')).jsiiVersion.split(' ')[0]")
-
-          npx jsii-pacmak@$jsii_version -v --outdir $PWD/dist --target js
+        run: npx projen package:js
```
eladb, line 97: nice!

src/build/build-workflow.ts
```diff
@@ -236,7 +248,6 @@ export class BuildWorkflow extends Component {
         uses: 'actions/download-artifact@v2',
         with: {
           name: artfiactName,
-          path: this.artifactsDirectory,
```
eladb, line 239: I am wondering if this will break in non-JSII workflows because jsii-release expects output in `dist/`.

eladb (changes requested):
see comment

src/build/build-workflow.ts
iliapolo: Not sure I follow. The output will still be in dist:

https://github.com/projen/projen/blob/dae6ab47cc021efbf1fbc997f9b939495d64a5b6/src/javascript/node-project.ts#L667-L671

This is the path to download the artifact to, since we are bundling up `.`, the download path should also be `.`, which is the default value for this action - so I removed it.

.github/workflows/build.yml
iliapolo: Yes a nice consequence

eladb (approved):

src/build/build-workflow.ts
eladb: Got it

PR #1406, opened 2021-12-28
fix(mergify): not all build jobs are considered

### LINKED ISSUE #1382: Mergify config doesn't require `package-` jobs to pass
Ever since #1359, packaging occurs in a different job than the build job.

These packaging jobs are what calls `jsii-pacmak`, which was previously called in the build job. 
However, mergify isn't configured to look at these jobs, allowing (automatic) breaking PR's to be merged.

For example:

https://github.com/cdk8s-team/cdk8s-plus/pull/226

eladb:
Fixes #1382 by synthesizing the filter with all the build jobs. This has to be done lazily because jobs can be added.

BREAKING CHANGE: `AutoMerge` doesn't take a `buildJob` any more. Instead, you can use `addConditions()` or `addConditionsLater()` to qualify the auto-merge.

---


src/github/auto-merge.ts
```diff
@@ -69,16 +64,46 @@ export class AutoMerge extends Component {
...
 
     this.project.addPackageIgnore('/.mergify.yml');
   }
+
+  /**
+   * Adds conditions to the auto merge rule.
+   * @param conditions The conditions to add (mergify syntax)
+   */
+  public addConditions(...conditions: string[]) {
+    this.conditions.push(...conditions);
```
iliapolo, line 85: Seems like we don't really need the `conditions` array if we already have the `lazyConditions`.

```suggestion
    this.lazyConditions.push({ renderConditions: () => conditions });
```

src/github/auto-merge.ts
```diff
@@ -69,16 +64,46 @@ export class AutoMerge extends Component {
...
+  }
+
+  /**
+   * Adds conditions that will be rendered only during synthesis.
+   * @param later The later
+   */
+  public addConditionsLater(later: IAddConditionsLater) {
```
iliapolo, line 92:
```suggestion
  public addLazyConditions(later: IAddConditionsLater) {
```

src/github/auto-merge.ts
```diff
@@ -69,16 +64,46 @@ export class AutoMerge extends Component {
...
+
+    for (const later of this.lazyConditions) {
+      output.push(...later.renderConditions());
+    }
+
+    return output;
+  }
+}
+
+export interface IAddConditionsLater {
```
iliapolo, line 107:
```suggestion
export interface ILazyConditions {
```

src/github/auto-merge.ts
```diff
@@ -69,16 +64,46 @@ export class AutoMerge extends Component {
...
+  renderConditions(): string[];
```
iliapolo, line 108:
```suggestion
  render(): string[];
```

src/github/auto-merge.ts
```diff
@@ -69,16 +64,46 @@ export class AutoMerge extends Component {
...
     });
 
     this.project.addPackageIgnore('/.mergify.yml');
   }
+
+  /**
+   * Adds conditions to the auto merge rule.
+   * @param conditions The conditions to add (mergify syntax)
+   */
+  public addConditions(...conditions: string[]) {
```
iliapolo, line 84: I guess we don't strictly need this method since we only add lazy conditions. Maybe drop it for now to keep it clean?

iliapolo (changes requested):

src/github/auto-merge.ts
eladb: I think that if we have `addConditionsLater()` it makes sense to also have `addConditions()`? It also helps signify the difference between them.

src/github/auto-merge.ts
eladb: Done.

src/github/auto-merge.ts
eladb: I actually prefer "addConditionsLater()" because the "later" applies to how we are adding the conditions and not the nature of the condition ("lazy condition" sounds weird). But that's really a matter of personal preference.

src/github/auto-merge.ts
eladb: Keeping with the "later" semantics (reasons above).

src/github/auto-merge.ts
eladb: Done.

iliapolo (approved):

Issue #1407, opened 2021-12-28
Latest releases are not publishing correctly after `0.46.3`


bmiller08:
After `0.46.3` releases the `projen` package seems to be published incorrectly or not at all for certain languages

- PyPi has `0.46.8` as expected
- Maven has `0.46.3`
- npm has `0.46.3`, but the latest version gets published as `0.0.0`

eladb:
@bmiller08 thanks for reporting. We are looking into this.

PR #1408, opened 2021-12-28
fix: release does not bump

iliapolo:
The approach proposed in https://github.com/projen/projen/pull/1384 fails the release workflow. The problem is that the `release` task will run `unbump`, reverting the `package.json` version to `0.0.0`.

Instead, we go back to the standard artifacts mechanism of uploading the contents of the `dist` directory. In jsii projects, the dist directory will contain the entire built repository, including the bump. This is done by resetting the default `package` task to simply copy over the repo into `dist`.

When we download the artifact, we have the built repository in the `dist` directory, and we run packaging and publishing from this directory, instead of the root. 

Note that for non-jsii projects, the artifacts will simply be the tarballs, so they will not have access to the repository.  

> This is all still too funky, and should be revisited when we think about the new/enhanced modeling. 

Also, intentionally flip to dry run releases so I can inspect the workflow and artifacts before possibly publishing another bad version. 

Fixes https://github.com/projen/projen/issues/1407

---


.github/workflows/build.yml
```diff
@@ -91,10 +88,13 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: build-artifact
+          path: dist
       - name: Install Dependencies
         run: yarn install --check-files --frozen-lockfile
+        working-directory: dist
       - name: Create js artifact
         run: npx projen package:js
+        working-directory: dist
```
eladb, line 97: Eventually jsii-release expects the output in `dist/<LANG>` - wouldn't this cause the output to be under `dist/dist/<LANG>`?

src/github/workflows.ts
```diff
@@ -184,6 +184,14 @@ function renderJobs(jobs: Record<string, workflows.Job>) {
     };
   }
 
+  function renderJobSteps(steps: workflows.JobStep[]) {
+    return steps.map(s => {
+      const res = { ...s, 'working-directory': s.workingDirectory };
```
eladb, line 189: I believe this capitalization fix should happen automatically

eladb (changes requested):
Something feels off (or I am missing something). It seems like jsii-release won't be able to find the artifacts.

I think the publish jobs should do this:

```sh
$ mv dist/ repo/
$ cd repo
$ jsii-pacmak --target boom
$ mv dist ../
```

This will basically result in pwd looking like this:

```
./dist/<lang>/<artifacts>
./repo
```

.github/workflows/build.yml
iliapolo: In this context, the `dist` directory is what contains the built repository from the build job. When `package:js` finishes, the directory structure of the job will be: `./dist/dist/js`

Then, `jsii-release` is executed from `dist` as well. So it sees the artifacts in the expected location -> `dist/js`.

src/github/workflows.ts
iliapolo: Unfortunately it didn't...why do you think it should happen?

.github/workflows/build.yml
iliapolo: https://github.com/projen/projen/blob/4eaf4d179f39c1aac69d4415e07e16d1c5d8caa9/.github/workflows/release.yml#L111-L117

Notice the `working-directory`.

iliapolo: Basically remember that the repository is now inside `dist`, we just need to switch to that directory when running pacmak and jsii-release.

iliapolo:
> Something feels off (or I am missing something). It seems like jsii-release won't be able to find the artifacts.
> 
> I think the publish jobs should do this:
> 
> ```shell
> $ mv dist/ repo/
> $ cd repo
> $ jsii-pacmak --target boom
> $ mv dist ../
> ```
> 
> This will basically result in pwd looking like this:
> 
> ```
> ./dist/<lang>/<artifacts>
> ./repo
> ```

It essentially what's happening, except without fussing with renmaing/moving directories. Just switching to the right one when needed.

.github/workflows/build.yml
eladb: Oh I see, but now there's more variance between how publishers work between JSII & non-JSII projects. I rather we keep the special-JSII behavior to the pre-publishing steps if possible, so that the rest of the job is kept unchanged.

iliapolo: Ok, let me check

iliapolo:
@eladb Ok, your way results in a smaller diff so the verdict is 👍 

Basically I implemented:

```console
$ mv dist/ repo/
$ cd repo
$ jsii-pacmak --target boom
$ mv dist ../
```

Note that this is implemented in the workflow, not the task. I think that makes sense because these shananigans are a result of the workflow, locally they are not needed.

eladb (approved):
Nice!

PR #1417, opened 2021-12-29
fix(workflows): mutation check did not consider unstaged changes

eladb:
We accidentally used `--staged` in the `git diff` call that checked for mutations without calling `git add .` before.

This change consolidates the mutation check into the patch upload "action" so this is done in a single place.

---


.github/workflows/build.yml
```diff
@@ -73,7 +70,7 @@ jobs:
         run: '[ -s ${{ runner.temp }}/.repo.patch ] && git apply ${{ runner.temp
           }}/.repo.patch || echo "Empty patch. Skipping."'
       - name: Found diff after build (update your branch)
-        run: git diff --staged --exit-code
+        run: git diff --exit-code
```
iliapolo, line 73: Applying the patch doesn't do `git add`, I think instead we just need to `git add .` - no?

iliapolo: @eladb The self mutation reverted this commit

src/build/build-workflow.ts
```diff
@@ -235,7 +235,7 @@ export class BuildWorkflow extends Component {
         }),
         {
           name: 'Found diff after build (update your branch)',
-          run: 'git diff --staged --exit-code',
+          run: 'git diff --exit-code',
```
eladb, line 238:
```suggestion
          run: 'git add . && diff --staged --exit-code',
```

src/release/release.ts
```diff
@@ -457,7 +457,7 @@ export class Release extends Component {
 
     // anti-tamper check (fails if there were changes to committed files)
     // this will identify any non-committed files generated during build (e.g. test snapshots)
-    releaseTask.exec('git diff --staged --ignore-space-at-eol --exit-code');
+    releaseTask.exec('git diff --ignore-space-at-eol --exit-code');
```
iliapolo, line 460: Should also have a `git add .`, which makes really uncool to run locally :\ .

Im ok addressing this later though.

iliapolo (approved):

PR #1424, opened 2021-12-29
Update README to be clear that git v2.28 or later is needed (to support git init -b) 

gherlein:
I stupidly missed the git init step on an Ubuntu 20.10 EC2 instance (quite a common instance type) and was confused when git init -b failed.  Updating the readme to hopefully help someone else.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ $ npx projen new PROJECT-TYPE
 ...
 ```
 
+NOTE:   If creating a new project fails while doing "git init -b" you probably skipped
+the "git init" step.  *projen* will do that for you but requires that you have at least 
+git version 2.28 installed in order for "git init -b <branch>" to be supported. 
+
 Currently supported project types (use `npx projen new` without a type for a
 list):
 


Chriscbr (changes requested):
> and was confused when git init -b failed

I think it might be possible to fix this directly instead to save future users' pain:

https://github.com/projen/projen/blob/09e8eced1f4e796db0c04b618f23374d29afb2c5/src/cli/cmds/new.ts#L248

What if we run `git init` followed by `git branch -m main` instead?

gherlein:
That would probably be better, since Ubuntu 20.04 is LTS and will be a very common platform for many

eladb (changes requested):
Ooouch. Would it make more sense to check the git version and use the previous set of commands as a fallback?

gherlein:
That's even better.  Maybe with a note to the user "older version of git found, using fallback command list; please consider upgrading to a version of git >= 2.28" or something like that

gherlein:
I'll submit a new PR with a fix

Issue #1431, opened 2021-12-30
github release for jsii projects that contain a `CHANGELOG.md` in the root is failing


iliapolo:
Apparently the existence of this file messes up the `rsync` [command](https://github.com/projen/projen/blob/main/src/cdk/jsii-project.ts#L193) we run for creating the artifact. 

The file conflicts with the `changelog.md` file created in the `dist` directory, making it so that artifact itself doesn't contain it.

[Causing](https://github.com/aws/constructs/runs/4665340309?check_suite_focus=true):

```console
Run errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt) --target $GITHUB_REF 2> $errout && true; exitcode=$?; if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then cat $errout; exit $exitcode; fi
  errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt) --target $GITHUB_REF 2> $errout && true; exitcode=$?; if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then cat $errout; exit $exitcode; fi
  shell: /usr/bin/bash -e {0}
  env:
    GITHUB_TOKEN: ***
    GITHUB_REPOSITORY: aws/constructs
    GITHUB_REF: refs/heads/10.x
open dist/changelog.md: no such file or directory
```

In fact, this exposes a bug in the GitHub publishing job, which doesn't do the necessary per publish steps that other publishing jobs do following https://github.com/projen/projen/pull/1408.

For projects that dont contain a `CHANGELOG.md` file, GitHub publishing only works by mistake, because of this unexpected behavior of `rsync`.

PR #1432, opened 2021-12-30
fix(jsii): github release not created

iliapolo:
Introduce a way to configure github pre-publishing steps so that jsii projects can properly prepare the working directory for the expected paths of the release job.

Fixes https://github.com/projen/projen/issues/1431

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,6 +59,10 @@ jobs:
         with:
           name: build-artifact
           path: dist
+      - name: Prepare Repository
+        run: mv dist .repo
+      - name: Collect GitHub Metadata
+        run: mv .repo/dist dist
       - name: Release
         run: errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R
           $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt)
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -59,6 +59,10 @@ jobs:
         with:
           name: build-artifact
           path: dist
+      - name: Prepare Repository
+        run: mv dist .repo
```
iliapolo, line 63: Same as we do for all other publishing jobs in jsii case.

.github/workflows/release.yml
```diff
@@ -59,6 +59,10 @@ jobs:
...
+      - name: Collect GitHub Metadata
+        run: mv .repo/dist dist
```
iliapolo, line 65: This will created a working directory of:

```console
dist/version.txt
dist/releasetag.txt
dist/changelog.md
```

eladb (approved):
Meh

src/release/publisher.ts
```diff
@@ -214,7 +225,7 @@ export class Publisher extends Component {
       return {
         name: 'github',
         registryName: 'GitHub Releases',
-        prePublishSteps: options.prePublishSteps ?? [],
+        prePublishSteps: options.prePublishSteps ?? this._gitHubPrePublishing,
```
iliapolo, line 228: @eladb wondering if you think a merge is more appropriate here.

iliapolo:
> Meh

I'm comforted by the fact the more `meh` we add the more red we should see when we refactor this :)

src/release/publisher.ts
eladb: yes, probably.

PR #1435, opened 2021-12-30
feat(node): do run "yarn/npm install" if package.json did not change

eladb:
Run `yarn/npm install` in post-synthesis only if `package.json` had actually changed. Otherwise, there is no need.

If `node_modules` doesn't exist in the project root, we also install to allow users to simply run `npx projen` from freshly cloned repositories.

[!](https://user-images.githubusercontent.com/598796/147762341-5cc24fa8-bf53-4f80-88c1-be632391a3fa.mov)

---


Chriscbr (approved):
Woah very cool!

One thought I had was whether there should be some kind of "--force" flag to override this behavior. But this might first require formalizing this post-installation behavior as an API.

I also wonder if it makes sense to also extend this behavior to the projenrc file? But for this to be reliable we would also probably have to check that tasks.json did not change and the projen dependency defined in `package.json` did not change.

eladb:
> One thought I had was whether there should be some kind of "--force" flag to override this behavior. But this might first require formalizing this post-installation behavior as an API.

I guess one can just run `yarn --force` no?

PR #1450, opened 2022-01-02
feat(eslint): use prettier if enabled

eladb:
When `prettier` is enabled, automatically configure `Eslint` to use it for code formatting, while disabling any rules that may conflict. Previously, users would have had to explicitly set `prettier: true` both at the project level and within `eslintOptions`.

BREAKING CHANGE: Prettier configuration is now under a `settings` option instead (instead of `prettierOptions: { bracketSpacing: false }`, write `prettierOptions: { settings: { bracketSpacing: false } }`).
* **prettier:** `prettier.config` is now called `prettier.settings` and DOES NOT include `overrides`.
* **node:** The `prettierIgnoreEnabled` option is now under `prettierOptions.ignoreFile` and can be accessed via `prettier.ignoreFile`. Similarly, `project.addPrettierIgnore()` is now `project.prettier.addIgnorePattern()`.

---


src/javascript/eslint.ts
```diff
@@ -356,11 +341,17 @@ export class Eslint extends Component {
...
-      rules: this.rules,
+      rules: () => ({ ...this._formattingRules, ...this.rules }),
       overrides: this.overrides,
     };
 
     new JsonFile(project, '.eslintrc.json', { obj: this.config, marker: false });
+
+    // if the user enabled prettier explicitly _or_ if the project has a
+    // `Prettier` component, we shall tweak our configuration accordingly.
+    if (options.prettier || Prettier.of(project)) {
```
iliapolo, line 352: Thats pretty cool

iliapolo (approved):

PR #1453, opened 2022-01-02
chore: make projen pretty(ier)

eladb:
Enable `prettify` for this project, and reformat all source files.

I contemplated if we should configure prettify to align a bit more closely with the previous eslint formatting rules (a glaring difference is quotes versus double-quotes), but I am thinking to just rip the band-aid and accept the defaults.

---


.prettierignore
```diff
@@ -0,0 +1 @@
+# ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
```
Chriscbr, line 1: Can we avoid generating this file if it's empty?

Chriscbr (approved):

Chriscbr:
Whoops, I probably should have approved this PR first. Feel free to submit a new PR / override as needed

.prettierignore
eladb: I'll follow up.

Issue #1179, opened 2021-10-27
Projen marker incorrectly refers to projenrc.js


Chriscbr:
Even if a project uses `.projenrc.ts`, the marker inserted in projen-generated files still says:

```Generated by projen. To modify, edit .projenrc.js and run "npx projen".```

pgollucci:
src/file.ts
4:import { PROJEN_MARKER, PROJEN_RC } from "./common";
92:        : `${PROJEN_MARKER}. To modify, edit ${PROJEN_RC} and run "npx projen".`;
184:    const containsMarker = contents?.includes(PROJEN_MARKER);


closable

mrgrain:
Or simply change to text to refer to `projenrc` and docs

mrgrain:
Fixed!

PR #1459, opened 2022-01-03
feat: eject from projen

### LINKED ISSUE #596: feature request: eject out of projen

Chriscbr:
Closes #596

Adds support for ejecting from projen with the `eject` task. See docs/eject.md for more information.

Note that ejecting comes with the limitation that not all scripts in `package.json` may work afterwards. Specifically, those that have "builtin" steps, referring to files in projen, are no longer valid since projen is uninstalled after ejecting. This also affects GitHub workflows that require on these tasks, such as GitHub workflows for bumping / releasing new versions of packages.

Fixing this would probably require some more complex bundling and/or changes to the "builtins" task feature.

In addition, `projen eject` has not been tested with non-node projects yet.

TODO:
- [x] add tests
- [x] add docs/eject.md
- [x] ~manually test that "ejected" github workflows still work~ (this is a bit too complicated at the moment, see comments above)

BREAKING CHANGE: ReactProject's `eject` task now ejects the project from projen instead of ejecting from React.
* `Project.defaultTask` is now optional as it may be undefined if a project is being ejected.
* `FileBase.PROJEN_MARKER` is now deprecated, please use `FileBase.marker` instead. This value will be undefined if the project is being ejected.

---


src/typescript/typescript.ts
```diff
@@ -393,11 +393,10 @@ export class TypeScriptProject extends NodeProject {
...
     };
 
-    const resolver = new TextFile(
-      this,
-      path.posix.join(PROJEN_DIR, "jest-snapshot-resolver.js")
-    );
-    resolver.addLine(`// ${TextFile.PROJEN_MARKER}`);
+    const resolver = new TextFile(this, "jest-snapshot-resolver.js");
+    if (!IS_PROJEN_EJECTING) {
+      resolver.addLine(`// ${TextFile.PROJEN_MARKER}`);
```
eladb, line 398: How about we get rid of the `PROJEN_MARKER` const and expose `project.marker?` which will be undefined if we are ejecting? This will intrinsically signal that the market shouldn't be emitted.

src/awscdk/lambda-function.ts
```diff
@@ -1,5 +1,6 @@
 import { basename, dirname, extname, join, relative, sep, posix } from "path";
 import { pascal } from "case";
+import { IS_PROJEN_EJECTING } from "../common";
```
eladb, line 3: Instead of a global const, let's propagate this through `project.ejected`.

See my general comment about modeling eject as a desired state as oppose to an action.

src/component.ts
```diff
@@ -22,4 +22,9 @@ export class Component {
    * Called after synthesis. Order is *not* guaranteed.
    */
   public postSynthesize() {}
+
+  /**
+   * Called when `projen eject` is run, after synthesizing all files.
+   */
+  public eject() {}
```
eladb, line 29: Not sure this is needed. Projen is a desired state system. As such it's preferable to model things as state as oppose to actions.

src/file.ts
```diff
@@ -34,6 +35,14 @@ export interface FileBaseOptions {
...
    */
   readonly executable?: boolean;
+
+  /**
+   * Adds the projen marker to the file. This should be managed by the class
+   * that subclasses from FileBase.
+   *
+   * @default true
+   */
+  readonly marker?: boolean;
```
eladb, line 45: Instead of making this an input option, I think we should just expose this as a property that includes the marker string itself, and it will be undefined if `project.ejected` is true.

src/file.ts
```diff
@@ -143,9 +165,39 @@ export abstract class FileBase extends Component {
       executable: this.executable,
     });
 
+    this.checkForProjenMarker();
+
     this._changed = true;
   }
 
+  private checkForProjenMarker() {
```
eladb, line 173: Why is this needed?

eladb:
I think I would have approached this slightly differently. Instead of treating "eject" as an action, I would model it as a project properly (`ejected`) and then downstream code can just reflect on this property and make decisions such as inclusion of the marker or whether a file needs to be read/write or readonly.

eladb (changes requested):

Chriscbr:
> I think I would have approached this slightly differently. Instead of treating "eject" as an action, I would model it as a project properly (`ejected`) and then downstream code can just reflect on this property and make decisions such as inclusion of the marker or whether a file needs to be read/write or readonly.

Wow yeah that sounds way simpler, thanks for the suggestion! I'll update this PR with take 2 when I have a chance 👍 

When I started this I thought adding an `eject()` method on components would have an upside of "localizing" all of the eject-related logic to one place, but after working through the changes, I noticed in some places that just wasn't really practical (e.g. changing the contents of `SourceCode` files after the file was synthesized). So I think in addition to dropping that, accessing the state through `project` makes a lot more sense as well.

src/typescript/typescript.ts
Chriscbr: Hmm, interesting idea - if the marker string is within `Project`, I think it might also be easier to fix https://github.com/projen/projen/issues/1179.

src/file.ts
Chriscbr: I've added a marker property within `Project`.

I could be off mark (pun intended?), but I think `marker` should still be an overridable option in files so that the marker can be "skipped" where needed. For example, `eslint` throws an error if `.eslintrc.json` has a field `"//": "Generated by projen..."`, so we configure the `JsonFile` with `marker: false`. In this PR I moved the property from `ObjectFile` to `FileBase` class since I thought it made more sense from a data modeling perspective. WDYT?

src/file.ts
Chriscbr: Mmm... I added this to try and make it easier to discover projen-generated files with missing markers (or that files that stil contain markers after ejecting). I'm ok with taking this out, though it could helpful for debugging, lmk what you think.

I've since updated the logging messages to be a bit less complainy

src/task-runtime.ts
```diff
@@ -26,7 +25,7 @@ export class TaskRuntime {
 
   constructor(workdir: string) {
     this.workdir = resolve(workdir);
-    const manifestPath = join(this.workdir, Tasks.MANIFEST_FILE);
+    const manifestPath = join(this.workdir, ".projen/tasks.json");
```
Chriscbr, line 28: Inlining here saves us from importing the `Tasks` class when this file gets bundled by esbuild

src/file.ts
eladb: Yes absolutely

src/task-runtime.ts
eladb: Would still be nice to avoid the duplicate. Would it make sense to move this const here and import it from `tasks.ts`?

src/typescript/typescript.ts
```diff
@@ -393,11 +393,10 @@ export class TypeScriptProject extends NodeProject {
       return path.join(dir, filename);
     };
 
-    const resolver = new TextFile(
-      this,
-      path.posix.join(PROJEN_DIR, "jest-snapshot-resolver.js")
-    );
-    resolver.addLine(`// ${TextFile.PROJEN_MARKER}`);
+    const resolver = new TextFile(this, ".projen/jest-snapshot-resolver.js");
```
eladb, line 396: Why not use `PROJEN_DIR`?

src/typescript/typescript.ts
```diff
@@ -393,11 +393,10 @@ export class TypeScriptProject extends NodeProject {
...
+    if (!this.ejected) {
+      resolver.addLine(`// ${this.marker}`);
+    }
```
eladb, line 399:
```suggestion
    if (!this.marker) {
      resolver.addLine(`// ${this.marker}`);
    }
```

src/web/react.ts
```diff
@@ -215,7 +215,9 @@ export class ReactComponent extends Component {
       project.addFields({ "config-overrides-path": overridesPath });
 
       const configOverrides = new SourceCode(this.project, overridesPath);
-      configOverrides.line(`// ${FileBase.PROJEN_MARKER}`);
+      if (!project.ejected) {
```
eladb, line 218:
```suggestion
      if (!project.marker) {
```

src/web/react.ts
```diff
@@ -243,12 +245,6 @@ export class ReactComponent extends Component {
 
     project.compileTask.exec(`${reactScripts} build`);
 
-    project.addTask("eject", {
```
eladb, line 246: Should we still make this available through `react:eject` or somethin?

eladb (changes requested):

Chriscbr:
Pls keep

src/web/react.ts
Chriscbr: I'm not sure if this makes sense since the `ReactProject` type is currently designed around using Create React App (adding react-scripts as a dep, setting up react-rewired, etc.), so if you _actually_ ran `react:eject`, projen would still think it's a CRA project and add the react-scripts dependency back etc. and it would be an awkward experience.

src/typescript/typescript.ts
Chriscbr: I can't remember why I did this. Reverted

src/project.ts
```diff
@@ -446,9 +481,43 @@ export class Project {
...
+  public get ejected(): boolean {
+    return this._ejected;
+  }
+
+  /**
+   * The projen marker, used to identify files as projen-generated.
+   *
+   * Value is undefined if the project is being ejected.
+   */
+  public get marker(): string | undefined {
```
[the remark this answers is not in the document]
Chriscbr: Hmm... I can see what you mean. I've made the change.

The only case where it's a bit awkward is if you're trying to use the marker while constructing the actual file class, as in `new SampleFile(p, { contents: [this.marker] } )`. But I'm OK with that tradeoff

src/project.ts
```diff
@@ -148,9 +154,18 @@ export class Project {
...
-  public readonly defaultTask: Task;
+  public readonly defaultTask?: Task;
+
+  /**
+   * This task ejects the project from projen. This is undefined if the project
+   * it self is being ejected.
+   *
+   * See docs for more information.
+   */
+  public readonly ejectTask?: Task;
```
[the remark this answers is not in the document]
Chriscbr: Updated

docs/eject.md
```diff
@@ -0,0 +1,30 @@
+# Ejecting
+
+If you aren't satisfied with managing your project configuration with projen,
+you can `eject` at any time.
```
eladb, line 4: Judgement isn't needed :-)

```suggestion
To stop from using projen to manage your project configuration,
you can `eject` at any time and continue to manage your configuration manually.
```

.projenrc.js
```diff
@@ -179,22 +181,48 @@ project.npmignore.exclude("/SECURITY.md");
...
+  integTask.spawn(project.tasks.tryFind("package:python"));
+  integTask.spawn(pythonCompatTask);
+
+  project.buildWorkflow.addPostBuildJobTask(integTask, {
+    tools: { python: { version: "3.x" }, go: { version: "1.16.x" } },
+  });
+}
+
+// build `run-task` script needed for "projen eject" functionality
+function setupBundleTaskRunner() {
```
eladb, line 203: Curious, could we not use `Bundler` for this?

src/file.ts
```diff
@@ -73,10 +81,16 @@ export abstract class FileBase extends Component {
...
-    this.readonly = options.readonly ?? true;
+    this.readonly = !project.ejected && (options.readonly ?? true);
     this.executable = options.executable ?? false;
     this.path = filePath;
 
+    // `marker` is empty if project is being ejected or if explicitly disabled
+    this.marker =
+      project.ejected || options.marker === false
+        ? undefined
+        : `${PROJEN_MARKER}. To modify, edit ${PROJEN_RC} and run "npx projen".`;
```
eladb, line 92: Maybe in the future we can ask the `project` about the name of the projenrc file (e.g. when it's not a .js file).

eladb (approved):
Great job!

docs/eject.md
Chriscbr: I was thinking of this sort of how a 5-star hotel might tell you "If you aren't totally satisfied with our service, we will gladly give you a complete refund" -- but I see what you mean about keeping things factual and to the point 👍

PR #1466, opened 2022-01-03
fix(awscdk): permission denied error for Lambda function with `prettier` option

### LINKED ISSUE #1457: Permission denied error for Lambda function with `prettier` option
I enabled the `prettier` option which was introduced in #1450. My projects includes a [Lambda function](https://github.com/projen/projen/blob/main/docs/awscdk.md#aws-lambda-functions). The Lambda Function construct is generated by projen (as a readonly file). 

`projen build` shows this error:

```bash
👾 build » test » eslint | eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern src test build-tools .projenrc.js

Oops! Something went wrong! :(

ESLint: 8.6.0

Error: EACCES: permission denied, open '/Users/julian/dev/projen-prettier-lambda/src/example-function.ts'
👾 Task "build » test » eslint" failed when executing "eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern src test build-tools .projenrc.js" (cwd: /Users/julian/dev/projen-prettier-lambda)
```

The generated CDK construct for the Lambda function is not formatted correctly. However, the file cannot fixed because it is readonly.

Projen version: 0.50.0

Example project: https://github.com/jumic/projen-prettier-lambda

Ideas:
- Change readonly files to writeable when projen executes eslint/prettier.
- Generate valid file in [lambda-function.ts](https://github.com/projen/projen/blob/2ef39f895b255dc4c844947f29fb0568eb6f4088/src/awscdk/lambda-function.ts#L163-L205) if `prettier` is enabled.

jumic:
Prettier cannot format the generated Lambda construct file because it is a readonly file.
If `prettier` is enabled, the source code file must be be added to `.prettierignore` to exclude it from prettier.

This PR adds a `preSynthesize` step for prettier that will add all `SourceCode` files to the `.prettierignore` file.

Fixes #1457.

---


src/awscdk/lambda-function.ts
```diff
@@ -204,6 +204,8 @@ export class LambdaFunction extends Component {
     src.close("}");
     src.close("}");
 
+    Prettier.of(this.project)?.addIgnorePattern(constructFile);
```
eladb, line 207: Wondering if this should be the other way around. I was thinking that the Prettier component can search simply add all `SourceFile`s to its ignorefile.

P.s this will need to happen in a preSynth hook.

eladb (changes requested):

src/awscdk/lambda-function.ts
jumic: Yes, probably that's the better solution. This solution can add all `SourceCode` files to `.prettierignore` and is not specific to Lambda functions source code.

eladb (approved):
Looks great

Issue #1256, opened 2021-11-25
Escape hatches do not work for `UpgradeDependencies` workflow or task


Chriscbr:
Related to #1103, #1250

PR #1476, opened 2022-01-03
feat(javascript): allow users to run post-upgrade tasks

### LINKED ISSUE #1348: upgrade workflow doesn't update integration test snapshots for bundled lambdas
When an AwsCdkConstructLibrary self-updates by cron, the PR doesn't catch integration test snapshot changes.

This is undesirable if you're using an integration test with an asset derived from the lambda bundling feature. When said asset depends on an external dependency, an update to this dependency will likely change the asset hash. When this happens, the upgrade workflow succeeds, producing a PR that fails to build when it tries to assert the integration test snapshots:

image

misterjoshua:
Allows users to add post-upgrade tasks. This is useful in case the user wants to run commands after the upgrade workflow, such as automatically updating integration test snapshots. Example:

```ts
project.upgradeWorkflow?.postUpgradeTask.spawn(project.tasks.tryFind('bundle'));
project.upgradeWorkflow?.postUpgradeTask.spawn(project.tasks.tryFind('integ:my-test:snapshot'));
```

Fixes #1348
Related #1256

---


src/javascript/upgrade-dependencies.ts
```diff
@@ -122,6 +127,12 @@ export class UpgradeDependencies extends Component {
     this.containerOptions = options.workflowOptions?.container;
 
     project.addDevDeps("npm-check-updates@^12");
+
+    this.postUpgradeTask =
+      project.tasks.tryFind("post-upgrade") ??
+      project.tasks.addTask("post-upgrade", {
+        description: "Runs after upgrading dependencies",
+      });
```
misterjoshua, line 135: I use `tryFind` here because `NodeProject` creates two instances of `UpgradeDependencies` in many `NodeProject`-based integration tests because projen [detects itself here.](https://github.com/projen/projen/blob/ee76669988ebf04a47367debcd00031473339aac/src/javascript/node-project.ts#L716)

I had considered moving the creation of this task up to `NodeProject`, but wasn't sure whether I should follow form with how `post-compile` is added to projects via `ProjectBuild`, since it has a similar nature to `post-upgrade` in `UpgradeDependencies`.

misterjoshua:
@Chriscbr Hey. Any ideas what's needed to move this PR along, discussion, code-change, or otherwise?

src/javascript/node-project.ts
```diff
@@ -455,6 +455,13 @@ export class NodeProject extends GitHubProject {
    */
   public readonly upgradeWorkflow?: UpgradeDependencies;
 
+  /**
+   * A task run after the upgrade task.
+   */
+  public get postUpgradeTask() {
+    return this.upgradeWorkflow?.postUpgradeTask;
+  }
```
Chriscbr, line 463: Let's remove this since the `upgradeWorkflow` property is public anyways - I'd prefer to minimize the API surface area we're adding.

src/javascript/upgrade-dependencies.ts
```diff
@@ -197,6 +208,8 @@ export class UpgradeDependencies extends Component {
     // run "projen" to give projen a chance to update dependencies (it will also run "yarn install")
     task.exec(this._project.projenCommand);
 
+    task.spawn(this.postUpgradeTask);
```
Chriscbr, line 211: Can we add a unit test? I think we just want to validate that is there is some way to run things after the upgrade task.

Chriscbr (changes requested):
Hi, thanks for the ping. I took one more look to see if there's any quick fixes for the upgrade task not being available for escape hatching, but since I couldn't see any, I'm cool with adding this for now. Just a few small comments 🙂

src/javascript/node-project.ts
misterjoshua: No problemo.

src/javascript/upgrade-dependencies.ts
misterjoshua: Okay, I've added a unit test that checks that the `upgrade` task spawns `post-upgrade` as its last step. In the test, I didn't add anything to the `post-upgrade` task, but I can if you feel it's necessary.

misterjoshua:
@Chriscbr Alright - I've made changes following your review. Let me know whether you need anything else! Thanks.

Chriscbr (approved):

Issue #1356, opened 2021-12-20
hard to initialize project with an older version of projen


iliapolo:
Its fairly complex to initialize a project with an older version of projen at the moment, i.e the non latest version at the time of initialization. 

Intuitively I would have though that doing the following would be suffice:

```console
npm install projen@0.28.19
./node_modules/.bin/projen new typescript
``` 

i.e, the projen dependency of the project will be set to the currently used version. 

However, the dependency is being set to:

```json
 "devDependencies": {
    "projen": "^0.45.3",
  },
```

This is because projen defaults to set itself as a dependency without any version specifier: 

https://github.com/projen/projen/blob/daf0a076f147969a55dcb493f939c9d4b64571f5/src/javascript/node-project.ts#L448

Causing the latest version to be used. I think this a bit counter intuitive. Was this the intent of the feature? 

As a workaround, i'm currently explicitly setting the projen version in `.projenrc.js`:

```js
const path = require('path');
const projenVersion = require(path.join(require.resolve('projen'), '..', '..', 'package.json')).version;

const project = new JsiiProject({
  projenVersion,
});
```

For some more context, my use case is that the `.projenrc.js` file is provided to me externally, from an environment with its own projen version, and i'd like to use the same version to generate the new project. Otherwise I might end up with conflicting versions and its going to be hard to maintain.

iliapolo:
A straightforward solution would be to simply apply the suggested workaround as the default behavior.

@eladb thoughts?

eladb:
Can you simply pass `--projen-version` to `projen new`?

iliapolo:
I actually only used `projen new` for easy repro. My workflow is bootstrapping a project from an existing projen file that is managed in an environment that has its own projen version. So not really... :\

Chriscbr:
> For some more context, my use case is that the .projenrc.js file is provided to me externally, from an environment with its own projen version, and i'd like to use the same version to generate the new project. Otherwise I might end up with conflicting versions and its going to be hard to maintain.

Could you copy the `.projenrc.js` and `package.json` into your new directory and then `npm i`? Or create your own project type and use the `Projects.createProject()` API programmatically maybe?

----

On one hand, there's a lot of cases where projen [the CLI] initializing projects the latest version of projen [the library] by default seems like a good thing, since it means people start off with the latest fixes and features, even if the CLI is out of date.

[...]

PR #1480, opened 2022-01-04
fix: exclude projen in upgrades when version is specified

dontirun:
Also fixes  https://github.com/projen/projen/issues/1356#issuecomment-1005046584 

---


eladb:
@iliapolo can you take a look?

src/javascript/node-project.ts
```diff
@@ -455,6 +456,11 @@ export class NodeProject extends GitHubProject {
    */
   public readonly upgradeWorkflow?: UpgradeDependencies;
 
+  /**
+   * The packages that are ignored during the upgrade workflow.
+   */
+  public readonly ignoredUpgrades: string[] = [];
```
iliapolo, line 462: This functionality actually already exists:

https://github.com/projen/projen/blob/1e0bdedcd2e3a0366ef3143d7b7207bf1d58d4b1/src/javascript/upgrade-dependencies.ts#L23-L29

You can configure it on a node project via:

```ts
depsUpgradeOptions: {
  exclude: ['projen']
}
```

src/javascript/node-project.ts
```diff
@@ -709,6 +715,14 @@ export class NodeProject extends GitHubProject {
       );
       ignoresProjen = upgradeDependencies.ignoresProjen;
       this.upgradeWorkflow = upgradeDependencies;
+      new YamlFile(this, ".ncurc.yml", {
+        obj: {
+          reject: this.ignoredUpgrades,
+        },
+      });
+      if (options.projenVersion) {
```
iliapolo, line 723: I agree that if a user specified `projenVersion` we should add `projen` to the ignore list. But this should be done here:

https://github.com/projen/projen/blob/1e0bdedcd2e3a0366ef3143d7b7207bf1d58d4b1/src/javascript/node-project.ts#L693

iliapolo (changes requested):
@dontirun Thanks for this. See my comments. 

We should repurpose this PR to simply ignore `projen` by default if `projenVersion` is defined. Also this needs a test.

src/javascript/node-project.ts
dontirun: I completely missed that 🤦. I'll redothe PR

src/javascript/node-project.ts
dontirun: If we specify it here we should make sure that it's not overwritten by the [deepMerge](
https://github.com/dontirun/projen/blob/1588d8c1ea69fd536ef1ab03e7df441850497d1f/src/javascript/node-project.ts#L714).

dontirun:
@iliapolo Is this something that is still wanted/needed?

src/javascript/node-project.ts
```diff
@@ -703,16 +703,31 @@ export class NodeProject extends GitHubProject {
           gitIdentity: this.workflowGitIdentity,
         },
       };
+      const depsUpgradeOptions = options.depsUpgradeOptions ?? {};
+      const projenExclude = depsUpgradeOptions?.exclude ?? [];
+      if (options.projenVersion && !projenExclude.includes("projen")) {
+        projenExclude.push("projen");
+      }
```
iliapolo, line 710: This feels a bit cluttered. As I see it, all we need to do is provide a default value to `ignoreProjen`, both in the dependabot config and the dep upgrade. 

So, I think we just need to add: 

```ts
ignoreProjen: !!options.projenVersion,
```

Here:

https://github.com/projen/projen/blob/fafd5056e572185359e19d50b685f37ed496d1aa/src/javascript/node-project.ts#L693-L695

And here:

https://github.com/projen/projen/blob/fafd5056e572185359e19d50b685f37ed496d1aa/src/javascript/node-project.ts#L703-L705

No?

src/javascript/node-project.ts
```diff
@@ -703,16 +703,31 @@ export class NodeProject extends GitHubProject {
...
+        ])
       );
       ignoresProjen = upgradeDependencies.ignoresProjen;
       this.upgradeWorkflow = upgradeDependencies;
     }
 
     // create a dedicated workflow to upgrade projen itself if needed
-    if (ignoresProjen && this.package.packageName !== "projen") {
+    if (
+      !options.projenVersion &&
```
iliapolo, line 727: If we apply the above suggestion, we won't need to add this condition here.

iliapolo (changes requested):
@dontirun Sorry for letting this hang for a while. 

See my comments.

PR #1494, opened 2022-01-08
fix(cleanup): Only clean up orphaned files

### LINKED ISSUE #1484: Files managed by projen are deleted and recreated
When I run `projen` in the root of the repo, it seems like all of the files under projen management get deleted, and then re-created again.

This causes VS Code's Source Control panel to generate a massive diff of all of the files it discovers, of which there are many, as when the `.gitignore` file gets deleted, VS Code scans the entire `node_modules` dir.

I'm using projen with sub-project, which is the main divination from the norm. I am not sure if this is the relevant part, or if it's just causing the issue due to the number of files under projen's management.

MarkMcCulloh:
To avoid unnecessary thrashing of projen managed files, we can skip cleaning up files that we know we're just going to be creating again during synthesis. This means that changed files will simply be overwritten during synthesis. Even better, unchanged files won't be written at all.
With this PR, an orphaned file is one that is present in `files.json` (from the previous synthesis) but not one of the files that are about to be synthesized.

Included some refactoring in `cleanup.ts` to make the control flow more (hopefully) clear with this additional logic.

fixes #1484 

---

### DIFF
--- a/src/cleanup.ts
+++ b/src/cleanup.ts
@@ -7,22 +7,28 @@ import * as logging from "./logging";
 
 export const FILE_MANIFEST = `${PROJEN_DIR}/files.json`;
 
-export function cleanup(dir: string, exclude: string[]) {
+export function cleanup(dir: string, newFiles: string[], exclude: string[]) {
   try {
-    for (const f of findGeneratedFiles(dir, exclude)) {
-      fs.removeSync(f);
+    const manifestFiles = getFilesFromManifest(dir);
+    if (manifestFiles.length > 0) {
+      // Use `FILE_MANIFEST` to remove files that are no longer managed by projen
+      cleanOrphanedFiles(dir, manifestFiles, newFiles);
+    } else {
+      // Remove all files managed by projen
+      removeFiles(findGeneratedFiles(dir, exclude));
     }
   } catch (e) {
     logging.warn(`warning: failed to clean up generated files: ${e.stack}`);
   }
 }
 
-function findGeneratedFiles(dir: string, exclude: string[]) {
-  let files = getFilesFromManifest(dir);
-  if (files.length > 0) {
-    return files;
+function removeFiles(files: string[]) {
+  for (const file of files) {
+    fs.removeSync(file);
   }
+}
 
+function findGeneratedFiles(dir: string, exclude: string[]) {
   const ignore = [
     ...readGitIgnore(dir),
     "node_modules/**",
@@ -30,7 +36,7 @@ function findGeneratedFiles(dir: string, exclude: string[]) {
     ".git/**",
   ];
 
-  files = glob.sync("**", {
+  const files = glob.sync("**", {
     ignore,
     cwd: dir,
     dot: true,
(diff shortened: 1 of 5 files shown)


src/cleanup.ts
```diff
@@ -7,30 +7,36 @@ import * as logging from "./logging";
...
-    for (const f of findGeneratedFiles(dir, exclude)) {
-      fs.removeSync(f);
+    const manifestFiles = getFilesFromManifest(dir);
+    if (manifestFiles.length > 0) {
+      // Use `FILE_MANIFEST` to remove files that are no longer managed by projen
+      cleanOrphanedFiles(dir, manifestFiles, newFiles);
+    } else {
+      // Remove all files managed by projen with legacy logic
+      removeFiles(findGeneratedFiles(dir, exclude));
     }
```
Chriscbr, line 19: Can we normalize these branches like so:
```suggestion
    if (manifestFiles.length > 0) {
      // Use `FILE_MANIFEST` to remove files that are no longer managed by projen
      removeFiles(findOrphanedFiles(dir, manifestFiles, newFiles));
    } else {
      // Remove all files managed by projen with legacy logic
      removeFiles(findGeneratedFiles(dir, exclude));
    }
```

src/project.ts
```diff
@@ -426,8 +426,12 @@ export class Project {
...
     }
 
-    // delete all generated files before we start synthesizing new ones
-    cleanup(outdir, this.excludeFromCleanup);
+    // delete orphaned files before we start synthesizing new ones
+    cleanup(
+      outdir,
+      this.files.map((f) => f.path.replace(/\\/g, "/")),
+      this.excludeFromCleanup
+    );
```
Chriscbr, line 434: (optional) I wonder how this will interact with subprojects. Is there any way we can test this perhaps?

src/project.ts
```diff
@@ -426,8 +426,12 @@ export class Project {
       this.addExcludeFromCleanup(subproject.outdir + "/**");
     }
 
-    // delete all generated files before we start synthesizing new ones
-    cleanup(outdir, this.excludeFromCleanup);
+    // delete orphaned files before we start synthesizing new ones
+    cleanup(
+      outdir,
+      this.files.map((f) => f.path.replace(/\\/g, "/")),
```
Chriscbr, line 432: Can we add a unit test for the scenario where a user:
- creates a `File` with content
- runs synth to generate the actual file
- modifies the `File` to have empty contents
- runs synth again so the file is cleaned up per the snippet below (even though it may still be listed under `this.files`)

https://github.com/projen/projen/blob/287b59ad82d8ae9ea83dd16b460db63a4f2c93e4/src/file.ts#L128-L131

src/project.ts
MarkMcCulloh: I suppose I can add another cleanup test with a subproject, but I'm not really sure what to test for since the handling of orphaned files is on a per-`Project` basis instead of the globbing method used previously.

src/project.ts
MarkMcCulloh: Ah, yeah that was an orphan situation this PR did not handle. I'm not sure how it could, because cleanup happens before file synthesis. 
[I added `removeSync` to that snippet](https://github.com/projen/projen/pull/1494/files#diff-bc9705d0f7a567399044dfc66ccc82d4d9aa1cff116842a0094d54e463c9ecbcR129-R134) to handle this case and a test for it as well. Philosophically, this isn't so much a cleanup as it is just another way to "synthesize" the undefined content to disk.

Chriscbr (approved):
LGTM!

PR #1496, opened 2022-01-08
feat(awscdk): allow users to integration test multi-stack stages

### LINKED ISSUE #1495: feat(awscdk): Support multi-stack CDK stages in integration tests.
I'd like to write an integration test for a multi-stack CDK `Stage` in a Projen AWS CDK Typescript App. When deploying the stacks in these stages via the CDK CLI, the CLI requires one to provide the stack names. (`--all` is insufficient.) However, Projen's `IntegrationTest` doesn't support this. I'd like an option that allows me to specify which stacks to operate on. i.e., `my-stage/*`

misterjoshua:
Allows users to add AWS CDK integration tests consisting of stages with multiple stacks.

```ts
new awscdk.IntegrationTest(project, {
  name: "my-stage",
  entrypoint: "test/my-stage.myinteg.ts",
  stacks: ["my-stage/*"],
  tsconfigPath: project.tsconfigDev.fileName,
  cdkDeps: project.cdkDeps,
});
```

Fixes #1495

---

### DIFF
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -1,10 +1,10 @@
-import { basename, dirname, join } from "path";
+import { dirname, join } from "path";
 import { Component } from "../component";
 import { DependencyType } from "../dependencies";
 import { Project } from "../project";
 import { Task } from "../task";
 import { AwsCdkDeps } from "./awscdk-deps";
-import { FEATURE_FLAGS, TYPESCRIPT_INTEG_EXT } from "./internal";
+import { FEATURE_FLAGS } from "./internal";
 
 export interface IntegrationTestCommonOptions {
   /**
@@ -21,9 +21,8 @@ export interface IntegrationTestCommonOptions {
 export interface IntegrationTestOptions extends IntegrationTestCommonOptions {
   /**
    * Name of the integration test
-   * @default - Derived from the entrypoint by removing ".integ.ts"
    */
-  readonly name?: string;
+  readonly name: string;
 
   /**
    * A list of stacks within the integration test to deploy/destroy.
@@ -84,7 +83,7 @@ export class IntegrationTest extends Component {
   constructor(project: Project, options: IntegrationTestOptions) {
     super(project);
     const entry = options.entrypoint;
-    const name = options.name ?? basename(entry, TYPESCRIPT_INTEG_EXT);
+    const name = options.name;
     const dir = dirname(entry);
 
     const deploydir = join(dir, ".tmp", `${name}.integ`, "deploy.cdk.out");
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -20,6 +20,18 @@ export interface IntegrationTestCommonOptions {
  */
 export interface IntegrationTestOptions extends IntegrationTestCommonOptions {
   /**
+   * Name of the integration test
+   * @default - Derived from the entrypoint by removing ".integ.ts"
+   */
+  readonly name?: string;
+
+  /**
+   * A list of stacks within the integration test to deploy/destroy.
+   * @default - The CLI is not told any one stack in particular
+   */
+  readonly stacks?: string[];
+
+  /**
    * A path from the project root directory to a TypeScript file which contains
    * the integration test app.
    *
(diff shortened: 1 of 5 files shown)


src/awscdk/integration-test.ts
```diff
@@ -19,6 +19,17 @@ export interface IntegrationTestCommonOptions {
  * Options for `IntegrationTest`.
  */
 export interface IntegrationTestOptions extends IntegrationTestCommonOptions {
+  /**
+   * Name of the integration test
+   */
+  readonly name: string;
```
misterjoshua, line 25: I've moved responsibility for naming the integration test to `IntegrationTest`'s creator. Here's why: If the user wants to create an integration test with specific `stacks`, they can't currently use `AutoDiscover`, as there's no mechanism to provide additional options on a per-test basis, and all of the task commands are rendered at construction time. But, if a user tries to manually add `IntegrationTest` to the project, then either: AutoDiscover finds it first due to the `**/*.integ.ts` glob and the user's `IntegrationTest` conflicts, or the user has to provide a different filename like `test/my-stage.myinteg.ts` and `IntegrationTest` fails to derive the name of the integration test.

misterjoshua:
@Chriscbr Here's another PR for you whenever you get a chance. This one helps integration test AWS CDK Stages.

src/awscdk/integration-test.ts
misterjoshua: Strictly speaking, this isn't the only solution. It should also be possible to implement a pragma system like the AWS CDK's main repository does with `!cdk-integ`, but where `IntegrationTest` parses a list of stacks to deploy from the entrypoint file before generating the build script. Something like:

**test/my-stage.integ.ts:**
```ts
const app = new cdk.App();
// ... add the constructs here
app.synth();

// Pragma follows:
// !awscdk-integ pragma:stacks=my-stage/*
```

Chriscbr:
I'll defer this one to Elad since I think he's a bit more familiar with these components 😅

src/awscdk/integration-test.ts
```diff
@@ -19,6 +19,17 @@ export interface IntegrationTestCommonOptions {
...
+
+  /**
+   * A list of stacks within the integration test to deploy/destroy.
+   * @default - The CLI is not told any one stack in particular
```
eladb, line 29: Can we make the default `"*"`? I believe this will deploy all stacks.

src/awscdk/integration-test.ts
eladb, line 25: Would it make sense to keep this optional with the current behavior as the default?

eladb (changes requested):

src/awscdk/integration-test.ts
misterjoshua: @eladb We can make the default `"*"`, but if the integration test contains a stage with stacks in it, that wildcard does not locate the stacks in the stage. For the following example, you need to use `"my-stage/*"`:

```ts
import { App, Stack, Stage } from "@aws-cdk/core";

const app = new App();
const stage = new Stage(app, 'my-stage');
new Stack(stage, 'stack1');
new Stack(stage, 'stack2');
app.synth();
```

If you try to use `cdk deploy "*"` here, CDK CLI will say, `No stacks match the name(s) *`

Here's a project I threw together to demonstrate: https://github.com/misterjoshua/projen-multi-stack

image

I'm not sure if this is a CDK CLI bug or something else - I had assumed there was a reason for "*" not to match the stacks.

src/awscdk/integration-test.ts
misterjoshua: Sure, we can do that.

misterjoshua:
Bah. I botched the merge to update. I'm going to rebase this PR from main.

eladb (approved):

misterjoshua:
Thanks for the help! :)

src/awscdk/integration-test.ts
mmcculloh-dms: The logic for "*" is intentional, you can use "**" for all stacks of all stages. I believe `--all` does the same as well.

eladb: `**` sounds like the right direction

misterjoshua: I agree. I didn't know about that option!

misterjoshua: I just confirmed. :) 
image

misterjoshua: Here's a PR that changes the default to `**`: https://github.com/projen/projen/pull/1524

Issue #1511, opened 2022-01-11
Unable to retrieve git version in mac when creating a new project


erlanggatjhie:
When I run this command in my macbook:

```bash
npx projen new awscdk-app-ts
```

Projen throws the following error:

```bash
TypeError: Invalid Version: 2.24.3 (A G-128)
    at new SemVer
```

After investigating further, it looks like the default git installation in my mac produce a slightly different output when running `git version` (notice the `(Apple Git-128)`) which it didn't get cater for when we sanitise the output:

```bash
git version 2.24.3 (Apple Git-128)
```

PR #1512, opened 2022-01-11
fix(new): fix issue with default git installation on mac

erlanggatjhie:
Fixes https://github.com/projen/projen/issues/1511

This PR is about an attempt to fix an issue where projen fails to retrieve git version in mac osx.

---


src/util.ts
```diff
@@ -302,6 +302,16 @@ export function formatAsPythonModule(name: string) {
...
 }
 
+/**
+ * Extract git version number from command line
+ *
+ * @param gitVersionOutput the output from `git version` CLI
+ * @returns the version of git
+ */
+export function getGitVersion(gitVersionOutput: string) {
+  return gitVersionOutput.replace(/[a-z,/s]|\(.*\)/g, "").trim();
```
eladb, line 312: Wouldn't it make more sense to use a regular expression that matches the version instead of omitting the non-version parts?

Eg https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39

eladb (changes requested):

src/util.ts
Chriscbr: @eladb I'm not I understand the benefit of this version check / what we get by using the new `git init -b` syntax. What if we just called `git branch -M main` every time?

eladb: The reason I've changed this to use `-b` was that newer git versions emit an annoying warning that `master` is being deprecated. 

I am okay to revert though. Seems like it's not worth it

erlanggatjhie: I have updated the code as per your suggestion @eladb and decided just to use a simple regex there. However, if you think that the revert is the best option then I will just close this PR then :).

eladb (approved):

lrevnic:
This fixed also the windows issue as well https://github.com/projen/projen/issues/1518. Thank you!

PR #1514, opened 2022-01-11
fix: cdkVersion documents a default but is required

kaizen3031593:
I've changed it so that `cdkVersion` is not required. However, I'm not sure if there was any original intention to make it required, in which case, I can go back and just remove the `@default` documentation.

---


Chriscbr:
Ahh... yeah this is another projen quirk. The reason it documents a default is because the CLI tool has been set up to automatically provide values for required fields when you are creating a new project -- and our signal for what value to provide to the user initially is `@default`.

You can see an example of a special macro as default, which gets evaluated separately:

https://github.com/projen/projen/blob/a71ea69e42c0c02679db375912004ff5ea4ed8ad/src/project.ts#L30

Maybe we should make this semantic distinction clear by annotating these values with `@projendefault` or `@initValue` instead? cc @eladb

eladb:
Yes, I've originally considered using a different doc tag for initial values and decided to reuse default. I deeply regret that now. What do you think about `@initial`?

PR #1536, opened 2022-01-18
fix(cdk8s): import location should be in the rootDir

### LINKED ISSUE #1534: bug(cdk8s): spec import emits inaccessible k8s resources
When Projen sets up a cdk8s app project, the `import` task emits the generated API resources outside of the TypeScript `rootDir`, so they're not available from the rest of the app's source code per the default tsconfig.

**To reproduce:**
```
npx projen new cdk8s-app-ts
yarn run import
```

You'll find that cdk8s has emitted `./imports/k8s.ts`.

misterjoshua:
Changes the default `cdk8s import` location to `./src/imports` so that imported resource specs can be accessed from the rest of the cdk8s app.

Fixes #1534

BREAKING CHANGE: The `import` task in the `Cdk8sTypeScriptApp` project now outputs to the `src` directory by default instead of the project root.

---

### DIFF
--- a/src/cdk8s/cdk8s-app-ts.ts
+++ b/src/cdk8s/cdk8s-app-ts.ts
@@ -168,7 +168,7 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
 
     this.addTask("import", {
       description: "Imports API objects to your app by generating constructs.",
-      exec: "cdk8s import",
+      exec: "cdk8s import -o src/imports",
     });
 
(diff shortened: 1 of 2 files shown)


Chriscbr (approved):
Thank you @misterjoshua ! 😄

I'm going to err on the side of caution and mark this as a breaking change in the PR description, since it's possible that existing `Cdk8sTypescriptApp` users are expecting the imports to be generated in one place and now it will be in another.

PR #1544, opened 2022-01-19
feat(stale): skip issues with "backlog" label

eladb:
From now on, the stale manager will skip any issues or PRs that have the `backlog` label associated with them.

This can be configured through the `exemptLabel` option.

---


src/github/stale.ts
```diff
@@ -70,9 +70,16 @@ export interface StaleBehavior {
...
+   * @default "stale"
    */
   readonly staleLabel?: string;
+
+  /**
+   * Label which exempt an issue/PR from becoming stale. Set to `""` to disable.
+   *
+   * @default "backlog"
+   */
+  readonly exemptLabel?: string;
```
Chriscbr, line 82: Would it make sense to make this an array instead?

Chriscbr (approved):

src/github/stale.ts
eladb: Done

Issue #1439, opened 2021-12-31
Caching dependencies in GitHub Actions


moltar:
There appears to be no caching of the deps in the GitHub Action Workflows.

Shall this be added? This can greatly speed up runs.

eladb:
Yes that will be amazing. Contributions?

moltar:
Does this need to support all of the package managers, right? Any tips on where to dig in within projen.

I am somewhat familiar with the GH Action caching setup already.

Just new to projen.

moltar:
Can we use this for NPM? https://github.com/bahmutov/npm-install

moltar:
If this is ok, I will create a PR

Chriscbr:
Is it possible to use the built-in functionality for caching provided in the `setup-node` action? https://github.com/actions/setup-node#caching-packages-dependencies

moltar:
Would this be the place to change?

https://github.com/projen/projen/blob/ade63d68f35fae65d3a90a03496f843c682043ea/src/javascript/node-project.ts#L849-L863

I think we need to set the correct package manager based on the properties in the constructor.

PR #1547, opened 2022-01-20
feat: Caching dependencies in GitHub actions

Dzhuneyt:
Closes https://github.com/projen/projen/issues/1439

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,6 +91,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
+          cache: npm
       - name: Download build artifacts
         uses: actions/download-artifact@v2
         with:
@@ -117,6 +118,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
+          cache: npm
       - name: Download build artifacts
         uses: actions/download-artifact@v2
         with:
@@ -139,6 +141,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
+          cache: npm
       - uses: actions/setup-python@v2
         with:
           python-version: 3.x
@@ -164,6 +167,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
+          cache: npm
       - uses: actions/setup-go@v2
         with:
           go-version: ^1.16.0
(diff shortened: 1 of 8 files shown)


Chriscbr (changes requested):
I think this might not be as simple as a one line change. 😉

Could you share how you tested this change? I'd like to understand what this does before we merge it in -- perhaps a comparison between a GitHub workflow run where caching wasn't enabled and one where caching was enabled. 

Caching should also be configurable (so it can be disabled if it's causing problems in a user's project).

It's possible to configure projen projects with a different packager than npm -- we should probably switch the "cache" field value appropriately.

Unit tests? docs?

moltar:
@Dzhuneyt I think the solution needs to set the correct package manager value based on the project

https://github.com/projen/projen/issues/1439#issuecomment-1018251364

Also, I'd make caching opt-in, with maybe defaulting to true?

How about exposing `with.node-version` and `with.architecture` to be able to be set by the user?

eladb (dismissed):

Dzhuneyt:
> I think this might not be as simple as a one line change. 😉
> 
> Could you share how you tested this change? I'd like to understand what this does before we merge it in -- perhaps a comparison between a GitHub workflow run where caching wasn't enabled and one where caching was enabled.
> 
> Caching should also be configurable (so it can be disabled if it's causing problems in a user's project).
> 
> It's possible to configure projen projects with a different packager than npm -- we should probably switch the "cache" field value appropriately.
> 
> Unit tests? docs?

I'm quite familiar with GitHub Actions and the caching part of the "setup-node" action, but...

I have now created two new example repositories, using the "node" template of Projen. I noticed that after my change, the "Release" GitHub workflow starts breaking but the error is quite cryptic and I can't see a clear way forward. It seems to be comparing the GitHub Actions workflow files using a command `release | git diff --ignore-space-at-eol --exit-code
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml`. I don't know what `a` and `b` stand for.

Here's a vanilla repo created with a clean Projen:
https://github.com/dzhuneyt-projen/projen-test-vanilla/actions/runs/1728131880
Here's after my change:
https://github.com/dzhuneyt-projen/projen-test-with-github-caching/runs/4894846708

moltar:
@Dzhuneyt I think these errors are because you haven't updated the projen-managed files. When GH Actions run, they call `projen`, which re-generates the managed files, and then it diffs for changes. If changes were found (git is dirty), then it throws.

When I clone your repo, and install & run projen, I get the following files changed:

screenshot-20220121T181549-4g1ljOy1

Dzhuneyt:
@moltar which command have you ran to see these changes?

moltar:
> @moltar which command have you run to see these changes?

`projen`

PR #1554, opened 2022-01-23
fix: Remove markmac from autodependency updates

### LINKED ISSUE #1553: markmac update causes projen update, and vice-verse
projen depends on https://github.com/cdklabs/markmac
markmac depends on projen

Due to auto dep updates, this means that they will unnecessarily update each other every day.

MarkMcCulloh:
Also removed unused imports in .projenrc.js.

Not sure if there's another way to resolve the linked issue, but I think not having publish a new version every day would be nice.

If it would be better to disable markmac from auto updating projen instead then this PR can be closed.

fixes #1553 

---

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -51,7 +51,7 @@ const project = new cdk.JsiiProject({
   ],
 
   depsUpgradeOptions: {
-    exclude: ["markmac"]
+    exclude: ["markmac"],
   },
 
   projenDevDependency: false, // because I am projen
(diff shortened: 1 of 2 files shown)


.projenrc.js
```diff
@@ -51,6 +50,10 @@ const project = new cdk.JsiiProject({
     "all-contributors-cli",
   ],
 
+  depsUpgradeOptions: {
+    exclude: ["markmac"],
```
eladb, line 54: Add a comment explaining why this is excluded

eladb (changes requested):

eladb (approved):

PR #1558, opened 2022-01-24
feat(cdk8s): import an explicit k8s spec version

### LINKED ISSUE #1557: (cdk8s): select the default k8s version to import
I'd like to be able to change the version of k8s spec that projen imports when you run `yarn run import`. Right now it uses the default `k8s` spec, which provides `1.22.0`, but I need to make sure my cdk8s app is compatible with `1.20.0`, so I want to import `1.20.0` instead.

misterjoshua:
Allows users to specify the `k8s` version to import.

Fixes #1557 

---

### DIFF
--- a/src/cdk8s/cdk8s-app-ts.ts
+++ b/src/cdk8s/cdk8s-app-ts.ts
@@ -190,7 +190,7 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
     this.postCompileTask.spawn(synth);
 
     const cdk8sImports = options.cdk8sImports ?? [];
-    const cdk8sSpec = options.k8sSpecVersion
+    const k8sSpec = options.k8sSpecVersion
       ? `k8s@${options.k8sSpecVersion}`
       : "k8s";
 
@@ -201,7 +201,7 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
       obj: {
         language: "typescript",
         app: `node lib/${appEntrypointName}`,
-        imports: [cdk8sSpec, ...cdk8sImports],
+        imports: [k8sSpec, ...cdk8sImports],
       },
     });
 
--- a/src/cdk8s/cdk8s-app-ts.ts
+++ b/src/cdk8s/cdk8s-app-ts.ts
@@ -194,13 +194,14 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
       ? `k8s@${options.k8sSpecVersion}`
       : "k8s";
 
-    const appEntrypointName = path.basename(this.appEntrypoint, ".ts") + ".js";
+    const appEntrypointBaseName = path.basename(this.appEntrypoint, ".ts");
+
     new YamlFile(this, "cdk8s.yaml", {
       committed: true,
       editGitignore: true,
       obj: {
         language: "typescript",
-        app: `node lib/${appEntrypointName}`,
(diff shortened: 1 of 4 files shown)


src/cdk8s/cdk8s-app-ts.ts
```diff
@@ -12,6 +12,13 @@ export interface Cdk8sTypeScriptAppOptions extends TypeScriptProjectOptions {
    */
   readonly cdk8sVersion: string;
 
+  /**
+   * Import a specific Kubernetes API version
+   *
+   * @default - Use the cdk8s default
+   */
+  readonly cdk8sApiVersion?: string;
```
Chriscbr, line 20: This name seems misleading, since it's not the version of cdk8s that is changing. How about `k8sVersion`?

src/cdk8s/cdk8s-app-ts.ts
```diff
@@ -173,9 +180,13 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
...
     });
 
+    const k8sSpec = options.cdk8sApiVersion
+      ? `k8s@${options.cdk8sApiVersion}`
+      : "k8s";
+
     const importTask = this.addTask("import", {
       description: "Imports API objects to your app by generating constructs.",
-      exec: "cdk8s import -o src/imports",
+      exec: `cdk8s import ${k8sSpec} -o src/imports`,
```
Chriscbr, line 189: I'm not sure if this is the ideal place to specify the k8s import. By default, `cdk8s import` (without specifying any spec files) will import all specs listed in `cdk8s.yaml` ([docs](https://cdk8s.io/docs/latest/cli/import/#crds)). Instead of updating this command, projen should auto-generate cdk8s.yaml with the correct list of specs instead.

Chriscbr (changes requested):

src/cdk8s/cdk8s-app-ts.ts
misterjoshua: `k8sSpecVersion` sounds good.

src/cdk8s/cdk8s-app-ts.ts
misterjoshua: I believe the app creates a `cdk8s.yaml` as `SampleCode`. I wonder if we should bring this file into projen management so that we can control the file via the project options?

Chriscbr: That sounds good 👍

misterjoshua:
@Chriscbr Hey. I've changed the new option to `k8sSpecVersion`, and I've added a `YamlFile` to generate `cdk8s.yaml`. Ready for another review when you get a chance.

src/cdk8s/cdk8s-app-ts.ts
```diff
@@ -173,21 +181,31 @@ export class Cdk8sTypeScriptApp extends TypeScriptAppProject {
...
+    this.addTask("import", {
       description: "Imports API objects to your app by generating constructs.",
       exec: "cdk8s import -o src/imports",
     });
 
-    for (const importSpec of options.cdk8sImports ?? []) {
-      importTask.exec(`cdk8s import ${importSpec} -o src/imports`);
-    }
-
-    this.gitignore.include("imports/");
```
misterjoshua, line 185: This line seems like it was left dangling from #1536, since now the imports are `src/imports` by default, so I removed it.

Chriscbr (approved):
Nice! 👍

PR #1566, opened 2022-01-26
chore: prepare projen docs for GitHub pages

### LINKED ISSUE #1507: API.md too big to show anymore
This is what I get when I'm looking up the API now in projen:

Screen Shot 2022-01-10 at 4 22 43 PM

The "view raw" is basically unusable as well. Wondering if this is a chance for a necessary documentation overhaul. Happy to help if so.

An additional problem is that the links in the "Getting Started" section (which I use heavily right now haha), would link directly to the relevant API, and now I have to search through the raw markdown to get the API reference I'm looking for.

Chriscbr:
Moving API.md to the docs directory so it can be rendered on GitHub pages.

Also deleted the empty `mutable-builds.md` (file contents were moved to `build.md` some eons ago)

Closes #1507

---


.projenrc.js
```diff
@@ -229,4 +229,6 @@ setupBundleTaskRunner();
 // but not all files are updated
 project.postCompileTask.spawn(project.defaultTask);
 
+project.tasks.tryFind("docgen").reset("jsii-docgen -o docs/api/API.md");
```
eladb, line 232: Instead of escape hatching, Let's add an option to control api.md output path, no?

eladb (changes requested):

.projenrc.js
Chriscbr: Done

eladb (approved):

PR #1575, opened 2022-01-28
feat(cdk8s): add integration tests

### LINKED ISSUE #1573: (cdk8s): Support integration tests
I'd like to write integration tests for cdk8s projects the same way that I write integration tests for AWS CDK projects.

For instance, when I add an app in a file named `test/my-test-name.integ.ts`, I envision that Projen would generate these tasks:

* `integ:my-test-name:deploy` - Synths the integration test app, applies it with kubectl, and emits a snapshot
* `integ:my-test-name:snapshot` - Synths the integration test app and emits a snapshot
* `integ:my-test-name:assert` - Synths the integration test app and diffs it with a snapshot, erroring if they're different

misterjoshua:
This PR adds integration testing for cdk8s projects. Projen will detect files in your test directory named `*.integ.ts` and will generate integration testing tasks for you. Additionally, Projen will assert that your integration test snapshots have not changed unexpectedly when you build your project.

For example, for an integration test app in a file named `test/my-test-name.integ.ts` Projen will create the following tasks:

* `integ:my-test-name:deploy` - Synths the integration test app, applies it with kubectl, and emits a snapshot
* `integ:my-test-name:snapshot` - Synths the integration test app and emits a snapshot
* `integ:my-test-name:assert` - Synths the integration test app and diffs it with a snapshot, erroring if they're different

Fixes #1573

﻿---


misterjoshua:
@Chriscbr Hey. I've created this PR which adds cdk8s integration tests. I've roughly followed the approach @eladb used to create awscdk integration tests. I also incorporated an equivalent of @jogold's `integ:snapshot-all`. Please let me know what you think. :)

src/cdk8s/integration-test.ts
```diff
@@ -0,0 +1,122 @@
...
+      project.deps.addDependency("ts-node", DependencyType.BUILD);
+    }
+
+    // Synth, deploy, and capture the snapshot
+    const deployTask = project.addTask(`integ:${name}:deploy`, {
+      description: `deploy integration test '${name}' and capture snapshot`,
+    });
+    deployTask.exec(`rm -fr ${deployDir}`);
+    deployTask.exec(`cdk8s synth --app "${app}" -o ${deployDir}`);
+    deployTask.exec(`kubectl apply -f ${deployDir}`);
```
Chriscbr, line 84: hmm I wonder if this is something users would want to customize in the future. e.g. "after `kubectl apply`, it should also check and wait for the rollout status to be done". Probably a good idea to wait for use cases before over-engineering this though, so let's leave this as is. 👍

src/cdk8s/auto-discover.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @default false
+   */
+  readonly integrationTestAutoDiscover?: boolean;
+}
+
+/**
+ * Automatically creates integration test tasks for all .integ.ts files under
+ * the test directory of the project.
+ */
+export class AutoDiscover extends Component {
```
Chriscbr, line 34: As long as this is a separate component from the one in the `awscdk` module, can we also add unit tests for it? :)

src/cdk8s/integration-test.ts
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * The path of the tsconfig.json file to use when running integration test cdk apps.
+   */
+  readonly tsconfigPath: string;
+}
+
+/**
+ * CDK8S integration test.
+ */
+export class IntegrationTest extends Component {
```
Chriscbr, line 37: I wonder if we can try and reduce the code duplication between cdk8s and awscdk integ tests. Perhaps add an `IntegrationTestBase` class and/or `AutoDiscoverBase` class to the `cdk` module and then extend them / override the behaviors as needed within cdk8s and awscdk? It might require moving some initialization to private methods or protected methods etc. Could be a bit of work but if it makes sense it would help unify the abstraction that integ testing shares some common interface between CDK types.

Chriscbr (changes requested):
Looks great @misterjoshua , I like the idea that integration / snapshot testing can be used across CDKs. 😄 I've added a few comments.

PR description looks good - can we add something similar to our high-level docs? For example `docs/cdk8s.md`

src/cdk8s/auto-discover.ts
misterjoshua: Sure, that shouldn't be a problem.

src/cdk8s/integration-test.ts
misterjoshua: Yes. I had also considered that users might want to use `--prune` with `--selector` and `--prune-whitelist`. I agree that we should wait for use cases before going too deep here.

src/cdk8s/integration-test.ts
misterjoshua: This had crossed my mind as well, but I didn't want to make a sweeping change without this type of discussion first. I'll see what I can come up with. :)

misterjoshua:
I've pushed up some WIP so I can resume work on this at home over the weekend.

misterjoshua:
By now I'm more or less happy with the shape of `IntegrationTestBase`, so I'll leave that alone until we review it. I'll see what I can do about the `AutoDiscoverBase` concept and the dangling `TYPESCRIPT_INTEG_EXT` consts in each project type soon.

misterjoshua:
Extending a base class with a protected abstract method for `AutoDiscover` in cea291f felt awkward when writing tests for auto-discovery, so in 983eca0 and 980442e I moved to a compositional structure for `AutoDiscover`.

I figure it makes more sense to have a unique `AutoDiscover` for each project type. This way, we can share smaller, re-usable components and add them to `AutoDiscover` as the project type needs. As of 930442e, this is what all this looks like:

image

```plantuml
@startuml

title Relationships - Class Diagram

class awscdk.AutoDiscover
class awscdk.IntegrationTestAutoDiscover
class awscdk.IntegrationTest
class awscdk.LambdaAutoDiscover
class awscdk.LambdaFunction

class cdk8s.AutoDiscover
class cdk8s.IntegrationTestAutoDiscover
class cdk8s.IntegrationTest

class cdk.AutoDiscoverBase
class cdk.IntegrationTestAutoDiscoverBase

cdk.IntegrationTestAutoDiscoverBase <|-down- awscdk.IntegrationTestAutoDiscover
cdk.IntegrationTestAutoDiscoverBase <|-down- cdk8s.IntegrationTestAutoDiscover

cdk.AutoDiscoverBase <|-down- awscdk.LambdaAutoDiscover
cdk.AutoDiscoverBase <|-down- cdk.IntegrationTestAutoDiscoverBase

awscdk.AutoDiscover -up-> awscdk.IntegrationTestAutoDiscover
awscdk.AutoDiscover -up-> awscdk.LambdaAutoDiscover
awscdk.IntegrationTestAutoDiscover -> awscdk.IntegrationTest
awscdk.LambdaAutoDiscover -left-> awscdk.LambdaFunction

cdk8s.AutoDiscover -up-> cdk8s.IntegrationTestAutoDiscover
cdk8s.IntegrationTestAutoDiscover -> cdk8s.IntegrationTest

@enduml
```

I took initiative here because I had some time to burn. I'm happy to revert back to cea291f and take any other approach. :)

src/cdk/auto-discover-base.ts
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * @example "test"
+   */
+  readonly projectdir: string;
+}
+
+/**
+ * Base class for auto-discovering and creating project subcomponents.
+ */
+export class AutoDiscoverBase extends Component {
```
Chriscbr, line 29: Let's make `AutoDiscoverBase` and `IntegrationTestBase` abstract for now, since they don't do much on their own (minimizing API surface area):
```suggestion
export abstract class AutoDiscoverBase extends Component {
```
This might require updating tests to use a stub child class

Chriscbr (changes requested):
> I figure it makes more sense to have a unique AutoDiscover for each project type. This way, we can share smaller, re-usable components and add them to AutoDiscover as the project type needs.

Yep this makes sense to me. If the public APIs need to change in the future we can always refactor as needed, but this looks pretty much good to ship. (very clean diagrams by the way!)

Just one more small comment :)

Chriscbr (approved):

PR #1602, opened 2022-02-08
fix(upgrade-dependencies): upgrade task cannot be overwritten

### LINKED ISSUE #1129: v0.29.2 / #1112 introduced major breaking changes to upgrade process
The v0.29.2 release / #1112 PR introduced major breaking changes to the upgrade process that cannot be worked around. 

After this change it's no longer possible to customize the upgrade task on `NodeProject`s. This is because you can no longer access the upgrade task. This is required to keep so we can work around #892 and have the upgrade workflow update all the subprojects. Can we revert #1112 please?

For example:
```typescript
const upgradeTask = project.tasks.tryFind('upgrade');
upgradeTask.reset();
```
Will produces the error:
```
TypeError: Cannot read property 'reset' of undefined
```

But if you try to add the `upgrade` task:
```typescript
const upgradeTask = project.tasks.addTask('upgrade');
```
Results in:
```
Error: A task with the name upgrade already exists. To override it, call removeTask first and then create the new task.
```

This even happens if you try remove the task before adding it:
```typescript
project.tasks.removeTask('upgrade');
const upgradeTask = project.tasks.addTask('upgrade');
```
Results in:
```
Error: A task with the name upgrade already exists. To override it, call removeTask first and then create the new task.
```

cgatt:
Small change that moves the `createTask` step of `UpgradeDependencies` into the constructor, so that it is added during project instantiation. This allows the task to be overwritten with `project.removeTask()` per the error message:
```
Error: A task with the name upgrade already exists. To override it, call removeTask first and then create the new task.
```
This previously did not work, as the error was actually being thrown by the `UpgradeDependencies.createTask()` function during the pre-synth step. I considered instead adding error handling during this step, but could not decide whether silently not creating a task if one already existed or throwing an error with a clearer message would be a better behaviour, and decided on this approach instead.

Also includes a couple of non-breaking typo fixes.

fixes #1129 

---

### DIFF
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -7868,7 +7868,7 @@ Add steps to execute a successful build.
 addPostBuildSteps(...steps: JobStep[]): void
 ```
 
-* **steps** (<code>[github.workflows.JobStep](#projen-github-workflows-jobstep)</code>)  worklfow steps.
+* **steps** (<code>[github.workflows.JobStep](#projen-github-workflows-jobstep)</code>)  workflow steps.
   * **continueOnError** (<code>boolean</code>)  Prevents a job from failing when a step fails. __*Optional*__
   * **env** (<code>Map<string, string></code>)  Sets environment variables for steps to use in the runner environment. __*Optional*__
   * **id** (<code>string</code>)  A unique identifier for the step. __*Optional*__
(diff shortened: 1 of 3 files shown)


Chriscbr (approved):
Thanks!

PR #1604, opened 2022-02-08
feat: enforce implementation of projen marker on subclasses of FileBase

cgatt:
This PR is to extend the functionality and address an issue in the the Eject from Projen functionality from #1459 

In this PR, a boolean option `marker` was added to `FileBaseOptions`, that states "Adds the projen marker to the file", however many subclasses of FileBase continued to silently ignore this setting, only exposing the projen marker under class.marker for manual inclusion. I decided to re-architect this slightly to enforce that subclasses of FileBase must explicitly address this option by adding the abstract method `addProjenMarker()` for children to implement. This has the side benefit of moving the logic for whether to include the marker into the base class, rather than repeating a ternary in every subclass.

To support this change, implementations of addProjenMarker have been added to all subclasses of `FileBase`. These copy the existing style for marker addition and produce identical snapshots in all cases bar one: the class `ReactComponent` was I believe incorrectly using the logic `if (!configOverrides.marker) { add marker }` instead of `if (configOverrides.marker) { add marker }`, so this has been corrected and the snapshot updated accordingly.

Classes that did not implement a marker have been configured to override the `marker` option to false, and will throw `Error("Method not implemented.")` should this somehow be set to true without being implemented. This has been documented as well as I could, but I'm not happy with the result and open to suggestions. JSII does not support re-declaring properties in a child interface, so I could not properly re-declare `marker` with overridden documentation, and instead settled for adding an `@property` docstring on the child interface itself, as in the below example from `ReactTypeDefOptions`:
```ts
/**
 * @prop {boolean} [marker=false] - Ignored by ReactTypeDef. Marker will never be added to file.
 */
export interface ReactTypeDefOptions extends FileBaseOptions {}
```

The two classes that received substantial changes as a result of this are `TextFile` and by extension `SourceCode`. These two classes did not implement adding the projen marker, instead leaving it to be added manually if needed for obvious reasons. However, I do not believe this is in line with what we need for `eject` to function properly, and so I have added two additional properties to these classes: `projenMarkerPrefix` and `projenMarkerAtStart`. Between these properties and `marker`, TextFile (and SourceCode by extension, which merely passes them through to TextFile) can take responsibility for adding the marker and correctly removing it upon ejection, rather than relying on all users of TextFile and SourceCode to correctly implement the ejection logic.

All uses of SourceCode and TextFile in the projen package have been updated to use these properties in a way that does not modify their snapshots, and tests have been added to confirm they work as expected. However, this is a breaking change for all users of these classes outside the projen package, as the `FileBase.marker` property will now be undefined by default (as TextFile now defaults `marker` to false to least affect those on progen <0.52). This will only affect those who have already migrated to 0.52.x, as those moving from <0.52 will already have to modify their code to allow for the removal of FileBase.PROJEN_MARKER

BREAKING CHANGE: `FileBase.marker` will now be undefined on `TextFile` and `SourceCode` classes by default
---


Chriscbr (commented):
Thanks for the pull request and looking to improve this API. 🙂  

Is there is any way for someone creating a `TextFile` (or another custom file component) to customize where the marker should go while they are constructing it instead of during addProjenMarker. I feel like this API probably will satisfy 95% of use cases, but there could be situations where someone wants it in the middle of their file, and trying to inject it during addProjenMarker might be difficult. Can we add this as a test case? (I wonder if this might be possible by just subclassing TextFile)

cgatt:
I spent a lot of time trying to think of a better way to accommodate more customization for TextFile, but I eventually concluded that it was better to not over-complicate the options API, and instead encourage people to subclass FileBase themselves if they require a specific behaviour for the projen marker (the JsonFile class is a good example of this, adding the marker as the "//" property).
I actually foresee this being more of an issue for `SourceCode`, which does not inherit from FileBase (or TextFile) but instead wraps it. A potential solution would be to modify SourceCode to inherit from FileBase, potentially via Textfile, unless there's a good reason it currently doesn't?

Chriscbr:
I might need some more time to think this over, since I'm hesitant to extend the API surface area of FileBase with new abstract methods. Currently the projen marker is less critical than it to be, since we used to rely on it for cleaning up files, but now files are tracked within `.projen/files.json`. I realize components "forgetting" to render the marker or rendering it correctly could be problematic, but I'm not sure how critical it is to enforce that.

These are just my current thoughts. I think it would be a good idea to revisit this PR at a later time.

cgatt:
Hey @Chriscbr, shall we add the backlog label to this PR, or would you prefer to let it close and re-address it later?

Chriscbr:
Let's close this for now

PR #1622, opened 2022-02-16
feat(awscdk): allow integration test path metadata

misterjoshua:
Adds a `pathMetadata` option to IntegrationTest and the integration test options to auto-discover classes so that users can enable path metadata for their integration tests.

---

### DIFF
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -15,8 +15,8 @@ export interface IntegrationTestCommonOptions {
   readonly destroyAfterDeploy?: boolean;
 
   /**
-   * Enables path metadata
-   *
+   * Enables path metadata, adding `aws:cdk:path` with the defining construct's
+   * path to the CloudFormation metadata for each synthesized resource.
    * @default false
    */
   readonly pathMetadata?: boolean;
--- a/src/awscdk/integration-test.ts
+++ b/src/awscdk/integration-test.ts
@@ -15,8 +15,8 @@ export interface IntegrationTestCommonOptions {
   readonly destroyAfterDeploy?: boolean;
 
   /**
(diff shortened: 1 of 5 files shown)


src/awscdk/integration-test.ts
```diff
@@ -13,6 +13,13 @@ export interface IntegrationTestCommonOptions {
    * @default true
    */
   readonly destroyAfterDeploy?: boolean;
+
+  /**
+   * Enables path metadata
```
Chriscbr, line 18: Can we give any description of what this is, or link to some doc (or piece of code) that explains what this is?

test/awscdk/integration-test.test.ts
```diff
@@ -196,3 +196,35 @@ test("synthesizing an integration test containing a multi-stack stage", () => {
...
+
+  expect(
+    output[".projen/tasks.json"].tasks["integ:my-stage:deploy"]
+  ).toMatchSnapshot();
+  expect(
+    output[".projen/tasks.json"].tasks["integ:my-stage:snapshot"]
+  ).toMatchSnapshot();
+  expect(
+    output[".projen/tasks.json"].tasks["integ:my-stage:watch"]
+  ).toMatchSnapshot();
```
Chriscbr, line 229: Let's also assert that the snapshot doesn't contain `--no-path-metadata`

Chriscbr (changes requested):

test/awscdk/integration-test.test.ts
misterjoshua: Ok. I've changed these. I was worried that checking for the absence by step index would be too rigid, so I used jest's "expect" matching instead.

Chriscbr (approved):

PR #1627, opened 2022-02-19
fix: @testing-library/react belonging to dep not devdep

mmuller88:
react
* run "npx create-react-app my-app --template typescript" and check package.json

---

### DIFF
--- a/src/web/react.ts
+++ b/src/web/react.ts
@@ -197,12 +197,7 @@ export class ReactComponent extends Component {
     // No compile for react app
     project.compileTask.reset();
 
-    project.addDeps("react", "react-dom", "react-scripts@^5", "web-vitals");
-    project.addDevDeps(
-      "@testing-library/jest-dom",
-      "@testing-library/react",
-      "@testing-library/user-event"
-    );
+    project.addDeps("react", "react-dom", "react-scripts@^5", "web-vitals", "@testing-library/jest-dom", "@testing-library/react", "@testing-library/user-event");
     if (this.typescript) {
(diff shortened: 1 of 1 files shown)


Chriscbr (changes requested):
I'm not sure I agree with this change.

Based on what I found [here](https://stackoverflow.com/questions/44868453/create-react-app-install-devdepencies-in-dependencies-section/44872787#44872787), it sounds like the authors of create-react-app do this is because they are designing for a single use case type -- specifically, those building an app that won't be consumed as a library so it doesn't matter whether these are dev or dev dependencies. But if this react project ended up being used as a library, it would not make sense for the @testing-library dependencies to be included in the bundle.

IMO it's okay to leave the current dependencies as they are, or perhaps even moving _more_ of them into devDependencies to make it clearer that there is no run-time component of the project.

mmuller88:
> I'm not sure I agree with this change.
> 
> Based on what I found [here](https://stackoverflow.com/questions/44868453/create-react-app-install-devdepencies-in-dependencies-section/44872787#44872787), it sounds like the authors of create-react-app do this is because they are designing for a single use case type -- specifically, those building an app that won't be consumed as a library so it doesn't matter whether these are dev or dev dependencies. But if this react project ended up being used as a library, it would not make sense for the @testing-library dependencies to be included in the bundle.
> 
> IMO it's okay to leave the current dependencies as they are, or perhaps even moving _more_ of them into devDependencies to make it clearer that there is no run-time component of the project.

Thanks for the good feedback. Yeah, no problem if we not ingesting that. In terms of performance, it is probably advised not to include it.

PR #1631, opened 2022-02-20
fix(jsii): `build` task pollutes working directory with repo duplication

iliapolo:
In https://github.com/projen/projen/pull/1408, we changed the packaging task for jsii projects to simply produce a built version of the repository to the `dist` directory. This enables proper parallelization of `jsii-pacmak` in subsequent jobs. 

It also created a quirk for dev environments since it skips packaging all together, and creates a duplicate of the repository in the `dist` directory, which can also mess up IDE usage since it now sees two files of each...

In addition, this also breaks external consumers who were relying on `npx projen build` to create the language bindings.
> For example: https://github.com/cdklabs/aws-delivlib/blob/epolon/integrity-checks/lib/package-integrity/handler/repository.ts#L74

This PR conditions jsii packaging with a `CI` env variable, commonly used to signal we are running inside a CI platform.
When `CI` is true, we do the same thing we do now, which is prepare the repository for packaging.
When `CI` is false, we simply invoke `package-all` to create all bindings.

I acknowledge the usage of the `CI` variable is not ideal, but seems good enough until we properly model local dev environments to align them with the CI workflows. 

---


src/cdk/jsii-project.ts
```diff
@@ -197,9 +197,16 @@ export class JsiiProject extends TypeScriptProject {
...
+    const prepareRepoForCI = [
+      `mkdir -p ${this.artifactsDirectory}`,
+      `rsync -a . ${this.artifactsDirectory} --exclude .git --exclude node_modules`,
+    ].join(" && ");
+
+    // when running inside CI we just prepare the repo for packaging, which
+    // takes place in separate tasks.
+    // outside of CI (i.e locally) we simply package all targets.
+    this.packageTask.reset(
+      `if [ ! -z \${CI} ]; then ${prepareRepoForCI}; else npx projen ${this.packageAllTask.name}; fi`
```
eladb, line 209: There's a method that you can call on `project` I think that returns the command to execute a projen task (instead of hard coding `npx projen xxx`).

eladb (changes requested):

src/cdk/jsii-project.ts
iliapolo: Done 👍

eladb (approved):

PR #1500, opened 2022-01-09
feat(build): support gitlab workflows (WIP)

eladb:
Allow users to select which workflow engine to use for their build workflows.

This is enabled by creating an abstract layer for defining workflows.

WIP: 

- [ ] Documentation
- [ ] Unit tests
- [ ] CodeCov support
- [ ] `build.yml` no longer in `.gitattributes`
- [ ] Migrate release workflow
- [ ] Restore disabled behavior (see comments)
- [ ] Add `workflowProvider` option to `Project`

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1637, opened 2022-02-22
feat!: remove GitHubProject class

Chriscbr:
TODO: write down list of deprecations and breaking changes

---


src/java/java-project.ts
```diff
@@ -197,6 +203,35 @@ export class JavaProject extends GitHubProject {
...
+  }
+
+  /**
+   * Access for Gitpod component.
+   *
+   * This will be `undefined` if gitpod boolean is false.
+   */
+  public get gitpod(): Gitpod | undefined {
+    return this._sharedComponents.gitpod;
+  }
```
Chriscbr, line 226: I was also thinking of deprecating these properties and instead pointing users to use the same `Gitpod.of(project)` pattern to access these components - not sure

src/github/shared.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * A collection of unrelated, language-agnostic components that projen includes
+ * in all project types besides the base `Project`.
+ *
+ * This exists mostly to reduce code repetition and to avoid using an extra
+ * layer of inheritance.
+ */
+export class SharedComponents extends Component {
```
Chriscbr, line 90: Sort of a bland abstraction but it works. Open to suggestions

eladb: Would it make sense to basically merge this with the `GitHub` class?

src/java/java-project.ts
eladb: +1 to deprecate

src/github/shared.ts
eladb: :-( we are basically wrangling the problem from side to side. Now instead of a base class, it's a component :-)

But maybe that's one step towards nirvana..

I personally think `vscode`, `gitpod` and `devContainer` can all be just components that people can add to their projects instead of "baked into" the project.

Chriscbr: (+1 on removing vscode/gitpod/devContainer as builtins)

Yeah, frankly I start to second guess myself here... the question on my mind is basically whether it makes sense to have some class in the OOP hierarchy between Project and NodeProject/PythonProject/JavaProject or not. My intuition says there shouldn't be one, but even if so, it's kind of a handwavy argument since the end goal is a bit fuzzy. Maybe I'll think on this a PR a bit longer...

I think finishing https://github.com/projen/projen/pull/1500/files might inform the problem a bit more, so maybe I can try to pick that up sometime...

PR #1650, opened 2022-03-01
fix(upgrade-task): Fixed upgrade task to include line ending in package.json

AminFazlMondo:
If the package manager is set to npm, the `npm install` command adds a new line to the end of package.json. This is a known issue: https://github.com/npm/npm/issues/18545
All of the other tasks do not do this currently, and this causes the upgrade workflow to fail for the first time and self mutation fixes that. To avoid that, added this new line to package.json anyway.

---


AminFazlMondo:
I would like to get some feedback on this before investing too much.
Is this approach in line with the vision of the repo?
With the failures in the tests, I was wondering if it is OK to update the test, or take other approaches.

Chriscbr:
I think this is ok if it fixes issues / makes a more natural experience when the package manager is set to npm. If the package manager is set to yarn, then I think the output should be left as is, without an extra newline.

AminFazlMondo:
Hi @Chriscbr,
Could you please have a look at this again?
Cheers,

Chriscbr (approved):

PR #1661, opened 2022-03-05
fix: corrupt changelog for prefixed releases

iliapolo:
When multiple version lines exist for a project, we use a [`releaseTagPrefix`](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-22/main/.projenrc.js#L50) property to distinguish between them. 

The problem was that this prefix wasn't being considered when generating the changelog, resulting in it simply accumulating commits, instead of doing proper diffs. 

> See https://github.com/cdk8s-team/cdk8s-plus/releases 

This PR fixes the issue by passing the `releaseTagPrefix` configuration down to `standard-version`, which generates the changelog. 

---


test/release/bump.test.ts
```diff
@@ -81,7 +81,7 @@ test("select latest, with prefix", async () => {
     commits: [
       { message: "first version", tag: "prefix/v1.1.0" },
       { message: "unrelated version", tag: "v2.0.0" },
-      { message: "second version", tag: "prefix/v1.2.0" },
```
iliapolo, line 84: This change makes the test properly consider prefixes because now this commit impacts both the changelog and the version.

test/release/bump.test.ts
```diff
@@ -91,6 +91,7 @@ test("select latest, with prefix", async () => {
   expect(result.changelog.includes("Bug Fixes")).toBeTruthy();
   expect(result.changelog.includes("another bug")).toBeTruthy();
   expect(result.changelog.includes("bug")).toBeTruthy();
+  expect(result.changelog.includes("feature 1")).toBeFalsy();
```
iliapolo, line 94: When `releaseTagPrefix` isn't passed on to `standard-version`, this commit ends up in the changelog because the reference tag will be `v2.0.0`, and not `prefix/v1.2.0`.

iliapolo:
@Chriscbr have another looksy?

PR #1671, opened 2022-03-08
[WIP] feat!: migrate to constructs

Chriscbr:
Just an experiment.

Known issues:
- Adding a static "of" method to access a project given a scope seems to be important for components to access common things like dependencies, gitignore, and other project-specific information. I wanted these to all just be named "of" but this causes a compilation error from JSII because the same static method name is used in subclasses but with different return types. See https://github.com/aws/jsii/issues/3410. So for now the names look like "ofTypeScriptProject" etc. which is a bit awkward but I'm not sure if I see a more natural solution.

Open questions:
- In the future (or even within this PR) we should add a linter that enforces a common constructor structure on components. Right now the component signatures look like `constructor(scope: Construct, props: Props)` in v1 of this PR.
  - Should the ID be required? as in `constructor(scope: Construct, id: string, props: Props)` like most other CDK libraries? There is a constraint in `constructs` that sibling nodes must have different names, and this could get awkward if the IDs are not customizable.
  - Should we allow other kinds of scopes (i.e. project-specific scopes) in the first argument? Or is it better to standardize that it's always just "Construct" (and instead, the constructor can narrow down functionality based on project types using `ProjectType.ofProjectType(scope)`? I think the latter might be better long term.

---


.projen/tree.json
```diff
@@ -0,0 +1,541 @@
+{
```
Chriscbr, line 1: I generated this for fun, just to see what the construct tree looks like. The paths look a tad ugly because "/" occurs a lot in file paths, but isn't allowed in construct paths, so they just get translated to "--". I wonder if there's a way to make this more natural, or maybe have a specific "projen-specific" path instead with a different delimiting scheme, etc

src/json.ts
```diff
@@ -17,10 +19,30 @@ export interface JsonFileOptions extends ObjectFileOptions {
...
+    const isObjectFile = (c: IConstruct): c is ObjectFile =>
+      c instanceof ObjectFile;
+    const absolutePath = path.isAbsolute(filePath)
+      ? filePath
+      : path.resolve(Project.ofProject(scope).outdir, filePath);
+    return scope.node
+      .findAll()
+      .filter(isObjectFile)
+      .find((file) => file.absolutePath === absolutePath);
+  }
```
Chriscbr, line 40: By switching to constructs, escape hatching is more flexible since anyone can directly access `construct.node`. However I think these "tryFind" methods are still useful if you don't know where a file is generated from.

I'm not sure if  this functionality should be on `Project`, or if it should be on the filetype classes instead. We could also just expose `tryFind`/`tryFindFile` in the `Project` class and just expose `isXxx` static methods on these classes instead.

(re: naming -- if we choose to make them static methods, we can't name them all `tryFind` because of https://github.com/aws/jsii/issues/3410 so they have to be a little verbose)

pgollucci:
you probably want a pr/do-not-merge on this one

Chriscbr:
Closing for now as stale

mrgrain:
Implemented by #2974

PR #1688, opened 2022-03-15
feat(node-project): Added ability to provide scoped private packages

AminFazlMondo:
Currently there is no way for the packages that have some dependencies that are private to use projen. Thia allows scoped packages (eg, `@myscope`) to have different registryUrl and have code artifact login.

---


src/javascript/node-project.ts
```diff
@@ -353,6 +354,70 @@ export enum AutoRelease {
...
+   * @returns object containing the (domain, accountId, region, repository)
+   */
+  private static extractCodeArtifactDetails(registryUrl: string) {
+    const match = registryUrl.match(NodeProject.codeArtifactRegex);
+    if (match?.groups) {
+      const { domain, accountId, region, repository } = match.groups;
+      return { domain, accountId, region, repository };
+    }
+    throw new Error("Could not get CodeArtifact details from npm Registry");
+  }
```
Chriscbr, line 376: Let's put these static methods and constants outside of the class (either in this file or in `util.ts`) so we can reuse them or use them for testing

test/node-project.test.ts
```diff
@@ -888,6 +888,220 @@ test("node project can be ejected", () => {
...
+  const repository = "MyRepository";
+  const scope = "@stub-scope";
+  const accessKeyIdSecret = "stub-access-key-id";
+  const secretAccessKeySecret = "stub-secret-access-key";
+
+  test("adds AWS Code Artifact Login step prior to install to build workflow", () => {
+    const project = new TestNodeProject({
+      scopedPackagesOptions: [
+        {
+          registryUrl: `https://${domain}-${accountId}.d.codeartifact.${region}.amazonaws.com/npm/${repository}/`,
```
Chriscbr, line 904: nit: Can we extract this to `const registryUrl = ...` above?

Chriscbr, line 904: Can we extract this to `const registryUrl = ...` above?

Chriscbr (changes requested):
Cool! I've added a few small comments. Could you please add a section in our hand-written docs to document this feature? Just a few sentences would be great. :) Perhaps in `docs/node.md`, or wherever you see most appropriate.

(This is just brainstorming, nothing that needs to be addressed). How does this experience work locally / how do you authenticate to install packages when projen isn't running in CICD? I wonder if we should also add some kind of `npx projen codeartifact:auth` command, or some kind of "install hook" that allows you to specify one or more commands to run before installing dependencies in any task. Or is there a simpler way, like generating a .npmrc file, etc...

AminFazlMondo:
> Cool! I've added a few small comments. Could you please add a section in our hand-written docs to document this feature? Just a few sentences would be great. :) Perhaps in `docs/node.md`, or wherever you see most appropriate.
> 
> (This is just brainstorming, nothing that needs to be addressed): How does this experience work locally / how do you authenticate to install packages when projen isn't running in CICD? I wonder if we should also add some kind of `npx projen codeartifact:auth` command, or some kind of "install hook" that allows you to specify one or more commands to run before installing dependencies in any task. Or is there a simpler way, like generating a .npmrc file, etc...

I was thinking about that too, but wanted to iterate on it with others' input. Will investigate that in future.

hoegertn:
Support for scoped registries is already there using the NpmRc class. Would it be possible to read this to generate the login in CICD instead of adding another way of configuring this?

Chriscbr:
@AminFazlMondo what do you think?

AminFazlMondo:
That is right, we probably could use the npmRc for the registry but the code artifact token is required for the private one, soe It is right to say that this is more to generate token in CICD pipeline.
Are we suggestion that some of the code can be changed?

charliehart-mondo:
> Support for scoped registries is already there using the NpmRc class. Would it be possible to read this to generate the login in CICD instead of adding another way of configuring this?

I think the local npmrc file option is ok if that's the correct route to go, but it still needs to have the token setup presumably as an environment variable in a workflow step.

This is from the docs: https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#create-and-check-in-a-project-specific-npmrc-file

The codeartifact-auth step in this PR currently does "npm config set" for the relevant details (registry, _authToken, always_auth), it (or a different action) could be changed to update an environment variable instead and combine that with an npmrc file (would this be auto generated/updated based on config?)

https://github.com/MondoPower/codeartifact-auth/blob/main/src/index.ts

Either way, I suspect most of the code in this PR is required to add a login step regardless of if it does "npm config set" or sets an env variable, so which approach seems to fit best?

src/javascript/node-package.ts
```diff
@@ -829,9 +897,45 @@ export class NodePackage extends Component {
       npmRegistryUrl: npmr.href,
       npmTokenSecret: defaultNpmToken(options.npmTokenSecret, npmr.hostname),
       codeArtifactOptions,
+      scopedPackagesOptions: this.parseScopedPackagesOptions(
+        options.scopedPackagesOptions
+      ),
```
Chriscbr, line 902: Can we pass in options.codeArtifactOptions so that the user doesn't have to specify the options in both scopedPackagesOptions and codeArtifactOptions ?

Chriscbr (changes requested):

Chriscbr:
Managing the config via npmrc (with the `NpmConfig` component) could make sense, but it might be awkward since projen can't completely manage the file if it's being updated whenever someone runs `aws codeartifact login` or runs `npm config set ...`. So I'm ok with this approach

AminFazlMondo:
Hi @Chriscbr ,

Thanks for you feedback, after that it made us re-think the solution.
Originally, we wanted to support for the situation in which the code artifact details to publish to be different from the ones specified in the scoped packages. With what you mentioned above, we came to the conclusion that it is not a common use case and used the same details for scoped packages for simplicity.
We also made a task (`ca:login`) to login to code artifact (thanks for your suggestion), which can also be used in local development.
We would appreciate you having another look and share your thoughts with us or if looks good to you approve this.
Cheers,

src/javascript/node-project.ts
```diff
@@ -353,6 +356,53 @@ export enum AutoRelease {
...
+      name: `AWS CodeArtifact Login ${scopedPackageOptions.scope}`,
+      uses: "MondoPower/codeartifact-auth@1.2",
+      with: {
+        domain: codeArtifactDetails.domain,
+        repository: codeArtifactDetails.repository,
+        scope: scopedPackageOptions.scope,
+        region: codeArtifactDetails.region,
+        accountId: codeArtifactDetails.accountId,
+      },
+    });
```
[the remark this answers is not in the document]
AminFazlMondo: Done

test/node-project.test.ts
```diff
@@ -888,6 +889,222 @@ test("node project can be ejected", () => {
...
+  const registryUrl = `https://${domain}-${accountId}.d.codeartifact.${region}.amazonaws.com/npm/${repository}/`;
+
+  test("adds AWS Code Artifact Login step prior to install to build workflow", () => {
+    const project = new TestNodeProject({
+      scopedPackagesOptions: [
+        {
+          registryUrl,
+          scope,
+          accessKeyIdSecret,
+          secretAccessKeySecret,
```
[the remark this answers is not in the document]
AminFazlMondo: Added test case for it

docs/node.md
```diff
@@ -69,3 +69,33 @@ TODO
...
+
+## Features
+
+### Scoped Private Packages
+
+Scoped private packages can be configured in this project and its ancestors.
+
+All npm packages have a name. Some package names also have a scope. A scope follows the usual rules for package names (URL-safe characters, no leading dots or underscores). When used in package names, scopes are preceded by an @ symbol and followed by a slash, e.g. `@somescope/somepackagename`
+
+This feature supports specifying options on how to access packages in each of the scopes, otherwise will try to install from default npm registry.
```
Chriscbr, line 81:
```suggestion
This feature supports specifying options on how package managers should access packages in each of the scopes. If no options are specified, npm or yarn will try to install scoped packages from thepublic npm registry.
```

src/javascript/node-package.ts
```diff
@@ -314,6 +322,23 @@ export interface CodeArtifactOptions {
   readonly roleToAssume?: string;
 }
 
+/**
+ * Options for scoped packages
+ */
+export interface ScopedPackagesOptions {
+  /**
+   * scope of the packages
```
Chriscbr, line 330:
```suggestion
   * Scope of the packages
```

src/javascript/node-package.ts
```diff
@@ -314,6 +322,23 @@ export interface CodeArtifactOptions {
...
+   *
+   * @example "@angular"
+   */
+  readonly scope: string;
+
+  /**
+   * Url of the registry for scoped packages
```
Chriscbr, line 337:
```suggestion
   * URL of the registry for scoped packages
```

src/javascript/node-package.ts
```diff
@@ -286,6 +287,13 @@ export interface NodePackageOptions {
    * @default - undefined
    */
   readonly codeArtifactOptions?: CodeArtifactOptions;
+
+  /**
+   * Options for privately hosted scoped packages
+   *
+   * @default undefined
```
Chriscbr, line 294:
```suggestion
   * @default - fetch all scoped packages from the public npm registry
```

docs/node.md
```diff
@@ -69,3 +69,33 @@ TODO
...
+
+Currently, it only supports AWS CodeArtifact, either by directly access via credentials or by assuming a role using the specified credentials.
```
Chriscbr, line 83:
```suggestion
Currently, it only supports fetching packages from AWS CodeArtifact, either by directly access via credentials or by assuming a role using the specified credentials.
```

src/javascript/node-package.ts
```diff
@@ -829,9 +870,71 @@ export class NodePackage extends Component {
...
+      !this.scopedPackagesOptions ||
+      this.scopedPackagesOptions.length === 0
+    ) {
+      return;
+    }
+
+    this.project.addTask("ca:login", {
+      requiredEnv: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"],
+      steps: [
+        { exec: "which aws" },
```
Chriscbr, line 919:
```suggestion
        { exec: "which aws" }, // check that AWS CLI is installed
```

src/javascript/node-project.ts
```diff
@@ -746,6 +758,63 @@ export class NodeProject extends GitHubProject {
     this.package.addKeywords(...keywords);
   }
 
+  /**
+   * Get steps for scoped package
```
Chriscbr, line 762:
```suggestion
   * Get steps for scoped package access
```

src/javascript/node-project.ts
```diff
@@ -746,6 +758,63 @@ export class NodeProject extends GitHubProject {
...
+          name: "Configure AWS Credentials",
+          uses: "aws-actions/configure-aws-credentials@v1",
+          with: {
+            "aws-access-key-id": secretToString(
+              parsedCodeArtifactOptions.accessKeyIdSecret
+            ),
+            "aws-secret-access-key": secretToString(
+              parsedCodeArtifactOptions.secretAccessKeySecret
+            ),
+            "aws-region": "us-east-2",
```
Chriscbr, line 790: Is there a reason this is hard-coded? Can we make it customizable?

src/javascript/node-package.ts
```diff
@@ -382,6 +407,13 @@ export class NodePackage extends Component {
    */
   readonly codeArtifactOptions?: CodeArtifactOptions;
```
Chriscbr, line 408: If I understand correctly, this field is being used to also provide access to Code Artifact for installing packages (not just publishing). Please update the description to reflect this

src/javascript/node-package.ts
```diff
@@ -829,9 +870,71 @@ export class NodePackage extends Component {
...
+
+    this.project.addTask("ca:login", {
+      requiredEnv: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"],
+      steps: [
+        { exec: "which aws" },
+        ...this.scopedPackagesOptions.map((scopedPackagesOption) => {
+          const { registryUrl, scope } = scopedPackagesOption;
+          const { domain, repository } =
+            extractCodeArtifactDetails(registryUrl);
+          const commands = [
```
Chriscbr, line 924: Let's add a comment for future maintainers.
```suggestion
          // reference: https://docs.aws.amazon.com/codeartifact/latest/ug/npm-auth.html
          const commands = [
```

docs/node.md
```diff
@@ -69,3 +69,33 @@ TODO
 ## Releases (CI/CD)
 
 TODO
+
+## Features
+
+### Scoped Private Packages
```
Chriscbr, line 75: Please update this section to also explain that for the feature to work on GitHub, the user must add secrets named `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to their repo (I believe this is true?)

Chriscbr (changes requested):
This looks really great @AminFazlMondo ! I've left just a few more comments, but I think we're close to getting this merged in. 🙂

src/javascript/node-project.ts
charliehart-mondo: Good pick up.

It's possible that a user will have multiple scopes each in a different region (and in fact account if they have access), so we need to ensure those details are passed to the ca:login script so it can generate an auth token in the correct region/account (via the --region and --domain-owner cli options) - I'll update that shortly.

However, the region you set for configure-aws-credentials can be independent of those if I'm not mistaken so long as you override it in the cli (it is a required property for that github action unfortunately ), We could grab the first repositoryurl from the array of scoped packages, parse the url and get the region from there - or we could leave it hard-coded.

I'm going to suggest we leave it hard-coded

charliehart-mondo: We could obviously also add another property to set it in the options, I'm not sure if it's best/helps though?

docs/node.md
charliehart-mondo: The user can set custom secret names so long as that is passed to the codeartifact options. From there configure-aws-credentials will use the default or custom secret names that get passed along. I'll update the docs to reflect that codeartifact options are required (will that be enough to prompt users to set those secrets given it's a requirement for that?)

Chriscbr (approved):
Looks great! Thanks @AminFazlMondo and @charliehart-mondo :)

PR #1689, opened 2022-03-15
chore: disable stale workflow by default

Chriscbr:
After talking with others and reviewing its impact on contributors of different GitHub projects, it feels like enabling the stale workflow may not be the right default for all projects. This PR changes the default to not adding it, so it's now opt-in.

BREAKING CHANGE: `stale.yml` GitHub workflow is no longer generated as the default behavior in projects. To keep generating it, specify `stale: true` or initialize a `github.Stale` component directly.

---


kaizen3031593 (approved):
I don't have a problem with this change, but I'm wondering if we should count it as a breaking change.

PR #1692, opened 2022-03-17
feat: set assignees for dependency upgrade PRs

Chriscbr:
Following the documentation of https://github.com/marketplace/actions/create-pull-request

---


src/javascript/upgrade-dependencies.ts
```diff
@@ -367,6 +367,8 @@ export class UpgradeDependencies extends Component {
           branch: branchName,
           title: title,
           labels: this.options.workflowOptions?.labels?.join(",") || undefined,
+          assignees:
+            this.options.workflowOptions?.assignees?.join(",") || undefined,
```
kaizen3031593, line 371: Just calling out that I found it odd that we were using `||` here but I guess that is because `[].join(",") === ""` and we want that to be `undefined` also. If there's more to it let me know, cuz it seems like we gravitate towards `??` more.

kaizen3031593 (approved):

src/javascript/upgrade-dependencies.ts
Chriscbr: Yeah I believe it's to handle the empty array case, but you're right we usually prefer ??

PR #1695, opened 2022-03-18
fix: python documentation refers to `projenrc.js` instead of `.projenrc.py`

rfrezino:
When the documentation was created it was not possible
to write the project setting in python file. Now, no
longer is necessary to use .projenrc.js as .projenrc.py
is available.

---


rfrezino:
This PR is stuck in the build step for long time

Chriscbr (approved):
Thanks!

PR #1697, opened 2022-03-20
fix: cannot use escape hatches on UpgradeDependencies workflows

### LINKED ISSUE #1256: Escape hatches do not work for `UpgradeDependencies` workflow or task

Chriscbr:
Fixes #1256

Implementation notes:

Previously, escape hatches could not be used on the workflow generated by `UpgradeDependencies` because they were generated during the `preSynthesize()` lifecycle step -- so when someone wrote `project.tryFindFile('upgrade.yml').addOverride(...)` the file would not be found. The main reason `UpgradeDependencies` was implemented this way is because it needs to generate a workflow per release branch, and it's possible users can add branches later by calling `project.release.addBranch()`.

In addition, the component also creates an task named `upgrade` whose exact details depends on the list of project dependencies, which also can be added later by calling e.g. `project.addDeps()`.

To address the first issue I added a hook mechanism to the `Release` component so that whenever `Release.addBranch` is called, the `UpgradeDependencies` component can perform its own logic to add a corresponding workflow. To address the second issue, the component now creates the `upgrade` task with a lazyily-evaluated list of tasks that is resolved at synthesis time.

---


RomainMuller (approved):
I'm mildly concerned that this might end up running into weird order-of-execution issues, but also that actually seems mostly fine in this case... I guess this is a "when all else failed" kind of feature, so...

PR #1709, opened 2022-03-23
feat!: GitHub API access via GitHub Apps in workflows

### LINKED ISSUE #1657: PROJEN_GITHUB_TOKEN endangers enterprise adoption
Per https://github.com/projen/projen/pull/1383 some of the workflows now require a `PROJEN_GITHUB_TOKEN` which is a PAT with `repo`, `packages` and `workflows` scopes. In an enterprise context, this means I will need to do one of the following:

- Create a personal PAT and then populate it on a per-repo basis (I've done this with a single repo to prove the functionality). This is not ideal since it creates a dependency on a given employee's continued presence at the company. It's also creates heartburn for the security team since the PAT's scope is not limited to the repo it is working on, and worse, from an auditing perspective, it's operating under the identity of an employee.
- I could create per-repo machine users. This makes security happy, but now my engineers need to create and add a machine user in order to add a repo. This destroys one of the major value propositions of projen: the ability for a DevOps team to enable developers to create their own repos for libraries and microservices without sacrificing consistency and without bottelnecking on the DevOps team. It also means that Finance will crucify me for melting our GitHub budget.
- Ask a GitHub Organization owner create the PAT and an Org level secret. This eliminates the manual steps involved in creating a PAT but magnifies the potential attack surface to the point where our security team turns into a swarm angry bees.

@Chriscbr mentioned a possible approach leveraging https://github.com/marketplace/actions/github-app-token which looks like might solve my problems around this. What I'm looking for is 
- reasonable security model
- minimal to zero manual interaction

Convincing GitHub to offer configuration around the scope of the `GITHUB_TOKEN` would be an ideal solution.

Workaround: live with a broken `update-main` or burn political capital with the security team.

Chriscbr:
Fixes #1657 

Allows projen workflows requiring GitHub API access beyond the permissions of GITHUB_TOKEN to specify access from either a personal access token or a GitHub App. It can be configured like so:

```ts
const { javascript, github } = require('projen');

  // ...other options
  githubOptions: {
    projenCredentials: github.GithubCredentials.fromApp({ ... }),
  },
});
```

See https://github.com/projen/projen/blob/rybickic/github-app-access/docs/github.md for more details.

This change also removes `projenUpgradeSecret` from NodeProjectOptions, which resulted in a cascade of other breaking changes. Since https://github.com/projen/projen/pull/1383 (v0.47.0) we've required provide a PROJEN_GITHUB_TOKEN for all projects, but NodeProject still assumed projects were using GITHUB_TOKEN by default, so projects were being created with separate workflows for upgrading projen and for upgrading other dependencies. I've removed that default, though it's still possible to create these kinds of workflows separately (details here: https://github.com/projen/projen/blob/rybickic/github-app-access/docs/node.md).

BREAKING CHANGE: The deprecated `projenUpgradeSecret` is removed, and GitHub API access is now customizable through `githubOptions.projenCredentials`.
* Separate tasks/workflows for upgrading projen (and approving those PRs) are no longer generated by default. `autoApproveProjenUpgrades`, `projenUpgradeAutoMerge`, and `projenUpgradeSchedule`have been correspondingly removed.

---


docs/github.md
```diff
@@ -4,13 +4,51 @@ By default, many projects are initialized with `GitHub` component to enabled Git
...
+[GITHUB_TOKEN]: https://docs.github.com/en/actions/security-guides/automatic-token-authentication
+
+### Personal Access Token
+
+Follow the [GitHub docs
+instructions](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
+for creating a personal access token. When creating the PAT, grant the token
+`repo`, `workflow` and `write:packages` and `admin:repo_hook` permissions.
+
+Add the token as a secret to your repo under the name `PROJEN_GITHUB_TOKEN`.
```
RomainMuller, line 22: Would it be possible at all to make a CLI experience to provision those in an assisted manner?

docs/tasks.md
```diff
@@ -176,3 +176,6 @@ command to the projen CLI:
 
 This means that when `yarn compile` or `npm run compile` are executed, the
 projen CLI will be invoked and the task will be executed.
+
+You can see a list of all steps in a task from the command line by passing
+the `--inspect` flag, e.g. `yarn compile --inspect`.
```
RomainMuller, line 181: Maybe make this `--dry-run`? I don't know...

src/github/api-access.ts
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Provide API access through a GitHub personal access token.
+   *
+   * The token must be added as a secret to the GitHub repo, and the name of the
+   * secret can be specified here.
+   *
+   * @see https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
+   * @default - a secret named "PROJEN_GITHUB_TOKEN"
+   */
+  public static fromPat(options: ApiAccessPatOptions = {}) {
```
RomainMuller, line 31: `fromPersonalAccessToken` -- I like to avoid abbreviating on method names, except for extremely common cases...

src/github/api-access.ts
```diff
@@ -0,0 +1,90 @@
...
+    const appIdSecret = options.appIdSecret ?? "PROJEN_APP_ID";
+    const privateKeySecret =
+      options.privateKeySecret ?? "PROJEN_APP_PRIVATE_KEY";
+
+    return new ApiAccess({
+      setupSteps: [
+        {
+          name: "Generate token",
+          id: "generate_token",
+          uses: "tibdex/github-app-token@7ce9ffdcdeb2ba82b01b51d6584a6a85872336d4",
```
RomainMuller, line 58: Some customers might have limited ability to use external actions... it's something that can be configured on an org-level, and security orgs usually see this particular use case as a significant risk.

Can we somehow internalize this? Or is this just too much effort?

src/github/workflows.ts
```diff
@@ -53,9 +54,9 @@ export class GithubWorkflow extends Component {
   public readonly file: YamlFile | undefined;
 
   /**
-   * The name of a secret that includes a PAT that can be used by workflows.
+   * Authentication method that's used by projen workflows.
    */
-  public readonly projenTokenSecret: string;
+  public readonly projenApiAccess: ApiAccess;
```
RomainMuller, line 59: The comment and type name here seem to convey somewhat different meaning... Credentials might be a better term to wrap both?

docs/github.md
Chriscbr: Looks like there is an API for creating GitHub apps automatically: https://docs.github.com/en/rest/reference/apps#create-a-github-app-from-a-manifest . I'll leave it for another PR but it's a good idea 👍

docs/tasks.md
Chriscbr: This is just some extra documentation I decided to slip in :)

src/github/api-access.ts
Chriscbr: Good point

src/github/api-access.ts
Chriscbr: It's not a crazy amount of effort to switch this to a projen "built-in" task (which references some `xxx.task.ts` file), but I'd rather just move fast and reimplement the logic later if people ask for it.

src/github/workflows.ts
Chriscbr: Ahh yeah "credentials" is the word I was looking for here. I went through a couple names like `projenApiAccess` and `projenWorkflowAuth`, but `projenGithubCredentials` or `projenApiCredentials` definitely feels like a solid improvement.

RomainMuller (approved):
Like it better like that :) GG

PR #1711, opened 2022-03-24
feat(awscdk-app): initialize sample code stack names based on project name

### LINKED ISSUE #1708: (awscdk) Use project name as stackId instead of an hardcode my-stack-dev and my-stack-prod
Sometimes I create a new cdk project and forget to change the stackId name, it´s not a big deal but I think a small change will be good to avoid problems in deploy

maybe something like `${this.project.name}-dev` and `${this.project.name}-prod` on code sample

Capturar

marciocadev:
closes #1708 

---

### DIFF
--- a/src/awscdk/awscdk-app-py.ts
+++ b/src/awscdk/awscdk-app-py.ts
@@ -5,7 +5,7 @@ import {
   CdkConfigCommonOptions,
   CdkTasks,
 } from ".";
-import { Component, DependencyType, SampleDir, TextFile } from "..";
+import { Component, DependencyType, SampleDir, SourceCode } from "..";
 import { Pytest } from "../python/pytest";
 import { PythonProject, PythonProjectOptions } from "../python/python-project";
 import { AwsCdkDepsPy } from "./awscdk-deps-py";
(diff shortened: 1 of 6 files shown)


Chriscbr (changes requested):
Looks good, just one small change

src/awscdk/awscdk-app-py.ts
```diff
@@ -113,36 +113,34 @@ class AppCode extends Component {
   constructor(project: AwsCdkPythonApp, fileName: string, cdkVersion: number) {
     super(project);
 
-    const src = new SourceCode(project, fileName);
+    const src = new TextFile(project, fileName, {
+      readonly: false,
```
Chriscbr, line 117: Let's add a `readonly` option to SourceCodeOptions instead so this can stay as SourceCode

marciocadev: About that, `marker` file are always `readonly` ? Cause I don't know if a marker file can be editable

If that's the case maybe `readonly` flag turns irrelevant and a read only file always have to take the `marker` flag

src/awscdk/awscdk-app-py.ts
```diff
@@ -113,36 +113,34 @@ class AppCode extends Component {
...
     }
-    src.line(`from ${project.moduleName}.my_stack import MyStack`);
+    src.line(`from ${project.moduleName}.main import MyStack`);
     src.line("");
     src.line("# for development, use account/region from cdk cli");
     src.open("dev_env = Environment(");
-    src.line("account=os.getenv('CDK_DEFAULT_ACCOUNT'),");
-    src.line("region=os.getenv('CDK_DEFAULT_REGION'),");
+    src.line("  account=os.getenv('CDK_DEFAULT_ACCOUNT'),");
+    src.line("  region=os.getenv('CDK_DEFAULT_REGION'),");
```
Chriscbr, line 131: Undo

Chriscbr (changes requested):

Chriscbr (approved):

PR #1714, opened 2022-03-25
feat: Add GithubActionTypeScriptProject

kaizencc:
Very, very, very rough draft on a project geared towards TypeScript github actions. My idea is for this to have the following features:

- Projen managed `action.yml` file
- automatic packaging to a single `/dist/index.js` file via the `@vercel/ncc` library.

I'm currently missing a bunch of docs and have absolutely zero tests; not even sure if it works. Ended up not having that much time to work on it today :(. There's also a lot of projen conventions that I fear I am stomping over, so happy to hear what mistakes I'm making.

---


Chriscbr (commented):
This looks great! 👍 Let me know when it's ready for another review

src/github-actions/actions.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Properties for creating a GitHubActionTypeScriptProject
+ */
+export interface GitHubActionTypeScriptOptions
+  extends TypeScriptProjectOptions {
+  /**
+   * Every GitHub Action must have a metadata file named `action.yml`.
+   * Projen will manage this file for you using the specifications of
+   * this property.
+   */
+  readonly metadata?: ActionMetadata;
```
kaizen3031593, line 15: I really want to do `Partial<ActionMetadata>` here but JSII won't let me :(.

My thought is that `ActionMetadata` should simply model the GitHub schema as an "L1", which means things like `name` are required. However, I don't really want to require `name` here in the "L2", cuz we can copy `this.name` over.

The only thing I can think of is creating another interface, `GitHubActionMetadata`, that mirrors `ActionMetadata` except it is all optional properties (i.e. basically `Partial<ActionMetadata>`). And then we have `readonly metadata?: GitHubActionMetadata;`. 

But before I do that I want to make sure there's no prior art for something like this. It seems fairly common.

src/github/actions-metadata-model.ts
```diff
@@ -0,0 +1,302 @@
...
+   * This property is required if you are using composite actions,
+   * and must be omitted otherwise.
+   */
+  readonly value?: string;
+}
+
+/**
+ * The GitHub Action runtime or composite/docker identifier.
+ */
+export enum RunsUsing {
```
kaizen3031593, line 126: I originally wanted to use an enum-like class here (like Lambda's `Runtime`) so that when `node18` comes out or whatever we can instantly support it. But I ended up having problems with that idea since `runsOn` needs to be a string. We'd need to post-process the enum-like class into a string, and I didn't want to deal with that, so I used enum instead.

kaizen3031593 (commented):
Some of my initial thoughts. Also, is 18k additions to `API.md` normal?

README.md
```diff
@@ -79,332 +80,3 @@ list):
...
 * [typescript](https://projen.io/api/API.html#projen-typescript-typescriptproject) - TypeScript project.
 * [typescript-app](https://projen.io/api/API.html#projen-typescript-typescriptappproject) - TypeScript app.
-<!-- </macro> -->
-
-> Use `npx projen new PROJECT-TYPE --help` to view a list of command line
-> switches that allows you to specify most project options during bootstrapping.
-> For example: `npx projen new jsii --author-name "Jerry Berry"`.
-
-The `new` command will create a `.projenrc.js` file which looks like this for
-`jsii` projects:
```
Chriscbr, line 89: Revert

The list in the README should get automatically updated by the projen build, no need to manually edit it.

src/github-actions/actions.ts
```diff
@@ -0,0 +1,56 @@
+import { ActionMetadata, RunsUsing } from "../github/actions-metadata-model";
+import { TypeScriptProject, TypeScriptProjectOptions } from "../typescript";
+import { YamlFile } from "../yaml";
+import { GitHubActionMetadata } from "./github-action-metadata";
+
+/**
+ * Properties for creating a GitHubActionTypeScriptProject
```
Chriscbr, line 7:
```suggestion
 * Properties for `GitHubActionTypeScriptProject`
```

src/github-actions/actions.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export interface GitHubActionTypeScriptOptions
+  extends TypeScriptProjectOptions {
+  /**
+   * Every GitHub Action must have a metadata file named `action.yml`.
+   * Projen will manage this file for you using the specifications of
+   * this property.
+   */
```
Chriscbr, line 15: Trying to tweak the API description to be a bit more intent based than implementation based

```suggestion
  /**
   * Options for the GitHub Action metadata stored in `action.yml`
   *
   * @default - an action named after the project `name` that runs from dist/index.js
   */
```

src/github-actions/actions.ts
```diff
@@ -0,0 +1,56 @@
...
+  readonly metadata?: GitHubActionMetadata;
```
Chriscbr, line 16: Suggestion to protect API surface area from future name conflicts / be more explicit
```suggestion
  readonly actionMetadata?: GitHubActionMetadata;
```

src/github-actions/actions.ts
```diff
@@ -0,0 +1,56 @@
...
+    // package as a single runnable .js file in /dist
+    this.addDevDeps("@vercel/ncc");
+    this.packageTask.reset("ncc build --source-map --license licenses.txt");
+
+    this.package.addField("main", "lib/index.js");
+    this.addGitIgnore("!/dist/");
+    this.annotateGenerated("/dist/**");
+
+    // Create metadata for projen managed `action.yml` file.
+    const defaultMetadataOptions: ActionMetadata = {
```
Chriscbr, line 40: I'm not sure if we need to define both interfaces `ActionMetadata` and `GithubActionMetadata`
```suggestion
    const defaultMetadataOptions: GithubActionMetadata = {
```

src/github-actions/actions.ts
```diff
@@ -0,0 +1,56 @@
...
+        main: "dist/index.js",
+      },
+    };
+
+    new YamlFile(this, "action.yml", {
+      obj: {
+        ...defaultMetadataOptions,
+        ...options.metadata,
+      },
+    });
```
Chriscbr, line 54: I think we should model `action.yml` file as its own component -- as an L1. (YamlFile and JsonFile etc. can all be thought of as L0s). Take a look at `typescript-config.ts` for an idea of how this should look. Since it would be an L1, the API for this component doesn't need to be intent based, and instead just creates a YamlFile with data that has been typed checked.

```ts
/**
 * Properties for `ActionMetdata`
 */
export interface ActionsMetadataOptions extends GithubActionsMetadata {}

/**
 * A GitHub Action Metadata file.
 */
export class ActionMetadata {
  constructor(project: Project, options: ActionsMetadataOptions) {
    const fileName = options.fileName ?? "action.yml";
    // ...
  }
}
```

This component can probably go in the `actions` folder.

src/github/actions-metadata-model.ts
Chriscbr: Ok

src/github/workflows-model.ts
```diff
@@ -307,6 +307,16 @@ export interface JobStep {
    */
   readonly env?: Record<string, string>;
 
+  /**
+   * TBD
```
Chriscbr, line 311: Description?

docs/api/API.md
```diff
@@ -37,6 +37,7 @@ Name|Description
 [Version](#projen-version)|*No description*
 [XmlFile](#projen-xmlfile)|Represents an XML file.
 [YamlFile](#projen-yamlfile)|Represents a YAML file.
+[actions.GitHubActionTypeScriptProject](#projen-actions-githubactiontypescriptproject)|Create a GitHub Action with TypeScript.
```
Chriscbr, line 40: This looks fine - but the API.md file in the root directory should not be getting generated (I'm not sure how it ended up in the PR). Can you manually remove it?

Chriscbr (changes requested):

src/github-actions/actions.ts
Chriscbr: In the future I think we'll auto-generate these L1s

Chriscbr:
Superseded by #1811

PR #1721, opened 2022-03-28
fix: Add support for upper case variables

marciocadev:
Fixing the issue of job variables rendering as snake case and don't respecting the original format

closes #1693 

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1738, opened 2022-04-06
fix: creating new project with node v12 fails

### LINKED ISSUE #1740: Creating new project with node v12 fails
Minimal repro:

```
$ mkdir -p /tmp/projen-test
$ cd /tmp/projen-test
$ npx projen@latest --version
0.54.8
$ node --version
12.22.0
$ npx projen@latest new node
npx: installed 65 in 3.405s
👾 Project definition file was created at /private/tmp/foo/.projenrc.js
yarn install v1.22.18
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
child_process.js:674
    throw err;
    ^

Error: Command failed: yarn install --check-files
error yaml@2.0.0: The engine "node" is incompatible with this module. Expected version ">= 14". Got "12.22.0"
error Found incompatible module.

    at checkExecSyncError (child_process.js:635:11)
    at Object.execSync (child_process.js:671:15)
    at Object.exec (/Users/rybickic/.npm/_npx/20800/lib/node_modules/projen/lib/util.js:15:19)
    at NodePackage.installDependencies (/Users/rybickic/.npm/_npx/20800/lib/node_modules/projen/lib/javascript/node-package.js:661:16)
    at NodePackage.postSynthesize (/Users/rybickic/.npm/_npx/20800/lib/node_modules/projen/lib/javascript/node-package.js:300:18)
    at NodeProject.synth (/Users/rybickic/.npm/_npx/20800/lib/node_modules/projen/lib/project.js:281:22)
    at evalmachine.<anonymous>:12:20
    at Script.runInContext (vm.js:130:18)
    at Object.runInContext (vm.js:293:6)
    at createProject (/Users/rybickic/.npm/_npx/20800/lib/node_modules/projen/lib/projects.js:74:8)
```

This is causing an integration test to fail, which is causing new development builds to fail as well (example: https://github.com/projen/projen/pull/1721).

Workaround: add the option `--package-manager="npm"` when running `projen new`.

It looks like `yarn` is not respecting that `yaml` should be a bundled dependency of projen. I think this is related to the fact that package.json is listing these dependencies as both `bundleDependencies` and regular runtime `dependencies`.

Chriscbr:
Currently new projen development builds are failing (example: https://github.com/projen/projen/pull/1721) because `yaml` published v2.0.0 to the `next` tag, in which they bumped the minimum node version they require from 12 to 14. Since new development builds are automatically picking this up and projen still supports node v12, this PR pins the version of `yaml` so the new version isn't picked up by `yarn install`.

This should fix #1740

We should also upgrade the minimum node version projen supports, but it's better to do that in a separate PR. (Issue: https://github.com/projen/projen/issues/1739)

---

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -175,7 +175,7 @@
     },
     {
       "name": "yaml",
-      "version": "next",
+      "version": "2.0.0-11",
       "type": "bundled"
     },
     {
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -27,7 +27,7 @@ const project = new cdk.JsiiProject({
 
   bundledDeps: [
     "conventional-changelog-config-spec",
-    "yaml@next",
+    "yaml@2.0.0-11",
     "fs-extra",
     "yargs",
     "case",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "semver": "^7.3.5",
     "shx": "^0.3.4",
     "xmlbuilder2": "^2.4.1",
-    "yaml": "next",
+    "yaml": "2.0.0-11",
     "yargs": "^16.2.0"
   },
   "bundledDependencies": [
--- a/test/__snapshots__/new.test.ts.snap
+++ b/test/__snapshots__/new.test.ts.snap
@@ -256,11 +256,12 @@ project.synth();",
 }
 `;
 
-exports[`projen new node --outdir path/to/mydir 1`] = `
+exports[`projen new node --outdir path/to/mydir --package-manager="npm" 1`] = `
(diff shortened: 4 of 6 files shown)


Chriscbr:
Ahh, that's really pesky. So the build is failing because of integration tests depending on the live version of projen (also just reported it here: https://github.com/projen/projen/issues/1740). Definitely something that needs to be fixed.

For now I've added a workaround to fix the integration test for now (changing the package manager for new installations).

kaizen3031593 (approved):

PR #1758, opened 2022-04-10
feat: Renovatebot configuration

### LINKED ISSUE #1667: feature request: support Renovatebot by generating renovate.json
It would be nice to have projen to support also Renovatebot configuration, as versions defined by projen should be exluded from updates.

Relates also to #1659.

Features expected:

- Write ignores to Projen managed versions
- Add "sensible defaults" to e.g. grouping of updates

Hi-Fi:
Allow configuring renovatebot to each project type with few configurable fields.

Fix #1667
---


Hi-Fi:
@eladb @Chriscbr does this look OK?

Sidenote: [CONTRIBUTING.md](CONTRIBUTING.md) is a bit vague with
> Shout out to collaborators.

test/javascript/node-project.test.ts
```diff
@@ -661,6 +661,24 @@ test("enabling dependabot does not overturn mergify: false", () => {
   expect(snapshot).toHaveProperty([".github/dependabot.yml"]);
 });
 
+test("enabling renovatebot does not overturn mergify: false", () => {
+  // WHEN
+  const project = new TestNodeProject({
+    renovatebot: true,
+    projenVersion: "0.50.0",
+    minNodeVersion: "14.0.0",
```
pgollucci, line 669: Will be default soon. Do we need the minVersion?

test/javascript/node-project.test.ts
```diff
@@ -661,6 +661,24 @@ test("enabling dependabot does not overturn mergify: false", () => {
...
+    mergify: false,
```
pgollucci, line 670: Does renovate bot work  with Mergify, Dependabot, or the new Github branch merge stuff?

pgollucci (commented):
I might separate out json5 from the rest but in general it looks good to me

test/javascript/node-project.test.ts
Hi-Fi: Might be that test would be fine without it. Added it to ensure that `.projen/deps.json` defines version for `@types/node`. 

Same thing for `projenVersion`. 

Do you see that those version definitions should be removed?

test/javascript/node-project.test.ts
Hi-Fi: I don't see why it wouldn't, as it "just" creates PRs like users and dependabot.

Hi-Fi:
> I might separate out json5 from the rest but in general it looks good to me

Main thing here is the possibility to add the comment (but only inside JSON, not for whole object), as Renovatebot parser fails if there's keys that are not recognized (like `"//"` used elsewhere).

Programmatic creation of JSON5 doesn't have possibility for comments, so adding that "projen created" comment is still pure string handling.

It could be done if needed also other way around, so first generate full JSON and then regexp change that `"//"` element to comment.

Hi-Fi:
@pgollucci do you think there's some changes needed or is this OK to be merged? Improving configuration options etc. is then easier when base is available (which hopefully would work users as is)

Hi-Fi:
@eladb @Chriscbr @pgollucci 

This has been some time already without comments or anything. Is there some issue that should be fixed, or could this be merged and taken to use so it's easier to see if some features to this are needed.

src/renovatebot.ts
```diff
@@ -0,0 +1,159 @@
...
+   * (e.g. different for weekdays and weekends). Multiple rules are
+   * handles as OR.
+   *
+   * Some normal scheduling values defined in enum `RenovatebotScheduleInterval`.
+   *
+   * See: https://docs.renovatebot.com/configuration-options/#schedule
+   *
+   * @default RenovatebotScheduleInterval.ANY_TIME
+   */
+  readonly scheduleInterval?: string[];
```
Chriscbr, line 22: In TypeScript an enum value can be interchangeable with a string, but when JSII translates this into other languages, this will not be valid. See https://aws.github.io/jsii/specification/2-type-system/#enum

We can either change the default to match the type, like `@default ["at any time"]`, or change the type to match the default, like `readonly scheduleInterval?: RenovateScheduleInterval;` (and convert it to a string array as an implementation detail). Another option is to create a union-like class so that the user can custom options like this: `RenovateScheduleInterval.custom("every third wednesday")` (example: https://github.com/aws/aws-cdk/blob/4cb2fe8ff410cbb769565453d43feaa820047acb/packages/%40aws-cdk/aws-apigatewayv2/lib/parameter-mapping.ts#L14)

src/renovatebot.ts
```diff
@@ -0,0 +1,159 @@
...
+   * List of labels to apply to the created PR's.
+   */
+  readonly labels?: string[];
+}
+
+/**
+ * How often to check for new versions and raise pull requests for version
+ * updates.
+ *
+ * From https://docs.renovatebot.com/presets-schedule/
```
Chriscbr, line 53:
```suggestion
 * @see https://docs.renovatebot.com/presets-schedule/
```

docs/api/API.md
```diff
@@ -23,6 +23,7 @@ Name|Description
 [ProjectBuild](#projen-projectbuild)|Manages a standard build process for all projects.
 [Projects](#projen-projects)|Programmatic API for projen.
 [Projenrc](#projen-projenrc)|Sets up a project to use JSON for projenrc.
+[Renovatebot](#projen-renovatebot)|Defines renovatebot configuration for projen project.
```
Chriscbr, line 26: Let's also mention this in `docs/node.md`.

test/javascript/node-project.test.ts
Chriscbr: I think we can remove the projenVersion and minNodeVersion safely. (unless there's a reason we don't want @types/node to be updated to the latest minor versions?) It looks like the component should ignore projen by default regardless

test/javascript/node-project.test.ts
```diff
@@ -662,6 +662,24 @@ test("enabling dependabot does not overturn mergify: false", () => {
...
+  });
+
+  // THEN
+  const snapshot = synthSnapshot(project);
+  // Note: brackets important, they prevent "." in filenames to be interpreted
+  //       as JSON object path delimiters.
+  expect(snapshot).not.toHaveProperty([".mergify.yml"]);
+  expect(snapshot).toHaveProperty(["renovate.json5"]);
+  expect(snapshot["renovate.json5"]).toMatchSnapshot();
```
Chriscbr, line 680: Can we also add an assertion on the list of ignored dependencies, by parsing the json5 file? e.g.

```suggestion
  expect(json5.parse(snapshot["renovate.json5"]).ignoreDeps).toEqual(...);
  expect(snapshot["renovate.json5"]).toMatchSnapshot();
```

src/renovatebot.ts
```diff
@@ -0,0 +1,159 @@
...
+   * Projen upgrades are covered through the `ProjenUpgrade` class.
+   *
+   * @default true
+   */
+  readonly ignoreProjen?: boolean;
+
+  /**
+   * List of labels to apply to the created PR's.
+   */
+  readonly labels?: string[];
```
Chriscbr, line 46: Where does this field get used? (this definitely seems like a useful option to have!)

src/json.ts
```diff
@@ -37,11 +39,15 @@ export class JsonFile extends ObjectFile {
...
 
-    if (this.marker) {
+    if (this.marker && !this.isJson5) {
       sanitized["//"] = this.marker;
     }
 
     let content = JSON.stringify(sanitized, undefined, 2);
+    if (this.marker && this.isJson5) {
+      content = content.slice(0, -1) + `// ${this.marker}\n}`;
+    }
```
Chriscbr, line 49: Can we add a unit test to `json.test.ts` that checks that json5 files have a marker added as a comment instead?

Chriscbr (changes requested):
Hey @Hi-Fi - looks great, I've got just a couple comments on this PR!

src/renovatebot.ts
Hi-Fi: Apparently I wrote API first, and forgot to take this to use. So just usage missing.

src/json.ts
Hi-Fi: Sure, will do that

test/javascript/node-project.test.ts
Hi-Fi: OK, I'll remove those. As projen's own upgrade is now included in `upgrade` script, I think it should be only ignored if user specifies the version in `.projenrc.js`.

It's then user's own responsibility to add mutation check to project, or check PR carefully if upgrade to projen itself is included.

src/renovatebot.ts
Hi-Fi: Noticed that JSII thing, as original was `RenovateScheduleInterval | string` type.

Maybe would go with string, and just mention that some basic values are available from enum.

test/javascript/node-project.test.ts
Hi-Fi: Added checking of ignores in the test

src/renovatebot.ts
Hi-Fi: Added usage and included in tests

docs/api/API.md
Hi-Fi: Added mention next to dependabot

Chriscbr (approved):
LGTM.

PR #1766, opened 2022-04-13
feat: refactor mergify APIs

Chriscbr:
The motivation for this PR is to simplify the APIs for `Mergify` and `AutoMerge`. Mergify configuration is currently getting configured from multiple places (both `mergifyOptions` and `autoMergeOptions`), which leads to confusion since it's likely many users will only discover one or the other.

In another sense, projen's API is exposing options both through the L2 `Mergify` (an intent based API for managing Mergify configuration files) and the L3 `AutoMerge` (and intent based API for managing the concept of auto-merging, which relies on `Mergify` as an implementation detail). I think this makes the design more complicated / confusing to users.

Instead this PR changes the mental model so we just expose the L3 API (`autoMerge` and `autoMergeOptions`) in base project options, so mergify has to be either configured through `AutoMerge`, or created and managed separately.

BREAKING CHANGE: `mergify` and `mergifyOptions` are no longer available to directly configure at the project level. Instead, use `autoMerge` and `autoMergeOptions`. If you would like to disable automerging, add `autoMerge: false` to your project constructor and create a new `Mergify` component directly.
* Mergify cannot be accessed through `Github` or `project.github` anymore, instead it is a property of `AutoMerge` or `project.autoMerge`.

---


pgollucci:
@Mergifyio rebase

docs/github.md
```diff
@@ -40,6 +40,19 @@ const project = new javascript.NodeProject({
 });
 ```
 
+## Automatic Merging
```
kaizen3031593, line 43: I feel like this doc is missing a code snippet or two. I feel like there should be an example for configuring automatic merging _as well as_ what to do if you want to configure `Mergify` in your project for something else.

src/github/github-project.ts
```diff
@@ -97,8 +80,17 @@ export interface GitHubProjectOptions extends ProjectOptions {
...
+   * supports using Mergify to merge pull requests. Has no effect if `github` is
+   * set to false.
+   *
+   * @default true
+   */
+  readonly autoMerge?: boolean;
+
+  /**
+   * Configure options for automatic merging pull requests on GitHub. Has no
+   * effect if `autoMerge` is set to false or `github` is set to false.
```
kaizen3031593, line 93: I'm not too familiar with projen customs, but should we simply error out if `autoMergeOptions` is set and `autoMerge` or `github` is set to false? Seems odd to just ignore this property. Same goes for `autoMerge`

src/github/github.ts
```diff
@@ -2,26 +2,11 @@ import { Component } from "../component";
...
 import { PullRequestLint, PullRequestLintOptions } from "./pull-request-lint";
 import { GithubWorkflow } from "./workflows";
 
 export interface GitHubOptions {
-  /**
-   * Whether mergify should be enabled on this repository or not.
-   *
-   * @default true
-   */
-  readonly mergify?: boolean;
```
kaizen3031593, line 16: Since this is a fairly big API transition and we're removing a few properties (esp the non-deprecated ones like these), I feel like we need to have strong _migration_ docs between projen versions for users who are broken.

I _think_ I understand the vision of projen where we have lots of major versions and you don't really need to upgrade projen in your project. But if I'm a die-hard projen user with some ideas on how to set up mergify in an old project (in projen v0.34 for example) and it now doesn't exist when I'm trying to set up a new project (in projen v0.59 or something), I'm going to want to have some docs that tell me what to do now. I don't really know if there's a right way to achieve this though.

test/github/auto-merge.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test("provides access to mergify", () => {
+    const project = createProject();
+
+    const autoMerge = new AutoMerge(project.github!, {
+      approvedReviews: 2,
+      blockingLabels: ["draft"],
+    });
+    autoMerge.mergify.addRule({
```
kaizen3031593, line 38: is exposing `mergify` a leaky abstraction? especially if we plan on providing other auto-merging backends in the future (per the docs you wrote).

kaizen3031593 (changes requested):
Some minor comments, mostly revolving around documentation.

src/github/github.ts
kaizen3031593: I guess the BREAKING CHANGE comments go into the changelog? but do people actually rely on the changelog for help?

Chriscbr: I think putting all breaking changes / migration suggestions in the changelog is probably okay since it's a single source of truth for high-level API changes and it's easy to ctrl+f for some API name. But I agree it's a good idea to link to detailed docs as well.

We don't generate a single changelog file anywhere right now, so maybe that should be addressed first...

test/github/auto-merge.test.ts
Chriscbr: Hmmmm, yeah you're totally right.

In that case I'm not sure how much the API in this PR makes sense anymore. 😅 Maybe I'll put this back on hold for now

PR #1768, opened 2022-04-14
chore: bump minimum node version to 14.x

### LINKED ISSUE #1739: Update projen to require node v14
Node v12 is EOL in April 30, 2022 (ref: https://nodejs.org/en/about/releases/). Updating this will ensure we're able to pick up fresh/stable dependencies that depend on node v14 and above, and that we aren't aren't susceptible to security vulnerabilities in old node versions.

Chriscbr:
Fixes #1739

BREAKING CHANGE: new versions of projen now require node v14, since node v12 reaches EOL on April 30, 2022.

---


Chriscbr:
Added the pr/do-not-merge label since I'm not sure if there's actually a pressing need to merge this in immediately. But Node v12 is deprecated at the end of this month and we have no interest in supporting that version after it has been unsupported, so enforcing it in new versions with the engine requirement seems like the right thing to do for the project's health. 👍

.projenrc.js
```diff
@@ -59,8 +59,8 @@ const project = new cdk.JsiiProject({
 
   projenDevDependency: false, // because I am projen
   releaseToNpm: true,
-  minNodeVersion: "12.7.0",
-  workflowNodeVersion: "12.22.0", // required by @typescript-eslint/eslint-plugin@5.5.0
+  minNodeVersion: "14.0.0",
+  workflowNodeVersion: "14.17.0", // required by @typescript-eslint/eslint-plugin@5.19.0
```
Chriscbr, line 63: projen needs a higher version of node on its GitHub actions runner in order to install all of projen's dev dependencies successfully and run full builds. But consumers of projen don't actually need those since we bundle all of our runtime dependencies. (For example, it's safe to run `npx projen new python` when the installed version of node is v14.0.0). Hence why these version numbers are different.

pgollucci: Node < 14.11.0 is vulnerable to CVEs. 14.x is in maintenance. Its 2yrs old this week. Should we really be supporting that?  This feels like overkill.

https://www.cvedetails.com/vulnerability-list/vendor_id-12113/Nodejs.html

pgollucci: the workflow separation, not the 12->14,

kaizen3031593 (approved):
I don't have a problem with this. I agree that the workflow separation seems odd but I think the point is that projen should support whatever min node version it can.

Chriscbr:
Before we merge this I think I need to undo some changes made in #1709 that removed the default "upgrade-projen" workflow. I'm realizing that I think having separate upgrade workflows/tasks for upgradingn projen and for upgrading other dependencies might actually be the happier path, and removing that was probably a mistake. projen does have breaking changes every now and then, and so including it with all other dependencies is a bit too chaotic / unreliable for most projects. In particular, this change may not get automatically picked up with the `upgrade.yml` workflow because installing the new version of projen on a workflow using node v12 will just fail (and not even attempt to create a PR).

PR #1771, opened 2022-04-15
feat: add `addPreBuildSteps` API to build workflow

kaizen3031593:
Exposing pre-build steps on the build workflow. `BuildWorkflow` has a property called `preBuildSteps` so I think this API was unnecessarily omitted.

In addition to `addPostBuildSteps`, `BuildWorkflow` also has APIs for `addPostBuildJob` and `addPostBuildJobTask`, but I can't see a use case for needing to add Jobs pre-build so I did not add the corresponding `pre` APIs.

---

### DIFF
--- a/test/javascript/node-project.test.ts
+++ b/test/javascript/node-project.test.ts
@@ -561,6 +561,41 @@ test("extend github release workflow", () => {
   );
 });
 
+describe("extend github build workflow", () => {
+  let project: TestNodeProject;
+  beforeEach(() => {
+    project = new TestNodeProject();
+  });
+
+  test("addPostBuildSteps", () => {
+    // WHEN
+    project.buildWorkflow!.addPostBuildSteps({
+      name: "hello-world",
+      run: "echo 'hello world'",
+    });
+
+    // THEN
+    const workflow = synthSnapshot(project)[".github/workflows/build.yml"];
+    expect(workflow).toContain(
+      "- name: build\n        run: npx projen build\n      - name: hello-world\n        run: echo 'hello world'"
+    );
+  });
+
+  test("addPreBuildSteps", () => {
+    // WHEN
+    project.buildWorkflow!.addPreBuildSteps({
+      name: "hello-world",
+      run: "echo 'hello world'",
+    });
+
+    // THEN
+    const workflow = synthSnapshot(project)[".github/workflows/build.yml"];
+    expect(workflow).toContain(
+      "- name: hello-world\n        run: echo 'hello world'\n      - name: build\n        run: npx projen build"
+    );
+  });
+});
+
 describe("scripts", () => {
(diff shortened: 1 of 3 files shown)


test/javascript/node-project.test.ts
```diff
@@ -561,6 +561,41 @@ test("extend github release workflow", () => {
   );
 });
 
+describe("extend github build workflow", () => {
```
kaizen3031593, line 564: I couldn't find a test file for `BuildWorkflow` but similar tests were done in this file so I stuck them here. But happy to get some direction on where I should be adding tests.

test/javascript/node-project.test.ts
```diff
@@ -561,6 +561,41 @@ test("extend github release workflow", () => {
...
+    // WHEN
+    project.buildWorkflow!.addPostBuildSteps({
+      name: "hello-world",
+      run: "echo 'hello world'",
+    });
+
+    // THEN
+    const workflow = synthSnapshot(project)[".github/workflows/build.yml"];
+    expect(workflow).toContain(
+      "- name: build\n        run: npx projen build\n      - name: hello-world\n        run: echo 'hello world'"
```
Chriscbr, line 580: Can we make this less whitespace / indentation sensitive? How about we `YAML.parse` the contents and make an assertion on that perhaps?

test/javascript/node-project.test.ts
Chriscbr: This is fine - but I agree with your intuition that this should belong in a separate file specific to the component. 🙂

Chriscbr (changes requested):
LGTM 👍 

It's worth calling out that any steps added through this API will only affect the `build.yml` workflow that runs on pull requests -- but not in other builds, like in `release.yml` and `upgrade.yml`. We actually have another API for doing this: `workflowBootstrapSteps`.

I wonder if it would make sense to rename the component `PullRequestBuildWorkflow` to make this more clear?

kaizen3031593:
> It's worth calling out that any steps added through this API will only affect the `build.yml` workflow that runs on pull requests -- but not in other workflows where the repo is built, like in `release.yml` and `upgrade.yml`. We actually have another API for doing this: `workflowBootstrapSteps`.

🤦is there really a use case for this then? Seems like `workflowBootstrapSteps` covers this and more. `addPreBuildSteps` is only a use case when someone wants a prebuild step in `build.yml` and not `release.yml` or `upgrade.yml`. I can't think of why...

This might be another case of "had trouble finding the property I needed" rather than "projen is missing an API"

Chriscbr:
@kaizen3031593 That's fair. I think it might still make sense to add this API though, if only to mirror the `addPostBuildSteps` API for feature parity. 🙂

PR #527, opened 2021-02-03
bug(dependencies): parsing package aliases is broken

smacintyre:
## Description

`Dependencies.praseDependency` doesn't parse dependencies with package aliases correctly.

This PR fixes the handling of package aliases.

## Example

Given `bar@npm:@bar/legacy` should return `{ name: "bar", version: "npm:@bar/legacy" }`.

Current it will return `{ name: "bar", version: "npm:" }`

## Reasoning

The current bug is blocking me from adding TailwindCSS to a `ReactTypescriptProject`. 

`create-react-app` doesn't support PostCSS 8 and so TailwindCSS has a PostCSS 7 compatibility version for React: `tailwindcss@npm:@tailwindcss/postcss7-compat`. It is installed as a package alias for the `tailwindcss` package. See: https://tailwindcss.com/docs/guides/create-react-app

The presence of the second `@` cause `Dependencies.praseDependency` to render the version as `npm:`. Which causes sadness.

Please accept this humble pull request.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1779, opened 2022-04-19
fix: cannot use file path dependencies

### LINKED ISSUE #1777: Error when using file dependencies
It should be possible to use local paths as dependencies as described in the [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#local-paths). For example:

```
project.addPeerDeps("jsii-docgen@file:../jsii-docgen")
```

But this currently causes an error message like this:

```
TypeError: Invalid comparator: file:../jsii-docgen
    at Comparator.parse (/home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/comparator.js:38:13)
    at new Comparator (/home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/comparator.js:22:10)
    at /home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/range.js:133:47
    at Array.map (<anonymous>)
    at Range.parseRange (/home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/range.js:133:35)
    at /home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/range.js:34:22
    at Array.map (<anonymous>)
    at new Range (/home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/classes/range.js:34:8)
    at Object.minVersion (/home/ec2-user/development/cdkhacks/node_modules/projen/node_modules/semver/ranges/min-version.js:6:11)
    at NodePackage.renderDependencies (/home/ec2-user/development/cdkhacks/node_modules/projen/lib/javascript/node-package.js:489:46)
```

Chriscbr:
Fixes #1777 

We've supported dependency versions that aren't strictly semver-strings for a while now (see for example: https://github.com/projen/projen/pull/527). This simply makes the logic a little more robust / less prone to causing sudden errors.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license


pgollucci:
👍

src/javascript/node-package.ts
```diff
@@ -1406,3 +1406,11 @@ function determineLockfile(packageManager: NodePackageManager) {
 
   throw new Error(`unsupported package manager ${packageManager}`);
 }
+
+function minVersion(version: string): string | undefined {
+  if (semver.validRange(version)) {
+    return semver.minVersion(version)?.version;
+  } else {
+    return version;
```
kaizen3031593, line 1414: question: what about when someone erroneously supplies an invalid semver version. We don't error out, and maybe they would expect that. Can we attach a warning or something to this? I'm guessing [Annotations](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/core/lib/annotations.ts) don't exist in projen, so might be wishful thinking.

kaizen3031593 (approved):
Approved, because I don't have any blocking comments.

src/javascript/node-package.ts
Chriscbr: If the semver version is invalid, it will be caught when `yarn install` or `npm install` fails.

test/javascript/node-package.test.ts
```diff
@@ -96,6 +103,14 @@ function mockYarnInstall(
   );
 }
 
+function minVersion(version: string): string | undefined {
```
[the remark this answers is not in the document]
Chriscbr: Fair point. Moved it to a shared location

PR #1781, opened 2022-04-19
feat: suppress cdk template using -q instead of /dev/null (#1780)

### LINKED ISSUE #1780: Use quiet option in cdk synth:silent
Currently the cdk synth:silent step suppresses the template by redirecting all output to ```/dev/null```. The cdk synth step might produce output that is useful, e.g. using when using cfn-nag or a MFA prompt for the AWS CLI.  The cdk synth command has a quiet option to suppress the CloudFormation template. 

Is there any reason no to use the --quiet flag instead of ```/dev/null```?

marnixdessing:
Fixes #1780 

---


Chriscbr (approved):
Makes sense, good catch @marnixdessing ! Thanks for submitting this PR 🙂

PR #1783, opened 2022-04-19
fix: cannot disable automerge and while keeping mergify enabled

### LINKED ISSUE #1765: No way to disable auto-merge
Looks like the mergify configuration comes with an auto-merge rule by default, and there's no way to disable it. The option `mergifyOptions` has the ability to add additional auto-merge rules, but no way to remove the default rule.

I'm looking specifically at the `AwsCdkTypeScriptApp` project, but I'm seeing that this issue applies to other projects as well.

Chriscbr:
Supersedes #1766
Fixes #1765

---


test/github/mergify.test.ts
```diff
@@ -13,9 +13,10 @@ describe("mergify", () => {
     expect(snapshot[".mergify.yml"]).toMatchSnapshot();
   });
 
-  test("with options", () => {
+  test("with automerge", () => {
```
kaizen3031593, line 16:
```suggestion
  test("with automerge options", () => {
```

src/javascript/node-project.ts
```diff
@@ -603,7 +604,8 @@ export class NodeProject extends GitHubProject {
       }
     }
 
-    if (this.github?.mergify && this.buildWorkflow?.buildJobIds) {
+    const autoMerge = options.autoMerge ?? true;
+    if (autoMerge && this.github?.mergify && this.buildWorkflow?.buildJobIds) {
```
kaizen3031593, line 608: nit, but how do you feel about just going `options.autoMerge !== false &&...` it just feels like we're creating a const for no reason as we don't use it outside of the if.

src/javascript/node-project.ts
```diff
@@ -339,6 +339,7 @@ export class NodeProject extends GitHubProject {
 
   /**
    * Automatic PR merges.
+   * @default true
```
kaizen3031593, line 342: default is boolean but here, `autoMerge` is of type `AutoMerge`

kaizen3031593: It seems odd to me that `autoMerge` as a property is boolean but `autoMerge` as the exposed reference is type `AutoMerge`. At a minimum this one should be documented better than "Automatic PR merges."

src/github/github-project.ts
```diff
@@ -96,9 +96,15 @@ export interface GitHubProjectOptions extends ProjectOptions {
    */
   readonly autoApproveOptions?: AutoApproveOptions;
 
+  /**
+   * Enable automatic merging on GitHub. Has no effect if `github.mergify`
+   * is set to false.
+   */
+  readonly autoMerge?: boolean;
```
kaizen3031593, line 103: Should we have `@default true` here (instead of the other place)?

kaizen3031593 (changes requested):
Some minor nits

src/github/github-project.ts
Chriscbr: Whoops, yes thanks for catching this.

src/javascript/node-project.ts
Chriscbr: Updating the docstring

src/javascript/node-project.ts
Chriscbr: I don't have a strong opinion on it, I guess I just prefer this since it's a pattern we've been using in the rest of projen (`const something = input ?? default`). But yeah I can see arguments for either side. I'll inline it to make it more concise.

kaizen3031593 (approved):
Yay!

PR #1786, opened 2022-04-21
chore(jsii): Remove `--no-fix-peer-dependencies`

matt9ucci:
The `--fix-peer-dependencies` flag of jsii is hidden and no longer has any effect. It should be removed to avoid redundancy.
See the commit: aws/jsii@d1ff3e678c9011df5e51816b37bec77767b6e8e2

---


Chriscbr (approved):
Thanks @matt9ucci !

PR #1809, opened 2022-04-28
fix: cannot create `PythonProject` with `.projenrc.js`

### LINKED ISSUE #1807: Python-based projects can no longer use existing .projenrc.js files
I've got a number of projects that are PythonProject based, and now with the latest updates an `npx projen` no longer runs because it's trying to find a .projenrc.py file. I see options to turn that off, as well as turning on a .json option, but I need to still use the existing .projenrc.js files and don't see the option to do it.

To make this work, I've had to go hacking at the default task:

`this.tasks.tryFind('default')?.exec('node .projenrc.js');`

which isn't something I feel I should have to do to maintain backwards compatability.

Chriscbr:
Fixes #1807

To manage a PythonProject with a JavaScript projenrc file, the project will need to have these options configured:
```js
projenrcJs: true,
```

Or pass the option in the command line when creating the project, e.g. `npx projen new python --projenrc-js`.

---

### DIFF
--- a/src/python/python-project.ts
+++ b/src/python/python-project.ts
@@ -1,8 +1,15 @@
 import { GitHubProject, GitHubProjectOptions } from "../github";
+import {
+  Projenrc as ProjenrcJs,
+  ProjenrcOptions as ProjenrcJsOptions,
+} from "../javascript/projenrc";
 import { ProjectType } from "../project";
 import { Pip } from "./pip";
 import { Poetry } from "./poetry";
-import { Projenrc as ProjenrcPython, ProjenrcOptions } from "./projenrc";
+import {
+  Projenrc as ProjenrcPython,
+  ProjenrcOptions as ProjenrcPythonOptions,
+} from "./projenrc";
 import { Pytest, PytestOptions } from "./pytest";
 import { PytestSample } from "./pytest-sample";
 import { IPythonDeps } from "./python-deps";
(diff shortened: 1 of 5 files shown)


src/python/python-project.ts
```diff
@@ -134,7 +141,23 @@ export interface PythonProjectOptions
...
-  readonly projenrcPythonOptions?: ProjenrcOptions;
+  readonly projenrcPythonOptions?: ProjenrcPythonOptions;
+
+  /**
+   * Use projenrc in javascript.
+   *
+   * This will install `projen` as a JavaScript dependency and add a `synth`
+   * task which will run `.projenrc.js`.
+   *
+   * @default true
```
kaizencc, line 152: Color me confused... how can both `projenrcJs` and `projenrcPython` default to true?

kaizencc: Wait, this default looks like it should be false from the code.

kaizencc (changes requested):
It seems like if a user wants to use the JavaScript dependency, then must explicitly set `projenrcPython: false`. Seems like a poor user experience, or am I missing something? I feel like setting `projenrcJs: true` (assuming the default is actually false, not true) should be enough signal to assume the user does not want the Python dependency projenrc.

src/python/python-project.ts
Chriscbr: Whoops! good catch

Chriscbr:
Good callouts, I agree that disabling `projenrcPython` in order to use the other options is unnecessary 👍

kaizencc (approved):

PR #1819, opened 2022-05-02
feat: `tryRemoveFile` escape hatch

Chriscbr:
Adds an escape hatch method to all projects that allows removing files.

In the future, if projen uses the `constructs` programing model, this probably won't be needed since constructs have their own `node.tryRemoveChild` method - but for now this should be able to substitute for most use cases.

---


kaizencc (approved):
Looks good to me

PR #1832, opened 2022-05-06
fix(typescript): eslint task lints projenrcts

### LINKED ISSUE #1831: eslint task doesn't lint projenrc when using projenrc-ts
When starting a typescript project using `--projenrc-ts` the eslint task isn't updated to lint `.projenrc.ts`:

```json5
    // ...
    "eslint": {
      "name": "eslint",
      "description": "Runs eslint against the codebase",
      "steps": [
        {
          "exec": "eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern src test build-tools .projenrc.js"
        }
      ]
    },
   // ...
```

andrestone:
fixes #1831 

BREAKING CHANGE: EslintOptions.lintProjenRc is now a string
---

### DIFF
--- a/src/javascript/eslint.ts
+++ b/src/javascript/eslint.ts
@@ -39,8 +39,15 @@ export interface EslintOptions {
   readonly ignorePatterns?: string[];
 
   /**
+   * Projenrc file to lint. Use empty string to disable.
+   * @default PROJEN_RC
+   */
+  readonly lintProjenRcFile?: string;
+
+  /**
    * Should we lint .projenrc.js
    * @default true
+   * @deprecated use lintProjenRcFile instead
    */
   readonly lintProjenRc?: boolean;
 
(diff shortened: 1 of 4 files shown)


src/javascript/eslint.ts
```diff
@@ -39,10 +39,10 @@ export interface EslintOptions {
   readonly ignorePatterns?: string[];
 
   /**
-   * Should we lint .projenrc.js
-   * @default true
+   * Projenrc file to lint. Use empty string to disable.
+   * @default PROJEN_RC
    */
-  readonly lintProjenRc?: boolean;
+  readonly lintProjenRc?: string;
```
Chriscbr, line 45: How about we add a new field `lintProjenRcFile` instead and deprecate the existing field? I think could help us avoid a breaking change for users.

andrestone: I've considered that but then I thought maybe not many users would use this API directly. Can definitely do it if it's more appropriate.

Chriscbr (approved):

PR #1850, opened 2022-05-13
fix: resolve issue with option fqns in java when they do not exist in…

agdimech:
This PR resolves an issue whereby the Java projenrc synthesizer was throwing an error when fqns couldn't be resolved to a type in the jsii manifest.

```
return [jsiiType.namespace, jsiiType.name].filter((x) => x).join(".");
                     ^
TypeError: Cannot read property 'namespace' of undefined
```

The use case for this happening is I have a custom project type that extends JavaProject. As part of the bootstrap options iterator, it tries to resolve projen specific fqns i.e: projen.github.* into types within the .jsii manifest which I generate as part of my package. Given projen types are not bundled as part of my .jsii file, this will error out as above.

I am not sure if this is related, but I did notice the projen submodules were all empty in my .jsii file as shown below:

`"projen": {
      "submodules": {
        "projen.awscdk": {},
        "projen.build": {},
        "projen.cdk": {},
        "projen.cdk8s": {},
        "projen.cdktf": {},
        "projen.github": {},
        "projen.github.workflows": {},
        "projen.gitlab": {},
        "projen.java": {},
        "projen.javascript": {},
        "projen.python": {},
        "projen.release": {},
        "projen.typescript": {},
        "projen.vscode": {},
        "projen.web": {}
      }`

Repro steps:

```
mkdir foo && cd foo
npx projen new --from aws-prototyping-sdk pdk-pipeline-java
```

---

### DIFF
--- a/src/java/projenrc.ts
+++ b/src/java/projenrc.ts
@@ -150,10 +150,9 @@ export class Projenrc extends Component {
 
     const optionFqns: Record<string, string> = {};
     for (const option of bootstrap.type.options) {
-      if (option.fqn) {
-        optionFqns[option.name] = toJavaFullTypeName(
-          jsiiManifest.types[option.fqn]
-        );
+      const fqnType = jsiiManifest.types[option.fqn];
+      if (option.fqn && fqnType) {
+        optionFqns[option.name] = toJavaFullTypeName(fqnType);
       }
     }
 
--- a/src/java/projenrc.ts
+++ b/src/java/projenrc.ts
@@ -191,11 +191,16 @@ export class Projenrc extends Component {
   }
 }
 
-export function generateJavaOptionNames(options: ProjectOption[], jsiiManifest: any) {
+export function generateJavaOptionNames(
+  options: ProjectOption[],
+  jsiiManifest: any
+) {
   const optionFqns: Record<string, string> = {};
   for (const option of options) {
     if (option.fqn && jsiiManifest.types[option.fqn]) {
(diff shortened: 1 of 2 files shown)


src/java/projenrc.ts
```diff
@@ -150,10 +150,9 @@ export class Projenrc extends Component {
...
     for (const option of bootstrap.type.options) {
-      if (option.fqn) {
-        optionFqns[option.name] = toJavaFullTypeName(
-          jsiiManifest.types[option.fqn]
-        );
+      const fqnType = jsiiManifest.types[option.fqn];
+      if (option.fqn && fqnType) {
+        optionFqns[option.name] = toJavaFullTypeName(fqnType);
       }
     }
```
Chriscbr, line 157: Can we add unit tests for this?

Perhaps we could start by extracting these 10 lines to a helper function like `generateJavaOptionNames(options, jsiiManifest)` and test it in isolation in `test/java/projenrc.test.ts`. (or feel free to test it a different way)

Chriscbr (changes requested):

src/java/projenrc.ts
agdimech: Sure, just posted a new revision with the incorporated feedback Had a small issue in the last PR where the manifest looking was throwing if a fqn didn't exist in the option. This is also fixed in the latest revision.

Chriscbr (approved):

agdimech:
Apologies,  just fixed eslint errors and posted a new revision.

PR #1851, opened 2022-05-13
fix: resolve issue where generated import prefix is incorrect in .pro…

agdimech:
I am seeing an error whereby the generated .projenrc.py does not have the correct module_name. The FQN is aws-prototyping-sdk... however the exported python moduleName is aws_prototyping_sdk.

This PR ensures that the declared module_name takes precedence.

Repro steps:

```
mkdir foo && cd foo
npx projen new --from aws-prototyping-sdk pdk-pipeline-py
npx projen # this will throw the error
```

---

### DIFF
--- a/src/python/projenrc.ts
+++ b/src/python/projenrc.ts
@@ -119,9 +119,12 @@ export class Projenrc extends Component {
 }
 
 export function resolvePythonImportName(jsiiFqn: string, jsiiManifest: any) {
-  const moduleName = jsiiManifest.targets.python.module;
+  const moduleName = jsiiManifest?.targets?.python?.module;
+
   // Module name prefix should take precedence in the event moduleName !== fqn prefix
-  return [moduleName, ...jsiiFqn.split(".").slice(1)].join(".");
+  return moduleName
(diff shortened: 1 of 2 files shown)


src/python/projenrc.ts
```diff
@@ -98,7 +98,11 @@ export class Projenrc extends Component {
       bootstrap.args
     );
 
-    emit(toPythonImport(jsiiFqn));
+    const moduleName = jsiiManifest.targets.python.module;
+    // Module name prefix should take precedence in the event moduleName !== fqn prefix
+    const importName = [moduleName, ...jsiiFqn.split(".").slice(1)].join(".");
+    emit(toPythonImport(importName));
```
Chriscbr, line 104: Can we add a unit test for this?

Chriscbr (changes requested):

src/python/projenrc.ts
agdimech: Added unit test in latest revision.

Chriscbr (approved):

PR #1864, opened 2022-05-18
feat(eslint): support eslint config in yaml form

kellertk:
Adds a configuration option to the ESLint module to support writing the
configuration file as yml instead of json. Still writes as json by
default. ESLint natively supports yml configuration files and they're
nicer to work with. See
https://eslint.org/docs/user-guide/configuring/configuration-files

---

### DIFF
--- a/src/javascript/eslint.ts
+++ b/src/javascript/eslint.ts
@@ -3,6 +3,7 @@ import { PROJEN_RC } from "../common";
 import { Component } from "../component";
 import { NodeProject } from "../javascript";
 import { JsonFile } from "../json";
+import { YamlFile } from "../yaml";
 import { Prettier } from "./prettier";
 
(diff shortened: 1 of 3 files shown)


src/javascript/eslint.ts
```diff
@@ -75,6 +76,12 @@ export interface EslintOptions {
    * @default true
    */
   readonly tsAlwaysTryTypes?: boolean;
+
+  /**
+   * Write eslint configuration as YAML instead of JSON
+   * @default false
+   */
+  readonly yamlConfiguration?: boolean;
```
Chriscbr, line 84:
```suggestion
  readonly yaml?: boolean;
```

Chriscbr (changes requested):

PR #1883, opened 2022-05-25
feat!: refactor Project into StandardProject

### LINKED ISSUE #1761: Refactor opinionated parts of `Project` into `StandardProject`
Minimize the API surface area of `Project` and keep it to its "abstract" core. A project is just a collection of components that can be synthesized into a directory.

This may also help resolve some cyclic import issues issues discovered while working on #1671 where `Project` depends on other components and several other components depend back on `Project`.

Chriscbr:
Closes #1761

Remove most of all L3 constructs from the base "Project" type. This PR also updates tests to make use of the now-empty Project, and to use the `Testing.synth()` utility class.

Right now this is primarily an architectural change, so there are a few technical oddities:
- When you run `npx projen new standard` or `npx projen new project` it doesn't include any projenrc file of any kind. Fortunately you can still manually create `.projenrc.js` and synth and things work fine
- The empty project still includes `.projen/files.json`. This feels a bit inelegant, so maybe it can be revisited later
- I wasn't able to remove a couple of abstract methods from `Project` (`addPackageIgnore` and `addGitIgnore`) because doing so resulted in conflicts with `FileBase`

In a separate change I think we could remove GithubProject, and bring any functionality that's still important into StandardProject. So right now the inheritance chain looks like this:

```
Project <- StandardProject <- GithubProject <- NodeProject ( <- TypescriptProject ...)
```

but it could be simplified into

```
Project <- StandardProject <- NodeProject ( <- TypescriptProject ...)
```

BREAKING CHANGE: `Project` no longer comes bundled with most components such as `Gitignore`, `Tasks`, `Dependencies`, etc. and no longer includes properties or methods for modifying these. If you want to use those components, you should replace `Project` with `StandardProject` in projen scripts.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license


src/project.ts
```diff
@@ -197,64 +103,22 @@ export class Project {
...
+    this.logger = new Logger({
+      ...options.logging,
+      prefix: `[${this.name} ]`,
+    });
 
     if (!this.ejected) {
       new JsonFile(this, FILE_MANIFEST, {
         omitEmpty: true,
+        editGitattributes: false,
+        editGitignore: false,
```
Chriscbr, line 121: These is needed to prevent a runtime issue where during the creation of JsonFile, it tries adding its path to the project's "gitignore" but if you're a `Project`, you don't have a gitignore so an error is thrown

test/__snapshots__/project.test.ts.snap
```diff
@@ -1,37 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
-
-exports[`renovatebot: true creates renovatebot configuration 1`] = `
```
Chriscbr, line 3: (This test was moved to standard-project.test.ts)

src/github/github-project.ts
```diff
@@ -235,17 +236,4 @@ export class GitHubProject extends Project {
...
-   * github-linguist attributes. Generated files do not count against the
-   * repository statistics and language breakdown.
-   *
-   * @param glob the glob pattern to match (could be a file path).
-   *
-   * @see https://github.com/github/linguist/blob/master/docs/overrides.md
-   */
-  public annotateGenerated(glob: string): void {
-    this.gitattributes.addAttributes(glob, "linguist-generated");
-  }
```
Chriscbr, line 250: This method was moved to `StandardProject`, which GitHubProject extends, hence its removal here

test/vscode/dev-env.test.ts
```diff
@@ -34,82 +28,28 @@ describe("dev environment constructor", () => {
...
 
     // THEN
-    const outdir = synthSnapshot(project);
+    const outdir = Testing.synth(project);
     expect(outdir[GITPOD_FILE]).toMatchSnapshot();
   });
 });
 
-describe("dev environment enable/disable", () => {
-  test("given gitpod and devContainer are false", () => {
```
Chriscbr, line 43: Moved these tests into standard-project.test.ts

rix0rrr:
This is a bit of an aside, but I wonder why all these features need to come from inheritance anyway. The problem with inheritance is that it starts to get very inflexible at scale--which you are running into already, from the looks of the problem description of this PR.

For example:

> In a separate change I think we could remove GithubProject, and bring any functionality that's still important into StandardProject. 

Why isn't `GithubProject` decoupled from the class at all?

```ts
const project = new Project({ ... });

// Like this
new Github(project, { ... });

// Or like this
Github.of(project).addWorkflow({ ... });

// Or even like this ('github' is a namespace here)
new github.Workflow(project, { ... });
```

All of these allow decoupling Github functionality from any specific `Project` class.

Right now, if you want to use `TypescriptProject` you are tied to GitHub functionality. Why is that though? Isn't that a completely separate axis? By not tying it to inheritance but having different classes, you can allow BitBucket/CodeCommit/GitLab to be substituted for GitHub.

And the same holds for all other projen features.

src/project.ts
```diff
@@ -57,70 +49,18 @@ export interface ProjectOptions {
...
    */
   readonly logging?: LoggerOptions;
-
-  /**
-   * Generate (once) .projenrc.json (in JSON). Set to `false` in order to disable
-   * .projenrc.json generation.
-   *
-   * @default false
-   */
-  readonly projenrcJson?: boolean;
```
rix0rrr, line 67: Have you done analysis on the blast radius of this backwards incompatible change?

iirc projen auto-updates itself, so this is going to affect every single projen user. Will their builds be broken? Will they know what to do immediately? Will they groan and curse projen which is forcing them to do work they don't necessarily want to do?

User trust is something you can only burn once, so make sure you're doing it for something worthwhile.

Chriscbr:
Thanks for the review @rix0rrr . Yeah, I think you're super right to call out how big of a breaking change this is.

FWIW, I'm not sure if there's a safer way of achieving this. There's actually two major changes happening here:
1. carving out an "unopinionated" base class from the existing classes (I proposed this as reusing the name `Project`, but you're totally right to suggest it can/should be a new class like `ProjectBase` to minimize breakage)
2. change the other core classes (`Component` and `FileBase`) to only depend on the unopinionated base class

Here's an example to make (2) more concrete. Previously, if I had a custom projen component, then I could do this:

```ts
class Foo extends Component {
  constructor(project: JavaProject) {
    super(project);
    
    // ...
  }
  
  doSomething() {
    // `this.project` is typed `Project`
    this.project.addTask();
  }
}
```

...because I'm guaranteed that `this.project` is a `Project`, which has an `addTask` method. But this PR would change the contract so that `this.project` refers to `ProjectBase` (which doesn't have an `addTask` method), so I would have to change the logic to save the `JavaProject` in a private field, and then I can call `javaProject.addTask()`.

To address that part of the breaking change, I'd have to either retain the existing coupling between `Component` and the opinionated `Project`/`StandardProject`, and possibly add a parent class to `Component` to achieve that goal, like so:

```ts
export class ComponentBase {
  constructor(public readonly projectBase: ProjectBase) {
    project._addComponent(this);
  }

  public preSynthesize() {}
  public synthesize() {}
  public postSynthesize() {}
}

export class Component extends ComponentBase {
  constructor(public readonly project: Project) {
    project._addComponent(this);
  }
}
```

But even after doing this, there might still be a breaking change lying around somewhere because of this juggling between `Project`, `ProjectBase`, `Component`, and `ComponentBase`.

All that said, you've given me enough pause to question whether this moves the needle far enough or unblocks enough use cases to be worth justifying breaking our consumers, so maybe it makes sense to revisit this later

rix0rrr:
FWIW, if you add the new classes and keep the existing definitions of `Project` and `Component` unchanged, I don't see how this would introduce any breaking changes, so it seems safe to me.

mrgrain:
@Chriscbr I've drafted the PR until we've addressed these complex issues.

I tend to agree with @rix0rrr's last suggestion - creating new classes will be a more straight forward transition. Maybe this can and should be combined with classes that extend `Construct`? Because we will have the same issue with those.

Chriscbr:
@mrgrain Got it, yeah I'm totally cool with marking this as draft.

I'm still thinking about other ways of managing logic in projen to avoid this crazy inheritance hierarchy. As @rix0rrr alluded to, it introduces a lot of complexity. Proponents of the liskov substitution principle would say that if class B extends class A, then any property of A should also hold for B -- or phrased differently, B should behave the same if used somewhere you expect an A. But I'm not sure this holds in projen -- e.g. a TypeScriptAppProject extends TypeScriptProject, but TypeScriptAppProject actually takes away tasks related to packaging that you would have on a TypeScriptProject. So this suggests we should model these behaviors in a different way.

The trait/interface suggestion from Rico could be one avenue for this, probably some more experimentation is needed :)

Also in agreement that combining the change with a change to using `constructs` may be appropriate.

mrgrain:
@Chriscbr That's a very good point you raise re liskov. projen feels much more like composition rather than inheritance. Which again reminds me of CDK: an App (Stack) is composed of multiple Resources. With projen a Project is composed of multiple Patterns(?), Paradigms (?), Thingys. But then that difficulty lies within defining these Patterns independently enough. Take eslint as an example, it's strongly tied to the TS vs JS decision.

Or maybe some sort of Middleware approach could work for some stuff. Aspects would be the constructs equivalent (possibly with the exception that they can add files). I could see a generic Eslint Aspect/Middleware working nicely.

corymhall:
> Which again reminds me of CDK: an App (Stack) is composed of multiple Resources. With projen a Project is composed of multiple Patterns(?), Paradigms (?), Thingys. 

I actually really like this idea. I think we should model projen more like we do the CDK. Each of the components (e.g. eslint) would be an L2 which could be composed with other L2s (e.g. typescript) to make an L3 (e.g. TypeScriptApp).

> But then that difficulty lies within defining these Patterns independently enough. Take eslint as an example, it's strongly tied to the TS vs JS decision.

Couldn't we look at this the same way we do AWS resources? A lot of the L2s we have only mean anything when they are composed together with other L2s. If I create an ecs.TaskDefinition it doesn't get me very far without also creating an ecs.Service and potentially an elbv2.ApplicationLoadBalancer which I then need to know how to wire together. But then there is the ApplicationLoadBalancerFargateService which I could use that wires it together for me.

phitoduck:
I'm stoked to see the conversation headed towards composition over inheritance--and generally making `projen` more like the AWS CDK (constructs, traits/aspects, etc.)

I'm imagining a use-case of using different tools for generating docs--maybe `doxygen`, `docusaurus`, and `sphinx`. It's *possible* to generate JS/TS docs with `sphinx`, but it's rarely done. A low-level set of L1-L2 constructs could help you do low-level operations with `sphinx`. They'd give you full, unopinionated control for placing attributes into the various `sphinx` config files.

I imagine higher-level constructs would get more opinionated as in the case of `ApplicationLoadBalancerFargateService` as @corymhall said. 

Following this pattern of "higher level = more opinionated", you might have `SphinxForPython`, `SphinxForJS`, `SphinxForTS`, etc. I suppose you might have `ESLintForJS` and `ESLintForTS` (I know the `ForX` suffixes are ugly--it's just to communicate an idea).

To implement a construct like `SphinxForPython`, the `SphinxForPython` construct would probably need to know *something* about the internals of `pythonProject`, such as what folder the source code is in so that an "API Reference" could be generated (is it `src/`? is it `package_name`? Both are common in the python world). 

Again, as @corymhall said, this information could be provided by the user just as in AWS CDK! Here's some pseudocode of what I have in my head. Sorry for butchering the `projen`/`TypeScript` concepts here:

```typescript
export interface SphinxForPythonProps {
    sourceCodeDirsToRenderAPIReferenceSectionFor: string[]
}

class SphinxForPython {

    private readonly sourceCodeDirs: string[];

    constructor(private readonly project: Project, props: SphinxForPythonProps) {
        this.outdir = project.outdir
        this.srcCodeDirs = props.sourceCodeDirsToRenderAPIReferenceSectionFor
        project.addComponent(this)
    }

    public synthFiles() {
        // write sphinx files at `${this.outdir.path}/docs`
    }
}

const pythonProject: SomePythonProject

new SphinxForPython(project, {
    sourceCodeDirsToRenderAPIReferenceSectionFor: project.srcDir 
})

// ^^^ SphinxForPython didn't know where the sourceCodeDirsToRenderAPIReferenceSectionFor was, 
// but now it does!
// So SphinxForPython can now synth the files to the correct spot for this project
// even if the author of "SomePythonProject" never imagined that Sphinx
// might be used with their project
```

PR #1886, opened 2022-05-26
feat: warn users if CDK v1 deps are installed in a CDK v2 project

Chriscbr:
Proactively help users avoid errors by logging a warning if a CDK v1 dep has been added to the projenrc file of a CDK v2 project.

projen doesn't have a notion of "Annotations" like AWS CDK, so for now this warning is logged during the preSynthesize hook, at which point most dependencies should have been to the project's `Dependencies`.

---

### DIFF
--- a/src/awscdk/awscdk-deps.ts
+++ b/src/awscdk/awscdk-deps.ts
@@ -173,6 +173,20 @@ export abstract class AwsCdkDeps extends Component {
     this.addV1DevDependencies(...(options.cdkTestDependencies ?? []));
   }
 
+  public preSynthesize(): void {
+    const v1Deps = this.project.deps.all
+      .map((dep) => dep.name)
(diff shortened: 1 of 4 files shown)


src/awscdk/awscdk-deps.ts
```diff
@@ -343,6 +357,28 @@ export abstract class AwsCdkDeps extends Component {
   protected abstract packageNames(): AwsCdkPackageNames;
 }
 
+function isV1OnlyPackage(packageName: string): boolean {
```
kaizencc, line 360: can we do even better and check if the special packages are version 2.x or 1.x, and return true if they are 1.x? Not sure if we have that info at this point.

kaizencc (approved):
Great :). Just one nit that is probably not worth it, but I'll leave that up to you.

PR #1891, opened 2022-05-28
feat(github): customize GitHub Actions versions

### LINKED ISSUE #1702: lock down GitHub Actions to specific SHA
Per GH security docs: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

I think we should consider locking down all GH Actions that we reference by SHA instead of by tag.

e.g. Before:
```
uses: actions/checkout@v3
```

After:
```
uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # v3.0.0
```

Thoughts?

### LINKED ISSUE #1844: Manage GitHub actions versions
When workflows are created using projen (whether it's through high-level L3 components like `BuildWorkflow` or through manually specifying steps in a L2 `GithubWorkflow`), there are often many GitHub actions used. 

It's a best practice to upgrade these actions for the latest performance and security fixes, but currently this is difficult in projen. Either the version needs to be updated in the upstream component you are using, or you need to update the list of "steps" you're providing to a `GithubWorkflow` with the new version of the GitHub action.

Perhaps we can manage GitHub actions in our dependencies, in a way that lets you change the versions centrally from your projenrc file.

As an example, this kind of PR probably shouldn't be necessary / it should just be handled by projen's `UpgradeDependencies` component: https://github.com/cdklabs/construct-hub-webapp/pull/963

Related to https://github.com/projen/projen/issues/1702

Chriscbr:
Make it possible to fix/update GitHub actions with a single line of code.

This would close #1844 and close #1702

For example, adding

```ts
project.github.actions.addAction("actions/checkout@v10");
```

will update the version of "actions/checkout" used everywhere in GitHub workflows to "v10"

I tried this out, but I'm not in love with the amount of effort the API requires if you're authoring your own GitHub workflow (since now you have to use `actions.use("actions/checkout")`), so I might try seeing if there's another way to implement this...

---

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -16,6 +16,6 @@ jobs:
       pull-requests: write
     if: contains(github.event.pull_request.labels.*.name, 'auto-approve') && (github.event.pull_request.user.login == 'cdklabs-automation')
     steps:
-      - uses: hmarr/auto-approve-action@v2.1.0
+      - uses: hmarr/auto-approve-action@v2.2.1
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       CI: "true"
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v10
         with:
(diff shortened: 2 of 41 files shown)


mrgrain:
Resolved by #2439 and as far as I can see this PR does not anything else to it.

PR #1895, opened 2022-05-31
feat(github): do not create `schedule` if `cron` is empty

matt9ucci:
If `schedule.cron` is empty (e.g. `UpgradeDependenciesSchedule.NEVER`), `schedule: []` is created in upgrade-main.yml.

But it shows the following error in GitHub Actions page:

```
Invalid workflow file: .github/workflows/upgrade-main.yml#L1
`schedule` accepts a list of one or more maps with the `cron` key set
```

---


Chriscbr (approved):
Good catch @matt9ucci :)

PR #1906, opened 2022-06-03
feat(publisher): Support GitHub deploy keys for golang

kichik:
Using a deploy key instead of a personal access token provides more fine grained access control. If the secret is stolen somehow, the attacker will only have access to the go repository instead of full control of the user's account.
Requires https://github.com/cdklabs/publib/pull/436

---


kichik:
Testing procedure:

1. Install my version of projen locally
2. Create a new CDK construct project with projen
3. Run `yarn link projen` to get my local projen
4. Edit `.projenrc.js` to publish go with new settings
5. Add overrides to `release.yml` so it uses unreleased patch of publib
6. Disable self mutation so it doesn't accidentally upgrade to published projen
7. Manually remove `git diff` from release so projen version mismatch is ignored (overrides didn't work on `tasks.json`)
8. Push to https://github.com/kichik/projen-deploy-key-test
9. Check it still pushes to https://github.com/kichik/projen-deploy-key-test-go successfully

It looks like publib aims to do everything with one command and some environment variables, but I felt like ssh-agent belongs on the projen side because it's more of a system setting. Someone might want to use publib with their existing SSH settings on the system. Let me know if you disagree and I can try to get it working with `GIT_SSH` environment variable. It will require writing the file to disk which is less secure, but it should work too.

Chriscbr:
Looks good to me! I wasn't familiar with deploy keys so TIL 😄

I added one comment on the publib PR - we can merge this in once support is added in publib.

kichik:
Done. FWIW TIL of GitHub service accounts and ended up using that instead of deploy keys. I was going to try and add deploy key capability to replace `PROJEN_GITHUB_TOKEN`, but it requires more than pushing so a token is required. Creating a service account that has only the required permissions was simpler and seems to be what CDK does too.

Chriscbr (approved):

PR #1919, opened 2022-06-08
feat(cdk-config): add methods for adding includes and excludes

### LINKED ISSUE #1898: (AwsCdkTypeScriptApp): Add exclude/include methods for cdk.json
In building a typescript app that creates a CodeCommit repository using files from inside the repository; it would be helpful if to have `addExclude` and `addInclude` methods on the `cdkConfig` allowing me to add additional includes and excludes to `cdk.json`. This would be similar to the methods found on `tsconfig`.

If there's a better way to go about this please let me know. If not, I can find the time to put in a PR if the team is open to reviewing it.

ryansonshine:
This change adds `addExcludes` and `addIncludes` to the `CdkConfig` allowing users to customize the includes and excludes values in `cdk.json`.

Closes #1898

---


src/awscdk/cdk-config.ts
```diff
@@ -85,10 +85,22 @@ export class CdkConfig extends Component {
...
 
+  /**
+   * List of glob patterns to be included by CDK.
+   */
+  public readonly include: string[];
+
+  /**
+   * List of glob patterns to be excluded by CDK.
+   */
+  public readonly exclude: string[];
```
Chriscbr, line 96: Please make these fields private. We don't want these collections to be mutated outside of the `addXxx` methods. (`readonly` only enforces that you can't reassign the variable to a new array)

src/awscdk/cdk-config.ts
```diff
@@ -107,15 +119,37 @@ export class CdkConfig extends Component {
...
+  /**
+   * Add includes to `cdk.json`.
+   * @param patterns The includes to add.
+   */
+  public addIncludes(patterns: string | string[]) {
+    if (!Array.isArray(patterns)) {
+      patterns = [patterns];
+    }
+    this.include.push(...patterns);
+  }
```
Chriscbr, line 141: Let's make this `string[]` instead:
```suggestion
  public addIncludes(...patterns: string[]) {
    this.include.push(...patterns);
  }
```

We use a tool named jsii to translate projen's API into other languages like Java - and since they cannot express type unions as easily, other users will just see a method that accepts "Object" instead of "string" or "Array\<string>".

src/awscdk/cdk-config.ts
```diff
@@ -107,15 +119,37 @@ export class CdkConfig extends Component {
...
+  /**
+   * Add excludes to `cdk.json`.
+   * @param patterns The excludes to add.
+   */
+  public addExcludes(patterns: string | string[]) {
+    if (!Array.isArray(patterns)) {
+      patterns = [patterns];
+    }
+    this.exclude.push(...patterns);
+  }
```
Chriscbr, line 152: See comment above
```suggestion
  public addExcludes(...patterns: string[]) {
    this.exclude.push(...patterns);
  }
```

Chriscbr (changes requested):

src/awscdk/cdk-config.ts
ryansonshine: Ah interesting, is there a way to make jsii throw an error or warning when union types are used?

test/awscdk/cdk-config.test.ts
```diff
@@ -18,3 +18,85 @@ describe("context values", () => {
...
+describe("includes", () => {
+  test("should retain their initialized value", () => {
+    const defaultIncludes = ["src/**/*.ts", "test/**/*.ts"];
+    const config = new CdkConfig(new TestProject(), {
+      app: "test includes",
+      watchIncludes: defaultIncludes,
+    });
+
+    // @ts-ignore test values are being set properly
+    expect(config.include).toEqual(defaultIncludes);
```
Chriscbr, line 31: Can we avoid using typescript error suppressions? Either of these options would work fine:

- We can expose `include` and `exclude` "safely" by creating a getter which returns a copy of the array. e.g.:

```ts
public get includes(): string[] {
  return [...this._includes];
}
```
- We can make an assertion on the snapshot instead:
```ts
const config = Testing.synth(project)["cdk.json"];
expect(snapshot["cdk.json"].watch.include).toEqual(...);
```

Chriscbr (changes requested):

src/awscdk/cdk-config.ts
Chriscbr: That's a good question - I'm not sure off the top of my head, but I'll take a look into it :)

Chriscbr (approved):

PR #1920, opened 2022-06-08
feat(ObjectFile): `addToArray` escape hatch method

tompahoward:
The `addToArray` method allows values to be added to existing arrays within the object.

Unlike the `addOverride` method, `addToArray` works against `this.obj` rather than `this.rawOverrides`. It needs to do this in order to add to the array. Other suggestion approaches are welcome.

I don't really like the name `addToArray`, but can't think of anything better. Suggestions are welcome.

Any other feedback is also welcome.

---


Chriscbr (commented):
Hmm - yeah I was thinking for a while about if there's some way to allow this to work with the existing `rawOverrides` field, which is used to make overrides compatible with lazy values.

I made a PR to your fork with a suggested way to do this. It's a bit hacky but I think that might be OK.  https://github.com/tompahoward/projen/pull/1 Curious to hear your thoughts!

Another option I was thinking about is whether we should use a different escape hatch mechanism, like the `JsonPatch` model used in CDK for Kubernetes, to support this kind of use case. See this example of how it works: https://github.com/cdk8s-team/cdk8s-core/blob/2.x/test/api-object.test.ts#L322-L361

Chriscbr (changes requested):
Looks good to me! Can you just add a brief note about this method in `docs/escape-hatches.md` showing how it's used?

tompahoward:
`JsonPatch` looks like a much cleaner solution. Would you like me to update the PR, adding `addJsonPatch()` and removing `addToArray()`?

At the same time, do you want me to deprecate `addOverride()` and `addDeletionOverride()` or just leave them as is?

Chriscbr:
@tompahoward Let's stick to this solution for now (and leave addOverride / addDeletionOverride as is). If there are more use cases for JsonPatch we can still add it in the future 👍

tompahoward:
@Chriscbr I think we're good now. Let me know if there's anything else needed.

Chriscbr (approved):

PR #1940, opened 2022-06-18
feat(gitlab): job sequences for reusable pipeline fragments

ThomasSteinbach:
Hi,

I am contributing here as the author of the author of the [Gitlab CI Python library (gcip)](https://gitlab.com/dbsystel/gitlab-ci-python-library) I also have shown in short on my last presentation on the AWS Summit Berlin 2022.

There I also heard (not for the first time) of projen and talked with different people like @hoegertn (contributor of projen) about projen and that you possibly could solve the same problems than the gcip with projen.

Like (one part) of projen with the gcip you could describe your Gitlab CI pipelines in code. But the gcip is more as just a representation of the Gitlab CI reference in code. It offers you some really powerful concepts of how to compose pipelines of 'containers of jobs'. Those 'containers' are called Sequences and you can use them to

* bundle and share jobs doing a common task
* apply common configuration on the sequence to modify a bunch of jobs
* add sequences to sequences for maximum reusability of pipeline fragments

However projen solves much more problems apart from Gitlab pipelines, which makes me really want to use it. When switching to projen I would also like to define my Gitlab pipelines like I have it done before with the gcip.

With this PR I am adding one of the biggest features of the gcip to projen: Sequences

I wouldn't like to repeat myself in describing what Sequences are and how they work. Especially because I have already extended the projen documentation with a chapter about a sequence quick start. So please refer to the [documentation change right in this PR](https://github.com/projen/projen/pull/1940/files#diff-86081f6b7a8a556349e10a2ce7de6cbebb4c99beec8c728f55c1819780bd4397) to get a first view on sequences in projen.

Please also consider looking at the unit tests for understanding how sequences work.

This is the first version of sequences. I have a couple of more features I'd like to add, like making it easier to connect multiple jobs/sequences with Gitlab `needs` relations. The gcip does that already, however the implementation would look a little bit different in projen and this was out of scope of my first PR. This PR is already relatively huge, so I hold some of my ideas back. I hope this PR gets into the main branch, thus I (and my colleagues currently working on the gcip) could make the Gitlab part of projen even cooler.

Last thing: This is also my first greater work in TypeScript. I am programming a lot in other languages, so I think I am in general 'experienced', albeit new to TS. So I am happy to every suggestion on how to make to code better.

___

@Chriscbr - I had 'pinged' you, because you are one of the active maintainers and this PR was a while open as "draft" without attention. Now this PR is ready to be reviewed.

___


src/gitlab/configuration.ts
```diff
@@ -126,7 +127,7 @@ export class CiConfiguration extends Component {
    * Groups jobs into stages. All jobs in one stage must complete before next stage is
    * executed. Defaults to ['build', 'test', 'deploy'].
    */
-  public readonly stages: string[] = [];
+  private readonly _stages: string[] = [];
```
mrgrain, line 130: Any reason this public property should be private now?
If so, could there be a way to keep the the current access and deprecate it in favour of the new API?

src/gitlab/configuration.ts
```diff
@@ -140,7 +141,7 @@ export class CiConfiguration extends Component {
   /**
    * The jobs in the CI configuration.
    */
-  public readonly jobs: Record<string, Job> = {};
+  public readonly jobs: Sequence = new Sequence();
```
mrgrain, line 144: Is there a way keep the the current interface, deprecate it and offer this feature as a new API?

src/gitlab/sequence.ts
```diff
@@ -0,0 +1,489 @@
+import { AllowFailure, Artifacts, Cache, CacheKeyFiles, Environment, Filter, Image, IncludeRule, Inherit, Job, JobWhen, Need, Parallel, Release, Retry, Secret, Service, Trigger } from "./configuration-model";
```
mrgrain, line 1: This feel looks like it's generated from a source. If so, would it be feasible to add the generation process in here, instead of committing the result? (And having to manually keep it up to date.)

mrgrain (changes requested):

pflorek:
@ThomasSteinbach I really like the idea to group jobs and apply modification on them ❤️ 

There is one important point. This GitLab module has in mind to be open to changes. Therefore the new feature should not be closed. A simple function like `Sequence.tryFindJob(name: string)` or just return the map `Sequence.jobs` would be sufficient.

This comes really handy as a replacement of the child pipelines or addition. For instance I have a child pipeline on schedule and another one on merge request. But the DAG doesn't display that setup. With this feature I can have the same experience and the benefits of keyword `needs`.

Initially I didn't understand the term `Sequence`. For me it's just a group of jobs. Maybe we can add some functional stuff like `Sequence.map`

robert-affinidi:
@ThomasSteinbach what is the advantage of sequences over GitLab own extension (via extends) mechanism and yaml anchoring? (Except adding abstraction to it)

pflorek:
> @ThomasSteinbach what is the advantage of sequences over GitLab own extension (via extends) mechanism and yaml anchoring? (Except adding abstraction to it)

@robert-affinidi 

One can consider it as a monad. The proposed sequence contains a set of jobs. The sequence offers modifiers that apply given values to the contained Set of jobs. For example add tag `shared` to all jobs in the sequence. It's a nice shortcut.

mrgrain:
stale.

PR #1950, opened 2022-06-24
fix(publisher): add manual publish condition

gradybarrett:
Adding a condition to the git publish command, which performs
the same check as the bump command. If the last commit was a release
commit, it doesn't make sense to try to publish again. Doing so
actually causes an error because bump is skipped which means there
is no version file to read.

---


yuth (approved):

yuth:
Thank you for your contribution 🙏

PR #1951, opened 2022-06-24
chore: projen new command in README.md

tmokmss:
We need to run `npx projen new <type>` instead of `npx projen <type>` to initialize a project.

Sorry, I'm not sure what is an appropriate title for this PR so it would be great if you could fix it. 🙏 

---


yuth:
> Sorry, I'm not sure what is an appropriate title for this PR so it would be great if you could fix it. 🙏

There was space as first char of the title. Thank you for your contribution

PR #1957, opened 2022-06-27
feat(github): support calling reusable workflows in github actions workflows

### LINKED ISSUE #1956: github: Support calling reusable workflows
Some fields like uses, with, secrets are missing from workflow.Job interface.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses

niraj8:
Add uses, with, secrets fields to be used in Github workflow Job, these fields are used to call a reusable workflow.

May resolve #1956 

---


src/github/workflows-model.ts
```diff
@@ -10,7 +10,7 @@ export interface Job {
    *
    * @example ["ubuntu-latest"]
    */
-  readonly runsOn: string[];
+  readonly runsOn?: string[];
```
yuth, line 13: It might be worth creating 2 interfaces. One for the regular job and one for reusable workflow job so compiler can warn for missing properties in the jib definition

niraj8: I tried added another interface called `JobCallingReusableWorkflow`, please review again.
Thanks!

yuth (approved):

PR #1970, opened 2022-07-06
feat: add smithy build support to projen

mayurm88:
For folks using [smithy](https://awslabs.github.io/smithy/index.html) in their project, it would be great to support generation of `smithy-build.json` file using projen.

The types defined in this PR could be refined further for type safety but this PR is a starting point for those changes. [This](https://awslabs.github.io/smithy/1.0/guides/building-models/build-config.html#projections) was used as a reference for defining types in smithy-build.ts file.

Shoutout to @eduardomourar for collaborating with me on this!

---


src/smithy-build.ts
```diff
@@ -0,0 +1,93 @@
+import { Component } from "./component";
```
Chriscbr, line 1: Can we move this file into a new submodule (`src/smithy/smithy-build.ts`)?

src/smithy-build.ts
```diff
@@ -0,0 +1,93 @@
...
+export type PluginsType = Record<PluginKey, PluginValue>;
+
+export interface SmithyBuildOptions extends SmithyCommon {
+  readonly version?: string;
+
+  readonly outputDirectory?: string;
+
+  readonly projections?: ProjectionsType;
+
+  readonly ignoreMissingPlugins?: boolean;
```
Chriscbr, line 33: Typically we include a docstring for each option, including a description of what the default value or behavior is, like `@default "1.2.3"` or `@default - no projections`. See https://github.com/projen/projen/blob/main/src/gitlab/configuration.ts for an example of how it's formatted. (You can also include a link to any relevant smithy docs if you think it would help someone reading the docs).

src/smithy-build.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export class SmithyBuild extends Component {
```
Chriscbr, line 36: We should include a docstring for the component and any public properties/methods.

src/smithy-build.ts
```diff
@@ -0,0 +1,93 @@
...
+  public readonly outputDirectory?: string;
+
+  public imports?: string[];
+
+  public readonly ignoreMissingPlugins?: boolean;
+
+  public readonly manifest: any;
+
+  public projections?: ProjectionsType;
+  public plugins?: PluginsType;
```
Chriscbr, line 48: For properties that are mutable like `imports`, `plugins`, and `projections`, we usually try to avoid publicly exposing the underlying data structures to prevent bugs where different constructs try to mutate the objects directly and it becomes hard to refactor code. See the design guidelines on this page for more info (it's for a totally different repo, but we try to also follow the spirit of it in projen): https://github.com/cdk8s-team/cdk8s-plus/blob/0babfd9aa72d96f1d364bbc28a9aa516729a5775/CONTRIBUTING.md#mutability Instead, it's preferred to include getter methods that return a copy of the data structure -- see the link for an example.

`manifest` can be private altogether (it does not need a getter method) since the user can just view the individual fields `version`, `outputDirectory, etc).

For example:
```ts
public get projections() {
  return { ...this._projections };
}
```

Chriscbr (commented):
This looks cool, it would be great to have smithy support in projen! :) Thanks as well for adding several test cases.

src/smithy-build.ts
mayurm88: Thank you for guiding me on this. Added docstring for each of these options including the default behaviour.

src/smithy-build.ts
mayurm88: done

src/smithy-build.ts
mayurm88: Makes complete sense. Added these getter methods and made `imports`, `plugins` and `projections` private. Also made `manifest` private.

src/smithy/smithy-build.ts
```diff
@@ -0,0 +1,158 @@
...
+ * Options for `SmithyBuild`
+ */
+export interface SmithyBuildOptions extends SmithyCommon {
+  readonly version?: string;
+
+  readonly outputDirectory?: string;
+
+  readonly projections?: ProjectionsType;
+
+  readonly ignoreMissingPlugins?: boolean;
```
Chriscbr, line 36: These options should also probably have docstrings (they can be the same as the ones defined in the class below)

Chriscbr (approved):

mrgrain (changes requested):
@mayurm88 Let's add these extra jsdocs and this is good to go!

src/smithy/smithy-build.ts
mayurm88: Thanks and apologies for not adding these earlier.

mrgrain (approved):
🚢

src/smithy/smithy-build.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Options for `SmithyBuild`
+ */
+export interface SmithyBuildOptions extends SmithyCommon {
+  /**
+   * Defines the version of smithy-build.
+   * @default "1.0"
+   */
+  readonly version?: string;
+
```
mrgrain, line 35:
```suggestion

```

src/smithy/smithy-build.ts
```diff
@@ -0,0 +1,178 @@
...
+   * @default "1.0"
+   */
+  readonly version?: string;
+  
+  /**
+   * Specifies a location where smithy projections are written.
+   * @default - no output directory
+   */
+  readonly outputDirectory?: string;
+
```
mrgrain, line 41:
```suggestion

```

src/smithy/smithy-build.ts
```diff
@@ -0,0 +1,178 @@
...
+   */
+  readonly outputDirectory?: string;
+  
+  /**
+   * Map of projections name to projection configurations
+   * https://awslabs.github.io/smithy/1.0/guides/building-models/build-config.html#projections
+   * @default - no projections
+   */
+  readonly projections?: ProjectionsType;
+
```
mrgrain, line 48:
```suggestion

```

mrgrain (changes requested):

mrgrain (approved):

src/smithy/smithy-build.ts
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Options for `SmithyBuild`
+ */
+export interface SmithyBuildOptions extends SmithyCommon {
+  /**
+   * Defines the version of smithy-build.
+   * @default "1.0"
+   */
+  readonly version?: string;
+
```
mrgrain, line 35:
```suggestion

```

PR #1982, opened 2022-07-19
feat(awscdk): custom Lambda runtimes

jogold:
Allow to use custom Lambda runtimes for bundled functions.

Remove the "coupling" with the enum from `aws-cdk-lib.aws_lambda`
so the runtime can be updated to a newer runtime without having to
update the minimum version of the AWS CDK to depend on.

---


mrgrain (approved):
Nice one!

PR #1994, opened 2022-07-25
feat(object-file): added ability to apply JSON patches to object files

tompahoward:
---


test/object-file.test.ts
```diff
@@ -306,3 +307,83 @@ describe("addToArray", () => {
     });
   });
 });
+
+describe("addJsonPatch", () => {
+  test("addJsonPatch(p, v) adds to an existing array", () => {
```
Chriscbr, line 312: I think the test names need to be updated based on the new API names

Chriscbr (commented):
I like the idea of shortening the name from `addJsonPatch()` to `patch()` 🙂 

Looks great, just found one nitpick

mrgrain (changes requested):
@tompahoward Looking forward to this if you'll find the time to get it over the line =)

test/object-file.test.ts
tompahoward: done... finally 🎉

mrgrain (approved):
Looking good now! Thanks for fixing up the final bits 😃 

:shipit:

PR #1995, opened 2022-07-25
feat: adding steps for upgrading project to .ts file

mbonig:
I added a section which details the steps needed to upgrade a project previously created with a .projenrc.js file to the new .projenrc.ts file.

---


mrgrain (approved):
Thanks for that! I was definitely in that situation before 😅

PR #2000, opened 2022-07-28
feat(release): add option to set the minimal major version

DanielMSchmidt:
This new option is really convenient when you have a generated project that is self updating like the cdktf pre-built providers. With the current setup it's impossible to increase the major version without manual intervention. Since breaking changes in a 0.x version only result in a minor version increase there is no way to get to a higher major version.

---

### DIFF
--- a/src/release/release.ts
+++ b/src/release/release.ts
@@ -79,6 +79,19 @@ export interface ReleaseProjectOptions {
   readonly majorVersion?: number;
 
   /**
+   * Minimal Major version to release
+   *
+   *
+   * This can be useful to set to 1, as breaking changes before the 1.x major
+   * release are not incrementing the major version number.
+   *
+   * Can not be set together with `majorVersion`.
+   *
+   * @default - start releasing with major version 0
+   */
+  readonly minMajorVersion?: number;
+
+  /**
    * Bump versions from the default branch as pre-releases (e.g. "beta",
    * "alpha", "pre").
    *
@@ -345,6 +358,7 @@ export class Release extends Component {
     this.defaultBranch = this._addBranch(options.branch, {
       prerelease: options.prerelease,
       majorVersion: options.majorVersion,
+      minMajorVersion: options.minMajorVersion,
       workflowName: options.releaseWorkflowName ?? "release",
       tagPrefix: options.releaseTagPrefix,
       npmDistTag: options.npmDistTag,
@@ -613,6 +627,11 @@ export interface BranchOptions {
    * The major versions released from this branch.
    */
   readonly majorVersion: number;
+
+  /**
+   * The minimum major version to release.
+   */
+  readonly minMajorVersion?: number;
 
   /**
    * Bump the version as a pre-release tag.
--- a/src/release/release.ts
+++ b/src/release/release.ts
@@ -87,7 +87,7 @@ export interface ReleaseProjectOptions {
    *
    * Can not be set together with `majorVersion`.
    *
-   * @default - start releasing with major version 0
+   * @default - No minimum version is being enforced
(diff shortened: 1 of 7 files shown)


mrgrain:
I like the idea of this. Could you confirm/clarify the following for me:

>With the current setup it's impossible to increase the major version without manual intervention. Since breaking changes in a 0.x version only result in a minor version increase there is no way to get to a higher major version.

To my understanding, if a project starts put at 0.x.x (which is the default), there still won't be a way to get to a  higher major version. But instead of setting `majorVersion`, users can now choose  to set `minMajorVersion: 1` to upgrade?


How do major version releases currently behave if there's a breaking change and `majorVersion` is set?

DanielMSchmidt:
To clear this up a I'll start with the second question :) 

> How do major version releases currently behave if there's a breaking change and majorVersion is set?

If a major version is set and a breaking change occurs the build errors.

> To my understanding, if a project starts put at 0.x.x (which is the default), there still won't be a way to get to a higher major version. But instead of setting majorVersion, users can now choose to set minMajorVersion: 1 to upgrade?

Yeah exactly. Without this and without setting a major version (which you can't set in general due to builds erroring if it's there and they want to release beyond it) each major release would according to the semver spec (and also happening in reality) only update the minor version. There is no way to update the major version when starting from the major version 0. This is why I think this attribute would help, by starting with 1 you could ensure breaking changes get a major version increase. and it's distinclty different from just majorVersion since that is aimed for version specific release branches :)

DanielMSchmidt:
I don't understand the failing test though, any ideas what I could be missing?

DanielMSchmidt:
@mrgrain I found the issue, took the wrong way to solve it :D Now the detection is untampered and we instruct standard version directly where we should do it

src/release/release.ts
```diff
@@ -78,6 +78,19 @@ export interface ReleaseProjectOptions {
    */
   readonly majorVersion?: number;
 
+  /**
+   * Minimal Major version to release
+   *
```
mrgrain, line 83:
```suggestion
```

test/release/bump.test.ts
```diff
@@ -192,6 +192,37 @@ test("customization to versionrc reflects to changelog", async () => {
...
+    options: { minMajorVersion: 1 },
+    commits: [{ message: "v0", tag: "v0.1.2" }, { message: "commit2" }],
+  });
+  expect(result.version).toStrictEqual("1.0.0");
+  expect(result.changelog.includes("## [1.0.0]")).toBeTruthy();
+  expect(result.bumpfile).toStrictEqual("1.0.0");
+  expect(result.tag).toStrictEqual("v1.0.0");
+});
+
+test("minMajorVersion keeps release version if version is equal", async () => {
```
mrgrain, line 206: Could you please add an additional test case with current version greater than `minMajorVersion`

src/release/release.ts
```diff
@@ -78,6 +78,19 @@ export interface ReleaseProjectOptions {
...
+   *
+   * This can be useful to set to 1, as breaking changes before the 1.x major
+   * release are not incrementing the major version number.
+   *
+   * Can not be set together with `majorVersion`.
+   *
+   * @default - start releasing with major version 0
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah it is, if you start a blank project then it starts with 0, but this is much more precise and clearer 👍

test/release/bump.test.ts
DanielMSchmidt: Yes, will do 👍

mrgrain (approved):
🚀

PR #2004, opened 2022-07-30
feat: add support for `cp` command on Windows using shx

kichik:
---

### DIFF
--- a/src/task-runtime.ts
+++ b/src/task-runtime.ts
@@ -152,7 +152,7 @@ class RunTask {
         const cmd = exec.split(" ")[0];
         if (
           platform() == "win32" &&
-          ["mkdir", "mv", "rm", "cp", "cd"].includes(cmd)
+          ["mkdir", "mv", "rm", "cp"].includes(cmd)
         ) {
           command = `shx ${exec}`;
         } else {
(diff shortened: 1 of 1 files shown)


kichik:
I'm [using `cp`](https://github.com/CloudSnorkel/cdk-github-runners/blob/b816ef0b81f0fe802bced0049790151eb461d42f/.projenrc.js#L119-L120) and it would be nice to have it automatically use `shx`.

src/task-runtime.ts
```diff
@@ -150,7 +150,10 @@ class RunTask {
         let command = "";
         let hasError = false;
         const cmd = exec.split(" ")[0];
-        if (platform() == "win32" && ["mkdir", "mv", "rm"].includes(cmd)) {
+        if (
+          platform() == "win32" &&
+          ["mkdir", "mv", "rm", "cp", "cd"].includes(cmd)
```
mrgrain, line 155: This is great addition, but I don't think `cd` is supported by `shx`. [1]

```suggestion
          ["mkdir", "mv", "rm", "cp"].includes(cmd)
```

[1] https://github.com/shelljs/shx#unsupported-commands

mrgrain (changes requested):

src/task-runtime.ts
kichik: Thanks. I misread the documentation.

mrgrain (approved):

PR #2008, opened 2022-08-02
fix(awscdk): fix pytest import path

patrickdean:
The python app file being generated by awscdk-app-py is main.py but the test_example.py is trying to import from my_stack, causing the first build to fail. 

---


mrgrain (approved):
Nice, thank you!

PR #2010, opened 2022-08-03
feat(github): Allow more conventional commit types in PullRequestLint type defaults

suhasgaddam-trueaccord:
Default types were taken from https://github.com/conventional-changelog/conventional-changelog/blob/bfe3bf1c49d4a125474b398b2d304749fd3b56c7/packages/conventional-changelog-conventionalcommits/writer-opts.js#L179-L192

---

---

Feel free to reject this PR if `projen` does not want all the defaults in standard-version/conventional-changelog in the repository.

This can always be set in the settings.


mrgrain:
Hi @suhasgaddam-trueaccord and thanks for your contribution. Like you've mentioned I'm inclined to keep the setting as is. The original reasoning was:

> Since the default is that only fix and feat are included in the change log, I rather we only allow chore as an additional type which basically indicates (not included in the change log).
>
> So I think the default should be [ 'feat', 'fix', 'chore' ].
[Source](https://github.com/projen/projen/pull/1224#discussion_r748806606)

PR #2026, opened 2022-08-10
feat(upgrade-dependencies): excluded npm-check-updates from list of dependencies

AminFazlMondo:
This package does not need to have `npm-check-updates` as a devDeps. as it can use npx whenever it needs to check for updated packages.
Benefits:

1. One less package to be installed all the time
2. One less package to update (the first step is not required anymore)
3. There is also an issue in npm 8 in which `npm install` step throws `ENOTEMPTY` error as it is trying to rename packages that were installed prior to uprade. This is due to the fact that it is not a clean install, and it tries to rename files.

---


mrgrain:
Hi @AminFazlMondo and thanks for your contribution!

I'm not strictly against this change, but would you please go into more details about the motivation behind this?

> There is also an issue in npm 8 in which npm install step throws ENOTEMPTY error as it is trying to rename packages that were installed prior to uprade. This is due to the fact that it is not a clean install, and it tries to rename files.

This appears to be the main driver for this change. Can you elaborate more? We should address the underlying issue here.

```
// update npm-check-updates before everything else, in case there is a bug
// in it or one of its dependencies. This will make upgrade workflows
// slightly more stable and resilient to upstream changes.
```
This comment and related code gets removed. How is it not relevant anymore? What is this PR doing to address the challenges mentioned in the comment?

Also, we are moving from a constraint and locked version `npm-check-updates@^15` to a complete unconstrained one. That's definitely something we cannot do.

mrgrain (changes requested):
see above

AminFazlMondo:
Thanks @mrgrain  for quick look at this.
the workflow of upgrade is:

1. npm ci
2. <some upgrades using npm-check-updates>
3. npm i

What is happening here is that if ther is a package already installed as a dependency of another dependency and there are some upgrades to the root dependency. NPM tries to rename the folder in node_modules and install the package.

Below is an example of this;
```
👾 upgrade | npm install
npm ERR! code ENOTEMPTY
npm ERR! syscall rename
npm ERR! path <PathToMyPackage>/node_modules/projen/node_modules/@iarna/toml/lib
npm ERR! dest <PathToMyPackage>/node_modules/.projen-Awm7zVpt/node_modules/@iarna/toml/lib
npm ERR! errno -39
npm ERR! ENOTEMPTY: directory not empty, rename '<PathToMyPackage>/node_modules/projen/node_modules/@iarna/toml/lib' -> '<PathToMyPackage>/node_modules/.projen-Awm7zVpt/node_modules/@iarna/toml/lib'
```

Please note that the same repo had no problem, it only started after I upgraded to node 16 and NPM 8.

mrgrain:
That seems odd @AminFazlMondo I haven't came across this in any of my repos nor am I seeing reports of this. Would you be able to split this out into a minimal reproducible repo?

The whole thing _feels_ like an issue with `npm` to. Because the flow `npm ci --> upgrade packages --> npm i` should be well supported by it.

AminFazlMondo:
I agree, this is not a problem with projen, it is an issue with npm and how it handles things. give me some time and I will shoot through a public repo that is reproduces this issue for you to have a look

AminFazlMondo:
Hi @mrgrain ,
I have tried to reproduce this issue in a public repo.
I originally had only public packages in this repo and the upgrade task was working fine. As soon as, I introduced a private package that is hosted in AWS code artifact in [this commit](https://github.com/AminFazlMondo/TestPublicProjenRepo/commit/90f10d689df3734b07d763bbd1c0bd03f38d072a), I could reproduce this issue.
Unfortunately, this package is not available publicly for you to test, but I think replacing that with a private package that you have access to, will have the same effect.

Thanks,

mrgrain:
@AminFazlMondo Thanks for the reproduction! 👍🏻 
The private registry would explain how it's not a wider issue yet. 

Can you confirm how you are handling the config for your private CodeArtifact registry?

- Do you proxy npm through CodeArtifact? I.e. all requests go through AWS
- Have you setup scoped access for it? I.e. all packages in the scope `@your-scope` use CodeArtifact, all others use `npmjs.com`.
- Something else?

Just want to make sure I understand exactly what's going on.

AminFazlMondo:
I have setup scoped packages to go through code artifacts, and the rest of the packages through npm.
This is setup via scoped packages options, in projen. I have not checked in those details as they were sensitive info.

AminFazlMondo:
My earlier comment was result of my investigation, I was not able to test that. But I have followed these steps by running them manually and it worked.

mrgrain:
> My earlier comment was result of my investigation, I was not able to test that. But I have followed these steps by running them manually and it worked.

@AminFazlMondo are you saying that you cannot reproduce the issue now? If so, is that in general or just not with the reproduction repo?

AminFazlMondo:
@mrgrain ,
I probably should have clarified myself. I definately been able to replicate this in the test repo. What I meant was I could not test thhe changes in this PR in that repo to see if they solve the problem, what I did do was to run each of the steps command one by one to see if it works and it did.
I think the trick is not to do npm ci prior to upgrade, which by not relying on the `npm-check-updates` as a dependency, might be achieved.

mrgrain:
Cleaning this up. Aside from the reproduction challenge, this would need to address the requested changes.

PR #2029, opened 2022-08-12
fix: build breaks after update of @types/semver

mrgrain:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61586 introduces stricter types for the `semver` package, which in turn caused https://github.com/projen/projen/pull/2028 to fail the build when attempting to update `@types/semver` to the latest version.

This fix removes the parameter that is not compatible with the new types. The parameter is unused by the respective code in `semver` and we are not exposing the returned semver object, so this is a safe change.

---

### DIFF
--- a/src/util/semver.ts
+++ b/src/util/semver.ts
@@ -90,7 +90,7 @@ export function toReleaseVersion(
   assemblyVersion: string,
   target: TargetName
 ): string {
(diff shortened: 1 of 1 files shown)


mrgrain:
Pushed to the upgrade PR instead

PR #2047, opened 2022-08-19
feat(vscode): support VS Code workspace settings

mrgrain:
Settings keys languages are not further typed as there are A LOT of
them and plugins can add any new values to it.

Setting values are not typed as they are essentially of type JSON which is
difficult to express in TS and a jsii compatible way.

---

### DIFF
--- a/src/vscode/settings.ts
+++ b/src/vscode/settings.ts
@@ -1,0 +1,60 @@
+import { Component } from "../component";
+import { JsonFile } from "../json";
+import { VsCode } from "./vscode";
+
+/**
+ * VS Code Workspace settings
+ * Source: https://code.visualstudio.com/docs/getstarted/settings#_workspace-settings
+ */
+export class VsCodeSettings extends Component {
+  private readonly content: any;
+
+  constructor(vscode: VsCode) {
+    super(vscode.project);
+
+    this.content = {};
+
+    new JsonFile(vscode.project, ".vscode/settings.json", {
+      omitEmpty: false,
+      obj: this.content,
+    });
+  }
+
+  /**
+   * Adds a workspace setting
+   *
+   * @param setting The setting ID
+   * @param value The value of the setting
+   * @param language Scope the setting to a specific language
+   */
+  public addSetting(setting: string, value: unknown, language?: string) {
+    if (language) {
+      this.content[`[${language}]`] = this.content[`[${language}]`] ?? {};
+      this.content[`[${language}]`][setting] = value;
+    } else {
+      this.content[setting] = value;
+    }
+  }
+
+  /**
+   * Adds a workspace setting
+   *
+   * @param settings Array structure: [setting: string, value: any, languages?: string[]]
+   */
+  public addSettings(
+    settings: Record<string, unknown>,
+    languages?: string | string[]
+  ) {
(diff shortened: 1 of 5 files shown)


test/vscode/settings.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    ],
+    [
+      ["javascript", "typescript"],
+      [
+        ["editor.defaultFormatter", "esbenp.prettier-vscode"],
+        ["eslint.format.enable", true],
+      ],
+    ],
+    [["python"], [["python.formatting.provider", "black"]]],
+  ]);
```
mrgrain, line 84: I'm looking for feedback on this syntax. It's some what inspired by eslint config syntax, but doesn't quite have the convenience of it yet.

The first block L65-75 seems okay to me.

The second block L76-84 is intended to make it easy to create any permutation of rules and languages, i.e. to add N rules for N languages in one "statement".

I considered adding languages as a third array element (`["files.autoSave", "afterDelay", ["javascript", "typescript"]]` but this would only allow us to create a single rule for multiple languages.

Adding languages as a second parameter to `addSettings()` might be an option, although it would require multiple `addSettings()` for the above.


```ts
```

moltar: Looks jarring at the first sight.

moltar:
Is there a way to import VS Code type file, perhaps? Maybe even just copy it over into the repo?

mrgrain:
> Is there a way to import VS Code type file, perhaps? Maybe even just copy it over into the repo?

I looked into it, the schema doesn't appear to be published as a whole but appears to be assembled locally when needed. And even if it where published, it wouldn't cover extensions.

test/vscode/settings.test.ts
moltar: Why does it have to be an array of arrays, that's what I am confused about.

Why can't it be:

```ts
project.vscode?.settings.addSettings({
  "files.autoSave": "afterDelay"
})
```

mrgrain: Yes you're right. The array of arrays is not required if each language group gets it's own statement.

Thanks for the feedback. I'll update the PR.

moltar (approved):

src/vscode/settings.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Adds a workspace setting
+   *
+   * @param settings Array structure: [setting: string, value: any, languages?: string[]]
+   */
+  public addSettings(
+    settings: Record<string, unknown>,
+    languages?: string | string[]
+  ) {
+    if (languages && Array.isArray(languages)) {
```
moltar, line 48: You can pass undefined into isArray

moltar:
Also consider the more general solution. Not sure what it'd look like. But in vs code there can be multiple files. Like recommend extensions file is a separate one.

mrgrain:
> Also consider the more general solution. Not sure what it'd look like. But in vs code there can be multiple files. Like recommend extensions file is a separate one.

Launch Configs already exist #2044 added support for more properties
Recommended Extensions are added via #2046 

The APIs are very similar, although they are simpler than the settings.

rix0rrr (approved):

PR #2074, opened 2022-08-30
fix: remove hardcoded Node version in jsii workflows

### LINKED ISSUE #2073: Node v14 hardcoded in GitHub workflows
When I change the `minNodeVersion` and `workflowNodeVersion` fields in projen to other node versions, the `build.yml` workflow still contains reference to installing Node 14.x in several jobs like `package:js`.

I think this should default to the value of `workflowNodeVersion`, or no node version if it isn't specified.

Chriscbr:
Fixes #2073 

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,7 +90,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 14.17.0
       - name: Download build artifacts
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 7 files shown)


mrgrain (changes requested):
Looks good, but I think we need to also pass it on here:
https://github.com/projen/projen/blob/f33baebee33a3ddd9f12b651a83bf6e183f76028/src/javascript/node-project.ts#L541-L556

mrgrain (approved):
👍🏻

mrgrain:
Looks good but I want to check on the change in the nextjs snapshot before merging this.
Seems like we're defaulting to unsupported Node.js version now.

Chriscbr:
@mrgrain I think I found the culprit: https://github.com/projen/projen/blob/a3c0556b69b29456ce61517ed534824bb67100b4/src/web/next.ts#L79

Maybe it makes more sense to fix the NextJS projects in a separate PR

mrgrain:
@Chriscbr I have the Next.js follow up PR ready
https://github.com/projen/projen/pull/2083

PR #2079, opened 2022-08-31
feat(pom): support adding phase to plugin execution

cogwirrel:
I'd like to be able to build a lambda deployment package using a `JavaProject`. For this I need to use the `maven-shade-plugin` as mentioned here: https://docs.aws.amazon.com/lambda/latest/dg/java-package.html

The plugin execution needs `<phase>package</phase>` in order to run as part of the build. This PR allows this to be optionally specified.

See also: https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Using_the_executions_Tag

---


mrgrain (approved):
Nice one! Thanks for this contribution 🎉 

:shipit:

PR #2098, opened 2022-09-09
fix(release)!: existing tags for prerelease not respected

gradybarrett:
This change fixes the behavior of the bump task when a prerelease suffix is set, there are not yet any prerelease tags, but there are other normal tags for the current major version. The current behavior results in "first release" behavior. What should happen, and what this change implements, is a correct semver increment into a new prerelease.

For example,

Given
- A major version of `1`
- a previous tag of `1.2.0`
- a new `feat` commit
- a `prerelease` setting of `beta`

Previously a bump would have resulted in `1.0.0-beta.0`.
Now a bump will result in `1.3.0-beta.0`, as expected.

---

### DIFF
--- a/src/release/bump-version.ts
+++ b/src/release/bump-version.ts
@@ -291,9 +291,12 @@ function determineLatestTag(options: LatestTagOptions): {
 
   let tags = stdout?.split("\n");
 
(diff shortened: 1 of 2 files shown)


mrgrain (approved):
👍🏻

PR #2123, opened 2022-09-24
fix(react-ts): `@types/express` (4.17.14) and `@types/express-serve-static-core` (4.17.31) break compilation

iliapolo:
See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/62300

---


mrgrain (approved):
Nice catch!

PR #2127, opened 2022-09-26
fix: ignore overriden dependencies in renovate

### LINKED ISSUE #2126: Overrides are not reflected in deps.json causing renovate trying to update those
Currently there's 3 overrides in our constructs: 
```json
  "overrides": {
    "@types/responselike": "1.0.0",
    "got": "12.3.1",
    "@types/prettier": "2.6.0"
  },
```
When renovate runs, it tries to offer updated for those, which causes failure as `projen` run would revert versions (and we check in pipeline that code is not mutated).

`upgrade` task seems to be ignoring overrides, but Renovate doesn't. 

Not sure which one works incorrectly, as at least in Renovate support to overrides was explicitly added: https://github.com/renovatebot/renovate/issues/15278.

# Expected result

Renovate should ignore overrides, or at least offer flag to ignore those.

As versions are not in `.projen/deps.json`, some additional logic is needed.

Hi-Fi:
Fixes #2126 

---


src/javascript/node-package.ts
```diff
@@ -708,6 +708,21 @@ export class NodePackage extends Component {
...
+    const fieldName = packageResolutionsFieldName(this.packageManager);
+
+    let overrides: any = this.file.returnRawOverrides();
+
+    fieldName.split(".").forEach((field) => {
+      overrides = overrides[field];
+    });
+
+    return overrides;
+  }
```
mrgrain, line 724: This is a clever approach! You might find it easier to change to behaviour of `addPackageResolutions()` to match the one of `addDeps()`. I.e. store the resolutions in a class property and render as part of `renderDependencies()` and/or `resolveDepsAndWritePackageJson()` instead of using `this.file.addOverride()`

Hi-Fi: Thank you, have to check if some other option would be good. That was first approach that came to mind with just (mostly) getters.

Hi-Fi: Figured out a bit simpler solution, which can be later extended (if needed) to replace at least partly object-file.ts. Pushing it soon.

.projen/tasks.json
```diff
@@ -392,7 +392,7 @@
...
+          "exec": "yarn upgrade @types/conventional-changelog-config-spec @types/fs-extra @types/glob @types/ini @types/jest @types/node @types/semver @types/yargs @typescript-eslint/eslint-plugin @typescript-eslint/parser all-contributors-cli esbuild eslint-config-prettier eslint-import-resolver-node eslint-import-resolver-typescript eslint-plugin-import eslint-plugin-prettier eslint jest-junit jest jsii jsii-diff jsii-docgen jsii-pacmak json-schema json5 npm-check-updates prettier standard-version ts-jest typescript @iarna/toml case chalk conventional-changelog-config-spec fast-json-patch fs-extra glob ini semver shx xmlbuilder2 yaml yargs zlib @types/prettier @types/responselike got"
```
mrgrain, line 395: @Hi-Fi Looks like we need to remove the overrides when generating the upgrade task. Do you agree?

src/javascript/upgrade-dependencies.ts
```diff
@@ -182,7 +183,12 @@ export class UpgradeDependencies extends Component {
     const ncuExcludes = [
       ...new Set(
         this.project.deps.all
-          .filter((dep) => dep.version && dep.version[0] !== "^")
+          .filter(
+            (dep) =>
+              dep.version &&
+              dep.version[0] !== "^" &&
+              dep.type !== DependencyType.OVERRIDE
```
mrgrain, line 190: But here you are removing it. Something is off. 🤔

mrgrain (changes requested):
@Hi-Fi This is great! Thanks for contributing and the good work.

I'm not sure what is going wrong, but somehow the overrides still ended up in the generated upgrade task for projen itself. Once that's resolved, we can merge.

src/javascript/upgrade-dependencies.ts
Hi-Fi: I noticed that those came at first, but then after that ignore those were not changing snapshots anymore... maybe have to recheck later.

.projen/tasks.json
Hi-Fi: Hmm... this is probably some issue with versions, as deps get those new dependencies, but is it then so that tasks are generated with old code... Or then I need to rerun projen

Hi-Fi: Thank you for noticing, I just looked tests for these.

Hi-Fi: Ah, I missed that one, package manager update is not handled correctly (this is for ncu)

src/javascript/upgrade-dependencies.ts
Hi-Fi: That removal was for ncu, not for upgrade task of package manager.

mrgrain (approved):
🥳

Issue #1144, opened 2021-10-18
`projen test` updates snapshots by default


rix0rrr:
Given a `TypeScriptAppProject`, `yarn test` does the following:

```
$ npx projen test
....
🤖 test | jest --passWithNoTests --all --updateSnapshot
```

The `--updateSnapshot` is dangerous: if I run this in CI, `yarn test` will not actually validate my snapshots, but happily update them in place even if my tests should by all rights have failed.

The answer is not "anti-tamper protection": that will tell me that files have changed, it will not give me the nice error message and diff that `jest` would have given me if the snapshot test was allowed to fail in the first place. Plus, you can't assume that the tests will always be run with the synthesized GitHub workflow that includes anti-tamper protection: it might equally well run in a CodePipeline/CodeBuild setup, and that job might not even have access to the git repo to compare for any diffs.

msessa:
+1 for exposing the behavior as an option. changing the default will probably lead to an  [XKCD#1172](https://xkcd.com/1172/) type situation.

caviliar:
Hi, we are having a similar issue. We have updated to the latest version of projen and I can't get `yarn test` to run without updating snapshots. `antitamper` has been removed and I don't think there is a similar flag. @rix0rrr did you manage to solve this?

```
"test": {
      "name": "test",
      "description": "Run tests",
      "steps": [
        {
          "exec": "jest --passWithNoTests --all --updateSnapshot"
        },
        {
          "spawn": "eslint"
        }
      ]
    },
```

caviliar:
i got annoyed enough (i do some code. run tests. they automatically update snapshots). kinda pointless. 

added this to projenrc:
```
project.tasks.tryFind('test').reset('jest --passWithNoTests --all')
```

not the prettiest or there could be a better way... but if you want the previous behaviour back then hacking the `test` task back to the previous projen version did the trick for us. (until i come across the recommended way).

jtnz:
What is the intention of `projen test`? I feel like I'm constantly fighting against this behavior. Are we supposed to run `yarn test` in CI, then manually fail _if_ git is dirty afterwards? That doesn't sound very nice to me. And it's not just limited to `jest` either; `eslint` has `--fix`, which means if you run `yarn test` with the default behavior, the snapshots will be updated _and_ `eslint` will fix things like missing semicolons and trailing commas.

Today I tried improving on @caviliar 's solution with:
```
project.testTask.steps.forEach(step => {
  if (step.exec) {
[...]

PR #2129, opened 2022-09-27
fix: avoid updating snapshots in yarn test

rajyan:
fixes https://github.com/projen/projen/issues/1144

## Description

* As a lot of users is noticing, updating snapshot in task "test" is a unexpected behavior and it should be separated as "test:update"
* I believe that snapshot should not be updated in the `buildWorkflow`, but would try to keep the current behavior to avoid breaking changes

## Fix

---

### DIFF
--- a/src/javascript/jest.ts
+++ b/src/javascript/jest.ts
@@ -731,12 +731,6 @@ export class Jest {
         ? ` -c ${this.file.path}`
         : "";
 
-    // since our build & release workflows have anti-tamper protection, it is
-    // safe to always run tests with --updateSnapshot. if a snapshot changes,
-    // the `build` workflow will either fail (on forks) or push the update and
-    // `release` workflows will fail.
-    jestOpts.push("--updateSnapshot");
-
     // as recommended in the jest docs, node > 14 may use native v8 coverage collection
     // https://jestjs.io/docs/en/cli#--coverageproviderprovider
     if (
@@ -760,7 +754,7 @@ export class Jest {
     if (!testUpdate) {
       this.project.addTask("test:update", {
         description: "Update jest snapshots",
-        exec: `jest --updateSnapshot${jestConfigOpts}`,
+        exec: `jest --updateSnapshot ${jestOpts.join(" ")}${jestConfigOpts}`,
       });
     }
   }
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -322,7 +322,7 @@
       "description": "Run tests",
       "steps": [
         {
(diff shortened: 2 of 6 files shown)


mrgrain:
Thanks for this change @rajyan ! I think this is the best way forward. Trying to get some feedback from the community on it, but expect to hear back quickly.

rajyan:
Sorry for pushing after making the PR ready for review. 
Added some unit tests!

src/javascript/jest.ts
```diff
@@ -746,7 +746,7 @@ export class Jest {
       if (!testUpdate) {
         this.project.addTask("test:update", {
           description: "Update jest snapshots",
-          exec: `jest --updateSnapshot ${jestOpts.join(" ")}${jestConfigOpts}`,
+          exec: `jest --updateSnapshot${jestConfigOpts}`,
```
mrgrain, line 749: What's the thinking behind this change?

mrgrain: I think there might be options like `--roots` or `--runInBand` that will still have to be included here.

rajyan: I thought that it might be better to run only run `--udpateSnapshot` on `test:update` (which is the current behavior of `test:update` in `main`), but changed my mind again with your comment, because there might some options that can be used with `test:update` too. Thank you for pointing out!

rajyan:
Sorry for the mess. I think this PR is finally ready now.

mrgrain:
>Sorry for the mess. I think this PR is finally ready now.

Don't worry. I'm still waiting for some more community feedback.

pgarbe:
I like adding this new property to make it explicit and easy to change. The default could be set to true whenever you use the given GitHub build workflow but false when you're not using it (which is technically a breaking change but should be with low impact)

mrgrain:
@rajyan Based on community feedback, I'd like to introduce two new options to only update snaphots locally or in CI. Not expecting you to implement this, but in preparation for it, would you mind changing the flag from a boolean to an enum that looks something like this:

```
updateSnapshots: ALWAYS    <--- default
udpateSnapshots: NEVER
```

A follow-up PR could than easiliy introduce without breaking too much.
```
updateSnapshots: LOCAL
udpateSnapshots: CI
```
--- 
I cannot see a clear consensus on the default setting yet. Maybe a slight edge towards not updating in CI. So let's keep it for now. We can always bikeshed on this later.

rajyan:
> would you mind changing the flag from a boolean to an enum that looks something like this:

Sure!

mrgrain (approved):
Thanks again for this contribution @rajyan  🥳

Issue #2134, opened 2022-09-30
getfacl not found in v0.63.0?


anh-rivergon:
In v0.63.0 (could have been any release made on Sep-30), a new step "Backup artifact permissions" was added to the github workflow build.xml and triggered this error for us. Anyone else seeing the same issue -"getfack not found" ?

Run cd dist && getfacl -R . > permissions-backup.acl
  cd dist && getfacl -R . > permissions-backup.acl
  shell: sh -e {0}
  env:
    CI: true
/__w/_temp/9[2](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:2)dd0f[3](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:3)0-d[4](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:4)6f-4688-a246-[5](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:5)dcbb7d5ac03.sh: 1: /__w/_temp/92dd0f30-d4[6](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:7)f-4688-a246-5dcbb[7](https://github.com/a-cdk/ngx-app-stack/actions/runs/3161689264/jobs/5147555465#step:10:8)d5ac03.sh: getfacl: not found
Error: Process completed with exit code 127.

mrgrain:
Yes that was released today. What os or image does the workflow run on?

mrgrain:
In hindsight we should have made this an opt-in step.

mrgrain:
Cc @jmalins

anh-rivergon:
The release.xml says "runs-on: ubuntu-latest" -- I ran with "npx projen new awscdk-construct" went with the out of the box configuration.

mrgrain:
Thanks for the info. That's certainly expected to work. 🤔

jmalins:
It looks like the default configuration of the `awscdk-construct` runs using the `jsii/superchain:1-buster-slim-node14` container image. We will need to add the `acl` package to the superchain image build in the JSII project. That should fix the default case.

That being said, you're right, there should also be an opt-out for the cases where users bring their own images.

I can submit a superchain PR this weekend, as that should be a one liner, but may not be able to get to a parameterized fix until Monday.

[...]

PR #2132, opened 2022-09-29
fix(release): artifact permissions are not preserved in github actions

jmalins:
Fixes #2103.

Due to a bug in GitHub actions, file mode permissions (such as `+x`) are lost between when the build artifact is uploaded at the end of the build job and download in each release job. This results in the published artifact tarballs having incorrect permissions. The main user impact is that bin scripts are no longer executable. NPM automatically fixes this during install, but other JSII toolchains, such as Python, do not.

See details of GitHub Actions bug [here](https://github.com/actions/upload-artifact/issues/38).

To workaround the issue, this PR backs up the permissions of the artifact tree to a file (`permissions-backup.acl`) prior to upload, then restores them after download using the Linux `getfacl` and `setfacl` CLI tools.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2136, opened 2022-10-01
fix: getfacl not found

mrgrain:
In #2132 we fixed an issue with file permissions not being preserved in GH Action workflows. The implementation uses the `acl` package to backup end restore permissions.  While we can ensure the package is installed in all default runners or containers, it might not be available in customized workflows.

This change marks the permissions backup and restore steps as allowed to fail. Since it is a workaround for GH Actions issue and only required in certain situations, we should not fail the whole build if the package is not available.

Fixes: #2134 

Additionally fixes an issue with invalid workflow files being generated when `continueOnError` is set on a Job Step.

---


MrArnoldPalmer (approved):

mrgrain:
@Mergifyio refresh

pflorek:
@mrgrain thank you

PR #2174, opened 2022-10-14
feat: require copyright period and owner for MPL-2.0 license

ansgarm:
Hi 👋
Legal requires us to include this in our licenses. I'm happy to change this into a non-breaking change instead and alter the logic to allow an optional inclusion of this line instead of requiring it, if you'd rather want it that way.

---

### DIFF
--- a/license-text/MPL-2.0.txt
+++ b/license-text/MPL-2.0.txt
@@ -1,3 +1,5 @@
+Copyright (c) $copyright_period $copyright_owner
+
 Mozilla Public License, version 2.0
 
 1. Definitions


mrgrain:
I'm not keen on modifying licenses in any way that is not stipulated by the SPDX source we use (see source for [MPL-2.0](https://github.com/spdx/license-list-data/blob/master/text/MPL-2.0.txt)). 

For context also: #830 and #832

--- 

Could you include a NOTICE file instead like suggested in #830

Alternatively, it's quite easy to add any custom license:
```ts
  licensed: false,
});

new (class extends License {
  synthesizeContent() {
    return "COPYRIGHT 2022\n\n" + super.synthesizeContent();
  }
})(project, { spdx: "MPL-2.0" });
```

mrgrain (changes requested):
See comment above.

ansgarm:
Thanks for the suggestion @mrgrain 👍
I can see that fiddling with the license file seems weird – I'll go with the custom route you proposed!

ansgarm:
@mrgrain
Hm, this does not really work, as the package.json now states `UNLICENSED` unfortunately 😅
(I put `licensed: false` and `license: MPL-2.0`).

Is there a preferred way that you'd accept a new option? (I'm thinking about something like `skipLicenseFile` or something with a better wording?)

mrgrain:
It's never easy, isn't it? 😆 

Instead of `licensed: false`, try adding `project.tryRemoveFile("LICENSE");` before the new License
Or keep it and use an override: `project.tryFindObjectFile("package.json").addOverride("license", "MPL-2.0");`

--- 

Definitely welcoming contributions making any of these escape hatches easier, but having `license`,`licensed` and `skipLicenseFile` doesn't feel like a great solution. 🤔 

From a writing the hatch perspective, something like this would be nice. What do you think?
```ts
const license = new License(project, { spdx: "MPL-2.0" });
license.prependLine("COPYRIGHT");
project.package.setLicense(license);
```

There are obviously a few things missing here to support this kind of escaping.

ansgarm:
Hi @mrgrain 👋

I used `project.addFields({license: ".."})` now, which worked fine 👍

I can still make a PR if you want, but from my perspective this workaround is good enough and I don't think there are many users that need this 😁

mrgrain:
That's great. Obviously contributions are always welcome, but I agree that this is not a huge priority for most people.

PR #2182, opened 2022-10-18
chore: bump default version of cdktf construct

DanielMSchmidt:
---


mrgrain (changes requested):
Obviously the change works, but is this something we'd actually want?

I know CDKTF is still in the initial dev phase, so maybe that's a good idea. But for CDK we deliberately keep the minimum version low for max compatibility.

I'm just asking for confirmation here. You're obviously the expert on this topic and I won't stay in your way. 😄

DanielMSchmidt:
In CDKTF you can not expect minor versions to be compatible with one another (as a 0.x package each minor change is a breaking change), therefore we want to keep the default at the latest until we are more stable (e.g. 1.x)

PR #2183, opened 2022-10-18
feat: made bundler esbuild loaders configurable (#2181)

### LINKED ISSUE #2181: Allow configuration of additional esbuild loaders
Currently the project bundlerOptions (same holds for bundler AddBundleOptions) has no way to include different loaders for different file types.

A use case I ran into is importing a template file as a text in a lambda bundled by Projen. As the accompanied files are not copied to the assets directory of the bundled lambda, importing the template during bundling made more sense. To accomplish this using `import template from './template.mustache'` I manually had to include the `--loader:.mustache=text` option in the esbuild command.

In my opinion it would be nice to be able to configure different loaders in the bunlderOptions (or in the addBundle method of the bundler directly), for example:
```javascript
  const p = new NodeProject({
    name: "test",
    defaultReleaseBranch: "main",
    bundlerOptions: {
      esbuildVersion: "^3",
      loaders: {
        'mustache': 'text', // Adds a loader for .mustache files
      }
    },
  });
```
I can probably implement this myself if this you agree this is a good solution.

marnixdessing:
Fixes #2181 
---


mrgrain (approved):
Very cool! Thanks for contributing this feature 🎉

:shipit:

PR #2188, opened 2022-10-20
feat: support specifying commit types and scopes that fail version bumping to skip a release

### LINKED ISSUE #2179: Support skipping releases
Hey 👋

We implemented a feature to skip releases for our CDKTF pre-built provider bindings if neither the underlying Terraform provider version nor the CDKTF version had been changed. This is mainly driven by the fact that releasing on every dependency update that was merged to `main` causes us to [run into limits in Registries such as PyPI](https://github.com/hashicorp/terraform-cdk/issues/2169).

However, our current approach uses some pretty ugly hacks and workarounds and I'd like to add proper support to Projen instead or at least properly expose options that enable us to set / override the required options.

We're currently doing this ([source](https://github.com/hashicorp/cdktf-provider-project/blob/ddc68e0a30396f5fb19fe9cca32759e422ec47bb/src/index.ts#L190-L206)):
```ts
    (this.tasks.tryFind("release")!.condition as unknown as any) =
      "node ./scripts/should-release.js";
    const releaseJobSteps: any[] = (
      this.github?.tryFindWorkflow("release") as any
    ).jobs.release.steps;
    const gitRemoteJob = releaseJobSteps.find((it) => it.id === "git_remote");
    assert(
      gitRemoteJob.run ===
        'echo ::set-output name=latest_commit::"$(git ls-remote origin -h ${{ github.ref }} | cut -f1)"',
      "git_remote step in release workflow did not match expected string, please check if the workaround still works!"
    );
    const previousCommand = gitRemoteJob.run;
    const cancelCommand =
      'echo ::set-output name=latest_commit::"release_cancelled"'; // this cancels the release via a non-matching SHA;
    gitRemoteJob.run = `node ./scripts/should-release.js && ${previousCommand} || ${cancelCommand}`;
    gitRemoteJob.name += " or cancel via faking a SHA if release was cancelled";
```

This basically consists of two parts:
1) Adding a `condition` to the `release` task to be able to skip building and bundling a release
2) Overriding the "check for new commits" Github workflow step in a way that it also aborts all further steps/jobs if a release was skipped

Part 2 was required because adding `condition` to the release did not cause the workflow (i.e. further workflow steps) to be cancelled because there seems no way for the workflow to know that it should be cancelled. This however, might be by design and I'm just misusing `condition` for something it was not designed to do.

Before I put up a PR, I'd need some guidance around the way you'd like to approach such a feature (I might be looking at this from a completely wrong angle).
My current idea would be the following:
1) Add some kind of option for a `skipRelease` task/script/executable (that works like `condition` on tasks)
2) If that option is set, add a pre-build step to the release Github Workflow which runs this step and records its exit code
3) If that option is set, add an `if` config to all following workflow steps & jobs to skip them if the release should be skipped.

What do y'all think about this? Would you propose a different approach? Do you think skipping a release is a worthwhile feature at all?

ansgarm:
Resolves #2179


test/release/bump.test.ts
```diff
@@ -252,6 +252,22 @@ test("minMajorVersion throws if set together with majorVersion", async () => {
...
+        { message: "second version", tag: "v1.2.0" },
+        { message: "docs: updated Readme" },
+        { message: "chore(deps): updated dependencies" },
+      ],
+    })
+  ).rejects.toThrowError(
+    /No new commits have been added that warrant a release. All new commits were ignored/
+  );
+});
+
```
ansgarm, line 270: TODO: I need to handle merge requests as well (they should be ignored too, if an ignore config is set). I should add a test for this.

src/release/release.ts
```diff
@@ -487,28 +504,77 @@ export class Release extends Component {
...
+      branch.skipConventionalCommitTypes &&
+      branch.skipConventionalCommitTypes.length > 0
+    ) {
+      checkReleaseEnv.SKIPPED_TYPES =
+        branch.skipConventionalCommitTypes.join(",");
+    }
+
+    // if this is the release for "main" or "master", just call it "release:check".
+    // otherwise, "release:check:BRANCH"
+    // TODO: this will conflict if the user configures a branch called "check"
```
ansgarm, line 544: Is this important to handle?
The existing check for `main` vs `master` also fails if users happen to use both 😄

src/release/release.ts
```diff
@@ -487,28 +504,77 @@ export class Release extends Component {
...
+    const checkReleaseTaskName =
+      branchName === "main" || branchName === "master"
+        ? "release:check"
```
ansgarm, line 547: @mrgrain
This seemed to fit better than a `bump:check` task at this place – what do you think?

src/release/release.ts
```diff
@@ -573,7 +632,7 @@ export class Release extends Component {
       releaseTask.spawn(publishTask);
     }
 
-    const postBuildSteps = [...this.postBuildSteps];
+    const postBuildSteps = [...this.postBuildSteps]; // TODO: we'd need to add an "if" to these, too 😅
```
ansgarm, line 635: @mrgrain this one feels tricky – maybe it's cleaner to use some Github API to just cancel the whole workflow from the `release:check` step?
The existing check for new commits doesn't share this problem because it only runs after these steps 😅

test/release/bump.test.ts
ansgarm: TODO: I need to update this test too, it needs to move into its own file as this behaviour isn't part of bump anymore.

ansgarm (commented):
@mrgrain 
Continued some work on this – mentioned you in some comments I'd like some feedback on 🙂

src/release/release.ts
```diff
@@ -119,6 +120,22 @@ export interface ReleaseProjectOptions {
...
+  readonly skipConventionalCommitTypes?: string[];
+
+  /**
+   * Skip releasing if only commits matching these scopes have been made
+   * to the default branch since the last release
+   *
+   * e.g. ["docs"] skips if e.g. the only commit was "fix(docs): fix docs typo"
+   */
+  readonly skipConventionalCommitScopes?: string[];
+
```
mrgrain, line 138: Ah yes we have types and scopes. I wonder if there's a way to combine these options for even more flexiblility.
Like: "Skip `fix(docs)` but don't skip `feat(docs)`. 

Something like this maybe:
```
[
  'chore',
  ['fix,' [docs'],
  'ci'
]
````

**Comment by mrgrain** [comment]
@ansgarm Note to us that we need to consinder Breaking Changes.

Should they be treated like any other commit or always be released, regardless of scope?

File: src/release/release.ts (hunk 2 above)
Comment (mrgrain) [reply src/release/release.ts:544]: Yes, I think we might have to call it `check-release` or `release-check` or `pre-release` (<-- I like this one) if we go with a pre branch model.

Comment (mrgrain) [reply src/release/release.ts:544]: Side bar: I don't like how this is currenlty done. It shouldn't check for a hardcoded branch name, but rather give the default branch the "nice" name.

File: src/release/release.ts (hunk 3 above)
Comment (mrgrain) [reply src/release/release.ts:547]: In general yes, it's just the double `:` that won't work. See comment above.

File: src/release/release.ts [hunk 6]
```diff
@@ -487,28 +504,77 @@ export class Release extends Component {
...
+        branch.skipConventionalCommitTypes.join(",");
+    }
+
+    // if this is the release for "main" or "master", just call it "release:check".
+    // otherwise, "release:check:BRANCH"
+    // TODO: this will conflict if the user configures a branch called "check"
+    const checkReleaseTaskName =
+      branchName === "main" || branchName === "master"
+        ? "release:check"
+        : `release:check:${branchName}`;
```
mrgrain, line 548: Sidebar 2: I don't think we actually need different tasks for this, at least in the current implementation.
All "paramteres" are passed as environment variables, which are not set locally and only available in the workflow (via `sharedEnv`, L575).

Which I think is how the `bump` tasks does it as well at the moment.

Sidebar to the sidebar: We might want a feature at somepoint to parameterize builtin task steps.

src/release/release.ts
mrgrain: Agreed. I think what we need is a separate job here.
I've put a minimal example together:  https://github.com/mrgrain/abandon-workflow-experiment/blob/main/.github/workflows/release.yml

src/release/release.ts
ansgarm: We could also use a similar syntax as when writing conventional commit messages:
```jsonc
[
  'chore', // type, any scope
  'fix:', // type, any scope
  'fix(deps)', // type && scope
  'fix(deps):', // type && scope
  '(docs)', // scope, any type
  ''
]
```
But I like your more explicit typed objects one better 😁 To me it's intuitive and explicit.

mrgrain: Sounds like objects are the winner then! 🎉

src/release/release.ts
ansgarm: Re `pre-release`:
I feel that `release-check` or `check-release` makes the current implementation a bit more obvious. I wouldn't expect that a failing `pre-release` task silently skips everything. Also, the `postBuildSteps` could be interpreted as pre release steps, too (if one interprets releasing = publishing).

test/release/bump.test.ts
ansgarm: fyi: I pushed a commit that adds support for detecting merge commits – there still might be edge cases though

ansgarm:
@mrgrain do you know a good way to e2e test the changes to Github Workflows? 😅

mrgrain:
> @mrgrain do you know a good way to e2e test the changes to Github Workflows? 😅

Yes, good-ish.

1. Create a new test project of the desired type with projen
2. Build your projen branch
3. Dump the `dist/js/projen@0.0.0.jsii.tgz` file into the root of the test project as `projen.tgz`
    Important: You need to get rid of the `@` in the filename. Apparently that confuses yarn.
4. Add this to the projenrc file in the test project:
```ts
project.addGitIgnore('!projen.tgz');
project.tryFindObjectFile('package.json')?.addOverride('devDependencies.projen', 'file:./projen.tgz');
```
5. Success

ansgarm:
Closing this for now, as I won't be able spend the time to finish this up. If someone else wants to pick this up, feel free to take my work and create a new PR.

mrgrain:
FYI going with a simplified approach in this PR:
https://github.com/projen/projen/pull/2758

Thanks for your work on this. Was super helpful to make decisions.

Issue #2194, opened 2022-10-22
set-output warning from main-upgrade workflow


rantoniuk:
The `::set-output` command from the block:
```
      - name: Find mutations
        id: create_patch
        run: |-
          git add .
          git diff --staged --patch --exit-code > .repo.patch || echo "::set-output name=patch_created::true"
```

is causing a warning on the Action run that this action is deprecated:
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/upload-artifactShow less
 
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

```

rantoniuk:
@mrgrain how this fix can be applied to an existing project?
The main README shows a snippet of `projen upgrade-projen` but that command doesn't exist anymore apparently.

mrgrain:
It's just `projen upgrade` now. But since you have an upgrade workflow defined, it should also self-upgrade with its next run.

Thanks for pointing out the outdated docs. ❤️

PR #2196, opened 2022-10-24
fix(github)!: set-output warning from main-upgrade workflow

mrgrain:
Fixes: #2194

See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#patching-your-actions-and-workflows

BREAKING CHANGE: Requires self-hosted GitHub Actions runners to be updated to version 2.297.0 or greater.

---


rix0rrr (approved):
Might break people with self-hosted runners that haven't updated yet. I'll leave it to you to decide whether or not that's a blocker.

Issue #1513, opened 2022-01-11
feat: It should be possible to use custom license templates


markussiebert:
At the moment it's not possible to use custom license templates. I would like to have the possibility to override the path where the lookup of the license templates happens, because I can imagine that there are use cases where you maybe have to use not so common ones.

something like:

```ts
cosnt lookupPath = props.lookupPath ?? `${__dirname}/../license-text`;
const textFile = `${lookupPath}/${spdx}.txt`;
```

or maybe an additional lookup path for custom license templates.

What do you think?

johnmarian:
I'm trying to use projen for private projects and want to be able to use the NPM license configurations or either:
`{
  "license": "UNLICENSED"
}`
or
`{
  "license" : "SEE LICENSE IN <filename>"
}`

Trying to specify either of these throws an exception with projen, but leaving the license field blank sets the default of Apache 2.0

mikejgray:
@johnmarian Would you please share what type of project you're trying to create where you aren't able to set `{"LICENSE": "UNLICENSED"}` successfully? If it's node-package, it allows you to use the above but you have to set the property `licensed` to `false` (it defaults to `true`). node-project doesn't have a similar option.

Either way, I'm working on an enhancement to allow both custom licenses and `UNLICENSED`.

johnmarian:
@mikejgray - Sorry, I found the `licensed: false` flag after I commented but forgot to update. That worked just fine. It still doesn't completely work what I need (which is a custom license) so thanks for working on the enhancement!

PR #2204, opened 2022-10-29
feat: custom license path

mikejgray:
Addresses #1513. Allows the user to pass a custom license path to projen.

Custom licenses can use the $copyright_owner and $copyright_period values as any other support license.

---

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4795,9 +4795,9 @@ npm-normalize-package-bin@^3.0.0:
   integrity sha512-g+DPQSkusnk7HYXr75NtzkIP4+N81i3RPsGFidF3DzHd9MT9wWngmqoeg/fnHFz5MNdtG4w03s+QnhewSLTT2Q==
 
 npm-package-arg@^10.0.0:
-  version "10.0.0"
-  resolved "https://registry.yarnpkg.com/npm-package-arg/-/npm-package-arg-10.0.0.tgz#a34f4a4208a937074b1fff0943a684fbacc83977"
-  integrity sha512-7dkh8mRp7s0KwVHKIVJnFCJQ2B34gOGnzgBjDGyprycmARq/82SX/lhilQ95ZuacP/G/1gsS345iAkKmxWBQ2Q==
+  version "10.1.0"
+  resolved "https://registry.yarnpkg.com/npm-package-arg/-/npm-package-arg-10.1.0.tgz#827d1260a683806685d17193073cc152d3c7e9b1"
+  integrity sha512-uFyyCEmgBfZTtrKk/5xDfHp6+MdrqGotX/VoOyEEl3mBwiEE5FlBaePanazJSVMPT7vKepcjYBY2ztg9A3yPIA==
   dependencies:
     hosted-git-info "^6.0.0"
     proc-log "^3.0.0"


mrgrain:
Hey @mikejgray thanks for the contribution, that's a great idea!

I'd like to dig into the API design with you. Wouldn't it be easier to make `licensePath` refer to a file (and I guess could also name it `licenseFile` then) and make `spx` optional as well? Plus maybe some custom validation to make sure users set one of them. What do you think?

I'd also like to see the `"SEE LICENSE IN <filename>"` be added as mentioned in this comment https://github.com/projen/projen/issues/1513#issuecomment-1238634003

mikejgray:
Hey @mrgrain , thanks for the feedback. Let me give that a go and see what happens. I may also need some assistance with the tests. They pass fine on my machine but apparently don't work on the GHA runners.

mikejgray:
Taking a look at your requests, I have a few observations:

1. Designing the API that way seems like it would add some complexity to extend later on. Keeping `spdx` mandatory and allowing it to pull from the contents of a file seems more flexible to my mind, plus not fundamentally alter the required parameters of the interface. The way I proposed here does make it more challenging to use the `UNLICENSED` option from the issue.
2. I saw that `node-project` and `node-package` aren't consistent in their implementation of licenses. `node-package` doesn't extend `License` from license.ts, while `node-project` does. I'd rather not lump a refactor into this same PR but is probably worth a followup. What do you think?

mrgrain:
With the current proposal, in order to use a license `FOOBAR`, a user would have to do something like:
```ts
{
  spdx: "FOOBAR",
  licensePath: "path/to/licenses"
}
```

Few things to note here:
- `FOOBAR` is not a spdx licence, so it would be semantically wrong
- `package.json` would contain `FOOBAR` as a license which would cause a warning (this can probably achieved with the current approach)
- projen gives up the guarantee that `spdx` is a always valid SPDX license name

---

The license interface could do with a bigger refactor to be more generic, with a special factory method `License.spdx('MIT')` to provide the shortcut for common licenses. That's a going to be though migration though that needs some thinking through.

--- 

> 2. I saw that `node-project` and `node-package` aren't consistent in their implementation of licenses. `node-package` doesn't extend `License` from license.ts, while `node-project` does. I'd rather not lump a refactor into this same PR but is probably worth a followup. What do you think?

Let's not worry too much about `node-package`  too much. `node-project` is what most users will interact with. But we need to make sure that this change doesn't break anything else, i.e. `node-package`  must produce a valid `package.json`.

mrgrain:
By the way, if this is really about escape hatching, maybe improvements like the ones I have proposed [here](https://github.com/projen/projen/pull/2174#issuecomment-1279031711) would be easier to achieve,

mrgrain:
> > What's the reasoning for providing a folder and not a file name by the way?
> 
> A misunderstanding by me about SPDX...it would allow minimal modifications to the existing code, but at the cost of guaranteeing valid SPDX license names, as you mentioned.

👍🏻 

> I like the improvement you proposed in the linked unmerged PR. What do you think about providing that escape hatch programmatically to the user if they pass in the filename? It would be specifically to add in the `"SEE LICENSE IN <filename>"` functionality requested in the issue.

Yes, I think adding `"SEE LICENSE IN <filename>"` automatically when a custom license is used would be the desired developer experience. 

> `{ "license": "UNLICENSED" }` is already available with existing code, so there's no reason to touch that at this time.
Agreed.

---

There are a few variations of ideas floating around now. Would you mind writing the proposed interface changes up in a comment? Just so we can agree on them before you go to work on it.

mikejgray:
@mrgrain I haven't forgotten this, just been busy - will try to post an update no later than this weekend. :)

PR #2242, opened 2022-11-16
fix(javascript): jest cli doesn't have an`--all` argument

philipmw:
This argument does not exist and hasn't existed since at least 22.x:

https://jestjs.io/docs/cli

https://archive.jestjs.io/docs/en/22.x/cli

This argument was added in Dec 2020 by a larger commit adding build tasks: 95112272c2b192144293c1064c77b2d8da354b8f

And before that, the whole file did not exist and the project has no mention of `--all`. So I cannot deduce the original reason for adding this argument.

---

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -322,7 +322,7 @@
       "description": "Run tests",
       "steps": [
         {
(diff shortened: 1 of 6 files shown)


mrgrain (approved):
Haha that's a little awkward 🙈 Nice find though! 🚀 Thanks @philipmw

Issue #2253, opened 2022-11-22
Enable markers for eslint configuration files


dj-rabel:
As I found in the docs of eslint, they do support comments for both, yaml as well as json configuration files. So theres nothing stopping us from adding the "generated by projen ..." comment to these.

Mentioned here: https://github.com/mozilla/addons-frontend/issues/942
In eslint docs here: https://eslint.org/docs/latest/user-guide/configuring/configuration-files#comments-in-configuration-files

Currently markers for both types of config files are disabled:
https://github.com/projen/projen/blob/6b7ab972023ea01988ea0693f115cec0f737be8c/src/javascript/eslint.ts#L382
https://github.com/projen/projen/blob/6b7ab972023ea01988ea0693f115cec0f737be8c/src/javascript/eslint.ts#L387

If you're fine with enabling it I would go ahead and open a pull request, changing both to true, as well as setting "allowComments: true" for the json variant to indicate it supporting comments. (adding "//": "~~ generated ..." to json, fails with an error. Might be why it was on false)
Reference to the flag im talking about: https://github.com/projen/projen/blob/6b7ab972023ea01988ea0693f115cec0f737be8c/src/json.ts#L34

Reason I come up with this: Found myself editing the file manually today, not realising it was managed by projen, as there was no comment indicating it inside the file.

mrgrain:
Yes please!

Inline Json comments are relatively new, so I guess that's why it wasn't enabled. Heads up, you might have to change some tests, since the default setup breaks with inline Json comments.

PR #2260, opened 2022-11-25
chore(eslint): enable markers for eslint config files

dj-rabel:
PR resolving issue https://github.com/projen/projen/issues/2253
____

As described in the referenced issue, this PR enables markers for yaml, as well as json config files for eslint.

Enabling it was easy, but as mentioned by @mrgrain, rewriting the tests ..... wasn't 🙈 

So there are two additional parts, I had to touch:
1. I've changed [the position the marker gets added to, inside a json file](https://github.com/projen/projen/pull/2260/files#diff-5b45d0800ed9310a182e74c50faf3883830b08588277690cab7d7cd82e7e5089R57). Before my change, it got added before the closing brace. I could not find any reference on why it is implemented this way. I checked references for eslint, typescript, jsonc and json5. All mention that comments can appear anywhere in the whole file. So theres no need to add it somewhere in between, where one will probably not see it. After my change now, it's added right in front of the opening brace. Had no issues with it in my tests.
2. I changed the way, the util/synth works. Theres a [flag](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R97) now, indicating that we want to have json comment support when parsing the json files for a directory snapshot. If this flag is set to true ([which it will be automatically, if the snapshot function was called with a project object given, with at least on JsonFile inside having supportsComments=true](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R60)), json files will not be parsed with nodes default JSON.parse method, but instead [with the one from the package comment-json](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R123). This allows us to access the comment sections from inside the json file, in code. You can see this in action inside the [unit tests](https://github.com/projen/projen/pull/2260/files#diff-7aea6ac1f2df4a619d919af7c66e875606cb6e0edf627111e8aff15946c49d0dR142), I adjusted. I tried to reduce the impact this could have on existing projects, as much as I could.

That's pretty much about it. Like always in development, coming up with a viable solution was the complex part here. Not so much the coding itself.
I hope this is somehow as you expected it to be or at least that you like the solution. In any case, let me know what you think.

_EDIT: Added some references to the actual code_
____


test/integ.test.ts
```diff
@@ -26,7 +26,10 @@ for (const projenrc of files) {
     sanitizeOutput(workdir);
 
     expect(
-      directorySnapshot(workdir, { excludeGlobs: ["node_modules/**"] })
+      directorySnapshot(workdir, {
+        excludeGlobs: ["node_modules/**"],
+        parseJson: false,
```
mrgrain, line 31: I did not expect that to be needed. Could you explain why that's the case?

mrgrain (changes requested):
This is amazing @dj-rabel 🤩 

Just one tiny clarification question, and it should be good to go. Thanks a lot for this!

mrgrain:
> 1. I've changed [the position the marker gets added to, inside a json file](https://github.com/projen/projen/pull/2260/files#diff-5b45d0800ed9310a182e74c50faf3883830b08588277690cab7d7cd82e7e5089R57). Before my change, it got added before the closing brace. I could not find any reference on why it is implemented this way. I checked references for eslint, typescript, jsonc and json5. All mention that comments can appear anywhere in the whole file. So theres no need to add it somewhere in between, where one will probably not see it. After my change now, it's added right in front of the opening brace. Had no issues with it in my tests.

Yes that makes sense. I don't recall having any specific reason for the current location.

> 2. I changed the way, the util/synth works. Theres a [flag](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R97) now, indicating that we want to have json comment support when parsing the json files for a directory snapshot. If this flag is set to true ([which it will be automatically, if the snapshot function was called with a project object given, with at least on JsonFile inside having supportsComments=true](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R60)), json files will not be parsed with nodes default JSON.parse method, but instead [with the one from the package comment-json](https://github.com/projen/projen/pull/2260/files#diff-2c1b89dbbc9ad18a45853cc2bc0a3fc879a0a384470f56358eb88f6c1a9a68a8R123). This allows us to access the comment sections from inside the json file, in code. You can see this in action inside the [unit tests](https://github.com/projen/projen/pull/2260/files#diff-7aea6ac1f2df4a619d919af7c66e875606cb6e0edf627111e8aff15946c49d0dR142), I adjusted. I tried to reduce the impact this could have on existing projects, as much as I could.

That's amazing and well done. It's pretty much what I had in mind for a refactor as well. I think alternatives could have been to use a simpler parsing library and just ignore all comments (but that makes testing tricky) or wrap JSON content into a helper class that can either access the data as an object or as a string (making the tests something like `expect(jsonContent.asLinesArray()[0]).toBe(MARKER)`.

---

But hypotheticals. This is great!

test/integ.test.ts
dj-rabel: Sure, there's a json file with marker comment inside some of the integration test samples. We could either use `parseJson: true, supportJsonComments: true,` now or `parseJson: false,`, as I did.
There is no content check on specific parts of the json files in this test, so I thought parsing every json file would be unnecessary.

dj-rabel:
Sorry for the late reply. Was busy moving appartment.

> This is amazing @dj-rabel 🤩
> 
> Just one tiny clarification question, and it should be good to go. Thanks a lot for this!

Glad to hear that! You're very welcome ❤️ 

> That's amazing and well done. It's pretty much what I had in mind for a refactor as well. I think alternatives could have been to use a simpler parsing library and just ignore all comments (but that makes testing tricky) or wrap JSON content into a helper class that can either access the data as an object or as a string (making the tests something like expect(jsonContent.asLinesArray()[0]).toBe(MARKER).

Exactly what I was thinking. I checked the overhead in tests with this library and it's minimal. Like a a few extra seconds for a full test run. So I thought this would be acceptable.

mrgrain:
@mergifyio update

mrgrain (approved):

PR #2293, opened 2022-12-12
fix(renovate): add options to overwrite configuration

### LINKED ISSUE #2292: renvovate options should be configurable
the options for renovate are [hardcoded](https://github.com/projen/projen/blob/133b10ae3013d7857baa4f58b56ade62a7bbafe7/src/renovatebot.ts#L141-L146). While this is good to have convenient default settings, there is a need to customize renovate settings. Every company / user probably has their own packageRules or extension. I was thinking first the make each property extensible, but when we take at at [schema json](https://docs.renovatebot.com/renovate-schema.json) of renovate, a generated model for typescript props would have 150k+ lines. So instead we can just leave it up to the user to override any property.

karlderkaefer:
closes #2292
---

### DIFF
--- a/src/renovatebot.ts
+++ b/src/renovatebot.ts
@@ -43,6 +43,10 @@ export interface RenovatebotOptions {
    * List of labels to apply to the created PR's.
    */
   readonly labels?: string[];
+
+  readonly overrideConfig?: any;
+
+  readonly marker?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mrgrain (approved):
🚀

karlderkaefer:
thank you, that speed of review and release process is awesome!

PR #2299, opened 2022-12-14
fix(awscdk): test sample code uses import based on srcdir

lexfelixpost:
I saw that the import of MyStack in the test file, generated by the typescript sample code component, doesn't use the srcdir parameter creating wrong imports if you change the srcdir.


mrgrain (approved):
Nice one!

Issue #1006, opened 2021-08-26
Github update dependencies fails with private registry


eriklztiqqe:
We have multiple repositories, which have dependencies towards AWS CDK constructs in private npm registries (Github packages, later possibly Cloudsmith).

To access these constructs we thus need NPM_TOKEN exposed, but the update dependencies workflow does not include this, only GITHUB_TOKEN.
This causes the workflow to fail each time and we get an email notification about it. Currently, that includes about 13-14 repositories, and there will be more.

This is kind of related to #815, which also concerns the lack of exposed variables, but a different use case.

jumic:
You can use `workflowBootstrapSteps` to add additional steps to your workflow. I used the following code snippet:
```ts
workflowBootstrapSteps: [{
  name: 'npmjs login',
  run: 'echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc',
  env: {
    NPM_TOKEN: '${{ secrets.NPM_TOKEN }}',
  },
}],
```
It creates file `.npmrc` including the `NPM_TOKEN`. In my example project I was able to use a private package from npm.js in my upgrade workflow.

Issue #2094, opened 2022-09-07
projen workflows cannot run docker commands by default


kaizencc:
**Problem**:

Sometimes docker commands are bundled into projen commands. For example, in https://github.com/cdklabs/awscdk-asset-awscli, we specify a pre-compile step that executes a `build.sh` script. This script runs a `docker build` command, which builds a local image. This encounters the following problem:

```bash
Run npx projen build
👾 build » default | node .projenrc.js
👾 build » pre-compile | layer/build.sh
>> Building AWS Lambda layer inside a docker image...
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: \
Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&shmsize=0&t=aws-lambda-layer&target=&ulimits=null&version=1": \
dial unix /var/run/docker.sock: connect: permission denied
```

Under the hood, projen by default is using the `jsii/superchain:1-buster-slim-node14` container image and docker commands are not set up to be run inside this image. Therefore, the following solution is necessary:

```js
workflowBootstrapSteps: [
    {
      name: 'Change permissions on /var/run/docker.sock',
      run: 'sudo chown superchain /var/run/docker.sock',
    },
  ],
```

Which gives the `superchain` user ownership of the docker socket.

**Proposal**:

This is a sharp edge that we had to figure out, and I can't imagine there aren't other scenarios where we want to run docker commands in a projen-managed repository. The proposal is to surface this command as a boolean property on the projen library construct, like so:

```ts
const project = new awscdk.AwsCdkConstructLibrary({
  ...
  allowDockerCommands: true,
});
```

----

And, because of https://github.com/aws/jsii/issues/3746, configuring passwordless sudo requires additional changes to `.projenrc.js`. We have been able to work around this via projen escape hatches:

```js
const buildWorkflow = project.tryFindObjectFile('.github/workflows/build.yml');
buildWorkflow.patch(JsonPatch.add('/jobs/build/container/options', '--group-add sudo'));
```

Until the issue in jsii is somehow addressed, this will need to be baked into this proposal as well.

berenddeboer:
[...]

PR #2311, opened 2022-12-20
fix(jsii): run workflow setup steps for language jobs

### LINKED ISSUE #1799: Adding npmrc for private registry with workflowBootstrapSteps fails on build.yml / package-js
In reference to #1006, we added the following snippet to our `.projenrc.ts`:

```
  //...
  workflowBootstrapSteps: [
    {
      name: 'Login to GitHub npm registry',
      id: 'npm_auth',
      run: 'echo -e "//npm.pkg.github.com/:_authToken=${NPM_TOKEN}" >> ~/.npmrc',
      env: {
        'NPM_TOKEN': '${{ secrets.NPM_PULL_SECRET }}'
      },
    },
  ],
});
```

This will work  for the initial jobs of the `build.yml` and `release.yml` workflows but will fail on the the following jobs as the `$HOME` directory is not shared between jobs.
Therefore the `Install Dependencies` steps of `build.yml/package-js` and `release.yml/release_npm` workflows/jobs will fail.

Pleas advise if this is a bug or if we are using the wrong API for what we need to do.

asdcamargo:
Fix #1799 #1006

Allow workflow setup steps to be used by extensions of node-project. 

This change will enable for example a step that sets NPM authentication for private repos before build and publish.

---

### DIFF
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -8427,6 +8427,7 @@ Name | Type | Description
 **package**🔹 | <code>[javascript.NodePackage](#projen-javascript-nodepackage)</code> | API for managing the node package.
 **packageManager**⚠️ | <code>[javascript.NodePackageManager](#projen-javascript-nodepackagemanager)</code> | The package manager to use.
 **runScriptCommand**🔹 | <code>string</code> | The command to use to run scripts (e.g. `yarn run` or `npm run` depends on the package manager).
+**workflowBootstrapSteps**🔹 | <code>Array<[github.workflows.JobStep](#projen-github-workflows-jobstep)></code> | <span></span>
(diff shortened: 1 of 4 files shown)


mrgrain:
@mergifyio rebase

mrgrain (approved):

cplee:
This change breaks the workaround in #2094 ... if I use the following to get access to Docker socket from the build job, it is also adding these steps to the package jobs which don't use the superchain image.

```workflowBootstrapSteps: [
    {
      name: 'Change permissions on /var/run/docker.sock',
      run: 'sudo chown superchain /var/run/docker.sock',
    },
  ],
```

kaizencc:
@cplee sorry I'm late to the party! But yes, it does. I think this needs to get reverted and/or exposed as an option and not the default. Will take it up with @mrgrain and arrive at a solution.

cplee:
I worked around the problem by using @mrgrain suggestion here: https://github.com/projen/projen/issues/2094#issuecomment-1378460172

PR #2312, opened 2022-12-20
chore(javascript): reduce the scope of the upgrade dependencies token

kylelaker:
The second step of this workflow that handles opening the pull request
is unnecessarily broad. Because the `PROJEN_GITHUB_TOKEN` (or the
otherwise configured token) will have `contents:write` and
`pull-requests:write` and that token is the one that is used to actually
push the content and open the pull request (as part of the
`peter-evens/create-pull-request` action), the `GITHUB_TOKEN` can be
used exclusively to checkout the branch with `contents:read`.

This change reduces the permissions of the `GITHUB_TOKEN` to the minimal
amount required and also reduces the places the configured token is used
to creating the pull request.

---

### DIFF
--- a/.github/workflows/upgrade-main.yml
+++ b/.github/workflows/upgrade-main.yml
@@ -42,14 +42,12 @@ jobs:
     needs: upgrade
     runs-on: ubuntu-latest
     permissions:
-      contents: write
-      pull-requests: write
+      contents: read
     if: ${{ needs.upgrade.outputs.patch_created }}
     steps:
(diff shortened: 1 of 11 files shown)


mrgrain:
PR looks good. I'm on leave at the moment and can't react if any thing goes wrong. So merging this in the New Year.

mrgrain:
@mergifyio rebase

mrgrain (approved):

PR #2314, opened 2022-12-21
feat: added github-actions checking options to dependabot

lexfelixpost:
I noticed that dependabot can also check github actions for version checking. But this configuration is missing in the dependabot component. 

Instead of reworking the whole structure to enable additional updates entries. I've opted for a simple boolean option to add the github-action updates entry enabling dependabot to check github-action versions.


mrgrain (changes requested):
This will conflict with synthesizing workflows. I.e. Dependabot will upgrade the actions version and the next time `projen` runs any workflows will be created with the old versions. Got any ideas how to work around this?

lexfelixpost:
Gave it a thought. I don't think there is a clean solution available. I'll close the PR. If such a solution does come up I'll circle back.

PR #2332, opened 2022-12-29
feat(tasks): add append methods to tasks

juho9000:
Allow appending to tasks in addition to prepending.

---

### DIFF
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -2687,6 +2687,57 @@ Name | Type | Description
 ### Methods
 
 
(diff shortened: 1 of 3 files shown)


mrgrain:
How are these different to `exec` `spawn` and `say`?

juho9000:
They're not, seems I've gone blind. I was only looking at the prepend-prefixed methods, and somehow completely missed that the regular spawn, exec and say are indeed appending.. :man_facepalming:

PR #2345, opened 2023-01-05
fix: use posix paths for autodiscovered lambda entrypoints

### LINKED ISSUE #2344: Auto-detected lambdas use platform specific path separators in entrypoint when generated on Windows
The exec step in `.projen/tasks.json` uses platform-specific path separators (i.e. `\`) when generated on Windows:

```
"exec": "esbuild --bundle src\rotate-secret.lambda.ts --target=\"node16\" --platform=\"node\" --outfile=\"assets\\rotate-secret.lambda\\index.js\" --tsconfig=\"tsconfig.dev.json\" --external:aws-sdk"
```

This causes the generated github workflows to fail with the following error:

```
 release » build » pre-compile » bundle » bundle:rotate-secret.lambda | esbuild --bundle src\rotate-secret.lambda.ts --target="node16" --platform="node" --outfile="assets\rotate-secret.lambda\index.js" --tsconfig="tsconfig.dev.json" --external:aws-sdk
Error: R] Could not resolve "srcrotate-secret.lambda.ts"
```

Disabling lambda auto-discovery and manually specifying a posix-pathed entrypoint is a functional workaround, but it would be convenient for this to be the default behavior.

jesse-grabowski:
Fixes #2344

Autodiscovered lambdas use back-slashes in their entrypoints when generated on Windows. This causes the generated github workflows to fail. Forcing posix path format for entrypoints fixes the workflows while still being compatible with esbuild on Windows.

---

### DIFF
--- a/src/cdk/auto-discover-base.ts
+++ b/src/cdk/auto-discover-base.ts
@@ -1,4 +1,4 @@
-import { join } from "path";
+import * as path from "path";
 import * as glob from "glob";
 import { Component } from "../component";
 import { Project } from "../project";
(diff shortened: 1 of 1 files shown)


mrgrain (approved):
Nice one, thanks!

PR #2357, opened 2023-01-07
fix(python): bump minimum python version to 3.7

### LINKED ISSUE #2356: projen requires Python ~=3.7, so it will not be satisfied for Python >=3.6,<3.7
The minimum version of Python supported by projen's PyPi package is 3.7, but projen defaults to 3.6 and has no easy mechanism to start a new project with an updated version.

PR forthcoming.

mikejgray:
BREAKING CHANGE: The default Python version used for Poetry and Setuptools was updated to `>=3.7` as required by each of these packages.

Fixes #2356 

---


mrgrain:
Hey @mikejgray Thanks for the PR. Given that Python 3.6 is EOL I'm wondering if we should just drop it everywhere.

Also have you seen the note [regarding this here](https://projen.io/python.html#poetry-dependencies):
> Poetry requires the python version to be specified as a dependency e.g. python@^3.7,<=3.9. When not specified, projen will set it to a default value of python@^3.6.

Do I understand this PR correctly that it would make the above the default?

mikejgray:
> Hey @mikejgray Thanks for the PR. Given that Python 3.6 is EOL I'm wondering if we should just drop it everywhere.
> 
> Also have you seen the note [regarding this here](https://projen.io/python.html#poetry-dependencies):
> 
> > Poetry requires the python version to be specified as a dependency e.g. python@^3.7,<=3.9. When not specified, projen will set it to a default value of python@^3.6.
> 
> Do I understand this PR correctly that it would make the above the default?

Hi @mrgrain You're welcome. The reason for this PR is that I couldn't initialize a higher-level PoetryPythonProject construct with poetry because projen was specifying 3.6 and there was no way I could find to pass a different version constraint on initialization. If you were to start with a basic PythonProject and change the parameters you could do it, but not on a fresh project. I figured once I sorted that out I could submit a separate PR to start a project with Poetry.

I updated from 3.6 to 3.7 everywhere I found in the code and just did a `grep -R '3.6' .` in `src/` and `test/` to confirm. To the best of my knowledge this PR will bump the minimum Python version in a PythonProject to 3.7

Probably worth noting that 3.7 will go EOL this summer so this process will potentially need to be repeated.

mikejgray:
Hey @mrgrain , are there any changes you'd like me to make to this PR besides merging the latest from main?

mrgrain:
> Hey @mrgrain , are there any changes you'd like me to make to this PR besides merging the latest from main?

Nothing from you. Also no need to update from main unless there are conflicts. I just need a bit more time to think about possible implications of this change.

mrgrain (approved):

mrgrain:
@mikejgray Thanks for the contribution! It's still quite annoying to create a new project for poetry because all the other options have to be disabled. I would be open to a PR that improves handling python tools.

Something along the lines of
```
if poetry => default pip, setuptools, venv to false
if poetry and pip => error
...
```

PR #2368, opened 2023-01-16
feat(cli): `projen new` can take simple list project options

mrgrain:
Re-enables and fixes tests related to creating projects from external modules. These were commented out due to a previously introduced breaking change. Now the affected external project types have caught up and I've replaced the incompatible packages with more recent ones to make the tests work again.

Additionally removes loading commands from a directory, as this has a known incompatibility with ESM and Deno and no 
practical benefit.

---


src/cli/cmds/new.ts
```diff
@@ -178,7 +207,7 @@ function commandLineToProps(
       }
     }
   }
-
+  console.log(props);
```
Naumel, line 210: Is this a left-over or intended?

Naumel (approved):

PR #2375, opened 2023-01-18
fix(python): creation of PythonProject with poetry

### LINKED ISSUE #1863: Cannot create python.PythonProject with Poetry.
In a new folder, execute:

```
projen new python --poetry true
```

You will see the following error:
```
projen new python --poetry true
/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/tasks.js:45
            throw new Error(`A task with the name ${name} already exists. To override it, call removeTask first and then create the new task.`);
            ^

Error: A task with the name install already exists. To override it, call removeTask first and then create the new task.
    at Tasks.addTask (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/tasks.js:45:19)
    at PythonProject.addTask (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/project.js:116:27)
    at new Poetry (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/python/poetry.js:19:36)
    at new PythonProject (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/python/python-project.js:76:28)
    at evalmachine.<anonymous>:1:28
    at Script.runInContext (node:vm:139:12)
    at Object.runInContext (node:vm:289:6)
    at createProject (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/projects.js:74:8)
    at Function.createProject (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/projects.js:27:9)
    at initProject (/Users/smckinney/.nvm/versions/node/v16.13.1/lib/node_modules/projen/lib/cli/cmds/new.js:241:25)
```

It looks like there's another package manager that's still being configured when the poetry option is supplied.

mikejgray:
Explicitly unsets pip, setuptools, and venv if poetry is set to true. This enables the creation of a PythonProject with the `--poetry true` flag.

Considered throwing errors if both poetry and pip were set to true, but opted instead to provide warnings in the interface instead. If there is a use case for poetry being set with one or more of pip, setuptools, and venv, we can re-evaluate this.

fixes #1863

---

### DIFF
--- a/src/python/python-project.ts
+++ b/src/python/python-project.ts
@@ -102,6 +102,9 @@ export interface PythonProjectOptions
    * Use poetry to manage your project dependencies, virtual environment, and
    * (optional) packaging/publishing.
    *
+   * Please note that this feature is incompatible with pip, setuptools, and venv.
+   * If you set this option to true, then pip, setuptools, and venv will be set to false.
+   *
    * @default false
    * @featured
    */
@@ -222,6 +225,15 @@ export class PythonProject extends GitHubProject {
       );
     }
 
+    let poetry = options.poetry ?? false;
+    let pip = options.pip ?? true;
+    let venv = options.venv ?? true;
+    let setuptools = options.setuptools ?? this.projectType === ProjectType.LIB;
+    // If we select poetry, then pip/venv/setuptools must be set to false
+    if (options.poetry) {
+      pip = venv = setuptools = false;
+    }
+
     // default to projenrc.py if no other projenrc type was elected
     if (options.projenrcPython ?? !anySelected(rcFileTypeOptions)) {
       new ProjenrcPython(this, options.projenrcPythonOptions);
@@ -231,15 +243,15 @@ export class PythonProject extends GitHubProject {
       new ProjenrcJs(this, options.projenrcJsOptions);
     }
 
-    if (options.venv ?? true) {
+    if (venv) {
       this.envManager = new Venv(this, options.venvOptions);
     }
 
-    if (options.pip ?? true) {
+    if (pip) {
       this.depsManager = new Pip(this);
     }
 
-    if (options.setuptools ?? this.projectType === ProjectType.LIB) {
+    if (setuptools) {
       this.packagingManager = new Setuptools(this, {
         version: options.version,
         description: options.description,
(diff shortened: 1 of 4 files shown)


src/python/python-project.ts
```diff
@@ -222,6 +225,15 @@ export class PythonProject extends GitHubProject {
...
     }
 
+    let poetry = options.poetry ?? false;
+    let pip = options.pip ?? true;
+    let venv = options.venv ?? true;
+    let setuptools = options.setuptools ?? this.projectType === ProjectType.LIB;
+    // If we select poetry, then pip/venv/setuptools must be set to false
+    if (options.poetry) {
+      pip = venv = setuptools = false;
+    }
```
mrgrain, line 235: I think I'd rather have this explicitly erroring if an invalid combination is provided.
```suggestion
    const poetry = options.poetry ?? false;
    const pip = options.pip ?? !poetry;
    const venv = options.venv ?? !poetry;
    const setuptools = options.setuptools ?? !poetry && this.projectType === ProjectType.LIB;
```

mrgrain (changes requested):
We also should tests the various combinations, to clearly describe the expected behavior.

The `@default` tags for each need to be updated as well.

src/python/python-project.ts
```diff
@@ -102,6 +102,9 @@ export interface PythonProjectOptions
    * Use poetry to manage your project dependencies, virtual environment, and
    * (optional) packaging/publishing.
    *
+   * Please note that this feature is incompatible with pip, setuptools, and venv.
+   * If you set this option to true, then pip, setuptools, and venv will be set to false.
```
mrgrain, line 106:
```suggestion
   * This feature is not compatible with pip, setuptools, or venv.
   * If you set this option to `true`, then pip, setuptools, and venv must be set to `false`.
```

mrgrain (changes requested):
Thanks @mikejgray  if you could add a few tests for error cases and we should be good to go.

mrgrain (approved):

PR #2380, opened 2023-01-22
feat(github): allow specifying permission scope for app token

kylelaker:
This updates `tibdex/github-app-token` to v1.7.0; newer versions of this
action support specifying the specific permissions that should be
granted to the created token.

This creates a new interface for `AppPermissions` following the
documentation to [create an installation access token][1]. Additionally,
these permissions can be `read`, `write`, or `admin`; however, there is
not a `none` options, so the `JobPermission` enum cannot be reused and a
new `AppPermission` enum is defined.

A test is also added to test specifying this.

Ideally the _default_ for `permissions` would be to just set
`pullRequests` and `contents` to `WRITE`; however, if users have added
steps onto the workflow that require additional permissions, those may
break if the default permissions are changed from "all available" to
"just PRs and contents".

[1]: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app

---


mrgrain:
Busy week for me, but looks good.

Is the App permission stuff a GitHub feature that previously wasn't exposed? Or something that this particular action came up with?

Asking because I don't want to tie ourselves too much to a particular action implementation.

kylelaker:
> Is the App permission stuff a GitHub feature that previously wasn't exposed? Or something that this particular action came up with?

The permissions for tokens are a GitHub thing! They're documented at https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app

This particular action (semi-)recently added support for actually specifying them to limit scope. Some other actions with a similar purpose and the Octokit SDK itself allow specifying them (https://github.com/octokit/auth-app.js). It should be fairly portable (maybe changing the name of the parameter within `with` or minor formatting changes to the value if the option were switched out).

mrgrain (approved):
🚀

mrgrain:
Thanks @kylelaker !

I would welcome a PR to reduce the default permissions to what is necessary. As long as we mark it as a breaking change and describe in the description how to revert to the previous behavior, we are good.

PR #2382, opened 2023-01-23
feat(github): allow basic customization of GitHub Action version strings

kylelaker:
This takes a stab at some basic support for customization of the version
component of GitHub Action `uses` definitions, primarily focusing on
specifying major/minor/SHA. Some organizations require that users
specify a SHA and some users like being able to automatically follow the
newest release of an action.

This setting is only available at the `GitHubOptions` level and is then
applied to _all_ Actions (or at least those defined via `GitHubAction`).
There isn't much to customize the representation of a _single_ `uses`
statement (beyond calling `toUsesString`). Additionally, users can still
provide a `string` directly for `uses` and that should be repspected
without modification.

An Action be specified by calling `findActionBy`, which primarily
contains the actions used within the Projen codebase itself; by creating
an instance of `GitHubAction` with the necessary fields defined, or by
providing the raw `uses:` string in the GitHub Actions workflow
format. `findActionBy` allows filtering by name and by version number,
choosing the first one found in the internal list.

I am unsure whether `findActionBy` should be exported; it creates an
inconsistent interface for end-users. How do they know whether
`findActionBy` will actually return an item or not?

Additionally, this interface might be a bit cleaner if it were more
directly part of the `Github` class; that might allow for more easily
overriding versions ("always use version `v4` of `actions/stale`") and
other potential changes. But hopefully at least having gotten (I think)
all the existing Actions usages into a single location will help with
any necessary refactoring for where `GitHubAction` itself lives and how
it's defined. Also, it may be ideal if users can specify their own
Action definitions within their `projenrc` and have `findActionBy` work
for actions defined that way too.

I tried a more class-based approach; however, that wasn't super viable
because of the call to `resolve` in `workflow.ts` which forbids objects
with constructors. That is the primary reason why this is data objects
with additional functions within the module, so I want to look more into
that to see if there's a better interface for users to define their own
Actions.

So basically, users would define something like the following in
`projenrc.js` to always use SHAs:

```js
{
  gitHubOptions: {
    versionResolutionOptions: {
      format: VersionRepresentation.SHA,
    },
  }.
}
```

Or to prefer major version numbers for trusted publishers:

```js
{
  githubOptions: {
    versionResolutionOptions: {
      format: VersionRepresentation.AUTO, // This is the default
      trustedPublishers: ["actions", "github"],
    }
  }
}
```

Which will use major version (if available) for "actions/" and
"github/"-published actions and then SHA for all others.

Also a better interface for `findActionBy` might be something like:

```ts
type GitHubActionFilter = { majorVersion?: string, minorVersion?: string };
findActionBy(name: string, opts?: GitHubActionFilter): GitHubAction | undefined;
```

In any case, hopefully this can be a starting point for some
conversations about the interface; but it does work in a minimal
capacity at the moment.

---

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -16,6 +16,6 @@ jobs:
       pull-requests: write
     if: contains(github.event.pull_request.labels.*.name, 'auto-approve') && (github.event.pull_request.user.login == 'cdklabs-automation')
     steps:
-      - uses: hmarr/auto-approve-action@v2.2.1
+      - uses: hmarr/auto-approve-action@de8ae18c173c131e182d4adf2c874d8d2308a85b
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,12 +15,12 @@ jobs:
       CI: "true"
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
         with:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c
         with:
           node-version: 14.18.0
       - name: Install dependencies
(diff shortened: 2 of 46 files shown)


kylelaker:
This was initially meant to target functionally like requested in #1702; but as it grew it sorta became something that did more and it's closer to #1891 in some ways but this has a _way_ worse interface.

This also makes the assumption that information about version and the SHAs for tags would be maintained directly in Projen and that support for multiple versions of an action may be desirable within a code base.

So it takes a different and less-Projeny approach than #1891. I will probably take a look at starting from that as a base for the next iteration on this though.

PR #2384, opened 2023-01-25
fix: prettier.addOverride was ignored

tinovyatkin:
`prettier.addOverride` was pushing into a temporary array and so effectively ignored. This PR fixes that.

---


mrgrain (approved):

mrgrain:
Thanks @tinovyatkin 🎉  

There's an issue with the jsii types here. I think it's enough to remove the `readonly`. You can verify this locally by running `yarn build`.

Would you also mind adding (or extending) a test for it so we don't run into this issue again?

mrgrain (approved):
🚀

tinovyatkin:
@mrgrain any particular reason we don't to merge it yet or just a forgotten `no-merge` label?

mrgrain:
> @mrgrain any particular reason we don't to merge it yet or just a forgotten `no-merge` label?

Thanks for following up on this. Yep, I just forgot to take of the label 🙈

PR #2386, opened 2023-01-26
fix: prettier overrides types uses wrong field name

tinovyatkin:
As per [documentation](https://prettier.io/docs/en/configuration.html#configuration-overrides), the override settings should be in the `options` field not in the `settings` as currently,

This is technically a breaking change but as projen is 0.x and the current `settings` is a pure bug I believe it's OK to release as a patch version.

---


mrgrain (approved):
🚀

PR #2388, opened 2023-01-26
feat: allow prettier configuration in yaml

tinovyatkin:
Prettier doesn't support comments in it's json configuration file and emits a warning on unknown field (i.e. `//`) key in config, so the only option to have a marker in the generated config is to either use json5 or yaml for the config.

This PR proposes to follow [ESlint component design](https://github.com/projen/projen/blob/main/src/javascript/eslint.ts#L85) and adds `yaml` option to the `prettierSettings` that will output Prettier config in YAML format with appropriate generation marker.

---


mrgrain (approved):
Awesome!

PR #2394, opened 2023-01-29
fix(eslint): remove json-schema redundant dependency

tinovyatkin:
The dependency `json-schema` is added by [Eslint component ](https://github.com/projen/projen/blob/83e3a5dba5d6bd0f1bf3f0b5b15e6ac245fa7526/src/javascript/eslint.ts#L163)for some undocumented reason.

However this dependency is not required as a peer by any other dependency in this component, neither it somehow used anywhere else within the projen project.

I'm assuming this is a leftover of something and proposing to remove this redundant dependency.

---


mrgrain (approved):
Good find!

PR #2395, opened 2023-01-29
feat(jest): convert Jest to Component

tinovyatkin:
Contrary to ESLint, Prettier, and most of the other components `Jest` is not extending the `Component` class but just a generic class that can only be accessed via `project.jest` property (if set)

This PR proposes converting the `Jest` class into the `Component` subclass for the following benefits:
* unification of approach with Eslint, Prettier, Projenrc, and other components
* providing `Jest.of(project)` API to be in line with Eslint and Prettier components and simplify accessing/extending Jest from outside components
* Facilitates future [migration to Constructs](https://github.com/projen/projen/pull/1671/files#diff-02bcf880af61bcbdba8cd4974d4a8ac84d32fcc0d4f4fb08d05cc053d3176c42) if that hopefully will happen one day

Additional small fixes in this PR:
* adding Jest external config to npmignore
* exposing the `file` property as public following [TypescriptConfig component](https://github.com/projen/projen/blob/main/src/javascript/typescript-config.ts#L405) design and simplifying JSON patching to support more properties.

---

### DIFF
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -7879,6 +7879,7 @@ Installs the following npm scripts:.
 
 __Submodule__: javascript
 
+__Extends__: [Component](#projen-component)
 
 ### Initializer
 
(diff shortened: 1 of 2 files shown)


mrgrain (approved):
Nice one!

PR #2399, opened 2023-01-31
fix(node): CodeArtifact auth with GitHub OIDC is missing workflow permissions

AminFazlMondo:
Currently if the auth provider for CodeArtifact is set to `GitHub OIDC`, the publish step does use GitHub identity and assumes role to publish the package to AWS CodeArtifact.
- Code Artifact optios are also used for private scoped packages, and will generate workflow steps to login to code artifact prioe to installation of the package. If `GitHub OIDC` is used in conjunction with scoped packages, the build and release workflows are not generated correctly to cater for this type of credential. This PR will generate that and not use the access key and secrets for the "Configure AWS Credentials" step.
- In order for the job to be able to use GitHub as Identity provider for AWS, the job needs to have access to id-token write ([read more](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-permissions-settings)) in the workflow. This code has been added to grant this permission to "build" and "release" workflow and jobs to achive this.

---


mrgrain (changes requested):
Thanks for this change, feels important!

I left a few comments/qustions inline.
Most importantly, can you also please add a description to the PR explaining the issue this is solving and how you are solving it. This helps me review things better and quicker, but also for future posterity.

src/release/release.ts
```diff
@@ -243,6 +248,12 @@ export interface ReleaseOptions extends ReleaseProjectOptions {
    * @default 14.x
    */
   readonly workflowNodeVersion?: string;
+
+  /**
+   * Additional permission that are required for the job
+   * @default - no additional permission
+   */
+  readonly additionalPermissions?: JobPermissions;
```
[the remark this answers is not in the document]
AminFazlMondo: I have renamed the file and class, hope it is more clear now

src/build/build-workflow.ts
```diff
@@ -95,6 +96,12 @@ export interface BuildWorkflowOptions {
    * @default "{ pullRequest: {}, workflowDispatch: {} }"
    */
   readonly workflowTriggers?: Triggers;
+
+  /**
+   * Additional permission that are required for the job
+   * @default - no additional permission
+   */
+  readonly additionalPermissions?: JobPermissions;
```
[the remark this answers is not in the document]
AminFazlMondo: I thought that might be openning up this for errors in future and missing the minimum permissions that it requires. But happy to be guided by you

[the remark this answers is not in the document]
mrgrain: Thanks for your input and I agree with you in principle. For this case, I reckon if we do it like I suggested it would be very hard to accidentally mess things up. One would have to explicitly set `contents: JobPermission.NONE` to achieve breaking the build and that's hardly something someone would do without purpose.

Let's go with `permissions` and my suggestion.

src/release/release.ts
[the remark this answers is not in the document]
mrgrain: Apologies for the confusion. We should _not_ rename the whole class and file. Please revert.

I meant the `additionalPermissions` option. With the other discussion in mind, it should be called `workflowPermissions`. Similar to `workflowNodeVersion` just above it.

mrgrain (changes requested):
See inline comments.

src/build/build-workflow.ts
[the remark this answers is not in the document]
AminFazlMondo: Thanks @mrgrain ,
I have made the changes you asked, and in order to avoid any confusion, I aslo made the original behavior as the default value for the passed in option to keep the interface the same, and if someone wants to override they can too.
Please let me know what you think?

src/build/build-workflow.ts
```diff
@@ -95,6 +96,12 @@ export interface BuildWorkflowOptions {
    * @default "{ pullRequest: {}, workflowDispatch: {} }"
    */
   readonly workflowTriggers?: Triggers;
+
+  /**
+   * Permission that are required for the job
+   * @default `{contents: 'write'}`
+   */
+  readonly workflowPermissions?: JobPermissions;
```
mrgrain, line 104: This one should still be called `permissions` only the one in release needs to workflow prefix.

```suggestion
  readonly permissions?: JobPermissions;
```

src/build/build-workflow.ts
```diff
@@ -95,6 +96,12 @@ export interface BuildWorkflowOptions {
    * @default "{ pullRequest: {}, workflowDispatch: {} }"
    */
   readonly workflowTriggers?: Triggers;
+
+  /**
+   * Permission that are required for the job
+   * @default `{contents: 'write'}`
```
mrgrain, line 102:
```suggestion
   * @default `{ contents: JobPermission.WRITE }`
```

src/build/build-workflow.ts
```diff
@@ -95,6 +96,12 @@ export interface BuildWorkflowOptions {
    * @default "{ pullRequest: {}, workflowDispatch: {} }"
    */
   readonly workflowTriggers?: Triggers;
+
+  /**
+   * Permission that are required for the job
```
mrgrain, line 101:
```suggestion
   * Permissions granted to the build job
   * To limit access to `contents`, the desired permissions have to be set explicitly: `{ contents: JobPermission.NONE }`
```

test/javascript/node-project.test.ts
```diff
@@ -1080,49 +1091,148 @@ describe("scoped private packages", () => {
...
+            name: "Install dependencies",
+            run: "yarn install --check-files --frozen-lockfile",
+          },
+        ])
+      );
+    });
+
+    test("adds id-token permission to release workflow", () => {
+      const buildWorkflow = yaml.parse(output[".github/workflows/release.yml"]);
+      expect(buildWorkflow.jobs.release.permissions).toEqual(
```
mrgrain, line 1230:
```suggestion
      const workflow = yaml.parse(output[".github/workflows/release.yml"]);
      expect(workflow.jobs.release.permissions).toEqual(
```

src/release/release.ts
```diff
@@ -243,6 +248,12 @@ export interface ReleaseOptions extends ReleaseProjectOptions {
    * @default 14.x
    */
   readonly workflowNodeVersion?: string;
+
+  /**
+   * Permission that are required for the job
+   * @default `{contents: 'write'}`
```
mrgrain, line 254:
```suggestion
   * Permissions granted to the release workflow job
   * @default `{ contents: JobPermissions.WRITE }`
```

mrgrain (changes requested):
Thanks @AminFazlMondo 
Got a few minor change requests. Getting close now!

AminFazlMondo:
Thanks @mrgrain ,

I have made these changes, could you please have a look again?

Cheers,

src/javascript/node-project.ts
```diff
@@ -812,8 +829,30 @@ export class NodeProject extends GitHubProject {
...
+      NodePackageCodeArtifactAuthProvider.GITHUB_OIDC
+    ) {
+      return [
+        {
+          name: "Configure AWS Credentials",
+          uses: "aws-actions/configure-aws-credentials@v1",
+          with: {
+            "aws-region": "us-east-2",
+            "role-to-assume": parsedCodeArtifactOptions.roleToAssume,
+            "role-duration-seconds": 900,
```
mrgrain, line 846: How can the region be hardcoded? Am I missing something here? 
I realize this is also hard coded a bit further down already.

(Same goes for `role-duration-seconds` but that's probably not the most pressing)

mrgrain (changes requested):
Hey @AminFazlMondo Looks good now.

On my final review round I noticed `aws-region` is hardcoded to `us-east-2` (and `role-duration-seconds` is as well). Is that on purpose?

If yes please explain it to me (maybe even add a comment) and I can approve.
If not on purpose, let's fix this and make both a parameter.

src/javascript/node-project.ts
AminFazlMondo: The decision back in #1688 was to reuse the `codeArtifactOptions` to figure out the role to assume and others. This class interface does not contain any information about the dwfault AWS region or session duration, so needed to either be hard coded or added to some entries, which I am not sure if `codeArtifactOptions` is the best place for it.
Also, as it is only using IAM to get token and following operations (such as `ca:loging`) will have region as part of the CLI commands. There is nothing that misses the region that would like to infer that from the envinonmnet variable, hence, it is not doing ny damage.
Having said that, if others are adding customised commands that need to infer the AWS region, this might be problematic.

mrgrain: Thanks for the explanation! That's annoying but it's also not getting any worse. 👍🏻

mrgrain (approved):

PR #2409, opened 2023-02-01
feat: convert TypescriptConfig to Component and allow comments

tinovyatkin:
This PR converts `TypescriptConfig` class into a subclass of `Component` and proposes omitting default values instead of hardcoding them.

The motivation for inheriting from `Component` is the same as per #2395 but as we can have several instances of `TypescriptConfig` per project we are not providing a static `.of` method. Yet this change allows us to easily feature-detect the presence of `TypescriptConfig` inside a project. This is incredibly handy in the case of using this class in some non-standard projects (i.e. Typescript monorepo root or as a base for `jsconfig.json`).

This PR also fixes the fact that `tsconfig.json` support comments. [1]

We also added a test for this component (we didn't have any) where we are loading generated config using the API of Typescript itself - that will catch any possible mistakes in `compilerOptions` as TS throws on unknown values.

[1] https://github.com/microsoft/TypeScript/pull/5450

---


src/javascript/typescript-config.ts
```diff
@@ -397,27 +398,36 @@ export interface TypeScriptCompilerOptions {
...
   public readonly fileName: string;
   public readonly file: JsonFile;
 
   constructor(project: NodeProject, options: TypescriptConfigOptions) {
+    super(project);
+
     const fileName = options.fileName ?? "tsconfig.json";
 
-    this.include = options.include ?? ["**/*.ts"];
-    this.exclude = options.exclude ?? ["node_modules"];
```
mrgrain, line 411: Changing these defaults makes me anxious. Can you please break them out into a separate PR.
I don't want the other changes to suffer from this discussion.

tsconfig.dev.json
```diff
@@ -1,3 +1,4 @@
+// ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
```
mrgrain, line 1: Do you have a reference for this? Mostly for the records.

tinovyatkin: https://github.com/microsoft/TypeScript/pull/5450

src/javascript/typescript-config.ts
tinovyatkin: Remove that change.

Also added a test of loading generated config by TypeScript, for the piece of mind regarding comments support.

mrgrain (approved):
🚀 Excellent work and thanks for adding the additional test!

PR #2414, opened 2023-02-04
feat(js-config): js config files (e.g. .eslintrc.json)

### LINKED ISSUE #2405: eslint: option to generate a js file for config
In order to make `eslint` work properly in a monorepo setup, I need to be able to configure the following:

```
parserOptions: {
  project: ['./tsconfig.json'],
  tsconfigRootDir: __dirname,
}
```

That kind of configuration is only possible using `.js` files, but currently projen will generate a `.json` file.

------

Motivation: I need to include the `tsconfigRootDir`, because if I don't ESLint will prefer the `./tsconfig.json` in the root of the repo, instead of the `./tsconfig` in the package directory. 

* I can't put an absolute path in the `json`, because this file goes into version control and it will be different on everyone's machine.
* I can't put a relative path in the `json` because it gets interpreted differently by the ESLint VSCode plugin (will resolve the path relative to the repo root) and ESLint when run from the command line (will resolve the path relative to cwd).

andrestone:
Introduces `JsConfigFile` which is compatible with `JsonFile` interface.

The class exposes the static method `JsConfigFile.identifier(str: string)` to allow rendering globals such as `__dirname`.

```typescript
  //...
  const eslint = new Eslint(project, {
    devdirs: ["foo", "bar"],
    dirs: ["mysrc"],
    jsConfig: true,
  });
  eslint.config.parserOptions = {
    project: ["./tsconfig.json"],
    tsconfigRootDir: JsConfigFile.identifier("__dirname"),
  };
  // ...
```

Implemented using a magic string, happy to reimplement if someone has a better idea.

Fixes #2405
---


src/js-config.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private replaceIdentifier(str: string): string {
+    const regex = new RegExp(
+      `(\"${JsConfigFile.jsConfigFileIdentifier})([a-zA-Z_$][a-zA-Z_$0-9]+)(\")`,
+      "gm"
+    );
+    return str.replace(regex, "$2");
+  }
+}
```
mrgrain, line 54: This file should probably live inside the javascript submodule.

test/js-config/eslintrc.test.ts
```diff
@@ -0,0 +1,29 @@
+import { Eslint } from "../../src/javascript/eslint";
+import { NodeProject } from "../../src/javascript/node-project";
+import { JsConfigFile } from "../../src/js-config";
+import { Projenrc } from "../../src/projenrc-json";
+import { synthSnapshot } from "../util";
+
+test(".eslintrc.js with identifier values", () => {
```
mrgrain, line 7: Nice! Would also need some tests dedicated to `JsConfigFile` directly.

src/js-config.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export interface JsConfigFileOptions extends JsonFileOptions {}
+
+/**
+ * Represents a JS configuratin file (e.g. .eslintrc.js).
+ * Suppor
+ */
+export class JsConfigFile extends ObjectFile {
+  static identifier(str: string): string {
+    return `${JsConfigFile.jsConfigFileIdentifier}${str}`;
```
mrgrain, line 17: Out of curiosity what else did you try?

Using a function or an object with a `toJSON()` comes to mind, but not sure if it's still run through `JSON.stringify` adding quotes afterwards.

mrgrain: I guess we could build our own recursive renderer that prints out `Json.stringify` unless the value is a certain instanceof (could use `Symbol` or something custom).

mrgrain: Here is a gist that does that: https://gist.github.com/mrgrain/a61408a4e8dd2006be5000a6523bf28e

But I wonder if that's enough or if people would want to start using imports as well. 🤔

src/javascript/eslint.ts
```diff
@@ -83,6 +84,12 @@ export interface EslintOptions {
    * @default false
    */
   readonly yaml?: boolean;
+
+  /**
+   * Write eslint configuration as JsConfig file instead of JSON
+   * @default false
+   */
+  readonly jsConfig?: boolean;
```
mrgrain, line 92: I guess at this point we should start replacing the boolean flags with a fileformat enum.

tinovyatkin:
I hope we consider all complex environments before merging this proposal.

Legacy ESlint JS-config is not supporting ESM, so, if `type: module` is set in `package.json` then this will fail - it should be `.cjs`. If we want to have Javascript based config for ESLint then it makes sense to implement new [Flat config support](https://eslint.org/blog/2022/08/new-config-system-part-2/) - that *only* can be `eslint.config.(mc)js` - that will be future proof, more performant and make way more sense.

If the whole proposal is just a workaround for setting  `tsconfigRootDir` then there are documented ways how to solve this problem - use [array of tsconfigs in monorepo](https://typescript-eslint.io/linting/typed-linting/monorepos) or globs or ESlint plugin settings - depending what you are trying top solve with `tsconfigRootDir`.

src/js-config.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  protected synthesizeContent(resolver: IResolver): string | undefined {
+    const json = super.synthesizeContent(resolver);
+    if (!json) {
+      return undefined;
+    }
+
+    const sanitized = JSON.parse(json);
```
tinovyatkin, line 38: NIT: (_as I hope this will not be merged_) but `replaceIdentifier` may be more robustly implemented as [JSON reviewer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter).

mrgrain:
Drafting this PR as I clearly needs more discussions.

> If the whole proposal is just a workaround for setting `tsconfigRootDir` then there are documented ways how to solve this problem - use [array of tsconfigs in monorepo](https://typescript-eslint.io/linting/typed-linting/monorepos) or globs or ESlint plugin settings - depending what you are trying to solve with `tsconfigRootDir`.

No I don't think it is. That might have been the issue that brought it up, but not being able to produce JS based configurations is certainly a gap in projen.

> Legacy ESlint JS-config is not supporting ESM, so, if `type: module` is set in `package.json` then this will fail - it should be `.cjs`. If we want to have Javascript based config for ESLint then it makes sense to implement new [Flat config support](https://eslint.org/blog/2022/08/new-config-system-part-2/) - that _only_ can be `eslint.config.(mc)js` - that will be future proof, more performant and makes way more sense.

This is very interesting. I wonder how that fits within our current Eslint component in general.

---

To me this now looks like we have two problems at hand:
- A generic `JsConfigFile` component that supports inlining code fragments into a generic configuration object. Needs to support both ESM and CJS.
- How to integrate the above with the existing Eslint component

src/javascript/eslint.ts
giseburt: :+1:

PR #2417, opened 2023-02-06
fix: add git config call to mark working directory as safe

DanielMSchmidt:
From git v2.35.2 onwards this config needs to be set, sothat
git operations are permitted in the directory.

This should fix issues like this:

```
Run git add .
fatal: detected dubious ownership in repository at '/__w/cdktf-provider-github/cdktf-provider-github'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/cdktf-provider-github/cdktf-provider-github
```

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,6 +80,8 @@ jobs:
         run: |-
           git config user.name "github-actions"
           git config user.email "github-actions@github.com"
+      - name: Set safe directory
+        run: git config --global --add safe.directory $(pwd)
       - name: Push changes
         run: |2-
             git add .
--- a/.github/workflows/upgrade-main.yml
+++ b/.github/workflows/upgrade-main.yml
@@ -60,6 +60,8 @@ jobs:
         run: |-
           git config user.name "github-actions"
           git config user.email "github-actions@github.com"
+      - name: Set safe directory
+        run: git config --global --add safe.directory $(pwd)
       - name: Create Pull Request
         id: create-pr
         uses: peter-evans/create-pull-request@v4
--- a/src/build/build-workflow.ts
+++ b/src/build/build-workflow.ts
@@ -329,6 +329,7 @@ export class BuildWorkflow extends Component {
           repository: PULL_REQUEST_REPOSITORY,
         }),
         ...WorkflowActions.setGitIdentity(this.gitIdentity),
+        ...WorkflowActions.setSafeDirectory(),
         {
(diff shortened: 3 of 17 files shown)


mrgrain:
Hey @DanielMSchmidt Do you have any references for this and or builds that fail due to this missing?

DanielMSchmidt:
Sure, this is a failing build: https://github.com/cdktf/cdktf-provider-github/actions/runs/4094672563/jobs/7061059203#step:7:11

mrgrain:
> From git v2.35.2 onwards this config needs to be set

And any references to this specifically? 

> Sure, this is a failing build: https://github.com/cdktf/cdktf-provider-github/actions/runs/4094672563/jobs/7061059203#step:7:11

Thanks!

I run some quick tests on my projects, and they both are still okay. 🤔 
https://github.com/mrgrain/streamlink-serverless/actions/runs/4103513404/jobs/7077803002#step:6:1
https://github.com/mrgrain/cdk-esbuild/actions/runs/4103626057/jobs/7078057342#step:7:1

---

I'm not convinced this is the right solution. Few thoughts why:
- The issue is clearly on going in the community and might be solved by GitHub Actions. So potentially this is only a temporary fix.
- I have plans to remove running the build inside an container anyway. It causes problems. This just adds to the pile.
- Should this be fixed at the source, i.e. superchain?
- If `git add` is the problem, I don't think we need to use it to create a patch file of current changes. We might be able to refactor this differently.

---

Related issues
https://github.com/actions/runner-images/issues/6775

mrgrain (changes requested):
Discussion needs to be resolved first.

DanielMSchmidt:
> And any references to this specifically?

https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765

> I'm not convinced this is the right solution. Few thoughts why:

Ok, I'm fine with not shipping this, especially if it's a temporary problem as you said. I was focused on unblocking our build / release process, but I could achieve this by adding a pre-build step so this PR was just me trying to be a good OSS citizen 👍

mrgrain:
> Ok, I'm fine with not shipping this, especially if it's a temporary problem as you said. I was focused on unblocking our build / release process, but I could achieve this by adding a pre-build step so this PR was just me trying to be a good OSS citizen 👍

Thanks, I appreciate this very much! Let's see if more people encounter this issue.
I do wonder what is different in your project that causes this. 🤔

PR #2424, opened 2023-02-08
feat(github): add support for workflows run-name

RomainMuller:
The `run-name` property allows configuring the name GitHub will generate for workflow runs, which is shown in the "Actions" tab in GitHub.

See: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name

---


mrgrain (approved):
🚀

PR #2433, opened 2023-02-10
feat(github): add support for the 'strategy' property in reusable workflows

nicholas-keers:
Hello, I wanted to use projen to generate a new project at my workplace, but I found that support for using a matrix in reusable workflows didn't exist, so I've created this small PR to add it. You can see that the 'strategy' key is now supported in the Github docs: https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow

I have moved the `strategy` property to the `CommonJobDefinition` as its now supported in `Job` and `JobCallingReusableWorkflow`. 

I've added a snapshot test following the convention I saw in `github-workflow.test.ts` and verified that the matrix renders as expected:

```
name: test-workflow
on: {}
jobs:
  use-a-matrix:
    name: Some reusable workflow
    permissions: {}
    uses: some-user/example-action.yaml@v1
    with:
      fruit: \${{ matrix.fruit }}
    secrets: inherit
    strategy:
      matrix:
        fruit:
          - apple
          - orange
          - banana
```
I don't believe this will break anything, tests passed on my machine ok - I hope that this is useful for others as well, and that it could be merged promptly assuming I've not missed anything!

---


mrgrain (approved):
Looks good, thank you!

PR #2435, opened 2023-02-10
fix(new): unable to use enum properties when creating external projects

### LINKED ISSUE #1932: Unable to change packageManager when using projen new
When running `npx projen new ... --package-manager=pnpm` I get the following error:

```
evalmachine.<anonymous>:7
  packageManager: javascript.NodePackageManager.PNPM,
                  ^

ReferenceError: javascript is not defined
```

This is due to the fact that the .projenrc file does not import the NodePackageManager from the javascript namespace.

mrgrain:
Fixes #1932 

When creating projects we run the initial project synth in side a node vm. By design this vm has limit context.
During this synth we evaluate option, some of which have enums values. The enums are imported from their respective (sub)module. For built-in projen project types this is fine, as the module of the project was already part of the vm context.

For external project types however, we might also refer to enums from projen (or other modules).
We already collect the necessary information to know which modules are used by these enum options.
But the modules were previously not made available in the vm context.

This changes adds all modules required for options into the vm. It adds them under a random name to avoid collision with possible real values. E.g. it's conceivable that a third-party project type exposes `projen` submodule.  

Also re-enables a previously disabled test for external project.

---


rix0rrr (approved):
You didn't really explain what the problem was or why the fix works, but I'll ship it.

mrgrain:
> You didn't really explain what the problem was or why the fix works, but I'll ship it.

Fixed! Thanks for calling me out 😃

PR #2439, opened 2023-02-11
feat(github): allow workflow action versions to be set

### LINKED ISSUE #1702: lock down GitHub Actions to specific SHA

### LINKED ISSUE #1844: Manage GitHub actions versions

mrgrain:
It is a good security practice to lock versions of GitHub Actions down to explicit commit hashes.
Ideally projen would manage and upgrade GitHub Actions versions like any other dependency.
However in absence of this we need a mechanism for users to change the version of workflows that come with projen.

This change implements to smallest possible solution.
Users can define explicit overrides for each used action if desired.
The API allows to override a specific action version, which can be useful if an action has breaking changes between major versions.
Alternatively all occurrences of an action can be overridden, irregardless of the action version.

Workflow creators continue to define actions as regular strings.
The recommendation would be to use commit hashes if they will be kept up to date, and major versions otherwise.
Sensitive workflow actions should always use commit hashes.

Closes #1844 and closes #1702

---


mrgrain:
@kylelaker Keen to get your opinion on this approach. It's obviously less feature-rich than your PR, but a workable minimal solution that can easily be extended in any direction.

My thinking is the next step would be to introduce GH Actions as a proper dependency type to projen.
Write the requested & resolved version to `deps.json`, read the resolved commit hash from there and introduce an update mechanism.

test/github/actions-provider.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      if (shouldBeUnchanged.length) {
+        describe("should be unchanged for", () => {
+          test.each(shouldBeUnchanged)("uses=%s", (uses) => {
+            expect(actions.get(uses)).toBe(uses);
+          });
+        });
+      }
+    }
+  );
```
kylelaker, line 62: A very basic addition (also written directly in the review comment) that semi-follows the style:

```suggestion
  );
  
  describe('version-specific override is preferred over general override', () => {
    const actions = new GitHubActionsProvider();
    const v1 = `${actionName}@v1`;
    const v1WithHash = `${actionName}@ffffff`;
    actions.set(v1, v1WithHash);
    actions.set(actionName, replacement);
    
    test(`uses=${v1} is replaced with uses=${v1WithHash}`, () => {
      expect(actions.get(v1)).toBe(v1WithHash);
    });
    describe("should be unchanged for", () => {
      test.each([any, 'v1.1', 'v2', 'main'])("uses=%s", (uses) => {
        expect(actions.get(uses)).toBe(replacement);
      }); 
    });
  });
```

kylelaker (commented):
I think this is really nice! It also potentially supports the use case of force-upgrading with `set('actions/checkout@v2', 'actions/checkout@v3')` which is super helpful and letting workflow authors just use strings is of course ideal. This covers all my personal use-cases for this capability, at least foundationally.

One small suggestion is to allow more specific overrides to take precedence (so prefer overrides for `@v2` more than overrides without a version). This should improve usability for users who may need to have two versions across workflows for compatibility reasons.

This should be a good base to build on top of though!

src/github/github.ts
```diff
@@ -7,6 +7,38 @@ import { GithubWorkflow } from "./workflows";
...
+    this.actions.set(action, override);
+  }
+
+  /**
+   * Resolve an action name to the version that should be used,
+   * taking into account any overrides.
+   */
+  public get(action?: string): string | undefined {
+    const parts = action?.split("@", 1);
+    return this.actions.get(parts?.[0]) ?? this.actions.get(action) ?? action;
```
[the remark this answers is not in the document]
mrgrain: Yes! Great ide 😃

src/github/actions-provider.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Manage the versions used for GitHub Actions used in steps
+ */
+export class GitHubActionsProvider {
+  private actions = new Map();
```
RomainMuller, line 5: I prefer the map types to be fully specified, as I'm not left to ponder what the key & value is inferred to be from reading the rest of the code.

```suggestion
  private actions = new Map<string, string>();
```

src/github/actions-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+   * Define an override for a given action.
+   * Setting an override for a specific action version will only override the exact same action version.
+   * To override all usages of an action, do not set version constraint.
+   *
+   * @example
+   * // Force any use of `actions/checkout` to use a specific commit
+   * project.github.actions.set("actions/checkout", "actions/checkout@ac59398");
+   *
+   * // Only replace usage of `v3` with a specific commit
+   * project.github.actions.set("actions/checkout@v3", "actions/checkout@ac59398");
```
RomainMuller, line 17: What happens if 2 conflicting declarations are registered? E.g: "replace this version with @foo; then replace all versions with @bar"? If this is "unspecified behavior", then this should be called out in the docs. And if it is specified behavior, then it also should be called out in the docs.

src/github/actions-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  public set(action: string, override: string): void {
+    this.actions.set(action, override);
+  }
+
+  /**
+   * Resolve an action name to the version that should be used,
+   * taking into account any overrides.
+   */
+  public get(action?: string): string | undefined {
```
RomainMuller, line 27: What is the use-case for `action` being `undefined`?

Also, it does not make sense to omit `action` here, even if we want to allow `undefined` as a convenience to avoid `null`-checking on call sites... Worst case scenario this should be:

```suggestion
  public get(action: string | undefined): string | undefined {
```

src/github/workflows.ts
```diff
@@ -295,7 +299,7 @@ function renderJobs(
       name: step.name,
       id: step.id,
       if: step.if,
-      uses: step.uses,
+      uses: actions.get(step.uses),
```
RomainMuller, line 302: If that's the one reason why `action` is optional, this could easily be changed to:

```suggestion
      uses: step.uses && actions.get(step.uses),
```

src/github/actions-provider.ts
mrgrain: Yes mostly to avoid unnecessary null checking. I see your point in keeping the public API clean though.

PR #2444, opened 2023-02-14
fix(new): cannot create a new project if the path has a space

### LINKED ISSUE #2443: Cannot init a new project if the path has a space
I've created a new empty Github repo (`test`). The aboslute path to the repository root folder is

`/home/myUser/job repos/test`

As you can notice, the `job repos` folder has a space in its name.

When I enter in the `ŧest` folder from terminal and run this command 

`npx projen new --from react-ts`

I first read that `projen` must be install (and I accept it). Soon after it fails with this error

```
/home/myUser/job
└── (empty)


Error: Command failed: npm ls --prefix=/home/myUser/job repos/test --depth=0 --pattern projen || npm install --save-dev -f --no-package-lock --prefix=/home/myUser/job repos/test projen@latest
```

If you notice, the path is cat at `job` keyword, and the space is read as the end of the path.

mrgrain:
Strings that might contain spaces need to be quoted when passed into `exec`.
This fixes the issue when the base path that `projen new` is run from contains a space.

Fixes #2443

---


Naumel (approved):
lgtm

PR #2455, opened 2023-02-17
feat: add support for Git LFS

rix0rrr:
Git LFS needs all workflow files to be updated, so it needs to  go into projen core.

---


test/__snapshots__/projects.test.ts.snap
```diff
@@ -16,8 +16,7 @@ project.synth();"
 `;
 
 exports[`createProject creates a project from an external project type, if it's installed 1`] = `
-"const { javascript } = require(\\"projen\\");
-const { AwsCdkTypeScriptApp } = require(\\"@pepperize/projen-awscdk-app-ts\\");
+"const { AwsCdkTypeScriptApp, javascript } = require(\\"@pepperize/projen-awscdk-app-ts\\");
```
rix0rrr, line 19: This got added for me, I don't know why, but it seems wrong.

src/github/github.ts
```diff
@@ -126,6 +140,24 @@ export class GitHub extends Component {
     if (options.pullRequestLint ?? true) {
       new PullRequestLint(this, options.pullRequestLintOptions);
     }
+
+    if (options.lfsPatterns && options.lfsPatterns.length > 0) {
+      const attrs = GitAttributesFile.of(this.project);
```
mrgrain, line 145: How about using `this.gitattributes`?

src/github/github.ts
```diff
@@ -94,12 +102,18 @@ export class GitHub extends Component {
    */
   public readonly actions: GitHubActionsProvider;
 
+  /**
+   * Whether LFS is enabled on the repository
+   */
+  public readonly lfs: boolean;
```
mrgrain, line 108: Not sure about this. We don't expose these kind of boolean flags elsewhere.

src/javascript/upgrade-dependencies.ts
```diff
@@ -299,11 +299,17 @@ export class UpgradeDependencies extends Component {
       contents: workflows.JobPermission.READ,
     };
 
+    const github = GitHub.of(this.project);
```
mrgrain, line 302: `this._project.github`

src/github/task-workflow.ts
```diff
@@ -145,9 +145,13 @@ export class TaskWorkflow extends GithubWorkflow {
...
     const preCheckoutSteps = options.preCheckoutSteps ?? [];
-    const checkoutWith = options.checkoutWith
-      ? { with: options.checkoutWith }
-      : {};
+
+    const checkoutWith: Record<string, any> = {};
+    Object.assign(checkoutWith, options.checkoutWith ?? {});
+    if (this.github.lfs) {
+      checkoutWith.lfs = true;
+    }
```
mrgrain, line 153: I see why we'd want this, but now `options.checkoutWith` does not provide a way to checkout without lfs anymore, which I think is a valid use case. 🤔

src/gitattributes.ts
```diff
@@ -7,6 +7,13 @@ import { Project } from "./project";
  * @see https://git-scm.com/docs/gitattributes
  */
 export class GitAttributesFile extends FileBase {
+  /**
+   * Return the gitattributes file for the given repository
+   */
+  public static of(project: Project): GitAttributesFile | undefined {
+    return project.root.gitattributes;
+  }
+
```
mrgrain, line 16: Every project has `project.gitattributes`. Feels not required.

src/github/github.ts
```diff
@@ -61,6 +62,13 @@ export interface GitHubOptions {
    * @deprecated - use `projenCredentials`
    */
   readonly projenTokenSecret?: string;
+
+  /**
+   * File patterns to mark as stored in Git LFS
+   *
+   * @default - Git LFS disabled
+   */
+  readonly lfsPatterns?: string[];
```
mrgrain, line 71: How about we move this to `Project` as a new `gitOptions` ?

mrgrain (changes requested):
Some minor inline feedback.

Not quite sure about the API for this. `GitAttributes` are currently attached to `Project` so I think it should stay there.

For the workflows, I think an explicit flag `downloadLfs` that defaults to `true` if any lfs patterns are present would be better.

src/github/github.ts
rix0rrr: Or does it not matter?

(Well I suppose `projen` creates `.gitattributes` for all monorepo projects right now anyway)

src/github/github.ts
rix0rrr: Should it not be a `Component`?

src/javascript/upgrade-dependencies.ts
mrgrain: You'd be calling a private property from a private method. The scope is well-defined here.

src/javascript/upgrade-dependencies.ts
```diff
@@ -299,11 +299,19 @@ export class UpgradeDependencies extends Component {
       contents: workflows.JobPermission.READ,
     };
 
+    const github = this._project.github;
+    if (!github) {
+      throw new Error("UpgradeDependencies must be used in a GitHub project");
+    }
```
mrgrain, line 305: Given that this workflow requires a NodeProject, we know we always have a `project.github` available.
```suggestion
```

mrgrain (changes requested):
Almost there!

mrgrain (approved):
🚀

PR #2468, opened 2023-02-26
Mrgrain/rfc/task runtime

mrgrain:
---


.replica.yaml
```diff
@@ -0,0 +1,219 @@
+# ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
+
+tasks:
```
moltar, line 3: Why YAML?

mrgrain: No reason. This is just to simulate an external task runner. In reality this would be whatever wireit or nx needs.

PR #2472, opened 2023-03-01
feat: allow overwriting the checkout settings during build

DanielMSchmidt:
We need to do a full checkout on build so that our jsii docgen generated golang docs have the correct major version in them


DanielMSchmidt:
I did not know about the override option beforehand, fixed it using that: https://github.com/cdktf/cdktf-provider-project/pull/306

PR #2486, opened 2023-03-08
feat(github): support all crud operations for github-workflow jobs

lexfelixpost:
In my custom projects I wanted to generate standard workflows that are also dynamic based on which components is enabled. While still allowing users to overwrite these defaults in case it is not to their liking. I saw this is not possible because the job property is private and we only have add functions in the workflow class. 

I've added the getJob, updateJob(s), removeJob functions so users can do this.

---


mrgrain (changes requested):
@lexfelixpost Some minor bits/questions. And your PR description MUST include the license footer.

```
---
```

src/github/workflows.ts
```diff
@@ -137,30 +137,78 @@ export class GithubWorkflow extends Component {
...
+   */
+  public removeJob(id: string) {
+    const updatedJobs = Object.entries(this.jobs).filter(
+      ([jobId, _job]: [
+        string,
+        workflows.Job | workflows.JobCallingReusableWorkflow
+      ]) => {
+        return jobId !== id;
+      }
+    );
```
mrgrain, line 209:
```suggestion
    const updatedJobs = Object.entries(this.jobs).filter(([jobId]: [string]) => jobId !== id);
```

mrgrain: Not sure if the type on the filter parameter `[jobId]: [string]` is needed at all here.

src/github/workflows.ts
```diff
@@ -137,30 +137,78 @@ export class GithubWorkflow extends Component {
...
+    jobs: Record<string, workflows.Job | workflows.JobCallingReusableWorkflow>
+  ) {
+    verifyJobConstraints(jobs);
+
+    const newJobIds = Object.keys(jobs);
+    const updatedJobs = Object.entries(this.jobs).map(
+      ([jobId, job]: [
+        string,
+        workflows.Job | workflows.JobCallingReusableWorkflow
+      ]) => {
```
mrgrain, line 185: Just wondering the type is necessary here, if so that's all good to keep

mrgrain (changes requested):

src/github/workflows.ts
lexfelixpost: The typings are not needed removed it from here and filter function on line 198.

mrgrain (approved):
🚀

PR #2489, opened 2023-03-09
fix(tasks): Handle non-string task env vars

icj217:
Project types that do not use `.projenrc.ts` cannot guard against end users providing numbers for environment variables. Casting them to strings before storing internally

If someone tries to do either of the following:
```js
// .projenrc.js
project.tasks.addEnvironment('LINT_THRESHOLD', 9);
```
```python
# .projenrc.py
project.tasks.add_environment('LINT_THRESHOLD', 9)
```

They'll get the following error on synthesis:
```
            if (value.startsWith("$(") && value.endsWith(")")) {
                      ^

TypeError: value.startsWith is not a function
    at RunTask.resolveEnvironment (/Users/.../node_modules/projen/lib/task-runtime.js:196:23)
    at new RunTask (/Users/.../node_modules/projen/lib/task-runtime.js:73:25)
    at TaskRuntime.runTask (/Users/.../node_modules/projen/lib/task-runtime.js:52:9)
    at Pip.installDependencies (/Users/.../node_modules/projen/lib/python/pip.js:51:17)
```

This PR makes the internal task components more defensive in terms of what kinds of values it might receive from end users. Added some tests to confirm behavior as well.

---

### DIFF
--- a/src/task-runtime.ts
+++ b/src/task-runtime.ts
@@ -255,7 +255,7 @@ class RunTask {
     const output: { [name: string]: string | undefined } = {};
 
     for (const [key, value] of Object.entries(env ?? {})) {
-      if (value.startsWith("$(") && value.endsWith(")")) {
+      if (String(value).startsWith("$(") && String(value).endsWith(")")) {
         const query = value.substring(2, value.length - 1);
         const result = this.shellEval({ command: query });
         if (result.status !== 0) {
--- a/src/task.ts
+++ b/src/task.ts
@@ -1,3 +1,4 @@
+import { warn } from "./logging";
 import {
   TaskCommonOptions,
   TaskSpec,
@@ -57,6 +58,7 @@ export class Task {
     this._locked = false;
 
     this._env = props.env ?? {};
+
     this._steps = props.steps ?? [];
     this.requiredEnv = props.requiredEnv;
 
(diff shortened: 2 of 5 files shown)


mrgrain:
I like this! Your implementation needs to be a bit more clever though in what values are accepted, see failing tests.

Also can we do an explicit check and emit a warning (via `this.project.logger.warn()`) if we encounter a conversion?

icj217:
> I like this! Your implementation needs to be a bit more clever though in what values are accepted, see failing tests.
> 
> Also can we do an explicit check and emit a warning (via `this.project.logger.warn()`) if we encounter a conversion?

@mrgrain I've fixed the tests (had to exclude `undefined` from being stringified) and added the warning messages.

src/task.ts
```diff
@@ -57,6 +58,7 @@ export class Task {
     this._locked = false;
 
     this._env = props.env ?? {};
+
```
mrgrain, line 61: Should we do the same warning in the constructor?

src/task-runtime.ts
```diff
@@ -255,7 +255,7 @@ class RunTask {
     const output: { [name: string]: string | undefined } = {};
 
     for (const [key, value] of Object.entries(env ?? {})) {
-      if (value.startsWith("$(") && value.endsWith(")")) {
+      if (String(value).startsWith("$(") && String(value).endsWith(")")) {
```
mrgrain, line 258: Nice!

mrgrain (changes requested):
Looking good. One last bit is missing.

src/task.ts
icj217: Good idea. I've added that private method

mrgrain (approved):

PR #2493, opened 2023-03-10
feat(cli): allow running `projen` in a subdirectory

### LINKED ISSUE #2418: Allow running 'projen' in a subdirectory
I would be helpful if `npx projen` would be able to run up the project tree and find `.projenrc.js`/`.projenrc.ts` in a parent directory and run itself there.

Current behavior:

```
$ npx projen
Error: Cannot find module '/path/to/project/some/subdirectory/.projenrc.js'
```

fynnfluegge:
This PR adds the feature to run `projen` commands from subdirectories in a projen project.

#### Closes #2418.
---

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,4 +1,5 @@
 import { resolve } from "path";
+import * as fs from "fs-extra";
 import * as yargs from "yargs";
 import newCommand from "./cmds/new";
 import { synth } from "./synth";
@@ -8,7 +9,7 @@ import * as logging from "../logging";
 import { TaskRuntime } from "../task-runtime";
 import { getNodeMajorVersion } from "../util";
 
-const DEFAULT_RC = resolve(PROJEN_RC);
+const DEFAULT_RC = findProjenRc(process.cwd());
 
 async function main() {
   const ya = yargs;
@@ -81,3 +82,22 @@ main().catch((e) => {
   console.error(e.stack);
   process.exit(1);
 });
+
+/**
+ * Run up project tree to find .projenrc.js
+ *
+ * @param cwd current working directory
+ * @returns path to .projenrc.js or undefined if not found
+ */
+function findProjenRc(cwd: string): string | undefined {
+  if (cwd === "/") {
+    return undefined;
+  }
+
+  const projenrc = resolve(cwd, PROJEN_RC);
+  if (fs.existsSync(projenrc)) {
+    return projenrc;
+  }
+
+  return findProjenRc(resolve(cwd, ".."));
+}
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -9,13 +9,15 @@ import * as logging from "../logging";
 import { TaskRuntime } from "../task-runtime";
 import { getNodeMajorVersion } from "../util";
 
-const DEFAULT_RC = findProjenRc(process.cwd());
+const DEFAULT_RC = resolve(PROJEN_RC);
 
 async function main() {
   const ya = yargs;
   ya.command(newCommand);
 
-  const runtime = new TaskRuntime(".");
+  let pathToProjenRc = findProjenRc(process.cwd());
+
+  const runtime = new TaskRuntime(pathToProjenRc ?? ".");
   discoverTaskCommands(runtime, ya);
 
   ya.recommendCommands();
@@ -96,7 +98,7 @@ function findProjenRc(cwd: string): string | undefined {
 
   const projenrc = resolve(cwd, PROJEN_RC);
   if (fs.existsSync(projenrc)) {
-    return projenrc;
+    return cwd;
   }
 
   return findProjenRc(resolve(cwd, ".."));
--- a/src/task-runtime.ts
+++ b/src/task-runtime.ts
@@ -35,7 +35,7 @@ export class TaskRuntime {
   public readonly workdir: string;
 
   constructor(workdir: string) {
-    this.workdir = resolve(workdir);
+    this.workdir = workdir;
     const manifestPath = join(this.workdir, TaskRuntime.MANIFEST_FILE);
     this.manifest = existsSync(manifestPath)
       ? JSON.parse(readFileSync(manifestPath, "utf-8"))
--- a/test/cli.test.ts
+++ b/test/cli.test.ts
@@ -35,13 +35,18 @@ test('running "projen" for projects with a "default" task will execute it', () =
   expect(directorySnapshot(project.outdir)["bar.txt"]).toStrictEqual("foo\n");
 });
 
-test('running "projen" in a subdirectory will execute .projenrc.js in parent directory', () => {
-  const dir1 = mkdtemp();
-  const dir2 = mkdtemp({ dir: dir1 });
+test('running "projen" in a subdirectory for projects with a "default" task will execute it from parent .projenrc', () => {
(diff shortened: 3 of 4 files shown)


mrgrain (changes requested):
I have a feeling if you add a test with a `.projenrc.ts` or `.projenrc.py` file this will not work.

fynnfluegge:
> I have a feeling if you add a test with a `.projenrc.ts` or `.projenrc.py` file this will not work.

True, I think I should rather search for toplevel `.projen` directory as you mentioned before

fynnfluegge:
Hey @mrgrain I modified `findProjenRcJs` to search for `ts`, `py`, `java`, `json` files, too. The approach to search for `.projen` directory worked also but the problem was if you run `npx projen` from root in a subproject it fails, since you need to find the most upper `.projen`. Also possible to traverse through all directories and return the last found `.projen` directory but I don't like that 😄
With `yarn link` I was able to run `npx projen` from any directory in my projen project with this PR. What's eventually missing are a few more tests.

mrgrain:
> Hey @mrgrain I modified `findProjenRcJs` to search for `ts`, `py`, `java`, `json` files, too.
That is definitely and improvement! Although some projects allow you to choose any arbitrary filename. 🤔 

>The approach to search for `.projen` directory worked also but the problem was if you run `npx projen` from root in a subproject it fails, since you need to find the most upper `.projen`. Also possible to traverse through all directories and return the last found `.projen` directory but I don't like that, what's your opinion? 😄 With `yarn link` I was able to run `npx projen` from any directory in my projen project with this PR. What's eventually missing are a few more tests.

Okay, feels like we are slowly getting to the core of the problem. I suppose part of the question is: What should happen when I run projen commands in a subproject? There seems to be a difference in running projen from a subdirectory of a project and running projen from a subproject.

At the moment I can `cd subproject && npx projen build` and it would only build this subproject. I'm not sure we can just take this away. What doesn't seems to work in the subproject directory is the default task. So maybe we need to fix that?

Let me set up an example repo.

fynnfluegge:
> At the moment I can `cd subproject && npx projen build` and it would only build this subproject. I'm not sure we can just take this away. What doesn't seems to work in the subproject directory is the default task. So maybe we need to fix that?
> 
> Let me set up an example repo.
> 
> Edit: Maybe something like this needs to be the default task for subprojects [mrgrain/projen-subproject-test@6e32206](https://github.com/mrgrain/projen-subproject-test/commit/6e322062079e56a726a645122953cf972fddb4bf)

Alright, guess I got it. So what we want is, when running `npx projen` in a subproject that this command should only run in the subproject. Hence, searching for `.projen` directory is the way as you initially mentioned. 

I was just confused about the failure of the `default` task in the root of a subproject and thought this should be handled by executing it from the parent project root and I switched back to search for the `.projenrc` to fix this. But if we say this is a bug this can be fixed separately and I change my PR back to searching for `.projen` which is really straight forward. 👍

mrgrain:
> Alright, guess I got it. So what we want is, when running `npx projen` in a subproject that this command should only run in the subproject. Hence, searching for `.projen` directory is the way as you initially mentioned.
> 
> I was just confused about the failure of the `default` task in the root of a subproject and thought this should be handled by executing it from the parent project root and I switched back to search for the `.projenrc` to fix this.

I mean I got confused too. Glad you started thinking about subprojects 😅 


> But if we say this is a bug this can be fixed separately and I change my PR back to searching for `.projen` which is really straight forward. 👍 What do you think?

Yes, that sounds like a good plan to me. We might have to sequence the changes correctly, i.e. first merging the fix for the default task.

fynnfluegge:
> Yes, that sounds like a good plan to me. We might have to sequence the changes correctly, i.e. first merging the fix for the default task.

Alright, created an issue for this, let's first fix #2497 ✅
I pushed a new commit that passes the directory of `.projen` the `TaskRuntime` so that projen commands can be executed an any subdirectory. But let's see how this will work out together with the fix for the default task in subprojects.

src/cli/index.ts
```diff
@@ -35,7 +37,7 @@ async function main() {
   ya.options("debug", { type: "boolean", default: false, desc: "Debug logs" });
   ya.options("rc", {
     desc: "path to .projenrc.js file",
-    default: DEFAULT_RC,
+    default: DEFAULT_RC ?? PROJEN_RC,
```
mrgrain, line 40:
```suggestion
    default: DEFAULT_RC,
```

src/cli/index.ts
```diff
@@ -81,3 +83,20 @@ main().catch((e) => {
   console.error(e.stack);
   process.exit(1);
 });
+
+/** Run up project tree to find .projen directory
+ *
```
mrgrain, line 88:
```suggestion
/**
 * Run up project tree to find .projen directory
```

src/cli/index.ts
```diff
@@ -81,3 +83,20 @@ main().catch((e) => {
...
+ * @returns path to .projen directory or undefined if not found
+ */
+function findProjenDir(cwd: string): string | undefined {
+  if (cwd === "/") {
+    return undefined;
+  }
+
+  if (fs.existsSync(resolve(cwd, PROJEN_DIR))) {
+    return cwd;
+  }
```
mrgrain, line 99: I think technically these need to be switched around if you want to allow `/.projen`
```suggestion
  if (fs.existsSync(resolve(cwd, PROJEN_DIR))) {
    return cwd;
  }

  if (cwd === "/") {
    return undefined;
  }
```

src/task-runtime.ts
```diff
@@ -35,7 +35,7 @@ export class TaskRuntime {
   public readonly workdir: string;
 
   constructor(workdir: string) {
-    this.workdir = resolve(workdir);
+    this.workdir = workdir;
```
mrgrain, line 38: Let's keep other changes to a minimum. Who knows where else we use `TaskRuntime` with a relative path.
```suggestion
    this.workdir = resolve(workdir);
```

mrgrain (changes requested):
Also needs tests 🚀

src/task-runtime.ts
fynnfluegge: I agree 👍

src/cli/index.ts
fynnfluegge: oops, forgot to undo this

fynnfluegge:
> Also needs tests 🚀

Will add some tests asap ✅

fynnfluegge:
Added a test that runs task in subdirectory ✅🚀

mrgrain (changes requested):
I reckon we need to more test cases here:

- In a nested subdirectory, projen finds the closest `.projen/task.json` and not the root one
- If there is no `.projen/task.json` file

Also - any chance these could use real projects and subprojects instead of mocks?

fynnfluegge:
> Also - any chance these could use real projects and subprojects instead of mocks?

Sure ✅

> In a nested subdirectory, projen finds the closest .projen/task.json and not the root one

Added ✅

> If there is no .projen/task.json file

Added ✅ But not sure if I got what you mean exactly 😄 That it terminates with an error and don't run in an infinite loop for example?

mrgrain:
@fynnfluegge Thanks for the work on this! PR looks good, but I want to test this myself first. Might take a few days before I get to it.

PR #2494, opened 2023-03-10
chore(node-package): Use tasks for installing dependencies

icj217:
Spawning a task allows end users to hook into the dependency installation process and increases flexibility

This aligns node and python projects on the same dependency installation API. For node projects, had to explicitly exclude the `install` and `install:ci` tasks from being included in `package.json` scripts to prevent infinite loops.

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -20,7 +20,8 @@ import { JsonFile } from "../json";
 import { Project } from "../project";
 import { isAwsCodeArtifactRegistry } from "../release";
 import { Task } from "../task";
-import { exec, isTruthy, sorted, writeFile } from "../util";
+import { TaskRuntime } from "../task-runtime";
+import { isTruthy, sorted, writeFile } from "../util";
 
 const UNLICENSED = "UNLICENSED";
 const DEFAULT_NPM_REGISTRY_URL = "https://registry.npmjs.org/";
@@ -478,6 +479,16 @@ export class NodePackage extends Component {
    */
   public readonly lockFile: string;
 
+  /**
+   * The task for installing project dependencies (non-frozen)
+   */
+  public readonly installTask: Task;
+
+  /**
+   * The task for installing project dependencies (frozen)
+   */
+  public readonly installCiTask: Task;
+
   private readonly keywords: Set<string> = new Set();
   private readonly bin: Record<string, string> = {};
   private readonly engines: Record<string, string> = {};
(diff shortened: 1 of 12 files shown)


icj217:
Was able to resolve all but 1 of the failing tests. For some reason, when synthesizing a `TypescriptProject`, the `TaskRuntime` component can't find the chalk package. I can't think of any reason for this that makes sense, given that the `TaskRuntime` for the project being created is using the project code and `chalk` is an installed dependency. Hoping someone else w/more familiarity with the codebase might be able to shed some light on this.

```
 FAIL  test/typescript/typescript.test.ts
  ● TypeScriptProject with default settings › compiles

    TypeError: chalk.underline is not a function

      113 |     if (envlogs.length) {
      114 |       this.logDebug(
    > 115 |         chalk.gray(`${chalk.underline("env")}: ${envlogs.join(" ")}`)
          |                             ^
      116 |       );
      117 |     }
      118 |

      at new RunTask (src/task-runtime.ts:115:29)
      at TaskRuntime.runTask (src/task-runtime.ts:76:5)
      at NodePackage.installDependencies (src/javascript/node-package.ts:1450:13)
      at NodePackage.postSynthesize (src/javascript/node-package.ts:863:12)
      at TypeScriptProject.synth (src/project.ts:572:14)
      at Object.<anonymous> (test/typescript/typescript.test.ts:23:13)
```

src/javascript/node-package.ts
```diff
@@ -588,6 +599,20 @@ export class NodePackage extends Component {
...
       this.license = options.license ?? "Apache-2.0";
     }
+
+    this.installTask = project.addTask("install", {
+      description:
+        "Install project dependencies and update lockfile (non-frozen)",
+      exec: this.installAndUpdateLockfileCommand,
+    });
+    // Necessary to prevent infinite loop
+    this.removeScript(this.installTask.name);
```
mrgrain, line 609: yes that's fine. I'd rather have them filtered out in `renderScripts` if you don't mind making that change

src/javascript/node-package.ts
```diff
@@ -588,6 +599,20 @@ export class NodePackage extends Component {
...
+
+    this.installCiTask = project.addTask("install:ci", {
+      description: "Install project dependencies using frozen lockfile",
+      exec: this.installCommand,
+    });
+    this.removeScript(this.installCiTask.name);
```
mrgrain, line 615: See above

src/task-runtime.ts
```diff
@@ -4,7 +4,7 @@ import { platform } from "os";
 import { dirname, join, resolve } from "path";
 import * as path from "path";
 import { format } from "util";
-import * as chalk from "chalk";
+import { gray, underline } from "chalk";
```
mrgrain, line 7: I came across this elsewhere. I think I need to make a PR forcing this everywhere.

mrgrain (changes requested):
I love this 😍 

Just a small request

src/task-runtime.ts
icj217: Glad to hear I'm not crazy :)

mrgrain (approved):

PR #2501, opened 2023-03-13
chore(ignore-file)!: Include comments in ignore files

icj217:
Several project types and components in the codebase try to include comments in their ignore file entries, but the component was excluding them from the file. 

Comments are useful for users to understand the entries in their ignore files, especially entries that they did not add directly/first-hand.

This is a breaking change and I'm definitely open to alternative implementatations to avoid that, but there wasn't a readily apparent alternative that didn't involve duplicating all the existing methods (e.g. `addPatterns()` ==> `addPatternsAndComments()`)

---


mrgrain:
Not against this, but it feels like an issue of personal preference?!

---

The root issue for this is the same as many others: A user wants to change the implementation of a component.
But since the component is nested deeply, we can't just expose options to make it configurable. Ideally we had Inversion of Control for this, but we don't.

---

Two other ideas:
- Introduce `ignorefileOptions` at the Project level and do the legwork to pass it through to every Ignorefile instance (or maybe we can have the ignorefile get the default options from project itself)
- Delay the filtering for comments to the synth phase and do some hacking with `preSynthesize` on the project and manipulate all the Ignorefile components

icj217:
@mrgrain Great point. I ran with your first bullet point suggestion and added it as a `gitIgnoreOptions` option on the base project interface, which gets passed to the ignore file component. Seems pretty natural that way and shouldn't preclude someone using `PythonProject` or `NodeProject` from setting it.

src/ignore-file.ts
```diff
@@ -1,31 +1,56 @@
...
+
 export class IgnoreFile extends FileBase {
   private readonly _patterns = new Array<string>();
+  public readonly filterCommentLines: boolean;
+  public readonly filterEmptyLines: boolean;
 
-  constructor(project: Project, filePath: string) {
+  /**
+   *
+   * @param project The proejct to tie this file to.
```
mrgrain, line 27:
```suggestion
   * @param project The project to tie this file to.
```

src/ignore-file.ts
```diff
@@ -1,31 +1,56 @@
...
-        continue;
-      }
-
-      this.normalizePatterns(pattern);
-
+      const isComment = pattern.startsWith("#");
+      const isEmptyLine = pattern === "";
+      if (isComment && this.filterCommentLines) continue;
+      if (isEmptyLine && this.filterEmptyLines) continue;
+      if (!isComment && !isEmptyLine) this.normalizePatterns(pattern);
```
mrgrain, line 53:
```suggestion
      if (isComment && this.filterCommentLines) {
        continue;
      }
      if (isEmptyLine && this.filterEmptyLines) {
        continue;
      }
      if (!isComment && !isEmptyLine) {
        this.normalizePatterns(pattern);
      }
```

mrgrain:
@icj217 Are you aware that this change file now only update `.gitfignore` files but no other ignorefiles?

src/ignore-file.ts
```diff
@@ -1,31 +1,56 @@
...
     for (const pattern of patterns) {
-      // skip comments
-      if (pattern.startsWith("#")) {
-        continue;
-      }
-
-      this.normalizePatterns(pattern);
-
+      const isComment = pattern.startsWith("#");
+      const isEmptyLine = pattern === "";
```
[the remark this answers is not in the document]
mrgrain: I think we should trim and the `Boolean` is just for good readability

mrgrain (changes requested):
Few minor inline comments

src/ignore-file.ts
mrgrain: typo

icj217:
> Few minor inline comments

Thanks yeah I accidentally clobbered your commit 😬 re-added the changes just now. I also exposed the ignore file options for the other files (prettierignore/npmignore)

mrgrain (approved):

PR #2510, opened 2023-03-15
feat(node)!: default to node16 and remove `jsii/superchain` from workflows

### LINKED ISSUE #1065: Explicit toolchain dependency (or stop using jsii/superchain)
Currently, all JSII projects use a "kitchen sink" container image called `jsii/superchain` which bundles all the tools needed to build & package JSII projects. This is a pragmatic solution, but it's not optimal. It means, for example, that users don't have much control over their build environment and that builds & releases are quite slow since they have to pull this huge image on every run.

Projen should offer a fine-grained model for toolchain dependencies and have the various workflows synthesize based on this model. This will also allow us to offer developer-side support for tool installations (via tools like homebrew for example).

Abstractly, project types should be able to declare their toolchain dependencies via the Dependencies API. Probably via some special type of dependencies (or repurpose one of the existing types).

Then, we need some kind of a provider pattern that will allow mapping toolchain dependency to a provider-specific definition. For example, GitHub workflows will use one of the `actions/setup-xxx` actions to install tools in workflows based on the project model. We will be able to then offer a local provider that will check that a certain tool is installed locally and even control it via tools like `nvm` or `rvm`, etc.

### LINKED ISSUE #2094: projen workflows cannot run docker commands by default

mrgrain:
BREAKING CHANGE: default to node16
To use any other Node version, explicitly provide the desired version number

BREAKING CHANGE: remove `jsii/superchain` image from AwsCdkConstructLibrary workflows
Using the `jsii/superchain` image provides no tangible benefit over installing dependencies with GitHub Actions. However AWS CDK Constructs often require to run docker commands, and thus using the image forces GitHub Action to exectute commands as Docker in Docker. This is does not work in many situations and generic fixes are not reliable (see #2094 & aws/aws-cdk#8799). Additionally the existing build and package workflows had an inconsistent usage of the image, causing more problems.

To restore the old behavior, set `options.workflowContainerImage` to the desired image.

Fixes #2094
Closes #1065

---


kaizencc (approved):
Yay!

PR #2511, opened 2023-03-15
fix(github): extend PullRequestOptions to include PushOptions

JackMoseley2001:
Hello 👋 firstly, thank you for such an intuitive and easy way to manage projects!

I found this slight issue whilst trying to create a workflow trigger to only occur when a PR is opened and changes occurred in a specific folder, but noticed this wasn't currently possible due to the API available.

The docs for `PushOptions` stated that the `pull_request` event can use these properties (and they are available to use as per the GitHub docs), but `PullRequestOptions` didn't extend `PushOptions`, instead only `PullRequestTargetOptions` did. This PR updates the doc string to include `pull_request_target` and makes `PullRequestOptions` extend `PushOptions`.

---

### DIFF
--- a/src/github/workflows-model.ts
+++ b/src/github/workflows-model.ts
@@ -1002,7 +1002,7 @@ export interface ProjectColumnOptions {
 /**
  * Pull request options
  */
-export interface PullRequestOptions {
+export interface PullRequestOptions extends PushOptions {
   /**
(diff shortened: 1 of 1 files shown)


mrgrain (approved):

JackMoseley2001:
Thank you @mrgrain for approving and merging 🎉

PR #2516, opened 2023-03-17
feat(javascript): Jest config support for addSetupfile() addSetupAfterEnv()

### LINKED ISSUE #2515: Additional Support for Jest setupFiles and setupFilesAfterEnv
**Current Functionality**
1) Both of these values an be passed into the initial package configuration but cannot be added to later by components or other plugin sorts of code.

2) Also, the documentation states that the default for both is an empty array, but in face the default is that they are not defined.

**My Use Case**
I am doing code generation using projen and need to add values to both of these arrays AFTER the projen constructor in a component that supports specific types of testing ([jest-dynalite](https://www.npmjs.com/package/jest-dynalite) in my case).

**Proposed Functionality**
1) Add a helper function to allow setupFiles and setupFilesAfterEnv to be added to the jest config. This already exists for addTestMatch(), addReporter() and a few others. This is just copy-pasta.

2) Make setupFiles and setupFilesAfterEnv default to an empty array to match the documentation (this is also helpful for the helper functions to work smoothly).

I will submit a PR proposal for this change soon.

cameroncf:
This adds two new helpers so that jest's setupFiles and setupFilesAfterEnv can both be appended after projen initialization.

Example Use case:
```typescript
// init a project, I cannot add setupFiles here, for reasons.
const project new TypeScriptProject({ ... })

// later... in my plugin / component....
export class MyComponent extends Component {
    constructor(  project: TypeScriptProject ) {
      super(project);

      // configure the project for test support
      project.jest!.addSetupFile('./mySetupFile.ts')
      project.jest!.addSetupFileAfterEnv('./mySetupFile.ts')
  }
}
```

**Note to maintainers / reviewer**
I've added two tests that pass. Several snapshots changed, all look good except one.

There is an unexpected change to snapshot [test/__snapshots__/projects.test.ts.snap](https://github.com/projen/projen/compare/main...cameroncf:feature/jest-add-setup-file?expand=1#diff-c06825c87bcdea4969ddb2e9e7605abae42ff155b190718d528f8c4163ed6e40L19) that I haven't been able to determine the cause of. This may need a quick review and with input am happy to make whatever modifications might be needed.

Resolves #2515 

---


src/javascript/jest.ts
```diff
@@ -680,6 +685,8 @@ export class Jest extends Component {
       watchPathIgnorePatterns: this.watchIgnorePatterns,
       testMatch: this.testMatch,
       reporters: this.reporters,
+      setupFiles: this.setupFiles,
+      setupFilesAfterEnv: this.setupFilesAfterEnv,
```
mrgrain, line 689: Can you tweak this so that empty arrays don't get added to the config?
I know we don't do this for the other props, but now is a good time as any to start!)

package.json
```diff
@@ -159,6 +159,8 @@
         }
       ]
     ],
+    "setupFiles": [],
+    "setupFilesAfterEnv": [],
```
mrgrain, line 163: Empty stuff shouldn't show up (see other comment)

test/__snapshots__/projects.test.ts.snap
```diff
@@ -16,8 +16,7 @@ project.synth();"
 `;
 
 exports[`createProject creates a project from an external project type, if it's installed 1`] = `
-"const { javascript } = require(\\"projen\\");
-const { AwsCdkTypeScriptApp } = require(\\"@pepperize/projen-awscdk-app-ts\\");
+"const { AwsCdkTypeScriptApp, javascript } = require(\\"@pepperize/projen-awscdk-app-ts\\");
```
mrgrain, line 19: Apologies, this change needs to be reverted. I'm aware the snapshot is erroneously updated locally and I'm looking for a fix. In the mean time, we just have to not commit this change.

mrgrain (changes requested):
Looks good. Two small requests and this is good to go.

src/javascript/jest.ts
```diff
@@ -766,15 +766,15 @@ export class Jest extends Component {
   }
 
   /**
-   * Adds a a setup file to setupFiles configuration array.
+   * Adds a a setup file to Jest's setupFiles configuration, written to package.json.
```
mrgrain, line 769: Just that. The user can choose where the config goes.
```suggestion
   * Adds a a setup file to Jest's setupFiles configuration
```

mrgrain (changes requested):

src/javascript/jest.ts
cameroncf: I removed the init for this as an empty array. It's undefined by default today so I've just moved this to a conditional inside addSetupFile*() to create the array if it's undefined.

package.json
cameroncf: No problem!

test/__snapshots__/projects.test.ts.snap
cameroncf: Revered. Build currently ongoing.

mrgrain (approved):
Nice one 🚀

PR #2520, opened 2023-03-17
feat(typescript): support `importsNotUsedAsValues`

jacklesliewise:
Allows us to set `importsNotUsedAsValues` to `error` instead of the default `remove` option - from the [docs](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues):

> This might be useful if you want to ensure no values are being accidentally imported, but still make side-effect imports explicit.

---

### DIFF
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -63,6 +63,29 @@ export enum TypeScriptModuleResolution {
 }
 
 /**
+ * This flag controls how import works, there are 3 different options.
+ *
+ * @see https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues
+ */
+export enum TypeScriptImportsNotUsedAsValues {
+  /**
+   * The default behavior of dropping `import` statements which only reference types.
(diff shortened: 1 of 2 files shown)


src/javascript/typescript-config.ts
```diff
@@ -158,6 +158,14 @@ export interface TypeScriptCompilerOptions {
    */
   readonly forceConsistentCasingInFileNames?: boolean;
 
+  /**
+   * This flag controls how `import` works, because you can use `import type` to explicitly create an `import` statement which should never be emitted into JavaScript.
+   *
+   * @see https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues
+   * @default "remove"
+   */
+  readonly importsNotUsedAsValues?: "remove" | "preserve" | "error";
```
mrgrain, line 167: Nice! The type must be an enum though. See other examples in this class.

mrgrain (approved):

mrgrain (approved):

PR #2521, opened 2023-03-17
feat: Allow TS-based projen files to be used in non-TS projects

icj217:
The end user UX of writing projen source files is much better in TS (for obvious reasons) than JS, so this PR reduces the TS-based RC file component's reliance/expectation of being instantiated in a TS project.

If the RC component is instantiated in a non-TS project, it'll use `npx` to invoke `ts-node` and will generate its own `tsconfig.projen.json` file.

- chore(typescript-config): Allow tsconfig to be created outside of NodeProject projects
- feat(projenrc-ts): Allow use of TS-based projenrc files in non-TS projects
- feat(python-project): Add support for using TS-based projenrc files

---


mrgrain:
Nice work so far @icj217 

First feedback: I think I'd rather have a second generic ProjenrcTs component than endless if statements in the old one. I think that should work and be much easier to manage in future.

icj217:
> Nice work so far @icj217
> 
> First feedback: I think I'd rather have a second generic ProjenrcTs component than endless if statements in the old one. I think that should work and be much easier to manage in future.

@mrgrain Good point. Created a separate component.

src/typescript/projenrc-ts.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * A projenrc file written in TypeScript
+ *
+ * This component can be instantiated in any type of project
+ * and has no expectations around the project's main language.
+ *
+ * Requires that `npx` is available.
+ */
+export class ProjenrcTs extends Component {
```
mrgrain, line 32: Please refactor to use the new abstract `ProjenrcFileBase`. Sorry about that... 😅

mrgrain (changes requested):
This looks awesome now! 

Left two small comments. One of my refactors ended up being merged first, sorry! 😅

src/typescript/projenrc-ts.ts
icj217: Done! Nice addition.

mrgrain (approved):

PR #2522, opened 2023-03-17
chore(javascript): fix invalid arg name in JSDocs

cameroncf:
Looks like I used an argument type instead of argument name in one spot of my recent PR #2515 in the JSDocs. I may look at building a test to catch this in the future but since it's Friday, here's a quick fix for the issue I created instead.

---


mrgrain:
Haha nice! I've noticed that as well, thanks for the quick fix.

It is emitted as a warning for some reason. I agree we should just turn it into an error.

mrgrain (approved):

PR #2539, opened 2023-03-21
feat(jsii): support for setting new `jsiiVersion`

### LINKED ISSUE #2165: (jsii) Remove Package Resolutions for `@types/prettier`
In #1855 we have added package resolutions to resolve an issue with `prettier`

Once the issue has been resolved at the source, we should remove the resolutions again.

Likely depends on upgrading typescript for jsii, which depends on https://github.com/aws/aws-cdk-rfcs/pull/373

### LINKED ISSUE #2264: (jsii) Remove Package Resolutions for `@types/babel__traverse`
This change in DefinitelyTyped drops release for TS3.9 which is needed for jsii: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63359

Pinning `@types/babel__traverse` to `7.18.2` works for now.

Once we can, we should remove the pin.

RomainMuller:
Now that jsii@5.0.0 is releases, we can allow users to explicitly set their jsii version.
For newer versions, we also no longer need to pin dependencies to versions compatible with TypeScript 3.9.

Closes #2264
Closes #2165

---


src/cdk/jsii-project.ts
mrgrain, line 1: As discussed.

We need to introduce a new filed `jsiiVersion` that takes an enum of jsii version lines.
Defaults to `1.x`. We should warn users about 1.x being deprecated soon.

We would than set the versioned dependencies accordingly, as well as the correct typescript version (unless it is set explicitly).
And all compat fixes are conditional on the version.

mrgrain (changes requested):

mrgrain (approved):

PR #2552, opened 2023-03-23
feat(inventory): Add overrides flag for parameters

abdsahin:
I had some challanges to create external project types with my own default values. As an example I would like to create **MyAwsCdkTypeScriptApp** that is extending **AwsCdkTypeScriptApp** with some custom logic as well as my own opiniated default values.

I first tried to use the [Omit Utility Type](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)

```typescript
interface MyAwsCdkTypeScriptAppOptions
        extends Omit<AwsCdkTypeScriptAppOptions, "fieldIwantToOverride">
{
        /**
         * default "myowndefaultvalue"
         */
        readonly fieldIwantToOverride: string
}

class MyAwsCdkTypeScriptApp extends AwsCdkTypeScriptApp {
        ...
        constructor(options: MyAwsCdkTypeScriptAppOptions)
        ...
}
```

Unfortunetely, JSII currently does not support that. Then I tried directly adding the field in my custom interface

```typescript
interface MyAwsCdkTypeScriptAppOptions
        extends AwsCdkTypeScriptAppOptions
{
        readonly fieldIwantToOverride: string
}

class MyAwsCdkTypeScriptApp extends AwsCdkTypeScriptApp {
        ...
        constructor(options: MyAwsCdkTypeScriptAppOptions)
        ...
}
```

JSII does not complain here as long as I keep everything about the property the same, but then this time Projen complains with the following error

```
Error: duplicate option "fieldIwantToOverride" in MyAwsCdkTypeScriptAppOptions (already declared in GitHubProjectOptions)
```

This PR changes the logic that inventory.ts uses for duplicates. I added a new directive called @overrides. If projen encounters 2 parameters with the same name, instead of generating an error, it honors the one with the **@overrides** flag.

This will only be used to create external project types, therefore I don't expect to have multiple @overrides directives for a type.

---


mrgrain (changes requested):
Hi @abdsahin and thanks for the contribution! This is a very clever idea and I _really_ like the idea of using a jsdoc tag for this!

> JSII does not complain here as long as I keep everything about the property the same, but then this time Projen complains with the following error

Would you mind clarifying something for me. If you "keep everything about the property the same", why do you still need to re-define it?

> instead of generating an error, it honors the one with the @OVERRIDES flag.
> This will only be used to create external project types, therefore I don't expect to have multiple @OVERRIDES directives for a type.

We'd still need to catch this case.

---

The PR is also missing test cases.

---

Aside from the above, you might be interested in this project: https://github.com/mrgrain/jsii-extend-interface
It should provide a lot more flexibility.

abdsahin:
> Hi @abdsahin and thanks for the contribution! This is a very clever idea and I _really_ like the idea of using a jsdoc tag for this!
> 
> > JSII does not complain here as long as I keep everything about the property the same, but then this time Projen complains with the following error
> 
> Would you mind clarifying something for me. If you "keep everything about the property the same", why do you still need to re-define it?

The type of the property and whether it is optional or not cannot be changed, otherwise JSII complains about it. But what I can change is the **default** and **featured** flags. If someone from my team then create a project using the my external project type, their projenrc file will be created accordingly.

> > instead of generating an error, it honors the one with the @OVERRIDES flag.
> > This will only be used to create external project types, therefore I don't expect to have multiple @OVERRIDES directives for a type.
> 
> We'd still need to catch this case.

Already done, if there are more than 1 property with the same name and **overrides** flag, an exception is thrown.

> The PR is also missing test cases.
I tried to find some similar test cases around **featured** directive but couldn't find, but I can try to create them.

> Aside from the above, you might be interested in this project: https://github.com/mrgrain/jsii-extend-interface It should provide a lot more flexibility.

That seems quite interesting, I need to look into it more but what I see is that it is linked to a project, not sure how to use it for the same purpose.

abdsahin:
I've added unit tests as well, but at that moment, not sure if we should go forward with this pull request due to an interesting behaviour of JSII that I've just realised.

I mentioned before that JSII does not complain when the same a parameter is defined in a child interface again as long as they have the same type and required/optional properties. This is not entirely true, followings are 2 examples

The following example produces a JSII error
```typescript
export interface ProjectB { 
  readonly testVarToOverride: boolean
}

export interface ProjectA  extends ProjectB{ 
    readonly testVarToOverride: boolean
}
```

The following example does not produce an error
```typescript
export interface ProjectB { 
    readonly testVarToOverride: boolean
}

export interface ProjectAOverrides { 
  readonly testVarToOverride: boolean
}

export interface ProjectA  extends ProjectB, ProjectAOverrides { 
}
```

So the **overrides** directive can still function and provide value, but I am not sure.

On the other hand, I had a look at the [jsii-extend-interface](https://github.com/mrgrain/jsii-extend-interface) and I believe it is fits the purpose nicely.

mrgrain:
> I mentioned before that JSII does not complain when the same a parameter is defined in a child interface again as long as they have the same type and required/optional properties. This is not entirely true, followings are 2 examples

Ah that's a shame. I've checked the jsii spec, but it's not entirely clear to me what the intended behavior is here.
But based on the code you've provided, I think that one of the examples will be bug:
- Either overriding properties is okay, and it shouldn't matter if it's via a subclass or a shared parent.
- Or it is not allowed and both cases should error.

I can go and try to find out.

> On the other hand, I had a look at the [jsii-extend-interface](https://github.com/mrgrain/jsii-extend-interface) and I believe it is fits the purpose nicely.

That's great to hear! I've accepted your PR and am working on some major improvement, including a rename and not requiring to use projen.
More feedback is welcome!

mrgrain:
Okay so the behavior is intended. It's a bit of a quirky in CSharp that the first (overriding) example isn't allowed, but the second (multiple interfaces) is fine.

I really liked the idea of `@overrides`, but given this only works for niche situations and would also only apply to defaults and docs, I don't think it is worth adding. I might even confuse people.

From a jsii perspective, creating a  new truct Interface is the expected thing to do. It's a bit annoying when you only need to update a default or docs, but that's how it is and technically the only way to uphold the liskov substitution contract. I reckon using a tool like [jsii-struct-builder](https://github.com/mrgrain/jsii-struct-builder) is unfortunately the best way to go about this.

If you're okay with that, I'll close the PR.

abdsahin:
I agree, hope it can be handled in the future, maybe with aws/jsii#2314

PR #2559, opened 2023-03-27
feat(vscode): support dev container features

tHyt-lab:
See: https://containers.dev/features

---

### DIFF
--- a/src/vscode/devcontainer.ts
+++ b/src/vscode/devcontainer.ts
@@ -16,11 +16,9 @@ const DEVCONTAINER_FILE = ".devcontainer.json";
 /**
  * devcontainer features options
  */
-export class DevEnvironmentFeature {
-  public readonly featureName?: string;
-  public readonly version?: string;
-
-  private constructor() {}
+export interface DevContainerFeature {
+  readonly name: string;
+  readonly version?: string;
 }
 
 /**
@@ -34,7 +32,7 @@ export interface DevContainerOptions extends DevEnvironmentOptions {
    * An array of VSCode features that specify the features that should be
    * installed inside the container when it is created.
    */
-  readonly features?: DevEnvironmentFeature[];
+  readonly features?: DevContainerFeature[];
 }
 
 export interface IDevContainerEnvironment extends IDevEnvironment {
@@ -44,7 +42,7 @@ export interface IDevContainerEnvironment extends IDevEnvironment {
    *
    * @param features featureName and version(optional default: latest)
    */
-  addFeatures(...features: DevEnvironmentFeature[]): void;
+  addFeatures(...features: DevContainerFeature[]): void;
 }
 
 /**
@@ -59,7 +57,7 @@ export class DevContainer
   private readonly postCreateTasks: Task[];
   private readonly ports: string[];
   private readonly vscodeExtensions: string[];
-  private readonly features: DevEnvironmentFeature[];
+  private readonly features: DevContainerFeature[];
 
   /**
    * Direct access to the devcontainer configuration (escape hatch)
@@ -72,7 +70,7 @@ export class DevContainer
     this.postCreateTasks = new Array<Task>();
     this.ports = new Array<string>();
     this.vscodeExtensions = new Array<string>();
-    this.features = new Array<DevEnvironmentFeature>();
+    this.features = new Array<DevContainerFeature>();
 
     this.dockerImage = options?.dockerImage;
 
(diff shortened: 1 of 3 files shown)


src/vscode/devcontainer.ts
```diff
@@ -13,23 +13,53 @@ import { Task } from "../task";
...
 
+/**
+ * devcontainer features options
+ */
+export class DevEnvironmentFeature {
+  public readonly featureName?: string;
+  public readonly version?: string;
+
+  private constructor() {}
+}
```
mrgrain, line 24: This should just be an interface.
Also I assume the `featureName` is always required?
And since it is only for dev containers rename to `DevContainerFeature`
```suggestion
export interface DevContainerFeature {
  public readonly featureName: string;
  public readonly version?: string;
}
```

test/vscode/devcontainer.test.ts
```diff
@@ -0,0 +1,127 @@
+import { synthSnapshot, TestProject } from "../util";
+
+const DEVCONTAINER_FILE = ".devcontainer.json";
+
+test("no devcontainer configured", () => {
```
mrgrain, line 5: Thanks for adding these tests! 🎉 🚀

test/vscode/devcontainer.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  // GIVEN
+  const project = new TestProject();
+
+  // WHEN
+  project.devContainer;
+
+  // THEN
+  expect(synthSnapshot(project)).not.toHaveProperty(DEVCONTAINER_FILE);
```
mrgrain, line 13:
```suggestion

  // THEN
  expect(project.devContainer).toBe(undefined);
  expect(synthSnapshot(project)).not.toHaveProperty(DEVCONTAINER_FILE);
```

src/vscode/devcontainer.ts
```diff
@@ -13,23 +13,53 @@ import { Task } from "../task";
  */
 const DEVCONTAINER_FILE = ".devcontainer.json";
 
+/**
+ * devcontainer features options
+ */
+export class DevEnvironmentFeature {
+  public readonly featureName?: string;
```
mrgrain, line 20: `featureName` feels very verbose. Either `name` or `feature` should be enough. What do you think?

mrgrain (changes requested):

src/vscode/devcontainer.ts
```diff
@@ -13,23 +13,51 @@ import { Task } from "../task";
  */
 const DEVCONTAINER_FILE = ".devcontainer.json";
 
+/**
+ * devcontainer features options
+ */
+export interface DevContainerFeature {
+  readonly name: string;
+  readonly version?: string;
```
mrgrain, line 21: Can you add a description to each of these please?

And `@default` for the `version`. Sorry I missed this in the first round.

src/vscode/devcontainer.ts
```diff
@@ -117,4 +160,21 @@ export class DevContainer extends Component implements IDevEnvironment {
...
+
+  private renderFeatures() {
+    if (this.features.length === 0) {
+      return undefined;
+    }
+
+    return this.features.reduce<{
+      [key: string]: () => { version: string };
+    }>((pv, feature) => {
+      if (feature.name) {
```
mrgrain, line 172: Now that `feature.name` is required, we can skip the if!

src/vscode/devcontainer.ts
```diff
@@ -117,4 +160,21 @@ export class DevContainer extends Component implements IDevEnvironment {
...
+        pv[feature.name] = () => ({
+          version: feature.version ?? "latest",
+        });
```
mrgrain, line 175: I don't think wrapping this into yet another function is necessary.

mrgrain (changes requested):
Thanks for the changes. Almost, there. Just some docs missing and a couple of opportunities to simplify code.

tHyt-lab:
Thanks for confirming! It is very informative.
I will fix it after work.

src/vscode/devcontainer.ts
tHyt-lab: Added description of DevContainerFeature.

src/vscode/devcontainer.ts
tHyt-lab: Changed from Function to Object.

tHyt-lab:
Thanks for checking with me again and again! I learned a lot.
Please check back with me now that the corrections are complete.

mrgrain (approved):

PR #2561, opened 2023-03-27
fix(file): annotate generated in sub-project creates incorrect patterns

### LINKED ISSUE #2560: Files in sub-projects are not correctly annotated as generated
When creating any sub-class of `FileBase`, the file is annotated as generated using the `linguist-generated` attribute in `.gitattributes` (see #629).

This is achieved by calling the `annotateGenerated` method on the root project, while passing the path relative to the sub-project.
https://github.com/projen/projen/blob/e236e583070096e933649b35d205c257dcbc0422/src/file.ts#L98-L100

There are two possible ways to fix this issue:
1. Call `annotateGenerated` on the sub-project instead
2. Reconstruct the path relative to the root project

My preference is option 1, and I am currently working on a pull request to make this change. However, I am open to discussion.

dippi:
Closes #2560

---

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -96,7 +96,7 @@ export abstract class FileBase extends Component {
     const globPattern = `/${this.path}`;
     const committed = options.committed ?? project.commitGenerated ?? true;
     if (committed && filePath !== ".gitattributes") {
-      project.root.annotateGenerated(`/${filePath}`);
+      project.annotateGenerated(`/${filePath}`);
     }
 
     this.absolutePath = path.resolve(project.outdir, filePath);
--- a/test/subproject.test.ts
+++ b/test/subproject.test.ts
@@ -191,6 +191,30 @@ test("subprojects use root level default task", () => {
   ]);
 });
 
+test("files are annotated as generated on subproject", () => {
+  // GIVEN
+  const parent = new Project({ name: "parent" });
+  const child = new Project({
+    parent,
+    name: "child",
+    outdir: path.join("packages", "child"),
+  });
+
+  const parentAnnotateGenerated = jest.spyOn(parent, "annotateGenerated");
+  const childAnnotateGenerated = jest.spyOn(child, "annotateGenerated");
+
+  // WHEN
+  new TextFile(child, "file.txt", { lines: ["Content"] });
+
+  // THEN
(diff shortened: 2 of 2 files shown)


mrgrain (changes requested):
Sounds good. Let's add a test case and this is good to go.

dippi:
I'm not sure where the test should be added:
- to `test/subproject.test.ts`, as the issue is only triggered in subprojects
- to some (or all) test suites regarding the sub-classes of `FileBase`, but it would be spread across multiple files
- to a new test suite for `file.ts`, but `FileBase` is an abstract class so we might need to create a `TestFile` sub-class to be able to instantiate it
- to a new test suite for `github-project`, as it's the only project implementing `annotateGenerated` (in all other cases we could just test `annotateGenerated` to be called, but not the effect on the `.gitattributes` file)

Let me know!

mrgrain:
It doesn't really matter, but I probably would pick `test/subproject.test.ts`

mrgrain (changes requested):
Thanks for the test. They are not wrong, but we prefer a different style here. You can look at  at how the other tests are doing it in the same file.

But basically instead of using spys and mocks, synth the project and inspect the output files.
In your case we'd expect the nested `.gitattributes` in the subproject to include the pattern, and the `.gitattributes` at the root to not include the pattern.

dippi:
That's the issue, `annotateGenerated` is an empty function on `Project`, so there is no `.gitattributes` to inspect. That's why I was hesitant on where to write the test.
https://github.com/projen/projen/blob/7c3fc859c1d22c9d11d486fe80f1d98927fac271/src/project.ts#L536-L545

So I took inspiration from here and used `spyOn`.
https://github.com/projen/projen/blob/e236e583070096e933649b35d205c257dcbc0422/test/javascript/node-package.test.ts#L157-L162

`GitHubProject` has the only implementation, does it make sense for me to use it in this test?
https://github.com/projen/projen/blob/5ecc4165dce59e6d3f89b6e395fa64a9e28e4027/src/github/github-project.ts#L239-L250

mrgrain:
Yes you can use GitHubProject.

mrgrain (approved):
Excellent! Thanks for implementing my test requirements 🙈😅

PR #2562, opened 2023-03-27
feat(javascript): add `addScripts` to `NodeProject`

jacklesliewise:
Allows us to set multiple scripts at once, rather than needing to call `setScript` multiple times, similar to `NodeProject.addFields`

> Note: we can't call it `setScripts` due to the following error:
> ```
> error JSII5001: Methods and properties cannot have names like "setXxx": those conflict with Java property setters.
> ```

---


mrgrain (changes requested):
Let's add a test and this is good to go

mrgrain (approved):

PR #2563, opened 2023-03-27
feat(cli): validate options when creating projects from external packages

### LINKED ISSUE #2487: JSDoc @default tag supports primitive types only
Using an array @default to inject comments into my projenrc file fails as the annotation is limited to primitive types only (https://jsdoc.app/tags-default.html).

Trying this:
```
  /**
     Default dependencies that lib-projen injects
     @default ['@cdktf/provider-aws@^12','constructs@^10.1.264','cdktf@^0.15.5']
     @featured
  */
  readonly defaultDeps: string[];
```

Fails with this error:
```
Error: required option "defaultDeps" with a @default must use primitive types (string, number or boolean). type found is: {"collection":{"elementtype":{"primitive":"string"},"kind":"array"}}
```

mrgrain:
Options validation for built-in project types is handled by yargs.
When using a project type from an external package via `--from`,  yargs is bypassed for options.
This meant that previously the validation and parsing behavior was incomplete.
Refactored code so that built-in and external types share much more codepaths.

Additionally align default value type checks in the inventory with options checks, fixes #2487

---


src/inventory.ts
```diff
@@ -235,9 +235,7 @@ function toProjectType(jsii: JsiiTypes, fqn: string): ProjectType {
     typename,
     pjid,
     fqn,
-    options: discoverOptions(jsii, fqn).sort((o1, o2) =>
```
mrgrain, line 238: Turns out `discoverOptions` is already sorting them. We double sort!

RomainMuller (approved):

src/cli/cmds/new.ts
```diff
@@ -168,6 +137,51 @@ function argType(
   return option.simpleType as "string" | "boolean" | "number";
 }
 
+/**
+ * Returns the description for a given project option
+ */
+function argDesc(option: inventory.ProjectOption): string {
+  let desc = [option.docs?.replace(/\ *\.$/, "") ?? ""];
```
Naumel, line 144:
```suggestion
  // remove the suffix made up of spaces ending in an and including the full stop
  let desc = [option.docs?.replace(/\ *\.$/, "") ?? ""];
```

Naumel (approved):
added non-blocking comments

PR #2565, opened 2023-03-28
feat: add generic type parameter to `Component` class

jacklesliewise:
If we allow `Component` to take a generic type parameter that extends `Project`, it means we can let a given `Component` access properties and methods that exist on a specific `Project` type.

For example, if we know that a `Component` is specifically for a `NodeProject`, we can pass `Component<NodeProject>` so that we can access `this.project.renderWorkflowSetup`.

Currently, if we want to do this we need to create a new custom `this._project` property with the extended project type manually - when we add this generic type parameter, we can simply access `this.project` instead.

We can make this generic type parameter optional and default to `Project` so that consumers won't be affected.

---

### DIFF
--- a/src/component.ts
+++ b/src/component.ts
@@ -3,8 +3,8 @@ import { Project } from "./project";
 /**
  * Represents a project component.
  */
-export class Component {
-  constructor(public readonly project: Project) {
+export class Component<T extends Project = Project> {
+  constructor(public readonly project: T) {
     project._addComponent(this);
   }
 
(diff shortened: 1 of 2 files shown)


mrgrain:
Heads up, while I like the I need to check what that means for jsii languages.

Best scenario is the generic will just be used for TS and jsii languages have the untyped version.
Worst scenario is that it breaks the generated types and we just can't do it.

mrgrain (changes requested):
"Request changes" until I now the above.

mrgrain:
Okay I checked. Unfortunately jsii doesn't support Generics right now. It probably should have failed to compile the code here, but doesn't. 

I know this is a common nuisance, but the work around isn't too bad.

jacklesliewise:
Ah that's a shame, no worries thanks for looking into it @mrgrain 👍

PR #2567, opened 2023-03-29
fix(jest): cannot set `preset` and `globals` config when using in a typescript project

riywo:
Some of `jestConfig` properties are not able to update through constructor
because of a bug by assigning hard coded default values.

This commit fixes the bug to assign hard coded values only if the same
properties are not set.

---

### DIFF
--- a/src/typescript/typescript.ts
+++ b/src/typescript/typescript.ts
@@ -20,6 +20,7 @@ import {
   ProjenrcOptions as ProjenrcTsOptions,
   TypedocDocgen,
 } from "../typescript";
+import { deepMerge } from "../util";
 
 export interface TypeScriptProjectOptions extends NodeProjectOptions {
   /**
@@ -445,12 +446,17 @@ export class TypeScriptProject extends NodeProject {
     );
 
     // add relevant deps
-    jest.config.preset = "ts-jest";
-    jest.config.globals = {
-      "ts-jest": {
-        tsconfig: this.tsconfigDev.fileName,
+    if (!jest.config.preset) {
+      jest.config.preset = "ts-jest";
+    }
+    jest.config.globals = deepMerge([
+      {
+        "ts-jest": {
+          tsconfig: this.tsconfigDev.fileName,
+        },
       },
-    };
+      jest.config.globals,
+    ]);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/typescript/typescript.ts
```diff
@@ -429,12 +429,16 @@ export class TypeScriptProject extends NodeProject {
...
-    };
+    if (!jest.config.preset) {
+      jest.config.preset = "ts-jest";
+    }
+    if (!jest.config.globals?.["ts-jest"]?.tsconfig) {
+      jest.config.globals = {
+        "ts-jest": {
+          tsconfig: this.tsconfigDev.fileName,
+        },
+      };
```
mrgrain, line 440: The if statement and the assignment don't really match up. I reckon we should deep merge the object

mrgrain (changes requested):
thanks for the fix. Your PR needs to include the license statement footer from the PR template (or see other PRs ).

riywo:
@mrgrain Ah, sorry! I created this PR by `gh pr create --fill` so that the PR template wasn't used. Updated the description by the footer.

mrgrain:
No worries! Really should have some automation for that. Although I'm surprised `gh pr` doesn't honor the template 🤔

src/typescript/typescript.ts
riywo: Ah, right. This breaks if there are some other options on `jest.config.globals`. Let me fix it.

Do we already have some deep merge libraries or functions in this project?

mrgrain: There's a `deepMerge` in `util.ts`. I guess it works. 😅

mrgrain: @riywo I'd love to see this fix merged. Is there anything else you need from me?

riywo: Just done (including rebase). Check this change: https://github.com/projen/projen/compare/ff3a1fd619d1035cb8a170cb32e71903a7e5a554..8b6d45d1bd8a42b97f04ee42489356b7124c8a51

riywo:
@mrgrain I addressed your comment. Check this diff as well https://github.com/projen/projen/compare/ff3a1fd619d1035cb8a170cb32e71903a7e5a554..8b6d45d1bd8a42b97f04ee42489356b7124c8a51

mrgrain (approved):

riywo:
Thank you!

PR #2569, opened 2023-03-30
feat(typescript): add `disableTsconfigDev` property to not generate the `tsconfig.dev.json`

gmukul01:
Allows us to not generate the `tsconfig.dev.json` file and use the `tsconfig.json` for all the tasks.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@typescript-eslint/eslint-plugin": "^5",
     "@typescript-eslint/parser": "^5",
     "all-contributors-cli": "^6.24.0",
-    "esbuild": "^0.17.14",
+    "esbuild": "^0.17.15",
     "eslint": "^8",
     "eslint-config-prettier": "^8.8.0",
     "eslint-import-resolver-node": "^0.3.7",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,56 +10,56 @@
     "@jridgewell/gen-mapping" "^0.1.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.12.13", "@babel/code-frame@^7.18.6":
-  version "7.18.6"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.18.6.tgz#3b25d38c89600baa2dcc219edfa88a74eb2c427a"
-  integrity sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.12.13", "@babel/code-frame@^7.18.6", "@babel/code-frame@^7.21.4":
+  version "7.21.4"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.21.4.tgz#d0fa9e4413aca81f2b23b9442797bda1826edb39"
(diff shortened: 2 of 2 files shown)


src/typescript/typescript.ts
```diff
@@ -264,24 +270,28 @@ export class TypeScriptProject extends NodeProject {
...
-          exclude: ["node_modules"],
-          compilerOptions: compilerOptionDefaults,
-        },
-        options.tsconfig,
-        options.tsconfigDev
-      )
-    );
+    if (options.disableTsconfigDev && !options.disableTsconfig) {
+      this.tsconfigDev = this.tsconfig!;
+    } else {
```
mrgrain, line 275: So we are disabling it, but then we are saying it is tsconfig?! I'm confused.

mrgrain (changes requested):
Great idea! Will need some test cases and questions answered.

Should we make sure that either tsconfig or tsconfigDev is available?
What happens with .projenrc.ts task when the dev config is gone?

PS: You can already do this by doing `disableTsConfig` and changing the name of the dev config. But that's a bad API.

gmukul01:
Should we make sure that either tsconfig or tsconfigDev is available? 
Answer: Yes I have make sure about this and even if you have passed both `disableTsconfigDev` and `disableTsconfig`  then it will definitely create the `tsconfig.dev.json`.

What happens with .projenrc.ts task when the dev config is gone?
Answer: Since I am assigning the default `tsconfig.json` to the `this.tsconfigDev`, all the tasks will consider the `tsconfig.json` for all the typescript commands.

gmukul01:
@mrgrain Added the test cases also. :)

test/typescript/typescript.test.ts
```diff
@@ -237,3 +237,57 @@ test("upgrade task ignores pinned versions", () => {
...
+    const prj = new TypeScriptProject({
+      name: "test",
+      projenrcTs: true,
+      defaultReleaseBranch: "main",
+      disableTsconfig: true,
+      disableTsconfigDev: true,
+    });
+
+    const snapshot = synthSnapshot(prj);
+    expect(prj.tsconfigDev.fileName).toBe("tsconfig.dev.json");
```
mrgrain, line 286:
```suggestion
    expect(prj.tsconfigDev.fileName).toBe("tsconfig.dev.json");
    expect(snapshot["tsconfig.dev.json"]).not.toBeUndefined();
```

test/typescript/typescript.test.ts
```diff
@@ -237,3 +237,57 @@ test("upgrade task ignores pinned versions", () => {
...
+describe("tsconfigDev", () => {
+  test("uses tsconfig.dev.json by default", () => {
+    const prj = new TypeScriptProject({
+      name: "test",
+      projenrcTs: true,
+      defaultReleaseBranch: "main",
+    });
+
+    const snapshot = synthSnapshot(prj);
+    expect(prj.tsconfigDev.fileName).toBe("tsconfig.dev.json");
```
mrgrain, line 250:
```suggestion
    expect(prj.tsconfigDev.fileName).toBe("tsconfig.dev.json");
    expect(snapshot["tsconfig.json"]).not.toBeUndefined();
    expect(snapshot["tsconfig.dev.json"]).not.toBeUndefined();
```

mrgrain (changes requested):
Nice  once @gmukul01 and thanks for the test cases. I've added a couple small improvement suggestions to them.

Re `disableTsconfigDev` AND `disableTsconfig`. I think we should at least log a warning if both are disabled. But wondering if we should simply prohibit that and throw an error instead. That makes it at least explicit. What do you think? I'll have a look how we do handle this in similar situations.

Either way, the docstrings will need to mention that as well.

Edit: I had a look. We should simply throw an error in  this case. It's what we do elsewhere.
Example `throw new Error("Cannot specify both 'contents' and 'source' fields.");`

test/typescript/typescript.test.ts
gmukul01: Done

mrgrain (changes requested):
Good to go. :shipit: 🚀

PR #2572, opened 2023-03-30
feat: add `workflowContainerImage` to `Publisher`

jacklesliewise:
Currently the `workflowContainerImage` doesn't get propagated to the `package` or `release` jobs

This PR makes sure that in `jsii-project` when creating the `package-[language]` jobs we set `container` in `extraJobOptions` if present.

It also makes sure when instantiating a `Release` object we can pass it to `release_[language]` jobs.

Both of these jobs currently propagate `workflowRunsOn` to them so the precedent is there.

---

### DIFF
--- a/src/release/publisher.ts
+++ b/src/release/publisher.ts
@@ -71,6 +71,13 @@ export interface PublisherOptions {
   readonly workflowNodeVersion?: string;
 
   /**
+   * Container image to use for GitHub workflows.
+   *
+   * @default - default image
+   */
+  readonly workflowContainerImage?: string;
+
+  /**
    * Version requirement for `publib`.
    *
    * @default "latest"
@@ -145,6 +152,7 @@ export class Publisher extends Component {
   private readonly dryRun: boolean;
 
   private readonly workflowNodeVersion: string;
+  private readonly workflowContainerImage?: string;
 
   constructor(project: Project, options: PublisherOptions) {
     super(project);
@@ -157,6 +165,7 @@ export class Publisher extends Component {
     this.condition = options.condition;
     this.dryRun = options.dryRun ?? false;
     this.workflowNodeVersion = options.workflowNodeVersion ?? "16.x";
+    this.workflowContainerImage = options.workflowContainerImage;
 
     this.failureIssue = options.failureIssue ?? false;
     this.failureIssueLabel = options.failureIssueLabel ?? "failed-release";
@@ -605,6 +614,11 @@ export class Publisher extends Component {
       ];
 
       const perms = opts.permissions ?? { contents: JobPermission.READ };
+      const container = this.workflowContainerImage
+        ? {
+            image: this.workflowContainerImage,
+          }
+        : undefined;
 
       if (this.failureIssue) {
         steps.push(
@@ -644,6 +658,7 @@ export class Publisher extends Component {
           if: this.condition,
           needs: [this.buildJobId],
           runsOn: this.runsOn,
+          container,
           steps,
         },
       };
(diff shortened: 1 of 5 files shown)


mrgrain:
Hi @jacklesliewise I'm not opposed to this. But could provide a bit more details to your use case?

Especially `JsiiProject` is a very targeted use case and I'm worried all these customizations are very specific and if we change `JsiiProject` it will just start breaking people.

Also are you aware of overrides?

jacklesliewise:
Hi @mrgrain - thanks for the response!

> But could provide a bit more details to your use case?

No worries, I perhaps should have made an issue first before jumping in, I'll make sure to do that next time to provide more context and gather alternative options (although I am enjoying working in this codebase anyway even if the PR gets closed 😄) 

We have custom docker images that we would like to use via `workflowContainerImage`, however these images already set up node and perform a bunch of SRE stuff so that our workflows can use `node`, `yarn` etc out of the box without needing to use `actions/setup-node`

We want a way to stop `projen` adding the `actions/setup-node` step to the generated workflows so that we can just rely on our docker image's node setup.

From what I can gather, for all the other languages `projen` only adds the appropriate language `tool` when that language is set as a target language, however because of certain node-based CLI utilities like `publib` you need to ensure that the `node` tool is set up regardless of whether `js` is a target language or not.

> Also are you aware of overrides?

I am not, could this suffice for our usecase?

mrgrain:
> No worries, I perhaps should have made an issue first before jumping in, I'll make sure to do that next time to provide more context and gather alternative options (although I am enjoying working in this codebase anyway even if the PR gets closed 😄)

Thanks! I appreciate that very much. An issue is perfect, but even explaining the why in the PR would be a good start.


> > Also are you aware of overrides?
> 
> I am not, could this suffice for our usecase?

I think it might! While it's not perfect it is the intended way to break out of the pre-defined behavior and do what ever you like.
Docs are here: https://projen.io/escape-hatches.html

But basically it means you an do stuff like:
```ts
project.github?.tryFindWorkflow('build')?.file?.patch(JsonPatch.remove('/jobs/build/steps/2'));
```

mrgrain:
> We have custom docker images that we would like to use via `workflowContainerImage`, however these images already set up node and perform a bunch of SRE stuff so that our workflows can use `node`, `yarn` etc out of the box without needing to use `actions/setup-node`

That makes sense. We already support `workflowContainerImage` for this use case.

> We want a way to stop `projen` adding the `actions/setup-node` step to the generated workflows so that we can just rely on our docker image's node setup.

Gotcha. I believe this might already be possible with the right combination of parameters (i.e. not setting either of `workflowNodeVersion` or  `minNodeVersion`. But that's a bad API and I guess not all workflows do support it. 

This is a worthwhile improvement I reckon. We just need to get the API right.

I wonder if we could have `actions/setup-node@v3` default to not request a specific version unless it's specified. 🤔 
This would technically "run" the action, but as per the docs it should use a node version from PATH:
```
The node-version input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.
```

> From what I can gather, for all the other languages `projen` only adds the appropriate language `tool` when that language is set as a target language, however because of certain node-based CLI utilities like `publib` you need to ensure that the `node` tool is set up regardless of whether `js` is a target language or not.

Yes, you need `node` available in every job.

jacklesliewise:
Makes sense!

I think what I'll do is strip down this PR, so that we just propagate `workflowContainerImage` down to `package` and `release` jobs, and then for our usecase we'll use an override 👍

src/release/publisher.ts
```diff
@@ -605,6 +614,11 @@ export class Publisher extends Component {
       ];
 
       const perms = opts.permissions ?? { contents: JobPermission.READ };
+      const container = this.workflowContainerImage
+        ? {
+            image: this.workflowContainerImage,
+          }
+        : undefined;
```
mrgrain, line 621:
```suggestion
        ? { image: this.workflowContainerImage }
        : undefined;
```

mrgrain (approved):
Excellent, thank you!

src/release/publisher.ts
mrgrain: Just an idea for next time (but I'm not even sure if the linter would be okay with that)

PR #2598, opened 2023-04-06
feat(jsii): new `JsiiProject`s default to version `~5.0.0`

mrgrain:
Introduces a new custom doctag `@pjnew` to force an initial value at project creating time.
This will allow us to change defaults we previously couldn't because the impact would be to complicated.

---

### DIFF
--- a/src/cdk/jsii-project.ts
+++ b/src/cdk/jsii-project.ts
@@ -127,6 +127,7 @@ export interface JsiiProjectOptions extends TypeScriptProjectOptions {
    * (e.g. `~5.0.0`).
    *
    * @default "1.x"
+   * @pjnew "~5.0.0"
    */
   readonly jsiiVersion?: string;
 }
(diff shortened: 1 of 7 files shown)


test/__snapshots__/new.test.ts.snap
```diff
@@ -44,9 +44,9 @@ project = PythonProject(
     author_name=\\"My User Name\\",
     module_name=\\"my_project\\",
     name=\\"my-project\\",
-    version=\\"0.1.0\\",
     project_type=ProjectType.LIB,
     projenrc_python=True,
+    version=\\"0.1.0\\",
```
mrgrain, line 49: Strict ordering is now enforced.

src/cdk/jsii-project.ts
```diff
@@ -127,6 +127,7 @@ export interface JsiiProjectOptions extends TypeScriptProjectOptions {
    * (e.g. `~5.0.0`).
    *
    * @default "1.x"
+   * @pjnew "~5.0.0"
```
mrgrain, line 130: `@pjnew` is modeled after the existing custom tag `@pjid`

Naumel (approved):

PR #2603, opened 2023-04-07
fix(node): typesVersions is added by jsii  and removed by projen

mrgrain:
`jsii@5.x` introduced a compatibility feature that updates the `typesVersions` field in `package.json`.
If the field is added is dynamic and depends on the package. A future version of jsii might also not require compat and remove the field again.
Previously projen would remove the contents of this field at synth time.
With this change existing content is kept as is and overrides can be added if required.

Expose `file` property on `NodePackage` as it is practice with other Components that are wrapping a File component.

Additionally this change does some cleanup on how other fields are managed.
Particularly it removes the usage of `addOverride` from the Component as this might interfere with user overrides.

Finally the use of `NodeProject.hasScript()` has been deprecated. This was an oversight as it is simply a proxy to an already deprecated method.
Also deprecates `NodePackageOptions.scripts` which has an odd side effect of adding tasks to the project. This is also a left-over from when npm scripts and tasks were strongly coupled.

---


src/javascript/node-package.ts
```diff
@@ -910,7 +915,7 @@ export class NodePackage extends Component {
       npmRegistryUrl = `https://${options.npmRegistry}`;
     }
 
-    const npmr = urlparse(npmRegistryUrl ?? DEFAULT_NPM_REGISTRY_URL);
+    const npmr = new URL(npmRegistryUrl ?? DEFAULT_NPM_REGISTRY_URL);
```
mrgrain, line 918: Replacing a deprecated node API.

test/javascript/node-package.test.ts
```diff
@@ -434,7 +434,7 @@ test("local dependencies can be specified using 'file:' prefix", () => {
 
 [NodePackageManager.YARN, NodePackageManager.YARN2].forEach(
   (packageManager) => {
-    test("yarn resolutions", () => {
+    test(`${packageManager} resolutions`, () => {
```
mrgrain, line 437: make tests names unique

Naumel (approved):

abdsahin:
This update breaks my CI/CD process. I created my own custom JSII Project type and I also use it inside the .projenrc.ts file as my project. Now I started to receive 

```
[2023-04-17T11:43:11.118] [ERROR] jsii/compiler - Compilation errors prevented the JSII assembly from being created
src/awscdk/awscdk-app-py.ts:1:43 - error TS2307: Cannot find module 'projen' or its corresponding type declarations.

1 import { awscdk, TextFile, release } from "projen";
```

Any thoughts around why? Followings are my dependencies

```JSON
  /* project dependencies */
  deps: [],
  peerDeps: [`projen@~0.71.5`],
  devDeps: ["@mrgrain/jsii-struct-builder@~0.3.0"],
```

mrgrain:
> This update breaks my CI/CD process. I created my own custom JSII Project type and I also use it inside the .projenrc.ts file as my project. Now I started to receive

I'm investigating. In the meantime, if you can you might want to try upgrading to `jsiiVersion: '5.0.x'` if you can.

mrgrain:
We've identified the issue and [this jsii PR](https://github.com/aws/jsii-compiler/pull/102) (plus a new release of projen)  should fix it.

PR #2607, opened 2023-04-13
feat(typescript): TypeScript TSConfig 5.0+ parameters, support for TSConfig extends.

BradenM:
This PR brings in the new parameters offered in [TypeScript 5.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html) and implements support for the `extends` field.

The logic for the `extends` maintains backwards compatibility for:

- Existing projects by not writing it if undefined / empty.
- Typescript <5.0 by writing a single relative path if/when only a singular extension is added.

While also allowing for the newly introduced support for multiple bases via an array.

##### Overview

1.  Support for the `bundler` module resolution strategy, which mimics the resolution patterns of modern bundlers.
2.  Addition of new TypeScript 5.0 configuration options and handling of deprecated options:
    - `allowImportingTsExtensions`: Allows TypeScript files to import each other with TypeScript-specific extensions (`.ts`, `.mts`, `.tsx`).
    - `allowArbitraryExtensions`: Suppresses arbitrary extension import errors with the assumption that a bundler will handle them.
    - `customConditions`: Specifies a list of additional conditions that should succeed when TypeScript resolves from an `exports` or `imports` field of a `package.json`.
    - `verbatimModuleSyntax`: Simplifies TypeScript's handling of import/export `type` modifiers.
    - `resolvePackageJsonExports`: Forces TypeScript to consult the `exports` field of `package.json` files if it ever reads from a package in `node_modules`.
    - `resolvePackageJsonImports`: Forces TypeScript to consult the `imports` field of `package.json` when performing a lookup that begins with `#` from a file that has a `package.json` as an ancestor.
3.  Implementation of the `extends` field in TypeScript configurations, allowing configurations to inherit settings from other configurations using a string, a list of strings, or `TypescriptConfig` instances.
4.  Unit tests to verify the extends functionality.

---

### DIFF
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -1,11 +1,10 @@
+import { readFileSync } from "fs";
 import * as path from "path";
 import * as semver from "semver";
 import { NodeProject } from ".";
-import { tryFindLocalBin } from "./util";
 import { Component } from "../component";
 import { JsonFile } from "../json";
 import { Project } from "../project";
-import { execOrUndefined } from "../util";
 
 export interface TypescriptConfigOptions {
   /**
@@ -632,16 +631,16 @@ export class TypescriptConfig extends Component {
     const tsVersion = semver.coerce(ts?.version, { loose: true });
     // early exit if we found dependency.
     if (tsVersion) return tsVersion;
-    // else, try looking for tsc executable.
-    const localTsc =
-      tryFindLocalBin("tsc", { cwd: this.project.outdir }) ?? "tsc";
-    const tscVersionStdout = execOrUndefined(localTsc, {
-      cwd: this.project.outdir,
-    });
-    if (!tscVersionStdout) return;
-    // tsc --version gives: "Version X.X.X"
-    const version = tscVersionStdout.match(/\d+(\.\d+)*/)?.[0];
-    return semver.coerce(version, { loose: true });
+    try {
+      const tsManifest = require.resolve("typescript/package.json", {
+        paths: [this.project.outdir],
+      });
+      const manifest = JSON.parse(readFileSync(tsManifest).toString());
+      return semver.coerce(manifest.version, { loose: true });
+    } catch (e) {
+      // no typescript found anywhere in resolution path.
+      return undefined;
+    }
   }
 
   /**
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -655,7 +655,7 @@ export class TypescriptConfig extends Component {
     if (!tsVersion) return;
     if (tsVersion.major < 5) {
       this.project.logger.warn(
-        "TypeScript < 5.0.0 is can only extend from a single base config.",
+        "TypeScript < 5.0.0 can only extend from a single base config.",
         `TypeScript Version: ${tsVersion.format()}`,
         `File: ${this.file.absolutePath}`,
         `Extends: ${this.extends}`
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -136,6 +136,22 @@ export interface TypeScriptCompilerOptions {
   readonly allowJs?: boolean;
 
   /**
+   * Allows TypeScript files to import each other with TypeScript-specific extensions (`.ts`, `.mts`, `.tsx`).
+   * Requires `noEmit` or `emitDeclarationOnly`.
+   *
+   * @default undefined
+   */
+  readonly allowImportingTsExtensions?: boolean;
+
+  /**
+   * Suppress arbitrary extension import errors with the assumption that a bundler will be handling it.
+   *
+   * @see https://www.typescriptlang.org/tsconfig#allowArbitraryExtensions
+   * @default undefined
+   */
+  readonly allowArbitraryExtensions?: boolean;
+
+  /**
    * Ensures that your files are parsed in the ECMAScript strict mode, and emit “use strict”
    * for each source file.
    *
@@ -154,6 +170,14 @@ export interface TypeScriptCompilerOptions {
    *
    */
   readonly declaration?: boolean;
+
+  /**
+   * List of additional conditions that should succeed when TypeScript resolves from an `exports` or `imports` field of a `package.json`.
+   *
+   * @see https://www.typescriptlang.org/tsconfig#customConditions
+   * @default undefined
+   */
+  readonly customConditions?: string[];
 
   /**
    * Emit __importStar and __importDefault helpers for runtime babel
@@ -190,8 +214,17 @@ export interface TypeScriptCompilerOptions {
   readonly forceConsistentCasingInFileNames?: boolean;
 
   /**
+   * Simplifies TypeScript's handling of import/export `type` modifiers.
+   *
+   * @see https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax
+   * @default undefined
+   */
+  readonly verbatimModuleSyntax?: string;
+
+  /**
    * This flag works because you can use `import type` to explicitly create an `import` statement which should never be emitted into JavaScript.
    *
+   * @deprecated For TypeScript 5.0+ use `verbatimModuleSyntax` instead.
    * @see https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues
    * @default "remove"
    */
@@ -413,6 +446,20 @@ export interface TypeScriptCompilerOptions {
    * @default true
    */
   readonly stripInternal?: boolean;
+
+  /**
+   * Forces TypeScript to consult the `exports` field of `package.json` files if it ever reads from a package in `node_modules`.
+   *
+   * @default true
+   */
+  readonly resolvePackageJsonExports?: boolean;
+
+  /**
+   * Forces TypeScript to consult the `imports` field of `package.json` when performing a lookup that begins with `#` from a file that has a `package.json` as an ancestor.
+   *
+   * @default undefined
+   */
+  readonly resolvePackageJsonImports?: boolean;
 
   /**
    * Modern browsers support all ES6 features, so ES6 is a good choice. You might choose to set
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -1,9 +1,11 @@
 import * as path from "path";
 import * as semver from "semver";
 import { NodeProject } from ".";
+import { tryFindLocalBin } from "./util";
 import { Component } from "../component";
 import { JsonFile } from "../json";
 import { Project } from "../project";
+import { execOrUndefined } from "../util";
 
 export interface TypescriptConfigOptions {
   /**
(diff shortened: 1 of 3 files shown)


src/javascript/typescript-config.ts
```diff
@@ -8,6 +9,12 @@ export interface TypescriptConfigOptions {
    * @default "tsconfig.json"
    */
   readonly fileName?: string;
+
+  /**
+   * Path or list of paths (TypeScript 5.0+) to base configuration(s) to inherit from.
+   */
+  readonly extends?: string | string[] | TypescriptConfig | TypescriptConfig[];
```
mrgrain, line 16: Unfortunately union types aren't great in other jsii languages, so we avoid them.

Instead we need a class with a bunch of static methods:

`fromPaths(paths: string[])`
`fromTypescriptConfigs(configs: TypescriptConfig[])`

Give it  a private constructor and a public `toJSON()` method to render.

src/javascript/typescript-config.ts
```diff
@@ -477,6 +547,24 @@ export class TypescriptConfig extends Component {
...
+
+    if (options.extends) {
+      const values = Array.isArray(options.extends)
+        ? options.extends
+        : [options.extends];
+      values.map((value) => this.addExtends(value));
+    }
+  }
+
+  public addExtends(value: string | TypescriptConfig) {
```
mrgrain, line 559:
```suggestion
  public addExtends(value: string | TypescriptConfig) {
```
Same as above re union type.

mrgrain (changes requested):
Nice one, thank you!

Just a small change required with regards to the union type.

src/javascript/typescript-config.ts
BradenM: > Unfortunately union types aren't great in other jsii languages, so we avoid them.
Ah, interesting. Can definitely where that could get complicated quickly. Good to know, thanks.

Have added a `TypescriptConfigExtends` to handle this :+1:

mrgrain:
FYI. The build failure is unrelated. Working on a fix, but might take some days as I'm on leave.

BradenM:
@mrgrain Was just about to comment asking if that indeed the case. :+1:

No rush -- please do enjoy your leave :smile:

src/javascript/typescript-config.ts
```diff
@@ -448,7 +509,41 @@ export interface TypeScriptCompilerOptions {
   readonly paths?: { [key: string]: string[] };
 }
 
+/**
+ * Container for `TypescriptConfig` `tsconfig.json` base configuration(s).
+ * Extending from more than one base config file requires TypeScript 5.0+.
```
mrgrain, line 514: Now that I think about it more, should we keep additional helper methods `fromPath()` and `fromTypescriptConfig()` that only take a single item. That way we could clearly mark them as TS4.X only.

src/javascript/typescript-config.ts
```diff
@@ -448,7 +509,41 @@ export interface TypeScriptCompilerOptions {
...
+ */
+export class TypescriptConfigExtends {
+  /**
+   * Factory for creation from array of file paths.
```
mrgrain, line 518: Depending on my other comment re TS4.x, this needs some explanation on what is allowed for which TS version.

src/javascript/typescript-config.ts
```diff
@@ -448,7 +509,41 @@ export interface TypeScriptCompilerOptions {
...
+   * @param paths Absolute or relative paths to base `tsconfig.json` files.
+   */
+  public static fromPaths(paths: string[]) {
+    return new TypescriptConfigExtends(paths);
+  }
+
+  /**
+   * Factory for creation from array of other `TypescriptConfig` instances.
```
mrgrain, line 526: Depending on my other comment re TS4.x, this needs some explanation on what is allowed for which TS version.

src/javascript/typescript-config.ts
```diff
@@ -182,9 +220,18 @@ export interface TypeScriptCompilerOptions {
...
+   *
+   * @see https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax
+   * @default undefined
+   */
+  readonly verbatimModuleSyntax?: boolean;
+
   /**
    * This flag works because you can use `import type` to explicitly create an `import` statement which should never be emitted into JavaScript.
    *
+   * @deprecated For TypeScript 5.0+ use `verbatimModuleSyntax` instead.
```
mrgrain, line 234: Let's turn this into a regular docstring remark that explains the situation. TS4.x is still a very valid choice at the moment and this might confuse people.

mrgrain (changes requested):
This looks good now, just a couple of docstrings that need clarifications.

Also, what do you think in regards of having separate `TypescriptConfigExtends` helpers for TS4.x and TS5.x?

src/javascript/typescript-config.ts
BradenM: The reason I went with `deprecated` originally was due to the language being used in the typescript docs (see: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#deprecations-and-default-changes)

That said, I do see where you are coming from and find myself agreeing :+1: 
Will update to provide more info.

mrgrain: Yes, it's tricky because ideally we would have a different interface per TS version but that's not very maintainable.
Also TS apparently never deprecates old versions, like you've correctly pointed out they have deprecated the use of certain flags with certain versions. I'm not sure what the best solution is here. Maybe `@deprecated` once TS5.5 is released.

src/javascript/typescript-config.ts
BradenM: This and related did cross my mind, but I wasn't too sure how 'strict' to be here or the level of introspection that should occur (i.e, looking through dependencies to validate the params against a specified typescript version and/or finding an installed version.)

If such introspection is already exposed somewhere / something you would consider, then I think throwing an error/warning when trying to use multiple extends on sub 5.0 would be a bit cleaner for the 'end user' than the possible confusion of the singular/plural methods.

Bit on the fence on that one though...

src/javascript/typescript-config.ts
BradenM: > Maybe @deprecated once TS5.5 is released.
Agreed. This makes the most sense.

src/javascript/typescript-config.ts
mrgrain: We do introspection in some places, but I'm not aware of that being setup for TypeScript yet. I agree it would be better, but likely has its own implications/challenges (the main one being that its possible to not specify and explicit version in code).

Alternatively we could play around with having a separate class, e.g. `TypescriptConfigV4Extend` to make it easier for users. But again this will have it's challenges to make jsii happy and avoid union types.

src/javascript/typescript-config.ts
```diff
@@ -479,11 +595,89 @@ export class TypescriptConfig extends Component {
...
+    if (!ts || !tsVersion) return;
+    if (tsVersion.major < 5) {
+      this.project.logger.warn(
+        "TypeScript >= 5.0.0 is required to extends from more than one base config.",
+        `TypeScript Version: ${ts.version}`,
+        `File: ${this.file.absolutePath}`,
+        `Extends: ${this.extends}`
+      );
+    }
+  }
```
BradenM, line 643: @mrgrain
What do you think about this non-strict validation instead of the extra methods and/or classes?

One alternative consideration I had here was throwing this message as an error and adding a `disableValidation` field to the config, but I felt that this would get the point across well enough without having to add more config.

BradenM:
@mrgrain The API docs need to be updated after [`aeb83f7` (#2607)](https://github.com/projen/projen/pull/2607/commits/aeb83f73ad107804826c5fefdf303ebca370cf30)

But I had just rebased before that commit, and now they don't seem to want to generate anymore... did something break with docgen recently?

src/javascript/typescript-config.ts
```diff
@@ -479,11 +595,89 @@ export class TypescriptConfig extends Component {
...
+  private validateExtends() {
+    // accept no extends or singular extends.
+    if (this.extends.length <= 1) return;
+    const ts = this.project.deps.tryGetDependency("typescript");
+    const tsVersion = semver.coerce(ts?.version, { loose: true });
+    // we assume latest by default, so return if not set (or not valid).
+    if (!ts || !tsVersion) return;
+    if (tsVersion.major < 5) {
+      this.project.logger.warn(
+        "TypeScript >= 5.0.0 is required to extends from more than one base config.",
```
mrgrain, line 637: The error sounds like it's the wrong way around:
```suggestion
        "TypeScript < 5.0.0 is can only extend from a single base config.",
```

mrgrain:
> I had just rebased before making that commit, and now they don't seem to want to generate anymore (had no issues in the commit right prior to the rebase)... did something break with docgen recently?

I'm not aware of anything, but that doesn't mean much.

src/javascript/typescript-config.ts
BradenM: I have added a fallback to this logic to look for a project local `tsc` executable and retrieve the current version from `tsc --version` output.

src/javascript/typescript-config.ts
```diff
@@ -479,11 +597,108 @@ export class TypescriptConfig extends Component {
...
+   * Attempt to resolve project TypeScript version.
+   * @private
+   */
+  private resolveTypeScriptVersion() {
+    const ts = this.project.deps.tryGetDependency("typescript");
+    const tsVersion = semver.coerce(ts?.version, { loose: true });
+    // early exit if we found dependency.
+    if (tsVersion) return tsVersion;
+    // else, try looking for tsc executable.
+    const localTsc =
```
mrgrain, line 636: Instead of executing tsc, let's copy the mechanism from node-package
https://github.com/projen/projen/blob/3c28310404f113ff71ccfc5e52e6267f10c716e6/src/javascript/node-package.ts#L1278

BradenM: :man_facepalming: 
Honestly, not sure where my head was when I decided to jump right to parsing tsc output…
Thanks for pointing out what should have been obvious from the start haha :smile:

Will push changes shortly.

mrgrain: It was a good idea!

mrgrain: Tbh, we probably should have a generic method for getting the actual version of a dependency.

BradenM: @mrgrain I was thinking the same thing while updating this.
Decided to split it out into a separate PR:
https://github.com/projen/projen/pull/2634

If #2634 looks good to you, then I will update this PR to utilize the new methods.

BradenM:
Blocked by #2634
Will need to be rebased and updated (ref: https://github.com/projen/projen/pull/2607#discussion_r1178682242) one #2634 is merged.

src/javascript/typescript-config.ts
```diff
@@ -468,6 +576,14 @@ export class TypescriptConfig extends Component {
...
       allowComments: true,
       obj: {
+        extends: () =>
+          // use string value for singular extension for TS<5.0 support;
+          // omit if no extensions.
+          this.extends.length === 1
+            ? this.extends[0]
+            : this.extends.length
+            ? this.extends
+            : undefined,
```
mrgrain, line 586: One last request: Turn this into a private helper method `renderExtends()`
```suggestion
        extends: () => this.renderExtends(),
```

mrgrain (changes requested):

src/javascript/typescript-config.ts
BradenM: :+1: I wasn't the biggest fan of the nested ternary anyways :smile:

mrgrain: Yes it's not... great. 😅

mrgrain (approved):

PR #2608, opened 2023-04-14
fix: remove deprecated zlib library

gmukul01:
Removing the deprecated `zlib`, as this functionality was folded into node.js core:
https://nodejs.org/docs/latest-v16.x/api/zlib.html#

Also the version being used has broken changes. 
https://github.com/kkaefer/DEPRECATED-node-zlib/issues/18

---


gmukul01:
Its really strange because it is not breaking on local
image

mrgrain:
We run the build with Node 14. Maybe there a some version compatibility issues with zlib the only got resolved in recent versions?

mrgrain:
Oh it might also be due to this: https://github.com/projen/projen/actions/runs/4653227759/jobs/8233918071

gmukul01:
Yes its failing on 16.13 but passing on 16.14, are we thinking to move to >=16.14 in future ?

mrgrain:
> Yes its failing on 16.13 but passing on 16.14, are we thinking to move to >=16.14 in future ?

Ah cool. In that case I think we can just change it here: https://github.com/projen/projen/blob/main/.projenrc.js#L67
I would welcome a PR for that if you want to speed things up.  😃

gmukul01:
Awesome raised [this](https://github.com/projen/projen/pull/2610) PR

mrgrain (approved):

PR #2610, opened 2023-04-15
chore: use node 16.14 version for github workflows

gmukul01:
Use Node 16.14 version for github workflows.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16.20.0
+          node-version: 16.14.0
       - name: Install dependencies
         run: yarn install --check-files
       - name: build
@@ -93,7 +93,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v3
         with:
-          node-version: 16.20.0
+          node-version: 16.14.0
       - name: Download build artifacts
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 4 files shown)


mrgrain (dismissed):

mrgrain:
Great. More weird failures. 🤷🏻‍♂️  /sarcasm off

Thanks for raising this though! Like before, I'm not sure about the cause of this.
`16.20` comes with quite the bump for the npm version as well - maybe we try the minimum and go to `16.14`?

BradenM:
@mrgrain @gmukul01 
Was being nosy and watching this PR go through :smile: 

Poking through, I noticed that the very first run of this PR somehow picked up on a cache:https://github.com/projen/projen/actions/runs/4709923731/jobs/8353278544#step:3:10

Considering this a node version change, thought it was a bit weird. Would have assumed no cache (or at most an invalidation). It may be worth pruning whichever github actions cache it seems to be picking up on (if this is indeed something unexpected).

gmukul01:
Yeah the errors are little weird. I have updated to `16.14`

mrgrain:
> @mrgrain @gmukul01 
> Was being nosy and watching this PR go through :smile: 
> 
> Poking through, I noticed that the very first run of this PR somehow picked up on a cache:
> https://github.com/projen/projen/actions/runs/4709923731/jobs/8353278544#step:3:10
> 
> Considering this a node version change, thought it was a bit weird. Would have assumed no cache (or at most an invalidation). It may be worth pruning whichever github actions cache it seems to be picking up on (if this is indeed something unexpected).
> 
> 

Good idea! Though if I interpret this correctly, the line you've linked is GH's node binary cache that's part of the runner. Basically they have a the latest version pre-installed and other versions have to be downloaded.

BradenM:
Ah, good point. That would make a lot more sense. :+1:

mrgrain (approved):
Yay, this worked!

PR #2614, opened 2023-04-17
feat(node): allow enforcement of licenses used by dependencies

mrgrain:
In some organizations only certain licenses are allowed to be used or a package want's to ensure it only depends on certain licenses.

This change adds a new feature to check the used licenses.
It will fail the build if any prohibited or not allowed license is encountered.

---


rix0rrr (approved):

mrgrain:
Example: https://github.com/projen/projen/pull/3194

PR #2621, opened 2023-04-19
fix(release): allow multiple tags on the same commit for separate branches

abdsahin:
Example use case: after testing a pre-release, if all good with the pre-release, it should be promotable to the next stage with the same commit.

---

### DIFF
--- a/src/release/bump-version.ts
+++ b/src/release/bump-version.ts
@@ -1,7 +1,7 @@
 import { promises as fs, existsSync } from "fs";
 import { dirname, join } from "path";
 import { Config } from "conventional-changelog-config-spec";
-import { compare } from "semver";
+import { compare, prerelease as getPreRelease } from "semver";
 import * as logging from "../logging";
 import { exec, execCapture } from "../util";
 
@@ -128,7 +128,15 @@ export async function bump(cwd: string, options: BumpOptions) {
 
   let skipBump = false;
 
-  if (currentTags.includes(latestTag)) {
+  const latestTagPrerelease = getPreRelease(latestTag);
+  // returns true if the latest tag on this commit is on the same release branch
+  const latestTagSamePrerelease =
+    (prerelease &&
+      latestTagPrerelease &&
+      latestTagPrerelease[0] === prerelease) ||
+    (!prerelease && !latestTagPrerelease);
+
+  if (currentTags.includes(latestTag) && latestTagSamePrerelease) {
     logging.info("Skipping bump...");
     skipBump = true;
 
@@ -166,7 +174,7 @@ export async function bump(cwd: string, options: BumpOptions) {
   exec(cmd.join(" "), { cwd });
 
   // add the tag back if it was previously removed
-  if (currentTags.includes(latestTag)) {
+  if (currentTags.includes(latestTag) && latestTagSamePrerelease) {
     exec(`git tag ${latestTag}`, { cwd });
   }
 
@@ -309,7 +317,11 @@ function determineLatestTag(options: LatestTagOptions): {
     const releaseTags = tags.filter((x) =>
       new RegExp(`^v([0-9]+)\.([0-9]+)\.([0-9]+)$`).test(x)
     );
-    if (releaseTags && compare(releaseTags[0], prereleaseTags[0]) === 1) {
+    if (
+      releaseTags &&
+      releaseTags.length > 0 &&
+      compare(releaseTags[0], prereleaseTags[0]) === 1
+    ) {
       tags = releaseTags;
     } else {
       tags = prereleaseTags;
--- a/test/release/bump.test.ts
+++ b/test/release/bump.test.ts
@@ -283,6 +283,68 @@ test("second prerelease after the first prerelease", async () => {
   expect(result.version).toEqual("1.0.1-beta.1");
 });
 
+test("same commit same prerelease on beta branch", async () => {
+  const result = await testBump({
+    options: {
+      prerelease: "beta",
+      majorVersion: 1,
+    },
+    commits: [
+      { message: "first version", tag: "v1.0.0" },
+      { message: "fix(test): testing", tag: "v1.0.1-beta.0" },
+      { message: "fix(test): testing2", tag: "v1.0.1-beta.1" },
+    ],
+  });
+
+  expect(result.version).toEqual("1.0.1-beta.1");
+});
+
+test("same commit prerelease on regular branch", async () => {
+  const result = await testBump({
+    options: {
+      majorVersion: 1,
+    },
+    commits: [
+      { message: "first version", tag: "v1.0.0" },
+      { message: "fix(test): testing", tag: "v1.0.1-beta.0" },
+      { message: "fix(test): testing2", tag: "v1.0.1-beta.1" },
+    ],
+  });
+
+  expect(result.version).toEqual("1.0.1");
+});
+
+test("same commit different prereleases", async () => {
+  const result = await testBump({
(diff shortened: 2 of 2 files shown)


mrgrain:
Hi @abdsahin Can you confirm the scenario with you. Do I understand this correctly, with this PR merged, if the release workflow (or task) is run multiple times in a row a new version is released everytime?

I'm also a bit confused how one would achieve the change without a new commit, since you have to add/remove/modify the `prerelease` property.

abdsahin:
Following is from the content of the .projenrc.ts
```typescript
defaultReleaseBranch: "main",
prerelease: "pre", //main branch is designed to be a pre-release of the major release
releaseBranches: {
  "release/0.x": {
    majorVersion: 0,
  },
},
```

The use case is our developers are creating features branches and merges to main. Ones all tests are done, we create another merge request to release/0.x.

Now, assume the merge strategy is fast-forward  without "squash commit" ,so no merge commits are created. release/0.x and main has the same commits.

I am trying to create a re-usable library, cannot be certain which teams will use merge commits and which not, therefore I thought this might be an improvement, but happy to hear your feedback.

mrgrain:
Thanks for the explanation @abdsahin ! This is very cool usage of the release branch feature. It didn't occur to me yet that you could do it that way! 🤯 

I think my main concern here is that current users might be working on the assumptions that releases are idempotent. By that I mean that running the release task/workflow on the same commit they will get exactly one version. 

It seems (but I'm not 100% sure) that this PR changes this behavior. So I'd like to understand better if that's a desirable change to the default behavior. For example if the re-releasing can only happen under very specific circumstances, we _might_ be okay merging it as a default. If not, we will need an option for it.

If any default behavior is changed, we should run campaign before applying it.

abdsahin:
For the above configuration I mentioned, if there is a **v1.0.0-pre.0** release on main branch and if you apply the same commit on **release/0.x** branch, you try to release **v1.0.0-pre.0** again on the wrong branch.

The PR changes only this behaviour by strictly checking the current branch.

Probably everyone is using merge commits and don't see that issue, so I can enforce using merge commits as well.

abdsahin:
@mrgrain , how about adding a new property named **branchIdenpotentRelease**( I am open to suggestions for the name), set it as **false** by default and if true then we can activate the changes in this PR?

I can update the PR accordingly.

mrgrain (changes requested):
Yeah I think a property would definitely be the safest option. But tbh I'm still trying to understand how this change is working. It might be okay to be released as a default.

src/release/bump-version.ts
```diff
@@ -128,7 +128,15 @@ export async function bump(cwd: string, options: BumpOptions) {
...
   let skipBump = false;
 
-  if (currentTags.includes(latestTag)) {
+  const latestTagPrerelease = getPreRelease(latestTag);
+  // returns true if the latest tag on this commit is on the same release branch
+  const latestTagSamePrerelease =
+    (prerelease &&
+      latestTagPrerelease &&
+      latestTagPrerelease[0] === prerelease) ||
+    (!prerelease && !latestTagPrerelease);
```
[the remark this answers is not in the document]
abdsahin: I tried to explain it a bit better.

abdsahin:
> Yeah I think a property would definitely be the safest option. But tbh I'm still trying to understand how this change is working. It might be okay to be released as a default.

@mrgrain, is it clear now? Do you still think we need a property to enable that behavior?

src/release/bump-version.ts
```diff
@@ -128,7 +128,19 @@ export async function bump(cwd: string, options: BumpOptions) {
...
 
-  if (currentTags.includes(latestTag)) {
+  // get the prerelease component(alpha, beta, pre) of the latestTag determined. Returns null if this is a regular release
+  const latestTagPrerelease = getPrereleaseComponent(latestTag);
+
+  const isLatestTagOnSameReleaseBranch =
+    (prerelease && // If we are bumping a prerelease now
+      latestTagPrerelease && // and the latest tag was also a prerelease
+      latestTagPrerelease[0] === prerelease) || // and the latest tag has the same prerelease component(alpha, beta, pre) as this prerelease
+    (!prerelease && !latestTagPrerelease); // If we are bumping a regular release( example: 1.0.5) now and also latest release was a regular release
```
mrgrain, line 138: Thanks for adding the comments. This makes it easier to understand.

Let's still refactor this into a well-named helper function that takes the `latestTag` and `options.prerelease` and uses early returns for readability.

src/release/bump-version.ts
```diff
@@ -309,7 +321,11 @@ function determineLatestTag(options: LatestTagOptions): {
     const releaseTags = tags.filter((x) =>
       new RegExp(`^v([0-9]+)\.([0-9]+)\.([0-9]+)$`).test(x)
     );
-    if (releaseTags && compare(releaseTags[0], prereleaseTags[0]) === 1) {
+    if (
+      releaseTags &&
+      releaseTags.length > 0 &&
```
mrgrain, line 326: Is this fixing a bug or something?

If it is required, let's also pull this out into a helper function.

mrgrain (changes requested):
Okay thanks for the explanations so far. I still have one question of understanding, but I'm sure you'll be able to clear that up for me.

We will need a new options for this that is `false` by default. Something along the lines of `allowMultipleReleasesOfSameCommit` and makes sure the docstring explains this scenario in detail.

mrgrain:
Alternatively taking a step back @abdsahin - Maybe what we actually want here is a `--force` option on the command that users can use to always force a new bump. You could then implement additional checks in userland.

I think I would very much prefer this alternative option. Do you think this would allow you to solve your problem in userland?

src/release/bump-version.ts
abdsahin: Yes this is for bug fixing, there are cases where ```releaseTags``` is returned as empty array and going directly into the if case rather than else.

src/release/bump-version.ts
```diff
@@ -128,7 +128,19 @@ export async function bump(cwd: string, options: BumpOptions) {
...
+  const latestTagPrerelease = getPrereleaseComponent(latestTag);
+
+  const isLatestTagOnSameReleaseBranch =
+    (prerelease && // If we are bumping a prerelease now
+      latestTagPrerelease && // and the latest tag was also a prerelease
+      latestTagPrerelease[0] === prerelease) || // and the latest tag has the same prerelease component(alpha, beta, pre) as this prerelease
+    (!prerelease && !latestTagPrerelease); // If we are bumping a regular release( example: 1.0.5) now and also latest release was a regular release
+
+  /**
+   * Skip the bump only if the current commit is already tagged within the same branch
```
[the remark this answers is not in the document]
abdsahin: Each branch is mapped to a specific prerelease component.

```
  releaseBranches:{
    "release/1.x.alpha": {
      majorVersion: 1,
      prerelease: "alpha"
    },
    "release/1.x.beta": {
      majorVersion: 1,
      prerelease: "beta"
    },
    "release/1.x.rc": {
      majorVersion: 1,
      prerelease: "rc"
    },
    "release/1.x": {
      majorVersion: 1
    }
  }
  ```
  
  So now
  - If it is a regular release and ```getPrereleaseComponent(latestTag)``` returns null meaning latest release was also a regular release.
  - If it is a prerelease, we compare the current commit prerelease component with the latest release one. As suggested I will try to add that into a helper function.

abdsahin:
@mrgrain , I've updated the branch, let me know if it looks better, we merge or close it accordingly.

PR #2632, opened 2023-04-26
feat(node-project): use resolution-mode=highest for pnpm by default

### LINKED ISSUE #2628: [node-package]: old package versions are pinned when using pnpm
When using `PackageManager.PNPM`, for any dependencies where we don't specify a version explicitly, we end up with projen pinning old versions of the package.

To reproduce:

* `npx projen new typescript --packageManager pnpm`
* Add `deps: ['aws-cdk-lib']`
* `pnpm projen`
* Notice `aws-cdk-lib@^2.0.0-alpha.0` is written to `package.json` (the latest is `2.76.0` at the time of writing!).

This seems to be due to `"*"` being written when no version is specified: https://github.com/projen/projen/blob/3c28310404f113ff71ccfc5e52e6267f10c716e6/src/javascript/node-package.ts#L1155

The [package.json docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies) mention that `*` means "any version", so it seems like it's actually an "implementation detail" that both `npm` and `yarn` both resolve the latest version, whereas `pnpm` does not 😞 .

It seems like Projen's intention is to have `yarn add`/`npm install`/`pnpm add` type behaviour for dependencies where the version isn't specified - ie. it resolves the latest version but pins it with a caret (`^`) in the `package.json`, so I do feel that this is a bug in projen.

I think the solution might be to use `latest` instead of `*` as the initial placeholder version, as I believe all 3 package managers will honour this. Though I can't see any official mention of `latest` in the [package.json docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies) so I'm not 100% sure if that's the right fix.

cogwirrel:
PNPM >= 8 resolves dependencies to the lowest matching version by default, which diverges from its previous behavior and and how NPM and Yarn resolve to the latest matching version.

Since projen uses `*` for dependencies where a version is not specified by the user, PNPM >= 8 resolves to the lowest published version for these which can cause problems, for example ts-node resolves to 0.2.0 on initial project synthesis which causes further calls to projen to fail.

With this change `.npmrc` file is always created for node projects. Reference the project's `npmrc` property to edit this instead of creating your own, or use `project.tryRemoveFile('.npmrc')` first.

Fixes #2628

---


mrgrain (changes requested):
Nice one @cogwirrel !

Can we change the breaking change to a fix of setting `omitEmpty` on the NpmConfig. The `.npmrc` is read-only. So having an empty file around is pointless.

cogwirrel:
No worries :) Ah nice one on the `omitEmpty`, much cleaner! :)

I think it's still a breaking change unfortunately as anyone who already does `new NpmConfig(nodeProject)` will get an error due to the file already being part of the project.

mrgrain:
> I think it's still a breaking change unfortunately as anyone who already does `new NpmConfig(nodeProject)` will get an error due to the file already being part of the project.

I see were you are coming from, but still don't think we should treat this as a breaking change. 
Basically any additional configuration added by users could be incompatible with changes to projen core. For example we could add a new workflow `foobar.yaml` and if someone had a workflow with the same name it would break.

test/javascript/npm-config.test.ts
```diff
@@ -66,5 +66,7 @@ class TestNodeProject extends NodeProject {
       defaultReleaseBranch: "master",
       ...options,
     });
+    // Remove default .npmrc
+    this.tryRemoveFile(".npmrc");
```
mrgrain, line 70: Do we still need this?

mrgrain (changes requested):

test/javascript/npm-config.test.ts
cogwirrel: Yes, since the NodeProject now creates it - even with `omitEmpty` it's still registered as a file so `new NpmConfig(project)` fails

cogwirrel:
Sure happy to not treat this as a breaking change if you prefer :) have updated the commit message

src/javascript/node-project.ts
```diff
@@ -732,6 +738,18 @@ export class NodeProject extends GitHubProject {
...
+    // Create the .npmrc file
+    this.npmrc = new NpmConfig(this, {
+      omitEmpty: true,
+    });
+
+    // For PNPM, the default resolution mode is "lowest", which leads to any non-versioned (ie '*') dependencies being
+    // resolved to the lowest available version, which is unlikely to be expected behaviour for users. We set resolution
+    // mode to "highest" to match the behaviour of yarn and npm.
+    if (this.package.packageManager === NodePackageManager.PNPM) {
+      this.npmrc.addConfig("resolution-mode", "highest");
```
mrgrain, line 751: Can you add a test to show that `project.npmrc.addConfig("resolution-mode", "lowest")` will override this default?

mrgrain (changes requested):

src/javascript/node-project.ts
cogwirrel: Sure :)

test/javascript/npm-config.test.ts
mrgrain: Oh I see why now. 🤦🏻

mrgrain (changes requested):
One last thing

mrgrain (approved):

cogwirrel:
Thank you! :)

PR #2634, opened 2023-04-27
feat(javascript): generic `tryResolveDependencyVersion` node-package method.

BradenM:
Ref: https://github.com/projen/projen/pull/2607#discussion_r1177611408

Defines:

- Generic `tryResolveDependencyVersion` javascript/util utility for locating and reading a package version from a given dependencies package.json manifest.
- A public method `NodePackage.tryResolveDependencyVersion` that checks for a semantically valid (i.e, not '\*') dependency version via `project.deps`. If not found it will fallback on the `tryResolveDependencyVersion` utility from above.

Also refactors `NodePackage.resolveDepsAndWritePackageJson` to utilize `tryResolveDependencyVersion` (should have no change in behavior in theory).

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -1301,20 +1301,14 @@ export class NodePackage extends Component {
         let desiredVersion = currentDefinition;
 
         if (currentDefinition === "*") {
-          try {
-            const modulePath = require.resolve(`${name}/package.json`, {
-              paths: [outdir],
-            });
-            const module = JSON.parse(readFileSync(modulePath, "utf-8"));
-            desiredVersion = `^${module.version}`;
-          } catch (e) {}
-
-          if (!desiredVersion) {
+          const resolvedVersion = this.tryResolveDependencyVersion(name);
+          if (!resolvedVersion) {
             this.project.logger.warn(
               `unable to resolve version for ${name} from installed modules`
             );
             continue;
           }
+          desiredVersion = `^${resolvedVersion}`;
         }
 
         if (currentDefinition !== desiredVersion) {
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -1303,7 +1303,7 @@ export class NodePackage extends Component {
         let desiredVersion = currentDefinition;
 
         if (currentDefinition === "*") {
-          const resolvedVersion = this.tryResolveDependencyVersion(name);
+          const resolvedVersion = tryResolveDependencyVersion(name);
           if (!resolvedVersion) {
             this.project.logger.warn(
               `unable to resolve version for ${name} from installed modules`
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -865,7 +865,9 @@ export class NodePackage extends Component {
    *
    * @param dependencyName Dependency to resolve for.
    */
-  public tryResolveDependencyVersion(dependencyName: string) {
+  public tryResolveDependencyVersion(
+    dependencyName: string
+  ): string | undefined {
     try {
       const fromDeps = this.project.deps.tryGetDependency(dependencyName);
       const version = semver.coerce(fromDeps?.version, { loose: true });
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -870,7 +870,7 @@ export class NodePackage extends Component {
       const fromDeps = this.project.deps.tryGetDependency(dependencyName);
       const version = semver.coerce(fromDeps?.version, { loose: true });
       if (version) return version.format();
-    } catch (e) {}
+    } catch {}
     return tryResolveDependencyVersion(dependencyName, {
       paths: [this.project.outdir],
     });
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -1,12 +1,17 @@
 import {
-  readFileSync,
   accessSync,
   constants,
   existsSync,
   readdirSync,
+  readFileSync,
 } from "fs";
 import { join, resolve } from "path";
-import { extractCodeArtifactDetails, minVersion } from "./util";
+import * as semver from "semver";
+import {
+  extractCodeArtifactDetails,
+  minVersion,
+  tryResolveDependencyVersion,
+} from "./util";
 import { resolve as resolveJson } from "../_resolve";
 import { Component } from "../component";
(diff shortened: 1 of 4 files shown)


src/javascript/node-package.ts
```diff
@@ -849,6 +854,28 @@ export class NodePackage extends Component {
...
+   *
+   * @remarks
+   * This method will first look through the current project's dependencies.
+   * If found and semantically valid (not '*'), that will be used.
+   * Otherwise, it will fall back to locating a `package.json` manifest for the dependency
+   * through node's internal resolution reading the version from there.
+   *
+   * @param dependencyName Dependency to resolve for.
+   */
+  public tryResolveDependencyVersion(dependencyName: string) {
```
mrgrain, line 868:
```suggestion
  public tryResolveDependencyVersion(dependencyName: string): string {
```

src/javascript/node-package.ts
```diff
@@ -849,6 +854,28 @@ export class NodePackage extends Component {
...
+    try {
+      const fromDeps = this.project.deps.tryGetDependency(dependencyName);
+      const version = semver.coerce(fromDeps?.version, { loose: true });
+      if (version) return version.format();
+    } catch (e) {}
```
mrgrain, line 873:
```suggestion
    } catch {}
```

src/javascript/node-package.ts
```diff
@@ -849,6 +854,28 @@ export class NodePackage extends Component {
...
+   * Otherwise, it will fall back to locating a `package.json` manifest for the dependency
+   * through node's internal resolution reading the version from there.
+   *
+   * @param dependencyName Dependency to resolve for.
+   */
+  public tryResolveDependencyVersion(dependencyName: string) {
+    try {
+      const fromDeps = this.project.deps.tryGetDependency(dependencyName);
+      const version = semver.coerce(fromDeps?.version, { loose: true });
+      if (version) return version.format();
```
mrgrain, line 872: I should really fix the eslint rule for that 🙈 
```suggestion
      if (version) {
        return version.format();
      }
```

src/javascript/node-package.ts
```diff
@@ -1274,20 +1301,14 @@ export class NodePackage extends Component {
...
-          try {
-            const modulePath = require.resolve(`${name}/package.json`, {
-              paths: [outdir],
-            });
-            const module = JSON.parse(readFileSync(modulePath, "utf-8"));
-            desiredVersion = `^${module.version}`;
-          } catch (e) {}
-
-          if (!desiredVersion) {
+          const resolvedVersion = this.tryResolveDependencyVersion(name);
```
mrgrain, line 1304: Let's use the one from utils here. Given that we already know projen doesn't know the version...
```suggestion
          const resolvedVersion = tryResolveDependencyVersion(name);
```

mrgrain (changes requested):

src/javascript/node-package.ts
BradenM: Strictly speaking, this should be `string | undefined`, not `string`.
On that note, I actually didn't type it purposely due to a bit of confusion on my part with how jsii handles union types (which you mentioned it had issues with earlier.).

Is that limitation specific to more complex unions such as what I had originally in #2607?

src/javascript/node-package.ts
BradenM: Not a fan? :smile: 
Personally, I find it a bit cleaner, but I can definitely see where others may not feel the same way.

src/javascript/node-package.ts
BradenM: Makes sense :+1:

src/javascript/node-package.ts
BradenM: Sorry, got ahead of myself re-requesting a review.
Still pending on what the correct solution to this one is. Thanks.

mrgrain: Gotcha. My ask is basically to explicitly add the expected return type, instead of implicitly relying on it.
If it's `string | undefined` than that's fine.

---

The jsii union types issue is mostly about user supplied types. Return types with undefined are fine.
See e.g. https://github.com/projen/projen/blob/b729a93146dc2a3f2a27a2ec67b48f8b24354964/src/tasks.ts#L114

BradenM: Thanks for the clarification :+1: 
Have added the return type.

BradenM:
@mrgrain The changes from https://github.com/projen/projen/pull/2634#discussion_r1178772369 caused the test failures. Just needed to respecify the project outdir as a resolution path since we are using the util directly at that point. Should be all good now. :+1:

mrgrain (approved):
🚀

BradenM:
@mrgrain Not sure what went wrong with `post-build-integ` job...
```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/feat/node-resolve-dependency*:refs/remotes/origin/feat/node-resolve-dependency* +refs/tags/feat/node-resolve-dependency*:refs/tags/feat/node-resolve-dependency*
  Error: fatal: unable to access 'https://github.com/BradenM/projen/': The requested URL returned error: 429
  The process '/usr/bin/git' failed with exit code 128
  Waiting 14 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/feat/node-resolve-dependency*:refs/remotes/origin/feat/node-resolve-dependency* +refs/tags/feat/node-resolve-dependency*:refs/tags/feat/node-resolve-dependency*
  Error: fatal: unable to access 'https://github.com/BradenM/projen/': The requested URL returned error: 429
  The process '/usr/bin/git' failed with exit code 128
  Waiting 13 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/feat/node-resolve-dependency*:refs/remotes/origin/feat/node-resolve-dependency* +refs/tags/feat/node-resolve-dependency*:refs/tags/feat/node-resolve-dependency*
  Error: fatal: unable to access 'https://github.com/BradenM/projen/': The requested URL returned error: 429
  Error: The process '/usr/bin/git' failed with exit code 128
```

429 sounds like rate limiting... maybe just needs a re-run?

PR #2645, opened 2023-05-01
fix(javascript): do rely on `package.json` being a valid entrypoint for dependency resolution.

### LINKED ISSUE #2644: Dependency resolution may fail on modules with customized entrypoint.
Dependencies that define/override entrypoints via the `exports` field in their package.json manifest file will fail to have their version resolved if `package.json` was not included as an export.

Even though the nodejs documentation suggests including `package.json` in `exports` (see: https://nodejs.org/api/packages.html#package-entry-points), many packages do not do so.

One big example (and the one that led me to this), is rollup:

```json
{
  "name": "rollup",
  "version": "3.21.1",
  "description": "Next-generation ES module bundler",
  "main": "dist/rollup.js",
  "module": "dist/es/rollup.js",
  "types": "dist/rollup.d.ts",
  "bin": {
    "rollup": "dist/bin/rollup"
  },
  "files": [
    "dist/**/*.js",
    "dist/*.d.ts",
    "dist/bin/rollup",
    "dist/es/package.json"
  ],
  "engines": {
    "node": ">=14.18.0",
    "npm": ">=8.0.0"
  },
  <- ... ->
  "exports": {
    ".": {
      "types": "./dist/rollup.d.ts",
      "require": "./dist/rollup.js",
      "import": "./dist/es/rollup.js"
    },
    "./loadConfigFile": {
      "types": "./dist/loadConfigFile.d.ts",
      "require": "./dist/loadConfigFile.js",
      "default": "./dist/loadConfigFile.js"
    },
    "./dist/*": "./dist/*"
  }
}
```

Because projen relies on resolving a `{dependencyName}/package.json` module id (and this behavior was carried through #2634), rollup (and many others) always fail to resolve.

BradenM:
Fixes #2644

---


src/util.ts
```diff
@@ -472,8 +472,9 @@ export function findUp(
   lookFor: string,
   cwd: string = process.cwd()
 ): string | undefined {
-  if (existsSync(path.join(cwd, lookFor))) {
-    return cwd;
+  const target = path.join(cwd, lookFor);
+  if (existsSync(target)) {
+    return target;
```
BradenM, line 477: This PR also resolves a bug I ran into trying to use the `findUp` method to resolve for a `package.json` file.

As it was, `findUp` always returned the directory of a found file, instead of the file path. 

I only found one other place where `findUp` was used (and it was for finding a projen directory), so this fix shouldn't impact anything else that I am aware of.

Please let me know if that is not the case.

BradenM: :thinking: Uh nevermind. Definitely broke the cli. Still don't believe this was the intended behavior (always returning the directory), but have removed for now and just handled in the resolution utilities.

mrgrain (approved):

PR #2653, opened 2023-05-05
fix: ensure env command is package manager specific

agdimech:
This PR significantly improves performance when using non-npm package managers.

When using for example PNPM - the `$(npx -c "node -e \\"console.log(process.env.PATH)\\"")` command is extremely slow and is preferred to use the package manager specific exec command. To put it into perspective, every time you call `pnpm projen XXX` this will call the exec command (in my case twice for a monorepo with a parent) which for lager projects like mine can take 5+ seconds.

View the below Before and After results to show the improvements this PR makes to build times which shows an approximate 60% improvement using PNPM:

## Before
`pnpm build --skip-nx-cache --parallel 8  2269.86s user 1166.58s system 417% cpu 13:43.63 total`

## After
`pnpm build --skip-nx-cache --parallel 8  1320.03s user 942.25s system 426% cpu 8:49.87 total`

---


agdimech:
Not sure why jsii cannot be found as locally the `node_modules/.bin` folder is added to the PATH.

mrgrain:
I love this!

mrgrain (approved):

PR #2656, opened 2023-05-05
fix(sample-file): unable to create empty sample directory

### LINKED ISSUE #1990: Unable to create empty sample directory
I was trying to create an empty sample directory for a project (just a directory whose contents aren't managed by projen) but found this did nothing:

```ts
new SampleDir(project, "foo", { files: {} });
```

This would be a bit odd if the directory is meant to be versioned controlled, because you cannot add empty directories to git, but in my scenario I had a build script / test that expected the directory to exist, and it was just for build storing artifacts (so I had it gitignored). Maybe the line above should create a directory instead of being a no-op?

Alternatively, maybe this is silly and we should just throw an error if `files` is empty.

andersonmgomes:
fixes #1990 

---

### DIFF
--- a/test/sample-file.test.ts
+++ b/test/sample-file.test.ts
@@ -1,6 +1,6 @@
 import * as fs from "fs";
 import * as path from "path";
-import { synthSnapshot } from "./util";
+import { mkdtemp, synthSnapshot } from "./util";
 import { Project, ProjectOptions } from "../src";
 import { SampleDir, SampleFile } from "../src/sample-file";
 
@@ -32,7 +32,8 @@ test("sample file from source", () => {
 
 test("sample empty directory", () => {
   // GIVEN
-  const project = new TestProject();
+  const tempDir = mkdtemp(); // dir auto removed after test
+  const project = new TestProject({ outdir: tempDir });
 
   // WHEN
   new SampleDir(project, "foo", {
@@ -43,8 +44,6 @@ test("sample empty directory", () => {
   // THEN
   project.synth();
   expect(fs.existsSync(fooPath)).toBeTruthy();
-  //removing the project directory
-  fs.rmSync(project.outdir, { force: true, recursive: true });
 });
 
 test("sample directory from files", () => {
--- a/test/sample-file.test.ts
+++ b/test/sample-file.test.ts
@@ -1,3 +1,4 @@
+import * as fs from "fs";
 import * as path from "path";
 import { synthSnapshot } from "./util";
 import { Project, ProjectOptions } from "../src";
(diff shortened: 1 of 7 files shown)


test/sample-file.test.ts
```diff
@@ -29,6 +29,18 @@ test("sample file from source", () => {
...
+test("sample empty directory", () => {
+  // GIVEN
+  const project = new TestProject();
+
+  // WHEN
+  new SampleDir(project, "foo", { files: {} });
+
+  // THEN
+  const snapshot = synthSnapshot(project);
+  expect(snapshot).toBeDefined();
```
andrestone, line 41: Shouldn't we test if whether or not the directory is created?

mrgrain (changes requested):

andersonmgomes:
@andrestone and @mrgrain ,
Absolutely yes! 
I fixed the test case and am now checking if the empty dir was created correctly.
Thank you for your review, and sorry about my fault in the first commit!

test/sample-file.test.ts
```diff
@@ -29,6 +30,23 @@ test("sample file from source", () => {
...
+  new SampleDir(project, "foo", {
+    files: {},
+  });
+  const fooPath = path.join(project.outdir, "foo/");
+
+  // THEN
+  project.synth();
+  expect(fs.existsSync(fooPath)).toBeTruthy();
+  //removing the project directory
+  fs.rmSync(project.outdir, { force: true, recursive: true });
```
mrgrain, line 47: Thanks for cleaning up afterwards! 

Instead of doing this manually, let's use the existing helper instead. It does automatically cleanup.

https://github.com/projen/projen/blob/main/test/util.ts#L59

mrgrain (changes requested):
No worries. One final adjustment is needed.

test/sample-file.test.ts
mrgrain: Or I think you can also keep using `synthSnapshot` and check if the directory exists. That helper also cleans up after itself.

andersonmgomes: mrgrain, I've tried to use synthSnapshot. The problem was checking if the "fooPath" existed before the cleaner did its job. On another side, if I do not call the synth method for the project, nothing happens, and any dir is created.
Can I change only line 47 to call the helper?

mrgrain: I see. In that case the `mkdtmp` helper should do it. It will cleanup after the test has run.

You'll need to create the directory first and pass it to the project as `outdir`. Then remove the manual cleanup. That should do it.

Issue #2300, opened 2022-12-14
PythonProject fails with `/bin/sh: python: command not found`


donatoaz:
Created a `PythonProject` and when running `npx projen` I got:

```
👾 default | ts-node --project tsconfig.dev.json .projenrc.ts
👾 Setting up a virtual environment...
Error: Command failed: python -m venv .env
/bin/sh: python: command not found
```

My terminal is `zsh` and does have `python` aliased to `python3` working fine.

I did note that `sh` itself really did not have the alias, and had to create a softlink to solve the issue. But perhaps it would be better if projen used the terminal of choice, instead of using `/bin/sh`.

Current env:
Mac OS 12.6.1
node v14.19.0
terminal: zsh

mrgrain:
It's very common to use a generic `/bin/sh` because starting to use terminals is a whole can of worms. We don't actually use it directly, its node's `child_process` that does that.

Either way, I think we attempt to detect `python3` or provide a way to customize the python binary.
Related issue: https://github.com/projen/projen/issues/2448

loganfarr:
For others, like myself, who found this after running into the issue after `npx projen new ...`
A note for a workaround: just run the venv command by itself and then try the projen command again. 

I.e.
```shell
python -m venv .env
npx projen new awscdk-app-py
```

The project will then install the right pip dependencies and set up successfully.

PR #2661, opened 2023-05-07
[WIP] feat(Python): python executable support python3

FonsBiemans:
When executing a python project, on new mac systems it fails because of lack of python 2.x. 
as a result of that the fix was to make the executable of python a variable with a default of python3. 

i'm not sure if it's all in the correct location (first issue), so please provide feedback. 

open issue : #2300

### DIFF
--- a/src/python/projenrc.ts
+++ b/src/python/projenrc.ts
@@ -22,6 +22,12 @@ export interface ProjenrcOptions {
    * @default - current version
    */
   readonly projenVersion?: string;
+
+  /**
+   * Python executable to use. default python3 alternative:  python.
+   * @default  "python3"
+   */
+  readonly pythonexec?: string;
 }
 
 /**
@@ -35,20 +41,21 @@ export class Projenrc extends ProjenrcFile {
    * The name of the projenrc file.
    */
   public readonly filePath: string;
+  public readonly pythonexec: string;
 
   constructor(project: Project, options: ProjenrcOptions = {}) {
     super(project);
 
     const projenVersion = options.projenVersion ?? PROJEN_VERSION;
     this.filePath = options.filename ?? ".projenrc.py";
-
+    this.pythonexec = options.pythonexec ?? "python3";
     project.deps.addDependency(
       `projen@${projenVersion}`,
       DependencyType.DEVENV
     );
 
     // set up the "default" task which is the task executed when `projen` is executed for this project.
-    project.defaultTask?.exec("python .projenrc.py");
+    project.defaultTask?.exec(`${this.pythonexec} .projenrc.py`);
 
     // if this is a new project, generate a skeleton for projenrc.py
     this.generateProjenrc();
--- a/src/python/venv.ts
+++ b/src/python/venv.ts
@@ -15,6 +15,12 @@ export interface VenvOptions {
    * @default ".env"
    */
   readonly envdir?: string;
+
+  /**
+   * Python executable to use. default python3 alternative:  python.
+   * @default  "python3"
+   */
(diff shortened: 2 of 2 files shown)


src/python/projenrc.ts
```diff
@@ -22,6 +22,12 @@ export interface ProjenrcOptions {
    * @default - current version
    */
   readonly projenVersion?: string;
+
+  /**
+   * Python executable to use. default python3 alternative:  python.
+   * @default  "python3"
```
mrgrain, line 28: My impression is that this is still the most portable default. Also if we change that it might be annoying for some.
```suggestion
   * @default  "python"
```

src/python/projenrc.ts
```diff
@@ -22,6 +22,12 @@ export interface ProjenrcOptions {
    * @default - current version
    */
   readonly projenVersion?: string;
+
+  /**
+   * Python executable to use. default python3 alternative:  python.
```
mrgrain, line 27:
```suggestion
   * Path to the python executable to use.
   *
   * @example "python3"
```

src/python/projenrc.ts
```diff
@@ -35,20 +41,21 @@ export class Projenrc extends ProjenrcFile {
    * The name of the projenrc file.
    */
   public readonly filePath: string;
+  public readonly pythonexec: string;
```
mrgrain, line 44:
```suggestion
  /** add docstring here */
  public readonly pythonExec: string;
```

src/python/projenrc.ts
```diff
@@ -35,20 +41,21 @@ export class Projenrc extends ProjenrcFile {
...
 
   constructor(project: Project, options: ProjenrcOptions = {}) {
     super(project);
 
     const projenVersion = options.projenVersion ?? PROJEN_VERSION;
     this.filePath = options.filename ?? ".projenrc.py";
-
+    this.pythonexec = options.pythonexec ?? "python3";
```
mrgrain, line 51:
```suggestion
    this.pythonExec = options.pythonExec ?? "python";
```

src/python/venv.ts
```diff
@@ -15,6 +15,12 @@ export interface VenvOptions {
    * @default ".env"
    */
   readonly envdir?: string;
+
+  /**
+   * Python executable to use. default python3 alternative:  python.
+   * @default  "python3"
+   */
+  readonly pythonexec?: string;
```
mrgrain, line 23:
```suggestion
  /** fixme same docstring as everywhere **/
  readonly pythonExec?: string;
```

mrgrain (changes requested):
Thanks for giving this a go @FonsBiemans !  I've left some minor inline feedback.

Also interesting that we need to have the option in two places. I guess it's because the components don't require a python project to work. I would have expected this to be an option on the Python Project though to be honest. Could you provide an example how it's used?

Ideas:
- Define on `PythonProject` have the components check if the project is a python project and use that exec by default.
- Support for `PYTHON` environment variable

src/python/venv.ts
```diff
@@ -15,6 +15,12 @@ export interface VenvOptions {
    * @default ".env"
    */
   readonly envdir?: string;
+
+  /**
+   * Path to the python executable to use.
+   * @default  "python"
```
mrgrain, line 21:
```suggestion
   * @default "python"
```

mrgrain (approved):

mrgrain (changes requested):
@FonsBiemans Last thing, you need to include the license statement footer in your PR description. See other PRs for examples.

mrgrain (approved):

PR #2663, opened 2023-05-08
feat(tasks): Add support for environment variables in task steps

icj217:
Currently tasks support environment variables, but only at the task level. These variables are resolved eagerly at the beginning of the task's execution. 

There are scenarios where having environment variables that are bound to a particular step can be quite useful, especially since its not possible to "pass state" from one step to another. The example below shows a scenario where a task starts a docker container and then needs to look up the container's ephemeral port and pass that info as an environment variable (`S3_ENDPOINT`) to another command. Currently, this environment variable has to be inline to the command, since it must be evaluated after the first spawn step finishes.

For example:
```typescript
project.tasks.addTask('tests:docker', {
            steps: [
                { spawn: 'start:docker:localstack' },
                { exec: `S3_ENDPOINT=http://localhost:$(docker compose port localstack 4566 | cut -d: -f2) docker compose run --rm tester` }
            ]
        })

```

With this change, environment variables can be defined at the individual step level (only resolved for `exec` and `builtin` steps), making the example above much cleaner to follow:

```typescript
project.tasks.addTask('tests:docker', {
            steps: [
                { spawn: 'start:docker:localstack' },
                { 
                    env: {
                        // Obtain value after start:docker:localstack finishes 
                        S3_ENDPOINT: '$(echo "http://localhost:$(docker compose port localstack 4566 | cut -d: -f2)")',
                    },
                    exec: `docker compose run --rm tester`
                }
            ]
        })
```

Features:
* Variables defined at step level will override task-level variables
* Can be defined in both upfront `steps` argument as well as in `task.exec()` and `task.builtin()` methods
* Variable values will be stringified (same as task/global variables)

---


src/task.ts
```diff
@@ -256,16 +274,24 @@ export class Task {
     }
   }
 
+  /**
+   * Ensure that environment variables are persisted as strings
+   * to prevent type errors when parsing from tasks.json in future
+   */
+  private getEnvString(name: string, value: any) {
+    if (typeof value !== "string" && value !== undefined) {
+      warn(
```
mrgrain, line 283: Looks like this will now warn twice for a wrong env type: First when `addEnv` is called, and then again when the task file is rendered via `_renderSpec`.

src/task-model.ts
```diff
@@ -118,6 +118,13 @@ export interface TaskStepOptions {
    * @default - no arguments are passed to the step
    */
   readonly args?: string[];
+
+  /**
+   * Defines environment variables for the execution of this step (`exec` and `builtin` only).
+   * Values in this map will be evaluated in a shell, so you can do stuff like `$(echo "foo")`.
+   * @default {}
```
mrgrain, line 125: Can a user also define simple values? If so, let's make this clear in the docstring. An `@example` tag would also be great.

```ts
env: {
  foo: "bar"
}
```

mrgrain (changes requested):
Nice one @icj217 ! 🚀 

Just got a couple of small suggestions.

src/task.ts
icj217: Good catch. I've decided to defer all value manipulation until writing to the manifest, so this should be resolved now.

icj217:
Friendly ping @mrgrain whenever you get a chance to re-review

mrgrain (approved):
🚀

PR #2685, opened 2023-05-18
fix(node-package): allow to remove resolutions

tinovyatkin:
This PR switches the source of truth for package resolutions from the internal structure `packageResolutions` to project dependencies (as in the case of all other dependencies).

This allows particularly to remove resolutions via simple `project.deps.removeDependency(name, DependencyType.OVERRIDE)` which is currently impossible (as resolution renders out of private `packageResolutions` and never syncs with the project dependencies after initial addition).

---


mrgrain (changes requested):
Nice one! Can you please add a test showing that we can remove a resolution using `removeDependency` and this should be good to go.

test/javascript/node-package.test.ts
```diff
@@ -490,6 +490,33 @@ test("npm overrides", () => {
...
+  });
+
+  pkg.addDeps("some-dep@1.0.0");
+  pkg.addPackageResolutions("some-dep@1.0.0", "other-dep@another-dep@1.2.3");
+  project.deps.removeDependency("some-dep", DependencyType.OVERRIDE);
+
+  const snps = synthSnapshot(project);
+  const pkgJson = snps["package.json"];
+
+  expect(pkgJson).toHaveProperty("dependencies.some-dep", "1.0.0");
```
mrgrain, line 508: I would have expected an assertion like this:
```suggestion
  expect(pkgJson).toHaveProperty("dependencies.some-dep", "1.0.0");
  expect(pkgJson).not.toHaveProperty("overrides.some-dep");
```

src/javascript/node-package.ts
```diff
@@ -1387,9 +1389,28 @@ export class NodePackage extends Component {
...
+          );
+          this.packageResolutions.delete(resolution);
+        }
+      }
+      for (const override of overridingDependencies) {
+        if (!this.packageResolutions.has(override))
+          this.project.logger.warn(
+            `Overriding dependency '${override}' lacks target. Make sure you adding overrides via addPackageResolutions method.`
+          );
+      }
```
mrgrain, line 1413: Is this whole bit even necessary? Looks like we are now storing package resolutions (aka overrides) in two places:
- the private `packageResolutions` field, and
- the project dependencies

And then we are checking that both places are actually in sync?! And debug/warn if not?
Feels like we can just get rid of the `packageResolutions` field and take the dependencies as source of truth.

tinovyatkin: I believe this is because current projen support for overrides is pretty limited (for example in NPM key for override can include version or even version range and value maybe an object, not just string) and this is some sort of future todo. Those more complicated cases can be solved via JsonPatch. I'll remove the `packageResolutions`.

mrgrain (approved):

mrgrain (changes requested):
Code looks good now. You've got a failing build due to snapshot updates.

mrgrain (approved):

PR #2686, opened 2023-05-18
feat(tasks): add conditions to task steps

icj217:
This will make the task runtime much more powerful/flexible and all the logic/functionality was basically already there.

A simple example (not the best/most logical but just for illustration):

```ts
project.addTask('docker:run': {
  steps: [
    { 
      name: 'Build (if necessary)', 
      exec: 'docker build -t my-image', 
      condition: 'docker image ls | (! grep my-image)' 
    },
    { name: 'Run', exec: 'docker run my-image' }
  ]
});
```

- feat(tasks): Add support for step conditions
- chore(docs): add info about step-level env/condition to tasks.md

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2694, opened 2023-05-23
feat(task): support adding condition to task

JeremyJonas:
Use Case:
- As a developer, I want to be able to add a condition to an existing task, so I can control how the task is executed during CI build

For instance, I want to add a condition to the "test" task, which is nested in the "build" task. During a CI release workflow, I want to skip tests during build, since they have already run in previous workflow job.

Currently to do this, I must reset the test task and recreate it.

---


icj217:
@JeremyJonas Funnily enough I also have a PR submitted for this enhancement: https://github.com/projen/projen/pull/2686

mrgrain (approved):

Issue #2434, opened 2023-02-10
Projen should allow setting project references for typescript projects


timoheinonen:
I have a monorepo project with multiple services (different aws cdk app projects) and a shared library for common constructs (own aws cdk construct library project). In future, I will probably move the aws cdk construct library to own repository and use it through npm registry, but for now, I would like to use it directly from the same repo. 

Typescript's project references (https://www.typescriptlang.org/docs/handbook/project-references.html) seems to be the best way to do this, but projen does't provide a way to set the tsconfig parameter "references" using TypescriptConfigOptions. Also, in the library project I would need to set tsconfig parameter "composite" to true, but the whole parameter is missing from TypescriptConfigOptions compilerOptions.

Would it be possible to add these to projen?

mrgrain:
Yes please! PRs are welcome.

In the meantime, you can easily use overrides:
```ts
      project.tsconfig?.file.addOverride('compilerOptions.composite', true);
      project.tsconfig?.file.addOverride('references', []);
```

skleinjung:
I have built this feature into a custom TypescriptProject subtype I use on my projects, but think it's something that could be submitted upstream pretty easily. However, there are two "flavors" of this and I'd like some feedback from the maintainers about which fits the project philosophy better.

1. The simplest and least "batteries included" version is to add a "references" property that takes a list of string, and directly sets it as the references.
2. A more advanced option (and what I use personally) is that "references" is a list of TypescriptProject references. The paths to  set in tsconfig are the relative paths between the referencing project's `outdir` and the referenced project's `outdir`. This requires all the projects to be projen-ified, but allows them to be synthed to any folder and the reference property doesn't need to care.
3. The most opinionated option, and what I will probably move to in my projects, is to not have a "references" property at all. It requires all projects (referencing and references) to share a common parent. Then, all the dependencies are checked against the parent's subprojects. If the dependency is a subproject of the common parent, a reference is added using the path logic in #2.

#3 supports the custom monorepo structure I have built around projen, and does what is to me the most sensible thing without custom properties. It might need to be expanded to include a full parent project hierarchy, although I personally don't nest projects more than one level deep.

If we can have a discussion about preferences on how this feature would be introduced, I should be able to find some time to work on it.

skleinjung:
[...]

PR #2695, opened 2023-05-23
[WIP] feat(typescript-config): add support for project references

skleinjung:
Add generic support for Typescript project references to the TypescriptConfig component in a manner similar to how TypescriptConfigsExtends are handled. Allows specifying references via path (absolute or relative), or by using a TypescriptConfig reference directly.

Once fully implemented, this PR would satisfy the baseline generic requirement expressed in Issue #2434. A follow-up phase would be to add opinionated logic to the Typescript project type to infer references from parent and dependency relationships. 

---

### DIFF
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -34,6 +34,17 @@ export interface TypescriptConfigOptions {
    * Compiler options to use.
    */
   readonly compilerOptions: TypeScriptCompilerOptions;
+
+  /**
+   * References to other Typescript modules in a multi-module project.
+   *
+   * If a project has any references, the `declaration` compiler option will be turned on and an
+   * error will be thrown if this option is disabled.
+   *
+   * @see https://www.typescriptlang.org/docs/handbook/project-references.html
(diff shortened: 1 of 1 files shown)


src/javascript/typescript-config.ts
```diff
@@ -554,6 +565,58 @@ export class TypescriptConfigExtends {
...
+
+  /**
+   * Factory for creation from array of other TypescriptConfig instances. This will automatically set the
+   * 'composite' compiler option on the referenced projects to true.
+   *
+   * @param paths Referenced `TypescriptConfig` instances.
+   */
+  public static fromTypescriptConfigs(configs: TypescriptConfig[]) {
+    throw new Error("Not implemented");
+  }
```
skleinjung, line 613: These methods don't let you specify the "prepend" option, mainly because I've personally never used it and I didn't want to go crazy with overloads. If needed, references with prepend set can be added one at a time via `addReference`, below. If anyone thinks its better to do so, I can add batch methods that accept that value as well.

mrgrain:
clean-up

PR #2711, opened 2023-05-31
fix(javascript): workflowPackageCache doesn't work with pnpm

### LINKED ISSUE #2710: `workflowPackageCache` doesn't work with `pnpm`
Follow up of https://github.com/projen/projen/pull/2709

Because "Setup Node.js" step runs before "Setup pnpm" step, it fails with:

```
Error: Unable to locate executable file: pnpm. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

Need to shuffle the order of the steps.

riywo:
To use `workflowPackageCache` option with `pnpm`,
"Setup pnpm" step must run before "Setup Node.js" step.

This order comes from pnpm's integration documentation: https://pnpm.io/continuous-integration#github-actions

Close #2710

---

### DIFF
--- a/src/javascript/node-project.ts
+++ b/src/javascript/node-project.ts
@@ -963,6 +963,14 @@ export class NodeProject extends GitHubProject {
     // first run the workflow bootstrap steps
     install.push(...this.workflowBootstrapSteps);
 
+    if (this.package.packageManager === NodePackageManager.PNPM) {
+      install.push({
+        name: "Setup pnpm",
+        uses: "pnpm/action-setup@v2.2.4",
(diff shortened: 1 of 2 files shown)


riywo:
I'm not sure where the `setup-node`'s error actually happens but probably here: https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/src/cache-utils.ts#L21

mrgrain:
Yes that makes a lot of sense. Unlike `npm` and `yarn`, `pnpm` isn't available by default.

PR #2719, opened 2023-06-04
fix(node): print PATH without escaping double quotes

### LINKED ISSUE #981: Task environment shell execution is not working on windows
When working on windows we get the following error during synthesize:
```
C:\local_projects\di-tools-fullload-service>npx projen
\"" was unexpected at this time.
🤖 Synthesis failed: unable to evaluate environment variable PATH=$(npx -c "node -e \"console.log(process.env.PATH)\""): unknown error
C:\local_projects\di-tools-fullload-service>
```

I traced it back to tasks.json where this line is add through the default task execution:
```
  "env": {
    "PATH": "$(npx -c \"node -e \\\"console.log(process.env.PATH)\\\"\")"
  },
```
When evaluating the value it will come up with the following statement:

```
npx -c "node -e \"console.log(process.env.PATH)\""
```
This works on mac but not on windows. On windows the escaping should be (double double quotes):
```
npx -c "node -e ""console.log(process.env.PATH)"""
```

I've seen other issues related to this. Would it not be a better idea to include some simple node based macros here?

i.e.
```
  "env": {
    "PATH": "#(console.log(process.env.PATH))"
  }
```
This way you keep control over the parsed content. Just dump it in a js evaluator or the system's node process.

matt9ucci:
Fix #981

To avoid escaping double quotes, use `--print` option instead of `-e console.log()`.

---


mrgrain (approved):
Nice one!

PR #2758, opened 2023-06-26
feat(release): support specifying which commits should be considered to cause a new release

### LINKED ISSUE #2179: Support skipping releases

### LINKED ISSUE #2751: Manual Pipeline Release broken due to pre-release logic introduced in PR #2609
PR https://github.com/projen/projen/pull/2609 attempted to fix incorrect prerelease version when there is a regular version. 

This change breaks manual releases when there is no releaseTags (empty array) as it's entering second half of if statement: 

https://github.com/projen/projen/blob/a970b968378796e246f7a91d045509ceabf14559/src/release/bump-version.ts#L323

From debugging I'm seeing:

```ts
the releases array = []

the prereleases array = [
  'v0.1.0-alpha.57', 'v0.1.0-alpha.56', 'v0.1.0-alpha.55',
  'v0.1.0-alpha.54', 'v0.1.0-alpha.53', 'v0.1.0-alpha.52',
  'v0.1.0-alpha.51', 'v0.1.0-alpha.50', 'v0.1.0-alpha.49',
  'v0.1.0-alpha.48', 'v0.1.0-alpha.47', 'v0.1.0-alpha.46'
]
```

which results in as `releaseTags[0] == undefined`
```ts
TypeError: Invalid version. Must be a string. Got type "undefined".
    at new SemVer (/Users/736574/workstation/ccoe-cdk/shared-aws-cdk/node_modules/projen/node_modules/semver/classes/semver.js:19:13)
    at compare (/Users/736574/workstation/ccoe-cdk/shared-aws-cdk/node_modules/projen/node_modules/semver/functions/compare.js:3:3)
    at determineLatestTag (/Users/736574/workstation/ccoe-cdk/shared-aws-cdk/node_modules/projen/lib/release/bump-version.js:165:49)
    at bump (/Users/736574/workstation/ccoe-cdk/shared-aws-cdk/node_modules/projen/lib/release/bump-version.js:33:58)
👾 Task "release » bump" failed when executing "/Users/736574/.asdf/installs/nodejs/18.15.0/bin/node /Users/736574/workstation/ccoe-cdk/shared-aws-cdk/node_modules/projen/lib/release/bump-version.task.js" (cwd: /Users/736574/workstation/ccoe-cdk/shared-aws-cdk)
```

## Possible Fix:

Tested Locally and changing statement to 

`releaseTags.length > 0 && ...` should fix it.

mrgrain:
Resolves #2179

Also fixes #2751, a bug with manual releases when there are no release tags yet.


src/version.ts
```diff
@@ -112,3 +124,72 @@ export class Version extends Component {
...
+   *
+   * This will only not release if the most recent commit is tagged with the latest matching tag.
+   *
+   * @param limitToProjectPath Only consider commits relevant to the path of the (sub-)project
+   */
+  static everyCommit(limitToProjectPath: boolean = false) {
+    const cmd = `git log --oneline $LATEST_TAG..HEAD`;
+
+    if (limitToProjectPath) {
+      return new ReleasableCommits(`${cmd} -- .`);
```
rix0rrr, line 143: Wouldn't we need the `cwd` here? Is this implicitly always guaranteed to be correct?

src/version.ts
```diff
@@ -112,3 +124,72 @@ export class Version extends Component {
...
+   * This will only release commit that match one of the provided types.
+   * All commits are required to match conventional commit spec and will be ignored otherwise.
+   *
+   * @param types List of conventional commit types that should be released
+   * @param limitToProjectPath Only consider commits relevant to the path of the (sub-)project
+   */
+  static ofType(types: string[], limitToProjectPath: boolean = false) {
+    const cmd = `git log --oneline $LATEST_TAG..HEAD..HEAD -P --grep '^(${types.join(
+      "|"
+    )}){1}(\([\w\-\.]+\))?(!)?: ([\w ])+([\s\S]*)'`;
```
rix0rrr, line 161: Oof! This looks scary to me. Are we sure everyone is going to be nice about their commit messages?

rix0rrr (approved):

src/version.ts
mrgrain: This relies on the fact `bump` is called with the owning (sub-)projects path as `cwd`. It's the only way to call a subprojects `bump` task. If one wanted to have a root level `bump` task that does the work for the subproject this wouldn't work and we'd have to use `ReleasableCommits.exec()`.

I guess we could make it a string instead?

src/version.ts
mrgrain: Added a reference and an even scarier version because `-P` is not guaranteed to exist.

No we can't be sure everyone is good about their commit messages. But in that case (i.e. no enforcement is implemented and for projen the PR linter has been explicitly disabled), user just shouldn't use this feature. They could either use `everyCommit()` or provide a custom regex.

src/version.ts
```diff
@@ -112,3 +124,72 @@ export class Version extends Component {
...
+export class ReleasableCommits {
+  /**
+   * Release every commit
+   *
+   * This will only not release if the most recent commit is tagged with the latest matching tag.
+   *
+   * @param limitToProjectPath Only consider commits relevant to the path of the (sub-)project
+   */
+  static everyCommit(limitToProjectPath: boolean = false) {
+    const cmd = `git log --oneline $LATEST_TAG..HEAD`;
```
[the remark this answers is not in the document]
mrgrain: I couldn't figure out a way to get the count without relying on other unix tools (`wc`), which I'd like to avoid due to OS compatibility (Windows support is already spotty).

I'll have a look if I can find a plumbing command.

src/version.ts
mrgrain: Changed to a string param. I like it better it also unlocks more patterns.

rantoniuk:
It's not obvious (only from the API docs) how to use this directly in `.projenrc.js`, I think I'm missing an import.
I tried to do:
```
  cdkVersion: '2.1.0',
  defaultReleaseBranch: 'main',
  releasableCommits: awscdk.ReleasableCommits.onlyOfType(['feat', 'fix']),
...
```
but it seems that class is not exported directly?

mrgrain:
It is exported at the top-level
```
import { ReleasableCommits } from 'projen';
```
But good shout re docs.

PR #2761, opened 2023-06-27
fix(node): unable to run `projen package` on windows

iliapolo:
TODO

---


src/cdk/jsii-project.package.task.ts
```diff
@@ -0,0 +1,44 @@
...
+  copyFiles(".", repoTempDir, [".git", "node_modules"]);
+  if (fs.existsSync(artifactsDirectory)) {
+    fs.rmdirSync(artifactsDirectory, { recursive: true });
+  }
+  fs.renameSync(repoTempDir, artifactsDirectory);
+} else {
+  child.execSync(`npx projen ${PACKAGE_ALL_TASK_NAME}`, {
+    stdio: ["inherit", "inherit", "inherit"],
+  });
+}
```
mrgrain: If this makes it any easier for you, I strongly believe that the `if CI` check should be done inside the GitHub Actions workflow (at which point it's not an if anymore) and not as part of the task.

But I realize you primarily want to fix this for windows. So this is still better than what we have today.

PR #2776, opened 2023-07-03
docs: document Selective Releases

mrgrain:
---

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -82,6 +82,25 @@ changelogs.
 releaseTrigger: ReleaseTrigger.scheduled({ schedule: '0 17 * * *' }),
 ```
 
+## Selective Releases
+
+It is possible to only bump the version on a subset of commits.
+For example you could only release a new version for every feature and fix that was added to the repo.
+
+```js
+releasableCommits: ReleasableCommits.featuresAndFixes(),
+```
+
+This check only runs according to the release trigger, but serves as an additional check to not create unnecessary releases.
+
+A custom check can be implemented `ReleasableCommits.exec()`.
+This command should return a list of commit hashes that are considered releasable.
+I.e. to not not bump the version, the command must print nothing and exit successfully.
+
+```js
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -82,6 +82,25 @@ changelogs.
 releaseTrigger: ReleaseTrigger.scheduled({ schedule: '0 17 * * *' }),
 ```
 
+## Selective Releases
+
+It is possible to only bump the version on a subset of commits.
+For example you could only release a new version for every feature and fix that was added to the repo.
+
+```js
+releasableCommits: ReleasableCommits.featuresAndFixes(),
```
rantoniuk, line 91: That still doesn't work for me:
```
const { awscdk, ReleasableCommits } = require('projen');
...

...
  releasableCommits: ReleasableCommits.featuresAndFixes(),
}
```
TypeError: Cannot read property 'featuresAndFixes' of undefined

Using import like you suggested also doesn't work because that is not a module:
```
const { awscdk } = require('projen');
import { ReleasableCommits } from 'projen';
```
(apart from the other thing that I don't want to [mix commonjs with ES](https://www.scaler.com/topics/nodejs/require-vs-import-nodejs/). 

Here are different versions I tried:
https://github.com/rantoniuk/cloudwatch-retention-setter/commits/main

mrgrain: Can confirm this works for me:

```js
const { typescript, ReleasableCommits } = require('projen');
const project = new typescript.TypeScriptProject({
  defaultReleaseBranch: 'main',
  name: 'test,
  releasableCommits: ReleasableCommits.featuresAndFixes(),
});
project.synth();
```

--- 

It looks like your projen version isn't up to date. You need at least [v0.71.113](https://github.com/projen/projen/releases/tag/v0.71.113), better [v0.71.115](https://github.com/projen/projen/releases/tag/v0.71.115).

https://github.com/rantoniuk/cloudwatch-retention-setter/blob/main/yarn.lock#L5253

PR #2785, opened 2023-07-08
fix(github): pull-request-lint

kennyg:
- adds githubBaseUrl to allow action to authenticate against github enterprise url

---


mrgrain (changes requested):
How about we always set this value to `${{ github.server_url }}`?

kennyg:
> How about we always set this value to `${{ github.server_url }}`?

I can use `${{ github.api_url }}`.  I'll update the PR.

mrgrain (changes requested):

mrgrain:
Looks good now!

mrgrain (approved):

kennyg:
> Looks good now!

@mrgrain my apologies. I used the wrong variable and didn't notice until I went to this this morning. I used `action_url` when I should have used `api_url`. PR coming up to fix.

mrgrain:
Interesting. It still works on public GitHub. I guess it fails for enterprise?

PR #2791, opened 2023-07-12
feat(typescript): use Node16 module & moduleResolution when appropriate

RomainMuller:
Instead of using `CommonJS` and `Node`, apply the `Node16` module and module resolution settings, which are preferred when targeting node >=16 as it allows dynamically choosing the correct algorithms based on file extension as well as package.json configuration.

Additionally, automatically configure `target` and `lib` based on the configured minimum node version (if present).

---


src/javascript/typescript-config.ts
```diff
@@ -47,28 +47,28 @@ export enum TypeScriptModuleResolution {
...
-  CLASSIC = "classic",
+  CLASSIC = "Classic",
 
   /**
    * Resolution strategy which attempts to mimic the Node.js module resolution strategy at runtime.
    *
    * @see https://www.typescriptlang.org/docs/handbook/module-resolution.html#node
    */
-  NODE = "node",
+  NODE = "Node",
```
mrgrain, line 57:
```suggestion
  NODE = "node",
```

Let's keep all of these lower case as per the docs. AFAIK they are case insensitive anyway. 

https://www.typescriptlang.org/tsconfig#moduleResolution
image

src/javascript/typescript-config.ts
```diff
@@ -78,6 +78,59 @@ export enum TypeScriptModuleResolution {
   BUNDLER = "bundler",
 }
 
+/**
+ * Sets the module system for the program.
+ *
+ * @see https://www.typescriptlang.org/docs/handbook/modules.html
+ */
+export enum TypeScriptModule {
```
mrgrain, line 86: Same here, let's stick with lowercase

mrgrain (changes requested):

package.json
```diff
@@ -121,6 +121,14 @@
...
   "main": "lib/index.js",
   "license": "Apache-2.0",
+  "typesVersions": {
+    "<=3.9": {
+      "lib/*": [
+        "lib/.types-compat/ts3.9/*",
+        "lib/.types-compat/ts3.9/*/index.d.ts"
+      ]
+    }
+  },
```
mrgrain, line 131: How come these are back now? They weren't needed before?

mrgrain:
Is this going to be a breaking change because the `module` field is now typed?

src/javascript/typescript-config.ts
RomainMuller: I found they look better capitalized properly, but yeah... it makes no difference.

package.json
RomainMuller: This is the issue I made https://github.com/aws/jsii-compiler/pull/204 for :)

RomainMuller:
> Is this going to be a breaking change because the `module` field is now typed?

Good question. I suspect possibly maybe? I can revert this to being `string` and replace the enum with a constants helper?

test/javascript/jest.test.ts
```diff
@@ -13,8 +13,10 @@ const compilerOptionDefaults = {
   experimentalDecorators: true,
   inlineSourceMap: true,
   inlineSources: true,
+  target: "ES2019",
   lib: ["es2019"],
-  module: "CommonJS",
+  module: "commonjs",
+  moduleResolution: "nNode",
```
mrgrain, line 19:
```suggestion
  moduleResolution: "node",
```

RomainMuller: 🤦🏻‍♂️

doublecompile:
Side note: I determined the proper configuration to get Projen and Jest working correctly with a TypeScript project where the `package.json` has `type` set to `module` (i.e. ESM instead of CommonJS), [check out this gist](https://gist.github.com/doublecompile/324afb33b6038526705dfebb3b529e42) for an example `.projenrc.ts`.

mrgrain:
Note to myself: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

jsii probably needs to update these as well

mrgrain:
closing stale PR

PR #2804, opened 2023-07-19
feat(upgrade-dependencies): allow customizing dependency types and semantic commit

iliapolo:
Users can now customize specific dependency types (e.g `dev`, `peer`, ...) as well as the semantic commit type (e.g `chore`, `feat`, ...)

This, along with the [`releasableCommits`](https://github.com/projen/projen/blob/2259df0eb7aea3f79e6aa97ede751c414bbdab97/src/version.ts#L131) feature will help us create separate workflows for runtime / non runtime dependencies and prevent non prod dependencies from triggering a release. 

---


src/javascript/upgrade-dependencies.ts
```diff
@@ -24,6 +24,36 @@ import { TaskStep } from "../task-model";
...
+   * `dependencies`
+   */
+  PROD = "prod",
+
+  /**
+   * `bundledDependencies`
+   */
+  BUNDLE = "bundle",
+}
+
```
mrgrain, line 56: Can use the generic types from there instead and map them later to the NodeJS types
https://github.com/projen/projen/blob/2259df0eb7aea3f79e6aa97ede751c414bbdab97/src/dependencies.ts#L287

mrgrain (changes requested):

src/javascript/upgrade-dependencies.ts
iliapolo: Done

src/dependencies.ts
```diff
@@ -322,4 +322,11 @@ export enum DependencyType {
    * Available for Node packages
    */
   OVERRIDE = "override",
+
+  /**
+   * An optional dependency that may or may not be installed by the package manager.
+   *
+   * Available for Node packages
```
mrgrain, line 329:
```suggestion
   * An optional dependency that may be used at runtime if available, but is not required.
   * It is expected to be installed by the consumer.
```

mrgrain (approved):

src/dependencies.ts
iliapolo: > It is expected to be installed by the consumer.

Not accurate. The package manager will install it - but if the installation fails it doesn't fail the install command

PR #2817, opened 2023-07-24
feat: add runner group support

### LINKED ISSUE #2814: Feature Request: Specify GitHub Runners by name
[Relevant GitHub Blog Post.
](https://github.blog/changelog/2022-11-01-github-actions-restrict-workflows-to-specific-runners-using-runner-group-names/)

Currently, the properties: `workflowRunsOn`, `githubOptions.pullRequestLintOptions.runsOn`, `depsUpgradeOptions.workflowOptions.runsOn`, `autoApproveOptions.runsOn`, `staleOptions.runsOns`, accessible in `.projenrc.ts`, accepts a list of strings; GitHub Runner labels. While this does allow _some_ Runner targeting, some teams or organizations may want to target Runners by their Runner Group. For example, Runner Groups may correspond to an specific clusters, environments, or teams, and labels alone may be insufficient.

GitHub introduced the ability to target Runners using Runner Group names (aforementioned blog post). 

Feature Example:
`.projenrc.ts`

```typescript
  workflowRunsOn: {
    group: "Default",
    labels: ["self-hosted", "Linux", "x64"],
  }
```
would render
```yaml
jobs:
  build:
    runs-on:
      group: Default
      labels:
        - self-hosted
        - Linux
        - x64
```

rami-husein:
Fixes #2814  

**[GitHub Actions: Restrict workflows to specific runners using runner group names](https://github.blog/changelog/2022-11-01-github-actions-restrict-workflows-to-specific-runners-using-runner-group-names/)**

`.projenrc.ts` properties:
- `workflowRunsOn`, 
- `githubOptions.pullRequestLintOptions.runsOn`
- `depsUpgradeOptions.workflowOptions.runsOn`
- `autoApproveOptions.runsOn`
- `staleOptions.runsOns`

...now all support specifying Runner Groups, by name.    

```
  workflowRunsOn: {
    group: "Default",
    labels: ["self-hosted", "Linux", "x64"],
  }
  ```
  
  The original support for specifying Labels in an array is still supported:
```
workflowRunsOn: ['ubuntu-latest', 'x86']
```

---

### CODE REVIEW

File: src/build/build-workflow.ts [hunk 1]
```diff
@@ -96,7 +96,7 @@ export interface BuildWorkflowOptions {
    * Github Runner selection labels
    * @default ["ubuntu-latest"]
    */
-  readonly runsOn?: string[];
+  readonly runsOn?: string[] | GroupRunnerOptions;
```
mrgrain, line 99: We can't use Union types unfortunately due to this not working well in other languages.

Can you change this to an additional new property `runsOnGroup` or similar that takes your new type.
And then extend the validation to make sure only one of the properties is set.

src/build/build-workflow.ts
```diff
@@ -112,6 +112,14 @@ export interface BuildWorkflowOptions {
   readonly permissions?: JobPermissions;
 }
 
+/**
+ * Options for WorkflowOptions.runsOn
+ */
+export interface GroupRunnerOptions {
+  readonly group: string;
+  readonly labels: string[];
```
mrgrain, line 120: I believe `labels` is optional. 
```suggestion
  readonly labels?: string[];
```

Source: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-groups-to-control-where-jobs-are-run

src/build/build-workflow.ts
```diff
@@ -431,7 +439,7 @@ export interface AddPostBuildJobTaskOptions {
    * Github Runner selection labels
    * @default ["ubuntu-latest"]
    */
-  readonly runsOn?: string[];
+  readonly runsOn?: string[] | GroupRunnerOptions;
```
mrgrain, line 442: For all the union types, see other comment.

mrgrain (changes requested):
Looks pretty good, thank you! We just have to sort out the union type issue and I'll have a closer look then.

src/build/build-workflow.ts
rami-husein: Ah, I see! I've fixed this in https://github.com/projen/projen/pull/2817/commits/b2ae37e270d5974c1293ab5b6e06f3954897a9df

src/build/build-workflow.ts
rami-husein: Good note! Fixed this in https://github.com/projen/projen/pull/2817/commits/b2ae37e270d5974c1293ab5b6e06f3954897a9df

src/build/build-workflow.ts
rami-husein: Fixed in https://github.com/projen/projen/pull/2817/commits/b2ae37e270d5974c1293ab5b6e06f3954897a9df

src/build/build-workflow.ts
```diff
@@ -164,8 +170,8 @@ export class BuildWorkflow extends Component {
   }
 
   private addBuildJob(options: BuildWorkflowOptions) {
-    this.workflow.addJob(BUILD_JOBID, {
-      runsOn: options.runsOn ?? this.defaultRunners,
+    let testing = {
```
mrgrain, line 173:
```suggestion
    const jobConfig = {
```

src/build/build-workflow.ts
```diff
@@ -321,19 +333,31 @@ export class BuildWorkflow extends Component {
...
     });
   }
 
   private addSelfMutationJob(options: BuildWorkflowOptions) {
+    let runsOnValue;
+    if (options.runsOnGroup) {
+      runsOnValue = { runsOnGroup: options.runsOnGroup };
+    } else {
+      runsOnValue = { runsOn: options.runsOn ?? this.defaultRunners };
+    }
```
mrgrain, line 356: Can we delegate this to the workflow instead of doing on on every workflow implementation?

src/build/build-workflow.ts
```diff
@@ -321,19 +333,31 @@ export class BuildWorkflow extends Component {
 
     steps.push({ run: commands.join("\n") });
 
+    const runsOnValue = options?.runsOnGroup
+      ? { runsOnGroup: options?.runsOnGroup }
+      : { runsOn: options?.runsOn ?? this.defaultRunners };
```
mrgrain, line 338: Can we delegate this to the workflow instead of doing on on every workflow implementation?

src/build/build-workflow.ts
```diff
@@ -267,14 +275,18 @@ export class BuildWorkflow extends Component {
     task: Task,
     options: AddPostBuildJobTaskOptions = {}
   ) {
+    const runsOnValue = options.runsOnGroup
+      ? { runsOnGroup: options.runsOnGroup }
+      : { runsOn: options.runsOn ?? this.defaultRunners };
```
mrgrain, line 280: Can we delegate this to the workflow instead of doing on on every workflow implementation?

src/github/auto-approve.ts
```diff
@@ -104,4 +110,22 @@ export class AutoApprove extends Component {
...
+      throw new Error(
+        "Both 'runsOn' and 'runsOnGroup' cannot be set at the same time"
+      );
+    }
+
+    return options.runsOnGroup
+      ? { runsOnGroup: options.runsOnGroup }
+      : { runsOn: options.runsOn ?? ["ubuntu-latest"] };
+  }
 }
```
mrgrain, line 131: Can we delegate this to the workflow instead of doing on on every workflow implementation?

src/build/build-workflow.ts
```diff
@@ -416,6 +440,24 @@ export class BuildWorkflow extends Component {
...
+    if (options.runsOnGroup && options.runsOn) {
+      throw new Error(
+        "Both 'runsOn' and 'runsOnGroup' cannot be set at the same time"
+      );
+    }
+
+    return options.runsOnGroup
+      ? { runsOnGroup: options.runsOnGroup }
+      : { runsOn: options.runsOn ?? ["ubuntu-latest"] };
+  }
```
mrgrain, line 460: Can we delegate this to the workflow instead of doing on on every workflow implementation?

mrgrain (changes requested):
Sorry for the late review. Finally got around to it.
Mostly about avoid code duplication and  responsibilities. You get the idea.

src/build/build-workflow.ts
rami-husein: Fixed in https://github.com/projen/projen/pull/2817/commits/f22c02fb5fcfba9b01ed50e9e32778c87a331ef8

src/github/workflows.ts
```diff
@@ -411,15 +424,4 @@ function verifyJobConstraints(
       );
     }
   }
-
-  // verify that job has a "runsOn" statement to ensure a worker can be selected appropriately
```
rami-husein, line 415: This verification was moved to [here](https://github.com/projen/projen/pull/2817/files#diff-2ef7bf776940d2981cb6052b7c3095b2a588ce40fd3aa53b6ad1405bf71bc274R35) because other, but related, verification needed to happen earlier on

src/build/build-workflow.ts
rami-husein: For this, and the others:
The logic can't occur after the creation of the Job, because the properties are readonly, and the creation of the Job occurs before the implementation of Workflow. If we want to have default runners, this logic needs to happen first.

However, in the interest of the DRY principle, I've moved the logic and verification into the [same file](https://github.com/projen/projen/blob/f22c02fb5fcfba9b01ed50e9e32778c87a331ef8/src/runner-options.ts) that holds the GroupRunnerOptions interface.

src/runner-options.ts
```diff
@@ -0,0 +1,48 @@
...
+    ? { workflowRunsOn: workflowRunsOn }
+    : { workflowRunsOn: ["ubuntu-latest"] };
+}
+
+function verifyJobConstraints(
+  runsOn?: string[],
+  runsOnGroup?: GroupRunnerOptions
+): void {
+  if (runsOn && runsOnGroup) {
+    throw new Error(
```
rami-husein, line 40: This may be not the best error to throw. Let me know if you have an idea.

src/github/workflows.ts
```diff
@@ -348,13 +349,25 @@ function renderJobs(
   }
 }
 
-function arrayOrScalar<T>(arr: T[] | undefined): T | T[] | undefined {
-  if (arr == null || arr.length === 0) {
+function arrayOrScalar<T>(
+  arr: T[] | T | GroupRunnerOptions | GroupRunnerOptions[] | undefined
+): T | T[] | GroupRunnerOptions | GroupRunnerOptions[] | undefined {
```
mrgrain, line 354: This should not be neccessary. The function is already generic. What error do you get?

src/runner-options.ts
mrgrain: Please make sure to mention the property names. I think the exact values here might be less relevant.

mrgrain (changes requested):

src/github/workflows.ts
rami-husein: I think was a lack of knowledge of how the typescript server resolves generics, on my part. I kept getting an error that type `GroupRunnerOptions` was not assignable to `string[]`. I thought the solution was to add `GroupRunnerOptions` to the types of the function.

After some more reading, I believe the Typescript server infers the type of the generic from a prior use and is incompatible with union types. So instead of calling `arrayOrScalar(job.runsOnGroup ?? job.runsOn)` I should have called `arrayOrScalar(job.runsOnGroup) ?? arrayOrScalar(job.runsOn)` so only a single type is given. This change was made in https://github.com/projen/projen/pull/2817/commits/a9245047a4df48ac72a931096f11bc4843ad2fc7

I also removed the `GroupRunnerOptions` stuff from the `arrayOrScalar()` function, but left in the Array type check, because `GroupRunnerOptions` doesn't have a `.length` property.

mrgrain:
Looks good now! Thanks for the contribution. Once the build is green, we are good to go!

mrgrain (approved):

PR #2820, opened 2023-07-26
fix(cli): improve handling of defaultBranch

trilom:
Consuming the construct `NodeProject` has a required property _defaultReleaseBranch_ (with default) of **main**.
If I would like to create a new `NodeProject` but rely on my environment's git config I get inconsistent results.

If I had the following .gitconfig at $HOME/.gitconfig
```config
[init]
	defaultBranch = trunk
```
Then attempt to run `projen new node --defaultReleaseBranch=trunk` I would receive github workflows for the **trunk** branch as expected, but the default branch for the git repository is always **main**.

This enhancement uses main if there is no globally set defaultReleaseBranch, or the config check returns with an error thus resolving the need to check versions.

After this enhancement, when I run the previous command `projen new node --defaultReleaseBranch=trunk` with my global git config set my default branch is called **trunk** as I would expect instead of **main** which was previously hardcoded with no options.

---


mrgrain (changes requested):
I love this change!

Can you extend this to add a new [macro](https://github.com/projen/projen/blob/e5899dd04a575209424a08fe90bde99e07ac6c7b/src/cli/macros.ts#L11) and change the default value for `defaultReleaseBranch` to it? Then you don't even have to provide the option anymore!

trilom:
> I love this change!
> 
> Can you extend this to add a new [macro](https://github.com/projen/projen/blob/e5899dd04a575209424a08fe90bde99e07ac6c7b/src/cli/macros.ts#L11) and change the default value for `defaultReleaseBranch` to it? Then you don't even have to provide the option anymore!

Sorry for the force push on this.

I added the macro as suggested but I don't feel _that_ happy about it and I think the majority of that centers around my need (using for other git repository sources instead of github).  I'd be happier with this solution if Project object took into account the default branch instead of Node Project.

The test originally created a "Project" but since this value is resolved only when the project has the defaultBranchName value (node, JSII etc) but since Project creates a "gitattributes" and other git things, I believe this should be there but that is objectively a heavier lift because a lot of tests use the "defaultBranchName" value and Project is closer to the core of the Projen project is than Node so I am frankly more hesitant on this change.

src/cli/cmds/new.ts
```diff
@@ -440,31 +436,17 @@ async function initProject(
...
-
   if (fs.existsSync(path.join(baseDir, "package.json")) && args.post) {
     exec("npm run eslint --if-present", { cwd: baseDir });
   }
-
   if (args.git) {
     const git = (cmd: string) => exec(`git ${cmd}`, { cwd: baseDir });
-    const gitversion: string = getGitVersion(
-      execCapture("git --version", { cwd: baseDir }).toString()
+    git(`init -b ${props.defaultReleaseBranch ?? "main"}`);
```
mrgrain, line 444: Yes, I guess semantically the issue is that `defaultReleaseBranch` is no the same as the default branch. 🤔

src/cli/cmds/new.ts
```diff
@@ -440,31 +436,17 @@ async function initProject(
...
-      logging.debug("default branch name set to main");
-    } else {
-      git("init");
-      git("add .");
-      git('commit --allow-empty -m "chore: project created with projen"');
-      logging.debug(
-        "older version of git detected, changed default branch name to main"
-      );
-      git("branch -M main");
-    }
```
mrgrain, line 467: Why is this no longer needed?

trilom: This was a condition set up to handle older Git versions.  The response is undefined on older versions so "main" will be used.

src/cli/cmds/new.ts
mrgrain: Okay, sorry about going back and forth on this but we should not use `defaultReleaseBranch` here. It is not was the property means. Instead we can use your new code to always select the configured default branch. I think it's fine to always use the the git config for this and fall back to `main` if not set (which is the case for version pre 2.28.0)

src/cli/cmds/new.ts
mrgrain: Yes, but it is required because versions <2.28.0 don't support `-b`:
see: https://git-scm.com/docs/git-init/2.27.0

mrgrain (changes requested):

PR #2835, opened 2023-07-29
fix(upgrade-dependencies): upgrade command doesn't consider requested types

iliapolo:
Given 

```ts
depsUpgradeOptions: {
  types: [DependencyType.PEER],
}
```

Would produce:

```console
npm-check-updates --dep peer --upgrade --target=minor
yarn upgrade
```

However, when `yarn upgrade` executes, the peer deps aren't upgraded because peer dependencies aren't installed by default. This causes [failures](https://github.com/cdk8s-team/cdk8s-hasura/actions/runs/5698912124/job/15447288139?pr=177) such as:

```console
build » compile | jsii --silence-warnings=reserved-word
Error: Error: Declared dependency on version ^2.24.0 of cdk[8](https://github.com/cdk8s-team/cdk8s-hasura/actions/runs/5698912124/job/15447288139?pr=177#step:5:9)s, but version 2.23.0 was found
```

Projen actually automatically adds peer dependencies as dev dependencies as well. This means we need to also produce: 

```console
npm-check-updates --dep dev --upgrade --target=minor --filter {peers}
```

Also, `yarn upgrade` is not the right command to run here because it will upgrade ALL transitive dependencies, instead of just for peers. What we need is `yarn upgrade {peers}`.

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -816,9 +816,8 @@ export class NodePackage extends Component {
     function upgradePackages(command: string) {
       return () => {
         return `${command} ${project.deps.all
-          .filter(
-            (d) => d.type !== DependencyType.OVERRIDE && types.includes(d.type)
-          )
+          .filter((d) => d.type !== DependencyType.OVERRIDE)
+          .filter((d) => types.includes(d.type))
           .map((d) => d.name)
           .filter((d) => (include ? include.includes(d) : true))
           .filter((d) => !exclude.includes(d))
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -808,21 +808,17 @@ export class NodePackage extends Component {
    * Render a package manager specific command to upgrade all requested dependencies.
    */
   public renderUpgradePackagesCommand(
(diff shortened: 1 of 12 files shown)


src/javascript/node-package.ts
```diff
@@ -808,21 +808,16 @@ export class NodePackage extends Component {
...
    */
   public renderUpgradePackagesCommand(
+    types: DependencyType[],
     exclude: string[],
     include?: string[]
   ): string {
     const project = this.project;
     function upgradePackages(command: string) {
       return () => {
-        if (exclude.length === 0 && !include) {
```
iliapolo, line 817: The aesthetics are no longer worth the added complexity.

src/javascript/node-package.ts
```diff
@@ -808,21 +808,16 @@ export class NodePackage extends Component {
...
-          // request to upgrade all packages
-          // separated for asthetic reasons.
```
iliapolo, line 819: The aesthetics are no longer worth the added complexity.

mrgrain (approved):

PR #2845, opened 2023-08-03
feat(upgrade-dependencies)!: switch to single commands that only use `--filter`

iliapolo:
Currently, we leverage 3 separate features in order to support all various configurations that an upgrade workflow can accept:

- `--reject` is used to facilitate the [`exclude`](https://github.com/projen/projen/blob/main/src/javascript/upgrade-dependencies.ts#L36) property.
- `--filter` is used to facilitate the [`include`](https://github.com/projen/projen/blob/main/src/javascript/upgrade-dependencies.ts#L43) property.
- `--dep` is used to facilitate the [`types`](https://github.com/projen/projen/blob/main/src/javascript/upgrade-dependencies.ts#L97) property.

This, in combination of the fact that `peerDependencies` are recorded as `devDependencies` as well, and `bundledDependencies` are recored as regular `dependencies` as well, created to much snowflaking in the code.

Since projen has knowledge of all project dependencies, we can actually make it a lot simpler and drop the use of `--dep` and `--reject`. Instead, we only build `--filter` list that is relevant to the specific dependency types that are requested.
That way, we only have positive commands that are also much easier to reason through. 

BREAKING CHANGE: the `renderUpgradePackagesCommand` has been removed from `NodePackage`. It should never have been used. Instead, configure the appropriate `depsUpgradeOptions` to control the upgrade task. If you render the upgrade command for a different purpose, you'll have to construct the command yourself according to the appropriate package manager. 

---


src/javascript/node-package.ts
```diff
@@ -807,21 +807,10 @@ export class NodePackage extends Component {
   /**
    * Render a package manager specific command to upgrade all requested dependencies.
    */
-  public renderUpgradePackagesCommand(
-    types: DependencyType[],
-    exclude: string[],
-    include?: string[]
-  ): string {
-    const project = this.project;
+  public renderUpgradePackagesCommand(include?: string[]): string {
```
mrgrain, line 810: This is a breaking change. We can do that. But I also wonder if we should instead 
- add a new private method that renders the command, only taking a list of packages to upgrade.
- deprecate the existing method.
- Make it use the new private method internally.

.projen/tasks.json
```diff
@@ -384,25 +384,19 @@
...
         },
         {
-          "exec": "yarn upgrade @types/conventional-changelog-config-spec @types/glob @types/ini @types/jest @types/node @types/semver @types/yargs @typescript-eslint/eslint-plugin @typescript-eslint/parser all-contributors-cli esbuild eslint-config-prettier eslint-import-resolver-node eslint-import-resolver-typescript eslint-plugin-import eslint-plugin-prettier eslint jest jest-junit jsii-diff jsii-docgen jsii-pacmak jsii-rosetta jsii npm-check-updates prettier standard-version ts-jest typescript @iarna/toml case chalk comment-json conventional-changelog-config-spec fast-json-patch glob ini semver shx xmlbuilder2 yaml yargs"
+          "exec": "yarn upgrade @iarna/toml case chalk conventional-changelog-config-spec fast-json-patch glob ini semver shx xmlbuilder2 yaml yargs"
```
mrgrain, line 399: I didn't expect to still see 2 ncu and 2 yarn upgrade commands. 🤔

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -27,17 +27,39 @@ test("allows configuring specific dependency types", () => {
...
+    [
+      {
+        "exec": "yarn upgrade npm-check-updates",
+      },
+      {
+        "exec": "npm-check-updates --upgrade --target=minor --filter=some-dep",
+      },
+      {
+        "exec": "npm-check-updates --upgrade --target=minor --filter=jest,jest-junit,npm-check-updates,projen,standard-version",
+      },
```
mrgrain, line 45: Same question - why twice?

src/javascript/node-package.ts
iliapolo: How common do you think its being used? it not exactly the golden path...

> Make it use the new private method internally.

Preserving the functionality of considering types you mean? I would like to move away from that...I can bring this back to accept `include` and `exclude`, like it was before https://github.com/projen/projen/pull/2835, and move the handling of `exclude` to this method exclusively.

.projen/tasks.json
iliapolo: Yeah i started with one command and then separated them based on dependency type so its easier to reason about. Granted, the command doesn't tell you which one belongs to which type (projen dependency type, not npm). 

I can add an `echo Upgrading ${type} dependencies` before each if you think its worth it :)

test/__snapshots__/integ.test.ts.snap
```diff
@@ -1159,25 +1159,31 @@ tsconfig.tsbuildinfo
...
         {
-          "exec": "yarn install --check-files"
+          "exec": "yarn upgrade @types/jest @types/node @typescript-eslint/eslint-plugin @typescript-eslint/parser aws-sdk eslint-import-resolver-node eslint-import-resolver-typescript eslint-plugin-import eslint jest-junit jest jsii-diff jsii-docgen jsii-pacmak npm-check-updates projen standard-version ts-jest typescript"
+        },
+        {
+          "exec": "yarn upgrade @aws-cdk/aws-apigateway @aws-cdk/aws-cloudwatch-actions @aws-cdk/aws-cloudwatch @aws-cdk/aws-dynamodb @aws-cdk/aws-ecs-patterns @aws-cdk/aws-ecs @aws-cdk/aws-elasticloadbalancingv2 @aws-cdk/aws-events-targets @aws-cdk/aws-events @aws-cdk/aws-lambda @aws-cdk/aws-rds @aws-cdk/aws-sns-subscriptions @aws-cdk/aws-sns @aws-cdk/aws-sqs @aws-cdk/core constructs"
+        },
+        {
+          "exec": "yarn upgrade @aws-cdk/aws-apigateway @aws-cdk/aws-cloudwatch-actions @aws-cdk/aws-cloudwatch @aws-cdk/aws-dynamodb @aws-cdk/aws-ecs-patterns @aws-cdk/aws-ecs @aws-cdk/aws-elasticloadbalancingv2 @aws-cdk/aws-events-targets @aws-cdk/aws-events @aws-cdk/aws-lambda @aws-cdk/aws-rds @aws-cdk/aws-sns-subscriptions @aws-cdk/aws-sns @aws-cdk/aws-sqs @aws-cdk/core"
```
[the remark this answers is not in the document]
iliapolo: Not exactly the same. The first also has `constructs`. Though i'm not sure which type would create this difference..checking

test/javascript/upgrade-dependencies.test.ts
iliapolo: https://github.com/projen/projen/pull/2845#discussion_r1283069338

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -323,8 +421,51 @@ test("upgrade task created without projen defined versions at NodeProject", () =
...
+        "projen",
+        "standard-version",
+      ],
+    },
+  });
+  const tasks = synthSnapshot(project)[TaskRuntime.MANIFEST_FILE].tasks;
+  expect(tasks.upgrade.steps[0].exec).toStrictEqual(
+    "echo No dependencies to upgrade."
   );
 });
```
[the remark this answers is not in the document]
iliapolo: Yes. And yes ouf of scope :)

src/javascript/node-package.ts
mrgrain: We totally can make a breaking change. It would just be _nicer_ to deprecate it first if we can easily. It's probably not used a lot. 

I don't think the method should have been public in the first place.

If we keep it breaking, can you please update the PR description

test/javascript/upgrade-dependencies.test.ts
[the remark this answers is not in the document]
mrgrain: What happens today? Does it run the ncu without any packages?

test/__snapshots__/integ.test.ts.snap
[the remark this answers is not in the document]
iliapolo: So apparently this is how `ConstructLibraryAws` sets up dependencies.

```json
  "dependencies": {
    "@aws-cdk/aws-apigateway": "^1.75.0",
    "@aws-cdk/aws-cloudwatch": "^1.75.0",
    "@aws-cdk/aws-cloudwatch-actions": "^1.75.0",
    "@aws-cdk/aws-dynamodb": "^1.75.0",
    "@aws-cdk/aws-ecs": "^1.75.0",
    "@aws-cdk/aws-ecs-patterns": "^1.75.0",
    "@aws-cdk/aws-elasticloadbalancingv2": "^1.75.0",
    "@aws-cdk/aws-events": "^1.75.0",
    "@aws-cdk/aws-events-targets": "^1.75.0",
    "@aws-cdk/aws-lambda": "^1.75.0",
    "@aws-cdk/aws-rds": "^1.75.0",
    "@aws-cdk/aws-sns": "^1.75.0",
    "@aws-cdk/aws-sns-subscriptions": "^1.75.0",
    "@aws-cdk/aws-sqs": "^1.75.0",
    "@aws-cdk/core": "^1.75.0"
  },

  "peerDependencies": {
    "@aws-cdk/aws-apigateway": "^1.75.0",
    "@aws-cdk/aws-cloudwatch": "^1.75.0",
    "@aws-cdk/aws-cloudwatch-actions": "^1.75.0",
    "@aws-cdk/aws-dynamodb": "^1.75.0",
    "@aws-cdk/aws-ecs": "^1.75.0",
    "@aws-cdk/aws-ecs-patterns": "^1.75.0",
    "@aws-cdk/aws-elasticloadbalancingv2": "^1.75.0",
    "@aws-cdk/aws-events": "^1.75.0",
    "@aws-cdk/aws-events-targets": "^1.75.0",
    "@aws-cdk/aws-lambda": "^1.75.0",
    "@aws-cdk/aws-rds": "^1.75.0",
    "@aws-cdk/aws-sns": "^1.75.0",
    "@aws-cdk/aws-sns-subscriptions": "^1.75.0",
    "@aws-cdk/aws-sqs": "^1.75.0",
    "@aws-cdk/core": "^1.75.0",
    "constructs": "^3.2.27"
  },

```

.projen/tasks.json
mrgrain: I'm not sure its's worth splitting the different types out.... How do you see this helping with reasoning?
I don't think I understand why it's been split in the first place.

test/javascript/upgrade-dependencies.test.ts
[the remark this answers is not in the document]
iliapolo: It just doesn't run ncu. It simply runs `echo No dependencies to upgrade`.

mrgrain (approved):

PR #2859, opened 2023-08-14
fix(file): incorrect projen command in markers

tinovyatkin:
Currently projen marker contains hard coded command to run projen `npx projen`.  However, it's not always the right command, particularly it looks completely wrong in JSII projects (such as Python). This PR fixes that by using proper existing `projenCommand` field in markers.

On side case of this change is generated files for projen itself, so this PR includes seconds commit with updates to to such files generated as result of `yarn build`

---


mrgrain (changes requested):
This looks good. When I experimented with this I came across a scenario where it didn't work. I'll need to dig that out to remind myself.

mrgrain (approved):
Approving. I think I just didn't like what this does to this repo. But it's actually accurate.

tinovyatkin:
@mrgrain label is blocking merge?

mrgrain:
> @mrgrain label is blocking merge?

Ah and the mergify config has changed, so I need to click the button manually. 🤦‍♂️

Issue #2875, opened 2023-08-22
Upgrade dependencies ignores peer dependency constraints 


mrgrain:
The upgrade dependency command currently does not obey peer dependency version constraints, because the underlying command `npm-check-updates` does not set the `--peer` flag. This leads to situation where the upgrade workflow incorrectly upgrades to a later version that is constrained as a peer dependency by one of it's dependencies:

See the difference in behavior:
```console
$ ncu npx npm-check-updates
Checking package.json
[====================] 2/2 100%

 ncu-test-peer-update     1.0.0  →  1.1.0
 ncu-test-return-version  1.0.0  →  2.0.0

Run npx npm-check-updates -u to upgrade package.json
```

```console
$  ncu npx npm-check-updates --peer
Checking package.json
[====================] 2/2 100%
[====================] 2/2 100%

 ncu-test-peer-update     1.0.0  →  1.1.0
 ncu-test-return-version  1.0.0  →  1.1.0

Ignored incompatible updates (peer dependencies):

 ncu-test-return-version  1.0.0  →  2.0.0  reason: ncu-test-peer-update requires 1.1.x

Run npx npm-check-updates --peer -u to upgrade package.json
```

Issue #2874, opened 2023-08-22
Peer dependencies are not updated anymore


mrgrain:
With https://github.com/projen/projen/pull/2845 we moved to a model of running a single command to update all deps.
Previously we had a separate command for every dependency type. As part of the simplification we accidentally dropped upgrading peer deps, because we completely dropped the use of `--deps` which by default does not include peer.

Issue #1601, opened 2022-02-08
Disabling `depsUpgrade` turns off related tasks


moltar:
When disabling `depsUpgrade` options the `projen upgrade` task is nuked as well.

However, the docs state:

> Use github workflows to handle dependency upgrades.

I totally assumed it will just remove the GH workflow, but keep the tasks for my manual local use.

cgatt:
I'm looking at ways to address this issue, however the current behaviour is a bit tangled due to the mutually exclusive options between `dependabot` and `depsUpgrade`. Will look into it further and raise a PR for a breaking change that moves dependabot setup for Node projects under the UpgradeDependencies class for consistency (As it is a form of dependency upgrade management and should be parallel to the workflow approach) and provides a more intuitive options interface to allow setup of local task, github workflow or dependabot independently.

In the meantime, the workflow can be disabled without removing the local task by including the following in your project config:
```ts
{
  depsUpgrade: true,
  depsUpgradeOptions: {
    workflow: false,
  },
}
```

moltar:
Thanks, Chris!

A bit off-topic, but don't know where the topic is:

> Will look into it further and raise a PR for a breaking change that moves dependabot setup for Node projects under the UpgradeDependencies class for consistency

Would be great if we could have a pluggable dependency update strategy. E.g. I use Renovate for deps, and currently, there is no support for it in projen. But I think it can be added because Renovate has an open-source solution, [that can run via GitHub Action as well](https://github.com/renovatebot/github-action).

PR #2883, opened 2023-08-25
feat(upgrade-dependencies)!: allow to configure version upgrade `target` and `satisfyPeerDependencies`

mrgrain:
Adds the `target` and `satisfyPeerDependencies` options to the `UpgradeDependencies` component.
`target` allows to define the target version to upgrade dependencies to (default: `"minor"`).
`satisfyPeerDependencies` will check peer dependencies of installed packages and filter updates to compatible versions (default: `true`).

BREAKING CHANGE: When upgrading dependencies, installed packages are now checked for peer dependencies and updates will be filtered to satisfy respective version constraints. The previous behavior was to always upgrade to the latest minor version and ignore incompatible peer dependency constraints. To use the previous behavior, set `satisfyPeerDependencies: false`.

BREAKING CHANGE: In [v0.72.0](https://github.com/projen/projen/releases/tag/v0.72.0) a bug was accidently introduced that causes peer dependencies to be always excluded from version upgrades. This releases fixes the issue. To not upgrade peer dependencies, only list desired dependency types in `types`.

Fixes https://github.com/projen/projen/issues/2874
Fixes https://github.com/projen/projen/issues/2875
Closes https://github.com/projen/projen/issues/1601

---


mrgrain:
`pr/do-not-merge` because I want to get a few breaking changes out in a single release.

src/javascript/upgrade-dependencies.ts
```diff
@@ -42,6 +42,27 @@ export interface UpgradeDependenciesOptions {
...
   readonly include?: string[];
 
+  /**
+   * Determines the target version to upgrade dependencies to.
+   *
+   * @see https://github.com/raineorshine/npm-check-updates#target
+   *
+   * @default "minor"
+   */
+  readonly target?: string;
```
iliapolo, line 52: Why no enum?

src/javascript/upgrade-dependencies.ts
```diff
@@ -225,7 +250,9 @@ export class UpgradeDependencies extends Component {
     const ncuCommand = [
       "npm-check-updates",
       "--upgrade",
-      "--target=minor",
+      `--target=${this.upgradeTarget}`,
+      `--${this.satisfyPeerDependencies ? "peer" : "no-peer"}`,
+      `--dep=${this.renderNcuDependencyTypes(this.depTypes)}`,
```
iliapolo, line 255: Why do we need this if we explicitly pass the packages to update?

src/javascript/upgrade-dependencies.ts
mrgrain: You can define any version tag there

src/javascript/upgrade-dependencies.ts
mrgrain: Because `peer` deps are not included by default.

So we now have a situation where I can specify `types: [DependencyType.PEER]` and my peer dependency doesn't get updated. Even worse: It will still upgrade a sev dependency of the same name, even though I said only peer deps please.

iliapolo: I'm still not following. If you say `types: [DependencyType.PEER]`, the code should filter out the packages that have this type, and explicitly put them in the filter list...

Can you elaborate even more? :)

vladcos:
Can confirm build crashes across all my projen-based projects due to peerDepencencies incompatibility after a running the upgrade command.

mrgrain:
> Can confirm build crashes across all my projen-based projects due to peerDepencencies incompatibility after a running the upgrade command.

Thanks for the report! Just to confirm, this happens with the current release and this PR should fix it. Correct?

src/javascript/upgrade-dependencies.ts
mrgrain: Yes thats right. The code will go through the list and find all peer dependencies.
Let's say we have a peer dependency on `packageA`.

Previously the generated command would be something like:
```
npm-check-updates --upgrade --target=minor --filter=packageA
```
However because the default value for the `--dep` flag is `prod,dev,optional`, no upgrades would be done at all.

But let's extend that scenario a bit and assume that we also have a dev dependency on `packageA` which is quite common. Now the generated command would still be exactly the same. And because of the default we would actually upgrade the dev dependency - despite explicitly specifying that only peer deps should be upgraded.

WIth this PR, the generated command would instead be

```
npm-check-updates --upgrade --target=minor --deps=peer --filter=packageA
```

and only the peer dependency would be upgrades, exactly as specified.

iliapolo: > However because the default value for the --dep flag is prod,dev,optional, no upgrades would be done at all.

Ah, that was my missing part, thanks 👍

iliapolo (approved):

vladcos:
> > Can confirm build crashes across all my projen-based projects due to peerDepencencies incompatibility after a running the upgrade command.
> 
> Thanks for the report! Just to confirm, this happens with the current release and this PR should fix it. Correct?

I haven't had a change to test it on my projects, but indeed those projects on 0.72 ver have the reported problem with the peer dependencies.

PR #2911, opened 2023-09-05
fix(typescript-project): add property to original ts options

agdimech:
This PR simply stores the original tsOptions as a public variable within the TypeScriptProject so it can be referenced later.

This is needed by my [`Monorepo`](https://github.com/aws/aws-pdk/blob/mainline/packages/monorepo/src/components/nx-configurator.ts#L380) construct as I want to set defaults to all Typescript subprojects of  `NodeNext` *unless* the user has specified settings for tsConfig which I would want to respect. Currently I have no way of knowing if the user has overriden these settings as I don't have a reference to the original options.

---

### DIFF
--- a/src/typescript/typescript.ts
+++ b/src/typescript/typescript.ts
@@ -177,6 +177,11 @@ export class TypeScriptProject extends NodeProject {
   public readonly testdir: string;
 
   /**
+   * Reference to passed in options.
+   */
+  public readonly options: TypeScriptProjectOptions;
+
+  /**
    * The "watch" task.
    */
   public readonly watchTask: Task;
@@ -197,6 +202,7 @@ export class TypeScriptProject extends NodeProject {
       },
     });
 
+    this.options = options;
     this.srcdir = options.srcdir ?? "src";
     this.libdir = options.libdir ?? "lib";
 
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -3704,7 +3704,7 @@ __Extends__: [cdk.ConstructLibrary](#projen-cdk-constructlibrary)
 new awscdk.AwsCdkConstructLibrary(options: AwsCdkConstructLibraryOptions)
 ```
 
-* **options** (<code>[awscdk.AwsCdkConstructLibraryOptions](#projen-awscdk-awscdkconstructlibraryoptions)</code>)  *No description*
+* **options** (<code>[awscdk.AwsCdkConstructLibraryOptions](#projen-awscdk-awscdkconstructlibraryoptions)</code>)  Reference to passed in options.
   * **name** (<code>string</code>)  This is the name of your project. 
   * **commitGenerated** (<code>boolean</code>)  Whether to commit the managed files by default. __*Default*__: true
   * **gitIgnoreOptions** (<code>[IgnoreFileOptions](#projen-ignorefileoptions)</code>)  Configuration options for .gitignore file. __*Optional*__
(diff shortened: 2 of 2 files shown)


src/typescript/typescript.ts
```diff
@@ -176,6 +176,11 @@ export class TypeScriptProject extends NodeProject {
    */
   public readonly testdir: string;
 
+  /**
+   * Reference to passed in options.
+   */
+  public readonly options: TypeScriptProjectOptions;
```
mrgrain, line 182: At least we have to give this a different name, since it will confuse people what this exactly is.

mrgrain (changes requested):
I don't think this really solves your problem in a generic way.

Imagine this project type:
```ts
class ModernDefaultsTypeScriptProject extends TypeScriptProject {
  public constructor(options: TypeScriptProjectOptions) {
    super({
      tsconfig: {
        compilerOptions: {
          target: 'foo',
          module: 'bar',
        }
      }
    });
  }
}
```

This would satisfy the [instanceof TypeScriptProject](https://github.com/aws/aws-pdk/blob/mainline/packages/monorepo/src/components/nx-configurator.ts#L374C11-L374C44) check, however `compilerOptions.target` & `compilerOptions.module` would appear as "set by the user".

In fact, the included `NextJsTypeScriptProject` behaves exactly like that already.


Could you provide a `ModernDefaultsTypeScriptProject` as part of your offering?

src/typescript/typescript.ts
agdimech: Sure will change to _userSettings and make private.

agdimech:
> I don't think this really solves your problem in a generic way.
> 
> Imagine this project type:
> 
> ```ts
> class ModernDefaultsTypeScriptProject extends TypeScriptProject {
>   public constructor(options: TypeScriptProjectOptions) {
>     super({
>       tsconfig: {
>         compilerOptions: {
>           target: 'foo',
>           module: 'bar',
>         }
>       }
>     });
>   }
> }
> ```
> 
> This would satisfy the [instanceof TypeScriptProject](https://github.com/aws/aws-pdk/blob/mainline/packages/monorepo/src/components/nx-configurator.ts#L374C11-L374C44) check, however `compilerOptions.target` & `compilerOptions.module` would appear as "set by the user".
> 
> In fact, the included `NextJsTypeScriptProject` behaves exactly like that already.
> 
> Could you provide a `ModernDefaultsTypeScriptProject` as part of your offering?

That is actually what I want :p I need a way to determine if a user has set these particular settings otherwise I will just apply modern defaults to all ts sub projects.

To give a more clear example:

If I have a monorepo with subproject a and b as children both of type tsProject: if subproject b has an override for moduleResolution and module, while subproject a has defaults (undefined for both) then this logic would apply nodenext to only subproject a. 

If subproject a actually wanted the default module and moduleResolution, they can provide explicit values for them I.e: node and I will not override them.

mrgrain:
Now that we are using constructs internally, maybe we can store this information in the node metadata instead?

tinovyatkin:
> This is needed by my [`Monorepo`](https://github.com/aws/aws-pdk/blob/mainline/packages/monorepo/src/components/nx-configurator.ts#L380) construct as I want to set defaults to all Typescript subprojects of `NodeNext` _unless_ the user has specified settings for tsConfig which I would want to respect. Currently I have no way of knowing if the user has overriden these settings as I don't have a reference to the original options.

This actually can be implemented with JsonPatch now:

```ts
 tsconfig.file.patch(JsonPatch.test('/compilerOptions/module', undefined), JsonPatch.add('/compilerOptions/module', 'NodeNext'))
```

PR #2917, opened 2023-09-09
fix(cli): respect git init.defaultBranch

tinovyatkin:
`projen new` by default initialises a git repository using hard coded default branch name `main`. Some organisations / users may have different defaults for git branch (like `mainline`) and so git since version 2.28 provides global option [init.defaultBranch](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch). This PR makes `projen new` to respect that setting when present.

---

### DIFF
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -457,6 +457,7 @@ async function initProject(
       execCapture("git --version", { cwd: baseDir }).toString()
     );
     logging.debug("system using git version ", gitversion);
+    // `git config init.defaultBranch` and `git init -b` are only available since git 2.28.0
     if (gitversion && semver.gte(gitversion, "2.28.0")) {
       const defaultGitInitBranch =
         execOrUndefined("git config init.defaultBranch", {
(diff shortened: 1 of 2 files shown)


tinovyatkin:
Noticed #2820 but it seems to be stale since July (and with broader changes) and we love this feature to be merged.

src/cli/cmds/new.ts
```diff
@@ -452,10 +458,16 @@ async function initProject(
     );
     logging.debug("system using git version ", gitversion);
     if (gitversion && semver.gte(gitversion, "2.28.0")) {
-      git("init -b main");
+      const defaultGitInitBranch =
+        execOrUndefined("git config init.defaultBranch", {
+          cwd: baseDir,
+        })?.trim() ||
+        args.defaultReleaseBranch ||
```
mrgrain, line 465:
```suggestion
        args.defaultReleaseBranch ||
```

Would you be okay to leave this out? I don't really want to mix project settings with cli settings.
The way to get `defaultReleaseBranch` involved is imo to update the default for it to also respect the git setting.

mrgrain (changes requested):
can you add a comment explaining why `main` doesn't need to change here 
(It's because git < 2.28 doesn't have `git config init.defaultBranch`)
```
git("branch -M main");
```

src/cli/cmds/new.ts
tinovyatkin: yes, sure

mrgrain (approved):
Thanks for your contribution! I'm glad we got there in the end. 🚀 
Also thanks to @trilom for the original PR.

PR #2926, opened 2023-09-14
feat(cli): interactive project creation (RFC #1365)

vladcos:
Experimenting with implementing the RFC https://github.com/projen/projen/blob/main/rfcs/project-creation-prompt.md

https://github.com/projen/projen/assets/135833592/bbe1d880-6859-4fa8-aa6f-f5f6c9e28f0f

I think it will improve the UX and lower the barrier of entry. I, as a relatively new user or projen, needed to invest time and effort in order to perform basic Projen configuration and turn on/off options i wanted, i had to read the whole super long API page just to find a couple of options I wanted. I think a little bit of guidance when scaffolding a new project might improve the popularity of this awesome tool.

For now I have hardcoded some options and had to integrate the yargs cli class with the prompt library in a hackish way because yargs lacks native support for it.

A few problems that need to be solved and some input from the community would be appreciated:
1.  Library choice: https://github.com/terkelg/prompts recommended in the RFC is poorly maintained. I found that https://github.com/SBoudrias/Inquirer.js is the most active and lightweight lib (although very basic).
2. Option source: It was recommended in the RFC to use the options flagged with @featured flag in the prompts. I think the options marked with this doctag are not very useful for prompts. I would suggest adding a new flag: i.e. @prompt 

---

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -135,6 +135,10 @@
       "type": "bundled"
     },
     {
+      "name": "@inquirer/prompts",
+      "type": "bundled"
+    },
+    {
       "name": "case",
       "type": "bundled"
     },
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -384,13 +384,13 @@
           "exec": "yarn upgrade npm-check-updates"
         },
         {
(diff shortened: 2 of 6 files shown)


mrgrain:
I love this! 😍  Looking forward to having this in projen!
The RFC is quite old now, and will need some overhauling.

> Library choice: https://github.com/terkelg/prompts recommended in the RFC is poorly maintained. I found that https://github.com/SBoudrias/Inquirer.js is the most active and lightweight lib (although very basic).

`prompts` doesn't look very maintained at the moment. 
`inquirer` looks good and is lightweight, but has comparatively many dependencies.
`enquirer` is my current favorite. Maintained, no additional dependencies, still small enough.

> Option source: It was recommended in the RFC to use the options flagged with `@featured` flag in the prompts. I think the options marked with this doctag are not very useful for prompts. I would suggest adding a new flag: i.e. `@prompt`

I'm okay with an extra tag. Although it makes me wonder what the difference between `@featured` and `@prompt` is supposed to be. 🤔 

The best starting point in my opinion is to prompt for required, but not provided options.  We don't even need to introduce `--interactive` for this.

> For now I have hardcoded some options and had to integrate the yargs cli class with the prompt library in a hackish way because yargs lacks native support for it.

And it feels awesome! An interesting problem to solve will be how we can encode the component kind of questions into code. Which components to group? When do we needd to follow-up with more questions (i.e. don't ask me follow-up questions on jest, if I didn't select it)

mrgrain:
closing stale PR

PR #2964, opened 2023-10-06
fix(node-package)!: components cannot change deps at preSynthesize stage

tinovyatkin:
NodePackage are synthesising all dependencies during `preSynthesize` phase rather than during `synthesize` (as projen philosophy dictates). This results that any components added after NodePackage (that means almost all of them in normal situation of using NodeProject) cannot make any changes to dependencies during their `preSynthesize` (that changes will be silently ignored).

However, ability for components to make dependencies changes during `preSynthesize` are crucial to not require specific components order. For example a component (i.e. implementing [BrowsersList](https://github.com/browserslist/browserslist) config) wants to add a dev dependency (i.e. [stylelint-no-unsupported-browser-features](https://github.com/ismay/stylelint-no-unsupported-browser-features)) only if another component (in this example - implementing [Stylelint](https://stylelint.io/user-guide/configure/) config) are present in the project. The logical place to do it without forcing components adding order is at `preSynthesize` that will be called when all components are added to the project.

This PR fixes the above mentioned fact by noticing that we are actually reading pre-existing `package.json` twice - once in the constructor and once again in the `renderDependencies`. So we assigning first read to a class private field and just re-use that at `renderDependencies`.

One existing unit test that were based on assumption at which moment pre-existing package.json will be read slightly adjusted to be free of that assumption.

Another unit test added to demonstrate that modifying dependencies from components during `preSynthesize` is now working as expected.

BREAKING CHANGE: This may potentially affect some edge cases that were relying on the internal implementation of the dependencies handling. If new dependencies appear/disappear after applying this fix, previous code `preSynthesize` handler should be inspected for any `.addDependency`-like calls - those were incorrectly ignored previously and should be removed if actually are not needed.

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -501,6 +501,7 @@ export class NodePackage extends Component {
   private readonly bin: Record<string, string> = {};
   private readonly engines: Record<string, string> = {};
   private readonly peerDependencyOptions: PeerDependencyOptions;
+  private readonly _prev?: Record<string, any>;
   private _renderedDeps?: NpmDependencies;
 
   constructor(project: Project, options: NodePackageOptions = {}) {
@@ -535,7 +536,7 @@ export class NodePackage extends Component {
 
     this.processDeps(options);
 
-    const prev = this.readPackageJson() ?? {};
+    this._prev = this.readPackageJson();
 
     // empty objects are here to preserve order for backwards compatibility
     this.manifest = {
@@ -562,11 +563,11 @@ export class NodePackage extends Component {
       license: () => this.license ?? UNLICENSED,
       homepage: options.homepage,
       publishConfig: () => this.renderPublishConfig(),
-      typesVersions: prev?.typesVersions,
+      typesVersions: this._prev?.typesVersions,
 
       // in release CI builds we bump the version before we run "build" so we want
       // to preserve the version number. otherwise, we always set it to 0.0.0
-      version: this.determineVersion(prev?.version),
+      version: this.determineVersion(this._prev?.version),
       bugs:
         options.bugsEmail || options.bugsUrl
           ? {
@@ -831,10 +832,9 @@ export class NodePackage extends Component {
   }
 
   // ---------------------------------------------------------------------------------------
-
-  public preSynthesize() {
-    super.preSynthesize();
+  public synthesize() {
     this._renderedDeps = this.renderDependencies();
+    super.synthesize();
   }
 
   public postSynthesize() {
@@ -1194,8 +1194,7 @@ export class NodePackage extends Component {
     this.manifest.bundledDependencies = sorted(bundledDependencies);
 
     // nothing further to do if package.json file does not exist
-    const pkg = this.readPackageJson();
-    if (!pkg) {
+    if (!this._prev) {
       return { devDependencies, peerDependencies, dependencies };
     }
 
@@ -1223,9 +1222,9 @@ export class NodePackage extends Component {
       }
     };
 
-    readDeps(devDependencies, pkg.devDependencies);
-    readDeps(dependencies, pkg.dependencies);
-    readDeps(peerDependencies, pkg.peerDependencies);
+    readDeps(devDependencies, this._prev.devDependencies);
+    readDeps(dependencies, this._prev.dependencies);
+    readDeps(peerDependencies, this._prev.peerDependencies);
 
     return { devDependencies, dependencies, peerDependencies };
   }
--- a/test/javascript/node-package.test.ts
+++ b/test/javascript/node-package.test.ts
@@ -2,7 +2,7 @@ import { readFileSync, writeFileSync, mkdirSync } from "fs";
 import { dirname, join } from "path";
 import * as semver from "semver";
 import * as YAML from "yaml";
-import { Project, DependencyType } from "../../src";
+import { Project, DependencyType, Component } from "../../src";
 import {
   NodePackage,
   NodePackageManager,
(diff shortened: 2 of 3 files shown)


test/javascript/node-package.test.ts
```diff
@@ -308,6 +302,12 @@ test("manually set devDependencies are not changed when a peerDependency is adde
     JSON.stringify(orig, undefined, 2)
   );
 
+  const pkg = new NodePackage(project, {
+    peerDependencyOptions: {
+      pinnedDevDependency: true,
+    },
+  });
+
```
mrgrain, line 310: Is this test change dictated by the fix? If so, I don't understand why the ordering matters here.
If not: Thanks for cleaning up! :D

tinovyatkin: Yes. This test was assuming the NodePackage will read pre-existing package.json in the `preSynthesize` so it was creating _pre-existing package.json_ after creating the component itself. We just re-arrange here to create that file before the component (as it would happen in the real life)

mrgrain (changes requested):
Oh nice, thanks for diving into this! I'm a bit surprised this isn't blowing anything up 🙈 

But I'm happy to merge it, if you mark this as a breaking change, and include some sort of guidance on how the original behavior could be replicated if needed.

src/javascript/node-package.ts
```diff
@@ -1194,8 +1194,7 @@ export class NodePackage extends Component {
     this.manifest.bundledDependencies = sorted(bundledDependencies);
 
     // nothing further to do if package.json file does not exist
-    const pkg = this.readPackageJson();
-    if (!pkg) {
+    if (!this._prev) {
```
mrgrain, line 1197: This is the breaking change and the required test change proof this.
And now that I'm thinking more about this, why did you get rid if the re-reading of package.json if the main goal was to move stuff to from `preSynthesize` to `synthesize`?

Projects are constructed first, components are usually constructed later. This applies to automatic components that the project creates, but also custom components added later in the rc file. Before your change, a component could manipulate `package.json` directly and any changes would be merged in. This not possible anymore.

mrgrain:
The other breaking changes would be down to the fact that previously all various steps performed by `renderDependencies()` would have been guaranteed to be completed by the time we hit `synthesize()` in a component. This order is not guaranteed anymore, even though `NodePackage` is likely to be one of the earlier components. 

Now the steps that are done in `renderDependencies` are quite obscure, so I'm happy to break that guarantee. As far as I can tell we'd loose the guarantee that the following has happened before `synthesize()`:
- Adding a devDep for every peerDep, including the semver check
- Resolving `file:` deps with the actual package name

Currently, these are only exposed via the deprecated `manifest` property. So again, happy to break it. But we should mark the version correctly and provide some guidance.

src/javascript/node-package.ts
mrgrain: Can we keep the re-reading of package.json?

tinovyatkin: "Component can change package.json directly" - do you mean "on the disk"? Because if component would do JsonPatch or addOverride (that means do changes to this file the projen way) - that functionality will stay in touch.

If a component were doing direct changes to the packages.json on the disk, those changes would be respected only in very limited part (literally at dependencies section, as for example `version` field would be read in the constructor) and those changes would be thrown when package.json would be removed at `preSynthesize` (as all generated files). So, do I understand correctly that you want to keep functionality of a theoretical projen components that were ignoring projen way to manage dependencies and were relying on some internal knowledge that projen is reading dependencies versions from `package.json` twice? Is that something we want to keep and name breaking change?

> Can we keep the re-reading of package.json

We can, but it will complicate implementation with no real benefits (see above).

tinovyatkin:
> Currently, these are only exposed via the deprecated `manifest` property. So again, happy to break it. But we should mark the version correctly and provide some guidance.

I can't agree that changes in edge functionality of private, non-documented methods can be considered a breaking change for `0.x` versioned library, but neither a see a huge topic for discussion here. Could you provide some guidance how to make this "breaking" mark - I assume adding something to the PR title?

src/javascript/node-package.ts
mrgrain: How does it complicate the implemenation?

mrgrain:
> I can't agree that changes in edge functionality of private, non-documented methods can be considered a breaking change for `0.x` versioned library, but neither I'm seeing a huge topic for discussion here. Could you provide some guidance how to make this "breaking" mark - I assume adding something to the PR title?

And likewise because we are at `0.x.`, to me there's no reason to hold back on new minor version releases.

We usually do it like this:
https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

Specifically including a section on how the previous behavior can be restored or worked around or why it can't.

tinovyatkin:
> Specifically including a section on how the previous behavior can be restored or worked around or why it can't.

I'm assuming this will be squashed on merge so I've added `!` to PR title and put a way of restoring "previous behavior" into PR description body. Would that be enough or I should reword the commit message itself?

mrgrain:
Commits are ignored. The PR title and description are what counts. Thanks for making the change!

src/javascript/node-package.ts
mrgrain: Okay, now I'm with you. Thanks for taking the time to explain!

mrgrain (approved):

moltar:
@tinovyatkin Sorry for the offtopic, but:

> (i.e. implementing [BrowsersList](https://github.com/browserslist/browserslist) config)

Are you implementing the BrowsersList? This was on my TODO for a while also. If so, will it be public? Thanks!

PR #2965, opened 2023-10-06
feat(vscode): allow to extend existing setting if exists

tinovyatkin:
VSCode has many settings which values are either array or object, like `files.exclude` or `prettier.documentSelectors`. Particular components normally want just to extend that setting (i.e. adding additional exclude, or some extra file extension to Prettier) rather than completely override it (as it would result in conflicts between components). Unfortunately we can't also reliable use JsonPatch as workaround as we are not always known where that settings already exits. 

This PR adds a flag that allows to extend existing array or object in case if it already exists. To make this a non-breaking change we are adding extra parameter defaulting to current behaviour.

Unit tests are extended to cover added functionality.

---

### DIFF
--- a/docs/api/API.md
+++ b/docs/api/API.md
@@ -12088,31 +12088,33 @@ new vscode.VsCodeSettings(vscode: VsCode)
 ### Methods
 
 
-#### addSetting(setting, value, language?)🔹 <a id="projen-vscode-vscodesettings-addsetting"></a>
+#### addSetting(setting, value, language?, extend?)🔹 <a id="projen-vscode-vscodesettings-addsetting"></a>
 
 Adds a workspace setting.
 
 ```ts
-addSetting(setting: string, value: any, language?: string): void
+addSetting(setting: string, value: any, language?: string, extend?: boolean): void
 ```
 
 * **setting** (<code>string</code>)  The setting ID.
 * **value** (<code>any</code>)  The value of the setting.
 * **language** (<code>string</code>)  Scope the setting to a specific language.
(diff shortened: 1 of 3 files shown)


mrgrain:
Nice! How about we have two new methods `extendSetting` and `extendSettings`?

tinovyatkin:
> Nice! How about we have two new methods `extendSetting` and `extendSettings`?

I'm not sure it's a right naming because it may confuse - this will still add settings if they didn't exist.

mrgrain:
> > Nice! How about we have two new methods `extendSetting` and `extendSettings`?
> 
> I'm not sure it's a right naming because it may confuse - this will still add settings if they didn't exist. Also, I'm not a native to English language but from my point of understanding `addSetting` means in case if setting already exists and it's an array then "add to array".

I'm not fuzzed about the naming. I'm fuzzed about methods with incredibly long parameter lists and particular multiple optional parameters at the end.

> I'm not a native to English language

Neither am I! But sometimes we are better place to find good naming 😆 

> from my point of understanding `addSetting` means in case if setting already exists and it's an array then "add to array".

`addX` is simply the naming convention in projen. That doesn't make it correct or better though. We don't have many examples for adding complex configurations, one existing one is [Eslint.addRules](https://github.com/projen/projen/blob/f743730eab66c96bd1350ccc397c902add486beb/src/javascript/eslint.ts#L422) which also doesn't merge rules together.

tinovyatkin:
Closing in favour of #2998  and #2997 - existing method `.addToArray` can be used to upsert array and add array values while fixed `JsonPatch.test` can be used to upsert deeply nested objects.

PR #2974, opened 2023-10-09
feat: Projects and Components are Constructs

### LINKED ISSUE #1763: Rearchitect projen to use constructs framework
Components in projen (such as `NodePackage`, `Eslint`, `JsonFile`, etc.) should extend `Construct` from the [constructs framework](https://github.com/aws/constructs) instead of extending our own custom `Component` type.

Motivations:

1. ~It should be possible for users to not only share/reuse project types, but also share and reuse individual projen components/constructs. I think the reason people don't do this right now is because it's not possible, but I'm not sure (this needs to be validated).~ (Currently this is possible without constructs, so I don't think this is deciding factor)
2. Allow for introspection into the "project tree". Currently all components are added directly to the `Project` class, so the components are not created into a tree. I think modeling components as a tree will make it easier for users to figure out how components or files are being generated since we can generate the "tree" as an artifact (like the `tree.json` generated by AWS CDK), and expose it for assorted purposes. This could enable features like e.g. a command `projen inspect file.txt` which could automatically determine which projen component generated a particular file.
3. Standardize constructor syntax for components. Right now different components take different kinds of lists of arguments when you construct them. Most of them take some kind of `Project` as their first argument, and some `options` object as their second argument, but not all of them. Standardizing this makes it easier to learn and use the framework, makes it easier to do code generation for, etc.
4. Open up future interoperability between "projen" components and other CDKs like cdktf, AWS CDK, etc.

`constructs` is fairly unopinionated and it doesn't bind projen to other technical decisions, such as whether or not the framework should use [jsii](https://github.com/aws/jsii/) for multi-language support or not etc.

Implementation POC here: https://github.com/projen/projen/pull/1671

mrgrain:
Resolves #1763 

Features:

- 🚧 `Project` & `Component` now extend `Construct` - making everything a construct!
- ✨ Magic AutoID feature for existing components, no hassle in the migration
- ℹ️ Basic Metadata added by default
- 📦 `Component` class now accept any construct as scope, and throw if no project can be found
- 📄 Core file types have been updated to implement thew new interface
- 🌳 An experimental `tree.json` component 

Next steps (after this PR):

- Update all components to accept any construct as scope
- Instantiate components with the current scope, instead of the project

---


moltar:
I am rooting for this change! I think this is _long_ overdue and would open up many use cases, such as Aspects (borrowing from CDK) and other graph tree ops.

mrgrain:
Only thing missing here is that `Component` currently takes a `Project`, not a `IConstruct` as scope. That needs to change, because otherwise all components are just direct descendants of the project. Not much of a tree then!

moltar:
> Only thing missing here is that `Component` currently takes a `Project`, not a `IConstruct` as scope. That needs to change, because otherwise all components are just direct descendants of the project. Not much of a tree then!

And projects aren't `IConstruct`, i.e. they do not extend `Construct` class?

mrgrain:
> And projects aren't `IConstruct`, i.e. they do not extend `Construct` class?

No projects are `IConstruct` and extend `Construct`. Not sure I understand what you are asking... 🤔

mrgrain:
Alright, `Component` accept now any `IConstruct` as scope, but fail if no project is found.
Changed a bunch of default implementations, but not going to do all in this PR.
Needs some tests and docs. 👀

mrgrain:
Added docs. Changes are already plenty tested by proxy.

src/component.ts
```diff
@@ -1,11 +1,39 @@
...
+import {
+  COMPONENT_SYMBOL,
+  isComponent,
+  findClosestProject,
+} from "./util/constructs";
+
+let currentId = 0;
+const componentId = () => {
+  return `AutoId${++currentId}`;
+};
```
moltar, line 12: Could this be scoped to a given `scope`, given that unique ID needs to be within the scope only?

src/project.ts
```diff
@@ -128,13 +134,30 @@ export interface GitOptions {
...
   /**
    * The name of the default task (the task executed when `projen` is run without arguments). Normally
    * this task should synthesize the project files.
    */
   public static readonly DEFAULT_TASK = "default";
 
+  /**
+   * Test whether the given construct is a project.
+   */
+  public static isProject(x: any): x is Project {
```
moltar, line 147: Can this be `unknown` type? It's a bit more idiomatic in TS.

src/project.ts
```diff
@@ -223,6 +246,20 @@ export class Project {
...
+    const autoId = `${new.target.name}#${options.name}@${path.normalize(
+      options.outdir ?? "<root>"
+    )}`;
+
+    if (options.parent?.subprojects.find((p) => p.outdir === outdir)) {
+      throw new Error(`There is already a subproject with "outdir": ${outdir}`);
+    }
+
+    super(options.parent as any, autoId);
+    Object.defineProperty(this, PROJECT_SYMBOL, { value: true });
```
moltar, line 259: I would co-locate this logic in `src/util/constructs.ts` in a util function in order not to leak implementation detail.

E.g.:

```ts
function tag(scope: IConstruct, tag: symbol) {
  Object.defineProperty(this, tag, { value: true });
}

function tagProject(scope: IConstruct) {
  tag(scope, PROJECT_SYMBOL)
}
```

rix0rrr (approved):

PR #2992, opened 2023-10-13
feat(gitlab)!: updates to configuration model

ctasada:
Added the missing keywords for Services
* pull_policy: https://docs.gitlab.com/ee/ci/yaml/#servicepull_policy
* variables: Added in Gitlab 14.5 https://docs.gitlab.com/ee/ci/services/index.html#available-settings-for-services

BREAKING CHANGE: All `variables` are now Records of strings, as per documentation. If you are currently passing in a number, please convert it to a string.

---


mrgrain:
Same as the other one, please run the build locally to add any mutations to your PR.

mrgrain (approved):

src/gitlab/configuration-model.ts
```diff
@@ -209,6 +209,20 @@ export interface Service {
   readonly entrypoint?: string[];
   /** Full name of the image that should be used. It should contain the Registry part if needed.*/
   readonly name: string;
+  /** The pull policy that the runner uses to fetch the Docker image */
+  readonly pullPolicy?: PullPolicy[];
+  /** Additional environment variables that are passed exclusively to the service.. */
+  readonly variables?: Record<string, number | string>;
```
mrgrain, line 215: In jsii languages this doesn't work very well. Let's just use a string here
```suggestion
  readonly variables?: Record<string, string>;
```

mrgrain (changes requested):

src/gitlab/configuration-model.ts
ctasada: @mrgrain this was copied from the `IncludeRule` section, where this value was already defined. I'm mentioning, because if we have a potential issue here, we may have it there too

mrgrain: Thanks @ctasada Acknowledged. The other definition will be wrong as well. Also confirmed that this indeed should be a string: https://docs.gitlab.com/ee/ci/yaml/#variablesvalue

Can you make the change for you PR and we can fix the other occurrence in a separate PR (would love if you did that, but not worries if not 😅 )

ctasada: @mrgrain Done. Fixed both definitions

src/gitlab/configuration-model.ts
```diff
@@ -212,6 +212,20 @@ export interface Service {
...
```
mrgrain, line 218:
```suggestion
  readonly variables?: Record<string, string>;
```
@ctasada thanks for the fixes! Looks like you've missed one.

mrgrain (changes requested):

src/gitlab/configuration-model.ts
ctasada: Good catch @mrgrain I applied the change and rebuild to update the API.md too.

mrgrain (approved):

PR #2997, opened 2023-10-14
feat(json-patch)!: each `ObjectFile.patch()` is an atomic operation and new failure behaviors for `JsonPatch.test()`

tinovyatkin:
`test` operation in JsonPatch are affecting ["patch as whole"](https://jsonpatch.com/#test) (`apply` operation of `fast-json-patch` actually throws when a test condition fails). However, `ObjectFile.patch` method are putting all patches into a single flatten array, discarding the concept of "operations" and actually disallowing use of `JsonPatch.test` in components (as it will broke whole file synthesis on a test failure).

This PR fixes the issue by considering all patches passed to a single call of `objectFile.patch` to be an atomic operation. That means either patches are applied, or none. When a `test` operation fails, only the current atomic operation fails. Any other will still be executed. This also changes the failure behavior of `JsonPatch.test` to fail silently, as opposed to the current exception. For this, we are also adding an optional parameter to `JsonPatch.test` to define the failure behaviour as either `SKIP` (new default behavior) or `FAIL_SYNTHESIS` (previous behavior). This allows to use JsonPatch for upsetting or conditionally changing some values based on current values (see added test):

Example:

```ts
 // adding object(s) if not exists - it's not possible to do something like this currently
objectFile.patch(
        JsonPatch.test("/first", undefined),
        JsonPatch.add("/first", {})
)
objectFile.patch(
        JsonPatch.test("/first/second", undefined),
        JsonPatch.add("/first/second", {})
)
objectFile.patch(
      JsonPatch.add("/first/second/field1", "value1"),
      JsonPatch.add("/first/second/field2", "value2"),
)
```

In additional to above changes this PR includes following complementary changes:
* updated version of `JsonPatch` that uses `deepClone` on patching operations
* exposing very useful `escapePathComponent` static method (imported from the same `fast-json-patch` dependency) to simplify operation on fields containing `/` and `~` (i.e. scoped dependencies overrides in `package.json`)

BREAKING CHANGE: A failure of `JsonPatch.test` does not fail synthesis anymore. To preserve previous behavior, pass `TestFailureBehavior.FAIL_SYNTHESIS` as third parameter to the `JsonPatch.test` call.

BREAKING CHANGE: Calls of `JsonPatch.test` are now only relevant to the scope of the list of patches passed to `ObjectFile.patch`. This should not break anything for truthy tests, and should not be relevant for failing tests. However, in case any issues are encountered, the previous behavior can be restored by placing all `JsonPatch`es in a single call of `ObjectFile.patch`.

---


src/object-file.ts
```diff
@@ -211,8 +211,22 @@ export abstract class ObjectFile extends FileBase {
    *
    * @param patches - The patch operations to apply
    */
-  public patch(...patches: JsonPatch[]) {
-    this.patchOperations.push(...patches);
+  public patch(...patches: (JsonPatch | JsonPatch[])[]) {
```
mrgrain, line 214: Doesn't work well in jsii languages. Also the nested arrays syntax is confusing.
What is the benefit over multiple `file.patch([])`?

mrgrain:
I disagree on this one. The JSON Patch RFC is in my opinion quite clear on this topic:

>   If a normative requirement is violated by a JSON Patch document, or
   if an operation is not successful, evaluation of the JSON Patch
   document SHOULD terminate and application of the entire patch
   document SHALL NOT be deemed successful.

https://datatracker.ietf.org/doc/html/rfc6902/#section-5

While it doesn't state it outright, to me this read like validation and not upserting.

But I understand the issue that patch operations are currently getting lost in the implementation because they are flattened into a single array. That's what we should fix.

I imagine calling `.patch()` multiple times as the way to express intent of different operations. Each of the operations should then by applied individually. Users can catch a test failure in userland if they are okay with it:
```ts
// This would synthesize just fine, operation not applied
try {
  file.patch([
    JsonPatch.test("/first/second", undefined),
    JsonPatch.add("/first/second", []),
  ]);
catch {}

// This would throw
file.patch([
  JsonPatch.test("/first/second", undefined),
  JsonPatch.add("/first/second", []),
]);
```

We could potentially make this easier by adding an optional parameter`.patch(patches, throwError = true)` (obviously need a better name) or maybe a second `.patchNotFail()` method.

The reason I'm reluctant to default to silently failing is that I struggle to find a uniform user experience. What should happen when it fails? Silent upsert? Print a warning? Hard fail? I think each of these cases have valid scenarios and drawbacks if a standard handling is introduced. The only way to leave this up to the user is to transparently throw an exception and let them handle it.

--- 

### Syntax 

I also don't really see the benefit of the nested array syntax over multiple calls to `.patch()`

---

> updated version of JsonPatch that uses deepClone on patching operations

Nice 👍🏻 

> exposing isTestOperation method

But why?

> exposing very useful escapePathComponent static method (imported from the same fast-json-patch dependency) to simplify operation on fields containing / and ~ (i.e. scoped dependencies overrides in package.json)

Nice 👍🏻

tinovyatkin:
> I disagree on this one. The JSON Patch RFC is in my opinion quite clear on this topic:

Not sure I understand where we disagree: the problem of current implementation of `.patch` that it doesn't retain the scope of "operation" as it flattens all calls of `.patch` into a single array. This way any `.test` are getting as "patch a whole" all calls of `.patch` on a given file. That is incorrect and this PR proposes to fix that by adding a feature to call `.patch` with array of patches - in this case all operations of array will be the "patch as whole" for the `.test` operation.

Now, for some reason you are mixing throw behaviour of our dependency with RFC of JsonPatch. It has to relation, `thow` is not part of RFC, it's an implementation details of the internal dependency. So, silently ignoring "test failure" throw (and only that) is perfectly aligned with RFC, because it does exactly what RFC dictates - if test operation fails patch as whole should not be applied. I don't understand where you are seeing a base of confusion here and which other behaviour may be applied here?

```ts
// This would synthesize just fine, operation not applied
try {
  file.patch([
    JsonPatch.test("/first/second", undefined),
    JsonPatch.add("/first/second", []),
  ]);
catch {}
```

This would not work with lazy values and `.patch` and `.addOverrides` are applying after resolve. If you do it differently to make `.test` working it will be way bigger blast radius.

mrgrain:
Thanks for looking into my suggestion. You're right, I've missed that the operation is only executed during synth phase. My bad. 🤦🏻 

My general concerns about the possible variations of error handling still stand though. Let's see if we can find a different solution. I'd still stick with the principle of `one call of .patch()` = `one atomic operation`. We just need to find a way to make the test fail behavior flexible.

My detour the JSON Patch RFC is in relation to the proposed error handling. I have indeed misinterpreted the references to error handling as prescriptive. Which it is - but just for a different RFC (the one for HTTP PATCH). That's also on me. Guess I shouldn't do reviews on a Sunday evening 😅 . 

At the end of the I am currently relying on the explicit failure behavior, and I think it should be possible to find a solution that allows for all three options: `SILENT`, `WARN` and `FAIL`.

tinovyatkin:
> I'd still stick with the principle of `one call of .patch()` = `one atomic operation`. 

> At the end of the I am currently relying on the explicit failure behavior, and I think it should be possible to find a solution that allows for all three options: `SILENT`, `WARN` and `FAIL`.

Implemented new approach (see updated PR description) that hopefully follows your suggestions. It seems to be even more scoped implementation while giving almost the same DX.

src/json-patch.ts
```diff
@@ -1,6 +1,35 @@
...
+
+export enum TestFailureBehavior {
+  /**
+   * Skip the patch operation and continue with the next operation.
+   */
+  SKIP_PATCH = "skip",
+  /**
+   * Throw an error and stop whole file synthesizes.
+   */
+  THROW = "trow",
```
mrgrain, line 25:
```suggestion
  THROW = "throw",
```

src/json-patch.ts
```diff
@@ -27,11 +56,31 @@ export class JsonPatch {
...
+              `Test operation failed at index ${e.index}: ${JSON.stringify(
+                op._toJson()
+              )}`
+            );
+            return document;
+          }
+        }
+      }
+      throw e;
+    }
```
mrgrain, line 83: Can we the failure handling into ObjectFile? I'd much prefer to use `project.logger()` but that won't be available here. It also avoids the `TEST_FAILURE_BEHAVIOR_SYMBOL` and keeps `JsonPatch` a thin wrapper around the library.

mrgrain (changes requested):
Almost!

src/json-patch.ts
```diff
@@ -89,8 +146,30 @@ export class JsonPatch {
...
+    });
+    return patch;
+  }
+
+  /**
+   * Escapes a json pointer path
+   * @param path The raw pointer
+   * @return the Escaped path
+   */
+  public static escapePathComponent(path: string): string {
```
mrgrain, line 171: This feels a bit less like leaking the underlying implementation through.

```suggestion
  public static escapePath(path: string): string {
```

src/json-patch.ts
tinovyatkin: Not sure I understand how that avoids the symbol? Do you mean we should somehow extend `ObjectFile.patch` method to add failure behavior there rather to `JsonPatch.test` static method?

tinovyatkin: Just to clarify my position here: `JsonPatch` class is our wrapping class that is only used at `ObjectFile.patch` method. Not every JsonPatch implementation library would throw on test failure (we already discussed that is implementation detail of `fast-json-patch`). So I'm seeing pretty logical that we are handling that details in the class that wraps that dependency.

As we already agreed, JsonPatch.test has nothing to do with throw, so it would be looking weird at `ObjectFile.patch`

tinovyatkin: An example of implementation of JsonPatch that is not throwing on test failure: https://www.npmjs.com/package/@typewriter/json-patch#test-failed

mrgrain: > Just to clarify my position here: JsonPatch class is our wrapping class that is only used at ObjectFile.patch method. Not every JsonPatch implementation library would throw on test failure (we already discussed that is implementation detail of fast-json-patch). So I'm seeing pretty logical that we are handling that details in the class that wraps that dependency.

I guess this is where our disagreement is: I like how `fast-json-patch` handles test failures. Because it is the _only_ way users can make an informed decision how to handle test failures themselves, by catching the error and reacting accordingly.

`JsonPatch` is only a class because that's the best way to implement free-floating top level functions in jsii. It's public interface only has static methods. It makes sense to keep the code there from an code structure perspective, from a user perspective I argue it doesn't matter precisely because JsonPatch is a wrapping class that is only used in ObjectFile.patch(). The problem is that now the wrapper class suddenly needs awareness of the project to log a warning.


>Not sure I understand how that avoids the symbol?

Sorry I brought this up. It's an implementation detail and not relevant to the wider point.

mrgrain: Attempt to rephrase my position: I consider `JsonPatch` a low-level implementation of the standard and thus should be flexible to support different use cases.

`ObjectFile.patch()` is our high-level abstraction (that merely uses the low-level implementation) and this is were we should focus the UX at. Including making the skip behavior the default, which I'm fully on-board with.

tinovyatkin: @mrgrain I'm struggling to invent a good DX for extending `ObjectFile.patch` or maybe misunderstanding your idea. Could you provide the method shape as you envisioning it?

mrgrain: How about this?
https://github.com/projen/projen/pull/2997/commits/546c3c7d9c507a687b4a741ef92448ddb9fb321a

tinovyatkin: Added documentation doesn't even mentioning the `log` option (and I agree it's pretty much useless). I may propose either to remove `log` option or changing the `JsonPatch.apply` method signature to allow pass logger there. What do you think? Moving the whole `try/catch` into `ObjectFile.synthesizeContent` will be a way more weird solution as we will need to pass symbol and we will not be able to pass custom error class from `fast-json-patch`, etc.

src/object-file.ts
```diff
@@ -1,5 +1,6 @@
+import { JsonPatchError } from "fast-json-patch";
```
tinovyatkin, line 1: This is bad practice codewise - we have our internal wrapper for `fast-json-patch` and here we are using extra knowledge of internal implementation of `JsonPatch` class to handle it's errors

src/object-file.ts
```diff
@@ -234,7 +235,33 @@ export abstract class ObjectFile extends FileBase {
...
+    let patched = resolved;
+    for (const ops of this.patchOperations) {
+      try {
+        patched = JsonPatch.apply(patched, ...ops);
+      } catch (error) {
+        if (!isTestOperationFailure(error)) {
+          throw error;
+        }
+        const currentOp = ops[error.index!];
+        switch (currentOp?._failureBehavior) {
```
tinovyatkin, line 248: This is violating bunch of OOP principals and bad practice code-wise - we are basically accessing one internal class properties from another place just because we know implementations details. This creates very bad "invisible" coupling between two classes.

src/object-file.ts
mrgrain: I agree, though we are already exposing the `fast-json-patch` exception anyway. Are you proposing we wrap all exceptions in a custom exceptions class (that either does the same thing, or takes information away from the user).

src/object-file.ts
mrgrain: Yep, you caught me. :P We need a `JsonPatchTestOperation` and I was too lazy to do it.

Saying that, the current `JsonPatch` class already is as far away from proper OOP as something can be.

src/json-patch.ts
mrgrain: > Added documentation doesn't even mentioning the log option

It does, I just renamed it to `WARN`.

> I may propose either to remove log option 

Yeah, I think I'm okay with that.

tinovyatkin: So, removing the `warn` we can rollback to have everything inside `JsonPatch` class and not spill around the implementation details, right?

src/object-file.ts
tinovyatkin: No, I proposed above to remove `warn` option and keep error handling inside the `JsonPatch` class.

mrgrain (approved):

PR #2999, opened 2023-10-15
feat(eslint): add additional overrides fields

tinovyatkin:
Adding missed fields declarations to ESLint [override](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns).

Practical use case:
```ts
ESLint.of(project)?.addOverride({
  files: ['*.graphql'],
  plugins: ['@graphql-eslint'],
  extends: ['plugin:@typescript-eslint/disable-type-checked'],
})
```
---


src/javascript/eslint.ts
```diff
@@ -97,6 +97,12 @@ export interface EslintOverride {
    */
   readonly files: string[];
 
+  /**
+   * Pattern(s) to exclude from this override.
+   * If a file matches any of the excluded patterns, the configuration won’t apply.
+   */
+  readonly excludedFiles?: string | string[];
```
mrgrain, line 104: In jsii languages this doesn't work very well. Let's just use an array here.

```suggestion
  readonly excludedFiles?: string[];
```

src/javascript/eslint.ts
```diff
@@ -106,6 +112,16 @@ export interface EslintOverride {
    * The overridden parser
    */
   readonly parser?: string;
+
+  /**
+   * Config(s) to extend in this override
+   */
+  readonly extends?: string | string[];
```
mrgrain, line 119: In jsii languages this doesn't work very well. Let's just use an array here.

mrgrain (changes requested):

src/javascript/eslint.ts
tinovyatkin: I changed as you suggested, but out of curiosity - in which languages this is problematic with JSII 5.x? I tested with JSII 5 and Python and it looks and validates just fine.

mrgrain (approved):

Issue #2980, opened 2023-10-10
Yarn Berry Configuration is Incomplete


blimmer:
# Glossary

| Term | Description |
| ---- | ----------- |
| Yarn Classic | Yarn v1 |
| Yarn Berry | Any version of Yarn > v1. At the time of writing, stable Yarn Berry is v3. v4 currently has a release candidate available |
| YARN2 | Projen refers to Yarn Berry as YARN2, which is a bit misleading |

# Description

`projen` currently supports specifying that we're using YarnV2: https://github.com/projen/projen/blob/38c6570c2808586bf9ec032214efdbb2907d86f0/src/javascript/node-package.ts#L1080-L1081

However, this switch simply updates the calls to global `yarn` to be compatible with Yarn > 1. It does not actually
ensure we're using Yarn Berry.

# Problems

## Init Uses Global Version Regardless of `--packageManager` flag

Even if you specify `--packageManager yarn2` at `init` time, Yarn Classic is used if it's set as the global version.

Reproduction steps:

```shell
> docker run -it --rm --entrypoint bash node:18
root@aa1079514eb5:/# yarn --version
1.22.19
root@aa1079514eb5:/# mkdir test-projen
root@aa1079514eb5:/# cd test-projen/
root@aa1079514eb5:/test-projen# npx projen@latest new typescript --packageManager yarn2
👾 Project definition file was created at /test-projen/.projenrc.ts
👾 Installing dependencies...
👾 install | yarn install
yarn install v1.22.19
info No lockfile found.
[1/4] Resolving packages...
warning standard-version > stringify-package@1.0.1: This module is not used anymore, and has been replaced by @npmcli/package-json
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
Done in 31.52s.
👾 Installing dependencies...
👾 install | yarn install
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
```

As you can see, the global version of Yarn (Yarn Classic) is used, even though we specified `--packageManager yarn2`.

## `package.json`, and `.gitignore` settings are not updated per Yarn Berry best practices

Yarn Berry configures itself in `package.json` and `.gitignore` files. Projen does not update these files
when using Yarn Berry.

[...]

PR #3014, opened 2023-10-21
feat: introduce `YARN_CLASSIC` and `YARN_BERRY` NodePackageManager options

blimmer:
Relates to #2980, specifically this comment: https://github.com/projen/projen/issues/2980#issuecomment-1759324495

This PR introduces two new package manager values: `YARN_CLASSIC` and `YARN_BERRY`. These names better align with the `yarn` ecosystem. It also deprecates `YARN`, warning that it might be repurposed in the future.

---


docs/node.md
```diff
@@ -42,7 +42,7 @@ according to the package manager being used. This can be configured:
 ```ts
   // ...
-  packageManager: javascript.NodePackageManager.YARN, // or NPM, PNPM, etc.
+  packageManager: javascript.NodePackageManager.YARN_CLASSIC, // or NPM, PNPM, etc.
```
blimmer, line 45: Leaving this as `YARN_CLASSIC` in the example until we have full support for Yarn Berry (see https://github.com/projen/projen/issues/2980)

src/javascript/node-package.ts
```diff
@@ -168,7 +168,7 @@ export interface NodePackageOptions {
   /**
    * The Node Package Manager used to execute scripts
    *
-   * @default NodePackageManager.YARN
+   * @default NodePackageManager.YARN_CLASSIC
```
blimmer, line 171: Still default to `YARN_CLASSIC` for now.

src/javascript/node-package.ts
```diff
@@ -513,7 +513,8 @@ export class NodePackage extends Component {
       ...options.peerDependencyOptions,
     };
     this.allowLibraryDependencies = options.allowLibraryDependencies ?? true;
-    this.packageManager = options.packageManager ?? NodePackageManager.YARN;
+    this.packageManager =
+      options.packageManager ?? NodePackageManager.YARN_CLASSIC;
```
blimmer, line 517: I considered writing a method (or large ternary) here that reworked `YARN` to `YARN_CLASSIC` and `YARN2` to `YARN_BERRY`. I didn't do that because:

1. It's a pain to do `Partial` types in JSII. If I could have my way with all TypeScript features I would write that `this.packageManager` cannot be `YARN` or `YARN2`.
2. I wasn't sure if other might be using/setting `this.packageManager` from this class.

docs/node.md
mrgrain: FYI, I think this should be NPM in future, mirroring Node.js (no offense to other package managers 😅 )

src/javascript/node-package.ts
mrgrain: I don't follow. The change here looks good to me though 👍🏻

src/javascript/node-package.ts
mrgrain: FYI, I think this should be NPM in future, mirroring Node.js (no offense to other package managers 😅 )

mrgrain (approved):
Thanks for that! Can you run `yarn build` and commit the remaining mutations to get the build pass?

PR #3015, opened 2023-10-22
fix(apache-variables): adding variables to license text

Liam-Johnston:


hoegertn:
This text is not meant to be replaced. It is meant as an example how to add the Copyright notice to your source code. License texts should not be modified...

mrgrain:
Thanks @Liam-Johnston it's what Thorsten said. Closing.

Liam-Johnston:
Oh no worries @mrgrain @hoegertn !

I guess the whole section in the `License` constructer should be removed then? https://github.com/projen/projen/blob/b4a8057942d2bc1d71c926e1afdef560bea1ade0/src/license.ts#L44

Unless is it ok for the MIT license type to have these values substituted?

Also should the file then be untracked by projen because currently once its generated its treated like any of the other synthesised files and is marked as read-only.

mrgrain:
@Liam-Johnston what's the problem you are trying to solve here?


MIT (and some others) license explicitly call for the copyright header to be adjusted when using the license. Apache 2.0 license does not. 

> Also should the file then be untracked by projen because currently once its generated its treated like any of the other synthesised files and is marked as read-only.

Yes that's correct, any file created by projen gets that treatment. If you want to manage the file yourself, just do that. 😁

Liam-Johnston:
Sorry a bit inexperienced when dealing with licenses, was hoping for a nice way to include the Apache license in a project I am working on substituting my name and the current year into the license instead of having the `[yyyy] [name of copyright owner]` part rendered.

All good, I will just include a license without using the License component 😄

hoegertn:
Can you elaborate on why you want to replace this text in the license file? It is NOT meant to be replaced...

Liam-Johnston:
Well currently when the License file gets generated it has the `[yyyy] [name of copyright owner]` text in the License [see here](https://github.com/projen/projen/blob/75283194163716080f2c75a40f5b5b934f3c9dee/license-text/Apache-2.0.txt#L190) and I was under the impression that this effectively makes the License invalid. 

I guess my concern is whether the License file that gets generated in the example projects [see here](https://github.com/projen/projen/blob/75283194163716080f2c75a40f5b5b934f3c9dee/src/javascript/node-project.ts#L1094) is voided because it doesn't have a name or year associated to it.

Just wondering the usefulness of synthesising these files if they aren't usable license files but they still remained tracked by projen and thus remain read only and any changes get overwritten.

Especially if downstream projects assume the license is valid which considering the arguments the License file accepts [(copyrightOwner & copyrightPeriod)](https://github.com/projen/projen/blob/75283194163716080f2c75a40f5b5b934f3c9dee/src/license.ts#L5) would be a reasonable assumption to make which you can see take place in the [projen-github-action-typescript](https://github.com/projen/projen-github-action-typescript) project already.

mrgrain:
This part is an appendix describing how you would add a license notification header to a file. That is something projen currently doesn't do.

Projen is aligned with other tools here, e.g. if you create a new GitHub repository with Apache 2.0 license:
https://github.com/mrgrain/shiny-broccoli/blob/main/LICENSE

Or the literal Apache webserver:
https://github.com/apache/httpd/blob/trunk/LICENSE#L189

Obviously you are welcome to have a different opinion and in the case simply don't use the projen feature. 😃

PR #3016, opened 2023-10-22
fix(object-file): allow boxed primitives, Set, Map and RegExp in values

tinovyatkin:
With current implementation ObjectFile (and all it subclasses) are throwing on encountering any perfectly valid Javascript boxed primitives in the object values (for example `new Sting('something')` or `new Boolean(something)`) even while those values may be stringified by `JSON.stringify` (the serialisation function that we are using) without any problem. This PR fixes that.

In addition, there are several JavaScript built-in objects that while having some limitations on JSII are very useful either for internal usage or for TS/JS-first usage of projen, however, those classes are missing built-in `.toJSON` method and so also resulting in throw from ObjectFile, particularly RegExp, Set and Map. This PR proposes to fix that by resolving RegExp (without flags) into its string source, Set into an array, and Map into an object. Note, that if we have a subclass of any of those built-in types that implements custom `.toJSON` then it will continue to work as with current version (see added tests).

As a simple demonstration of value of stringifiying `Set` I've included a small commit into this PR that switches implementation of ESLint fields that are expected to have non-repeated values (`plugins` and `extends`) from Array to Set using the fix implemented in this PR.

---

### DIFF
--- a/src/_resolve.ts
+++ b/src/_resolve.ts
@@ -22,7 +22,11 @@ export function resolve(value: any, options: ResolveOptions = {}): any {
   // that by default would be stringified as empty objects ('{}')
   // as they are missing a `toJSON` implementation.
   switch (true) {
-    case types.isRegExp(value) && !value.flags:
+    case types.isRegExp(value):
+      if (value.flags)
+        throw new Error(
+          "RegExp with flags should be explicitly converted to a string"
+        );
       return value.source;
 
     case types.isSet(value):
--- a/src/_resolve.ts
+++ b/src/_resolve.ts
@@ -1,3 +1,4 @@
+import { types } from "util";
 import { ResolveOptions, IResolvable } from "./file";
 
 function isResolvable(obj: any): obj is IResolvable {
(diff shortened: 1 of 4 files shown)


src/_resolve.ts
```diff
@@ -17,6 +18,23 @@ export function resolve(value: any, options: ResolveOptions = {}): any {
     return resolve(resolved, options);
   }
 
+  // Special resolution for few JavaScript built-in types
+  // that by default would be stringified as empty objects ('{}')
+  // as they are missing a `toJSON` implementation.
+  switch (true) {
+    case types.isRegExp(value) && !value.flags:
+      return value.source;
```
mrgrain, line 26: I think this is making to many assumptions. Specifically it is a different implementation then [RegExp.prototype.toString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/toString).

Happy to accept an implementation using `toString` instead. But also it seems okay to me to expect a user to call `toString()` themselves. For the Set & Map the conversion is less straightforward, so I agree the benefits outweigh the ambiguity.

mrgrain (changes requested):

src/_resolve.ts
mrgrain: Yes, hence my reluctance. In ESLint and Jest these are just strings that happen to be used as a RegExp. 
`/^tada$/.source` is equally easy to write.

tinovyatkin: `/^tada$/` seems to be even easier to write. But right now just throws with very very cryptic error message. I can keep that behaviour as is if you believe that is better DX than I'm proposing in this PR.

mrgrain: 🤦🏻 Of course. Can we throw with a better error in that case?

mrgrain (approved):
Sorry this took so long. Approved and should be good to ship once the conflict is resolved.

PR #3037, opened 2023-10-26
fix(node): yarn fails with `projen has unmet peer dependency "constructs@^10.0.0"`

mrgrain:
constructs is a peer-dependency of projen. yarn (and older versions of npm) don't install peer deps automatically. Therefore constructs needs to be explicitly added as a dev dependency.

Issue was introduced in https://github.com/projen/projen/pull/2974

---


vinayak-kukreja (approved):

mrgrain:
@moltar FYI Addressed your feedback in here because https://github.com/projen/projen/pull/2974 got merged before I could stop it 😅

PR #3048, opened 2023-10-27
feat: add Yarnrc component for YARN_BERRY

blimmer:
## Description

This PR significantly improves `projen`'s ability to use Yarn Berry.

Relates to #2980 

## Known Limitations
- There's not a way to pass `plugins` yet. When you run `yarn plugin import <plugin name>`, it writes to `.yarnrc.yml` and stores the plugin code in `.yarn`. We'll have to figure out the best way to handle that. The good news is that Yarn v4 includes the most common plugins by default again.

---


docs/node.md
```diff
@@ -111,6 +111,42 @@ You can also use dependabot or renovatebot to get Pull requests on dependency up
...
+});
+```
+
+#### Yarn Berry Corepack
+
+Following [Yarn's recommendation](https://yarnpkg.com/getting-started/install), we expect that
+[`corepack`](https://nodejs.org/api/corepack.html) is enabled on your machine. `corepack` transparently manages the
+`yarn` version based on the `packageManager` field in `package.json`.
+
+See [Yarn's installation documentation](https://yarnpkg.com/getting-started/install) for more information.
```
blimmer, line 148: Per your thread(s) in #2980 , it felt like this is an OK assumption to make.

src/javascript/node-package.ts
```diff
@@ -1483,6 +1499,50 @@ export class NodePackage extends Component {
...
+    project: Project,
+    { yarnBerryOptions }: NodePackageOptions
+  ) {
+    const {
+      version = "4.0.0",
+      // TODO: figure out how to use top-level NPM settings here (e.g., the result of `this.parseNpmOptions`)
+      // Some of these settings have overlap with `yarnrc.yml`. Questions:
+      // - Do we remove these options from the Yarnrc options interfaces and rely on them being set in the top-level options?
+      // - Do we add a parsing method, similar to `this.parseNpmOptions` that check for conflicts in the two configs?
+      // - Do we duplicate the logic, and set the relevant options here, too? (e.g., if this.npmRegistryUrl is set, set the relevant yarnrc option)
```
blimmer, line 1513: @mrgrain - curious to hear your thoughts on this question.

src/javascript/node-package.ts
```diff
@@ -1552,6 +1612,33 @@ export enum NpmAccess {
...
+   * Learn more at: https://yarnpkg.com/features/caching#zero-installs
+   *
+   * @default false
+   */
+  readonly zeroInstalls?: boolean;
+}
+
 interface NpmDependencies {
   readonly dependencies: Record<string, string>;
   readonly devDependencies: Record<string, string>;
```
blimmer, line 1644: Please expand the file below, it's the main thing this PR adds.

src/javascript/yarnrc.ts
```diff
@@ -0,0 +1,525 @@
+import { Component } from "../component";
+import { Project } from "../project";
+import { YamlFile } from "../yaml";
+
+/** https:/v3.yarnpkg.com/configuration/yarnrc#checksumBehavior */
+export enum YarnChecksumBehaviorV3 {
```
blimmer, line 6: There's a new checksum behavior (`RESET`) that exists in V4, so I have two separate enums. Thoughts?

src/javascript/yarnrc.ts
```diff
@@ -0,0 +1,525 @@
...
+/** https://yarnpkg.com/configuration/yarnrc#logFilters */
+export interface YarnLogFilter {
+  readonly code?: string;
+  readonly text?: string;
+  readonly level?: YarnLogFilterLevel;
+  readonly pattern?: string;
+}
+
+/** https://v3.yarnpkg.com/configuration/yarnrc#networkSettings */
+export interface YarnNetworkSettingV3 {
```
blimmer, line 44: Ditto here - anything that's suffixed `V3` is `V3` specific.

src/javascript/yarnrc.ts
```diff
@@ -0,0 +1,525 @@
...
+  SYMLINKS = "symlinks",
+}
+
+export enum YarnWorkerPoolMode {
+  ASYNC = "async",
+  WORKERS = "workers",
+}
+
+/** Configuration for .yarnrc.yml in Yarn Berry v3 */
+export interface YarnrcOptionsV3 {
```
blimmer, line 169: Some options were added/removed in V4, so I have two separate interfaces.

src/javascript/yarnrc.ts
```diff
@@ -0,0 +1,525 @@
...
+  /** https://v3.yarnpkg.com/configuration/yarnrc#cacheFolder */
```
blimmer, line 170: Copilot and ChatGPT helped me quite a bit in producing these, but a lot of the work was still manual. I was kinda lazy and just linked to the upstream docs instead of trying to script pulling the description in.

src/javascript/yarnrc.ts
```diff
@@ -0,0 +1,525 @@
...
+  /** https://v3.yarnpkg.com/configuration/yarnrc#unsafeHttpWhitelist */
+  readonly unsafeHttpWhitelist?: string[];
+  /** https://v3.yarnpkg.com/configuration/yarnrc#virtualFolder */
+  readonly virtualFolder?: string;
+  /** https://v3.yarnpkg.com/configuration/yarnrc#yarnPath */
+  readonly yarnPath?: string;
+}
+
+/** Configuration for .yarnrc.yml in Yarn Berry v4 */
+export interface YarnrcOptionsV4 {
```
blimmer, line 335: This one's V4-specific. I can also just call this `YarnrcOptions` without the `V4` if that makes more sense.

blimmer (commented):
@mrgrain - I had some time today to work on this. Could you take a look at the inline questions on this draft PR review? Once I get confirmation on how you'd like to proceed, I can finish this up and get it ready for review.

TYIA!

src/javascript/node-package.ts
```diff
@@ -1483,6 +1499,59 @@ export class NodePackage extends Component {
       : this.installTask;
     runtime.runTask(taskToRun.name);
   }
+
+  private configureYarnBerry(project: Project, options: NodePackageOptions) {
+    const {
+      version = "4.0.1",
```
blimmer, line 1505: `corepack` requires a fully qualified version for the `packageManager` property. If you provide values like `yarn@4`, `yarn@stable`, `yarn@~4`, etc. you get this error:

```
Usage Error: Invalid package manager specification in package.json; expected a semver version

$ yarn ...
```

So, unfortunately, we have to specify a full version here.

src/javascript/node-package.ts
```diff
@@ -1483,6 +1499,59 @@ export class NodePackage extends Component {
...
+
+    // Set the `packageManager` field in `package.json` to the version specified. This tells `corepack` which version
+    // of `yarn` to use.
+    this.addField("packageManager", `yarn@${version}`);
+    this.configureYarnBerryGitignore(zeroInstalls);
+
+    new Yarnrc(project, yarnRcOptions);
+  }
+
+  private checkForConflictingYarnOptions(
```
blimmer, line 1519: I figured this was OK to start off: https://github.com/projen/projen/pull/3048#discussion_r1375096442

src/javascript/node-package.ts
mrgrain: Ugh. That's annoying. How about we make the version required?

src/javascript/node-package.ts
```diff
@@ -1552,6 +1621,33 @@ export enum NpmAccess {
...
+   * @default - 4.0.1
+   */
+  readonly version?: string;
+
+  /**
+   * The yarnrc configuration.
+   *
+   * @default - a blank Yarn RC file
+   */
+  readonly yarnRcOptions?: YarnrcOptionsV3 | YarnrcOptionsV4;
```
mrgrain, line 1640:
```suggestion
  readonly yarnRcOptions?: YarnrcOptions;
```
This should be an enum like class:

```ts
YarnrcOptions.v3(options: YarnrcOptionsV3);
YarnrcOptions.v4(options: YarnrcOptionsV4);
```

Alternatively we can forgo the strict typing here, have a single `YarnrcOptions` interface and validate differently in the `v3` / `v4` static methods.

mrgrain: Or maybe it's even just the same interface and we validate it based on the version that is set. Differences between v3/v4 can be documented.

mrgrain (changes requested):
Good stuff! Obviously we should automated the options, but we don't have infrastructure for this. 🤷🏻

src/javascript/node-package.ts
blimmer: How would that work if people didn't provide `yarnBerryOptions` at all? e.g.,

```ts
  // ...
  packageManager: javascript.NodePackageManager.YARN_BERRY,
});
```

Would we throw a runtime error?

src/javascript/node-package.ts
blimmer: I went ahead and used the same interface. I marked removed options with a `@deprecated` JSDoc annotation and added validation logic to warn users about invalid config usage.

src/javascript/node-package.ts
mrgrain: Yeah that's annoying. I think we might be able to turn `javascript.NodePackageManager` into an enum like class. With static props for the existing values and `YARN_BERRY()` a static method that requires a version?

Which feels like what we would really need to do for all corepack package managers?! So we'd end up having an interface like this:

```ts
  // Corpack
  packageManager: javascript.CorePack.yarn('1.23.1'),
  packageManager: javascript.CorePack.yarn('4.0.1'),
  packageManager: javascript.CorePack.pnpm('8.0.0'),
  packageManager: javascript.CorePack.npm('100.0'),

 // pre-installed 
  packageManager: javascript.NodePackageManager.NPM,
  packageManager: javascript.NodePackageManager.PNPM,
  packageManager: javascript.NodePackageManager.YARN_CLASSIC,
  packageManager: javascript.NodePackageManager.YARN_BERRY,
});
```
The pre-installed ones maybe just all need to be deprecated right away? Or at least for `YARN_BERRY` which recommends corepack.

But let's do this separately. Happy to keep the default version for now.

blimmer: Sounds good - I filed https://github.com/projen/projen/issues/3070 to track that. FYI, `npm` is the only one that cannot be managed by `corepack` (lol). If we're cool leaving the default version for now, this is ready for review & merge.

mrgrain: Yep. I'll give it a final review tomorrow.

blimmer:
Btw, although this isn't directly related, we might be getting some nice Corepack reliability improvements in berry: https://github.com/yarnpkg/berry/issues/5912

mrgrain (approved):

Issue #3056, opened 2023-10-29
Projen doesn't use `components` tree internally


tinovyatkin:
The change at #2974 suppose to make projen based on `constructs`, but while adding inheritance from `Construct` it actually completely ignoring most of utilities built-in with constructs disallowing re-using of existing developers knowledge while working with `constructs`.

Some important aspects that are completely missing:

1. `node.addValidation` are silently ignored
2. `node.tryRemoveChild` behaviour is unpredictable
3. `project.node.addChild` is just ignored
4. It would be impossible to implement [dependencies](https://github.com/aws/constructs/blob/4d7391641b7a2f3f448b50def4586cd9684f39bc/src/construct.ts#L314) between components (which component should be synthesised first) in such implementation.

2-4 (and many other) are happens because implementation-wise we kept all previous logic and maintaining our own `_files` and `_components` flat arrays of components rather than relying on `constructs` utility for tree walking.

Some potentially great features (like `Directory` component that would automatically scope all `Files` inside it) that may be available almost automatically with `constructs` in fact are still impossible because we don't use nodes tree at all.

The concept of `defaultChild` (that is essential for escape hatches in `constructs` based composition, and I would expect for example for ESLint/Jest/Prettier component `defaultChild` to be the config file - this way we don't need to expose `file` property everywhere) are also completely ignored due to fact that we basically thrown out whole `node id` concept.

mrgrain:
Thanks for the detailed write up.

Validations I missed. Should be easy to add in now.

I'm mostly aware of 2-4 due to the lack of the internals actually using the tree and noted that follow up work will be required. I'm assuming the issue with `addNode` and `tryRemoveNode` are relating to this.

Re default nodes, that's definitely something we should consider. Right now most components still don't place the file inside their own scope, but rather scoping it to the project.

mrgrain:
First follow-up PR: https://github.com/projen/projen/pull/3061

I think dependency-ordered tree traversal should be next. This would mostly change the current implementation of `Project.synth()`. Ideally the dependency ordered synth would allow us to get rid of the separation between (sub)project and component and just automatically get the correct order from the tree.

After that it will be easy to add in validation. My current thinking here is that after `preSynth` but before `synth`, we first `lock` all constructs and then run `validate()`.

PR #3061, opened 2023-10-30
fix: construct tree not used internally

mrgrain:
Change the implementations of `project.subprojects`, `project.root`, `project.components` and `projects.files` to operate of the construct tree instead of arrays.

Addresses parts of #3056

---

### DIFF
--- a/src/project.ts
+++ b/src/project.ts
@@ -359,7 +359,7 @@ export class Project extends Construct {
    * The root project.
    */
   public get root(): Project {
-    return this.node.scopes.find(isProject) ?? this;
+    return isProject(this.node.root) ? this.node.root : this;
   }
 
   /**
@@ -378,12 +378,7 @@ export class Project extends Construct {
    * Returns all the subprojects within this project.
    */
   public get subprojects(): Project[] {
-    return this.node
-      .findAll()
-      .filter(
-        (p): p is Project =>
-          isProject(p) && p.parent?.node.path === this.node.path
-      );
+    return this.node.children.filter(isProject);
   }
 
   /**
--- a/src/project.ts
+++ b/src/project.ts
@@ -22,7 +22,12 @@ import { Renovatebot, RenovatebotOptions } from "./renovatebot";
 import { Task, TaskOptions } from "./task";
 import { Tasks } from "./tasks";
 import { isTruthy } from "./util";
-import { isProject, findClosestProject, tagAsProject } from "./util/constructs";
+import {
+  isProject,
+  findClosestProject,
+  tagAsProject,
+  isComponent,
+} from "./util/constructs";
 
 /**
  * Options for `Project`.
@@ -180,11 +185,6 @@ export class Project extends Construct {
   public readonly outdir: string;
 
   /**
-   * The root project.
-   **/
-  public readonly root: Project;
-
-  /**
    * Project tasks.
    */
   public readonly tasks: Tasks;
@@ -235,8 +235,6 @@ export class Project extends Construct {
    */
   public readonly commitGenerated: boolean;
 
-  private readonly _components = new Array<Component>();
-  private readonly _subprojects = new Array<Project>();
   private readonly tips = new Array<string>();
   private readonly excludeFromCleanup: string[];
   private readonly _ejected: boolean;
@@ -271,10 +269,6 @@ export class Project extends Construct {
     }
 
     this.outdir = outdir;
-    this.root = this.parent ? this.parent.root : this;
-
-    // must happen after this.outdir, this.parent and this.root are initialized
-    this.parent?._addSubProject(this);
 
     // ------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


mrgrain:
@tinovyatkin This still doesn't implement dependencies or `validate()`, but should fix the issues with tree manipulation.

src/project.ts
```diff
@@ -602,13 +602,13 @@ export class Project extends Construct {
 
     this.preSynthesize();
 
-    for (const comp of this._components) {
+    for (const comp of this.components) {
       comp.preSynthesize();
     }
```
tinovyatkin, line 607: I'm assuming pre-synthesize stage in projen should be similar to pre-synth stage in CDK, and so, at this stage nodes tree still can be changed by a components (sort of Aspects concept).  Can it? If it can, then we should walk the tree differently (as `this.components` returns static array rather than live tree branch). If  it can't then we should throw on attempt and probably document.

src/project.ts
```diff
@@ -361,26 +355,46 @@ export class Project extends Construct {
...
+      .findAll()
+      .filter(
+        (p): p is Project =>
+          isProject(p) && p.parent?.node.path === this.node.path
+      );
   }
 
   /**
-   * All files in this project.
+   * All files in this project and all subprojects.
```
tinovyatkin, line 390: That may be a breaking change.

src/project.ts
```diff
@@ -361,26 +355,46 @@ export class Project extends Construct {
...
   }
 
   /**
    * Returns all the subprojects within this project.
    */
-  public get subprojects() {
-    return [...this._subprojects];
+  public get subprojects(): Project[] {
+    return this.node
+      .findAll()
```
tinovyatkin, line 382: You don't need `findAll` just to compare `p.parent.node.path === this.node.path`. There is a `node.children` for that

src/project.ts
```diff
@@ -361,26 +355,46 @@ export class Project extends Construct {
...
    */
-  public get components() {
-    return [...this._components];
+  public get components(): Component[] {
+    return this.node
+      .findAll()
+      .filter(
+        (c): c is Component =>
+          isComponent(c) && c.project.node.path === this.node.path
+      );
```
tinovyatkin, line 374: This will return only direct children of Project. Wouldn't we allow scoped components?

src/project.ts
```diff
@@ -361,26 +355,46 @@ export class Project extends Construct {
     }
   }
 
+  /**
+   * The root project.
+   */
+  public get root(): Project {
+    return this.node.scopes.find(isProject) ?? this;
```
tinovyatkin, line 362: Why can't we use `node.root` for that?

mrgrain: We can. This was trying to be more in the spirit of constructs, since technically any Construct can be the root of the tree, not just a project. In practice you would have to do some trickery to get something else than a project as the root.

src/project.ts
mrgrain: Ah you're right. I got confused between this and `tryFindFile`.

src/project.ts
mrgrain: This should return all components that are scoped to this project, but not to any of the subprojects.
I've added a test to demonstrate this behavior.

src/project.ts
mrgrain: Similar to `.root`. Trying to make this correct for cases where somebody somehow managed to create a subproject that's scoped to something else.

src/project.ts
tinovyatkin: I believe that the only root node in the spirit of constructs can be the Construct that we allow to establish with undefined scope. It's only Project in our case (as `Stage` / `App` in CDK). So, `node.root` must be a root Project and can't be anything else.

src/project.ts
tinovyatkin: Not sure I understand. Here you are doing `findAll` (whole tree walk) but then `filter` where you checking `p.parent?.node.path === this.node.path` - that means you basically converting whole tree walk into `node.children` that is way faster. This would not make it work with scoped subprojects.

src/project.ts
tinovyatkin: I believe we you are repeating the same error by anchoring to our pre-saved `project` field rather than walking up `node.scopes`. I would suggest we should implement static `Project.of(IConstruct)` (that would be similar to `Stack.of` in CDK) and use it here.

mrgrain: `Stack.of` actually works the same because it's also caching the found Stack instance.

I'm not aware of a way to change the scope of a construct, so the pre-saved  `project` should always be correct.

src/project.ts
mrgrain: Consider this structure (lines represent scope, `parent` the parent property)
```
Project
└── Component
    └── ScopedProject(parent: Project)
```

The current code would find `ScopedProject` and correctly identify it as a subproject of `Project`. But you would have  to seriously hack to code to make this happen (if it's even possible). So like with `root`, I'm happy to simplify the code here.

src/project.ts
mrgrain: My goal on this iteration was to keep the current functionality the same. I believe this already a constraint of the existing implementation, so while we are not gaining anything (just yet), we also don't loose anything. I'm not sure if you are saying that this changes the current behavior.

---

> pre-synth stage in CDK
Do you mean Aspects?

`preSynthesize` is similar to Aspects in CDK, and like the CDK adding [another Aspect at this stage just isn't possible](https://github.com/aws/aws-cdk/blob/dd912daf2b91a4a32064341e92863afbd9eeebdd/packages/aws-cdk-lib/core/lib/private/synthesis.ts#L241)(*). We definitely can add a warning here.


(*) It probably is possible, but complex. I think what we need to do is evaluate the order, preSynth the component, remember that we have preSynth'ed the component, lock the construct, re-evaluate the order, and continue with the first component that hasn't been preSynth'ed yet.

tinovyatkin: I mean Prepare stage (as it called in [the official docs](https://docs.aws.amazon.com/cdk/v2/guide/apps.html#lifecycle)), but yes, basically when aspects runs.

It's definitely possible and constructs provides a mechanism for that, but I'm perfectly OK to leave that for the next iterations (probably dependencies will require that anyway).

src/project.ts
tinovyatkin: NIT: What I mean, that saved references to `parent` and `project` are mostly artefacts of previous way to maintain the tree. In this case we can easily use that `constructs` mechanisms and avoid coupling with those fields:

```ts
static of(scope: IConstruct) {
    if (isProject(scope)) return scope;
    if (!scope.node.scope)
      throw new Error(
        `${construct.constructor?.name ?? "Construct"} at '${
          Node.of(construct).path
        }' should be created in the scope of a Project, but no Project found`
      );
    return Project.of(scope.node.scope);
  }

  /**
   * Returns all the components within this project.
   */
  public get components(): Component[] {
    return this.node
      .findAll()
      .filter((c): c is Component => isComponent(c) && Project.of(c) === this);
  }

  /**
   * Returns all the subprojects within this project.
   */
  public get subprojects(): Project[] {
    return this.node
      .findAll()
      .filter((p): p is Project => isProject(p) && Project.of(p) === this);
  }
```

tinovyatkin (approved):

kaizencc (approved):

PR #3076, opened 2023-11-01
fix(typescript): bump @types/node version to 18

matt9ucci:
Node.js 16 reached EOL on September 11th, 2023.
https://nodejs.org/en/blog/announcements/nodejs16-eol
https://nodejs.org/en/about/previous-releases

---


mrgrain:
Thanks for that @matt9ucci ! Is it just the `@types/node` or are we still defaulting to 16 in other places?

matt9ucci:
Thank you for the comment @mrgrain! It is just the `@types/node`.

It seems that the legacy JSII (1.X) is not compatible with `@types/node` v18. Some test cases [failed to compile it](https://github.com/projen/projen/actions/runs/6717467045/job/18255368027?pr=3076#step:5:721).
Therefore, I added [the lines](https://github.com/projen/projen/pull/3076/files#diff-c5619f357a419a8975c6a32de334c4348d90397ad0102eacd5c5a27300c49015R428-R432) which keep v16 for legacy JSII projects.

src/cdk/jsii-project.ts
```diff
@@ -425,6 +425,11 @@ export class JsiiProject extends TypeScriptProject {
 
       // https://github.com/projen/projen/issues/2264
       this.package.addPackageResolutions("@types/babel__traverse@7.18.2");
+
+      this.addDevDeps(
+        // https://github.com/projen/projen/pull/3076
+        `@types/node@^${major(this.package.minNodeVersion ?? "16.0.0")}`
+      );
```
mrgrain, line 432: Can you make this conditional? Only do this if we are on 1.x
```suggestion
      if ((options.jsiiVersion ?? "1.x").startsWith('1.x')) {
        this.addDevDeps(
          // https://github.com/projen/projen/pull/3076
          `@types/node@^${major(this.package.minNodeVersion ?? "16.0.0")}`
        );
      }
```

mrgrain (changes requested):
See suggestion

src/cdk/jsii-project.ts
matt9ucci: Fixed.

mrgrain (approved):

mrgrain (approved):

PR #3084, opened 2023-11-03
feat(python): upgrade pytest default version

scottschreckengaust:
The default version was failing with the default:
`npx projen new awscdk-app-py`

```
from projen.awscdk import AwsCdkPythonApp

project = AwsCdkPythonApp(
    author_email="user@example.com",
    author_name="Example User",
    cdk_version="2.1.0",
    module_name="module-name",
    name="name",
    version="0.1.0",
    pytest_options={"version": "7.4.3"},
)

project.synth()
```

---


src/python/pytest.ts
```diff
@@ -6,7 +6,7 @@ export interface PytestOptions {
   /**
    * Pytest version
    *
-   * @default "6.2.1"
+   * @default "7.4.3"
```
mrgrain, line 9: This will need to change in line 32 as well

mrgrain (changes requested):

src/python/pytest.ts
scottschreckengaust: Added the snapshots too

mrgrain (approved):

PR #3090, opened 2023-11-06
feat(yarn): define Yarn Berry plugins

blimmer:
In #3048 we didn't implement support for `plugins` yet. This is a follow-up to introduce a basic interface.

Relates to #2980 

---

### DIFF
--- a/src/javascript/yarnrc.ts
+++ b/src/javascript/yarnrc.ts
@@ -133,6 +133,12 @@ export interface YarnSupportedArchitectures {
   readonly libc?: string[];
 }
 
+export interface YarnPlugin {
+  readonly checksum?: string;
+  readonly path: string;
+  readonly spec: string;
+}
+
 /** https://yarnpkg.com/configuration/yarnrc#cacheMigrationMode */
 export enum YarnCacheMigrationMode {
   REQUIRED_ONLY = "requied-only",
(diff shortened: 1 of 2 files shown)


src/javascript/yarnrc.ts
```diff
@@ -353,6 +359,22 @@ export interface YarnrcOptions {
...
+   * 2. Make a note of the `plugins` section of `.yarnrc.yml`. Copy the `spec`, `path`, and (optionally) `checksum` values to this config.
+   * 3. Run `yarn projen` to update your config.
+   *
+   * This process could be improved in the future. This is just an MVP solution to allow using plugins with `projen`
+   * and Yarn Berry.
+   *
+   * Note that Yarn v4+ includes all official plugins (e.g., workspace-tools, interactive-tools) by default: https://yarnpkg.com/blog/release/4.0#breaking-changes
+   *
+   * See https://yarnpkg.com/cli/plugin/import for more details.
+   */
```
blimmer, line 376: In a perfect world, it'd be nice if people could pass a list of plugins that we could call `import` on for them. For instance:

```ts
plugins: ["https://github.com/foo/example-plugin"];

plugins.forEach((plugin) => { execSync(`yarn plugin import ${plugin}`); }
```

then observe the changes to `.yarnrc.yml` and commit them.

However, this is more complex than I have time to work on right now. Therefore, this felt like a good MVP option.

mrgrain:
@blimmer Is this option really so much better than using a low-level patch? I'm worried we are locking ourselves into an API that we then have to spend extra work on to deprecate and replace.

blimmer:
I can definitely see the argument for a low-level patch instead of exposing this. I don't really know what the level of effort would be to run the `plugin import` and observe the changes to `.yarnrc.yml`. Is that hard? Are there examples of this kind of thing already?

mrgrain:
`package.json` does something similar. The steps are basically:

- Load the current plugin data from the file
- Synth the file, including the current plugin data
- Establish if plugin data has changed (new plugin added/old plugin removed) or maybe if a plugin needs installing
- If so, make the file readable
- Then run `yarn plugin import` for any missing plugin
- Make sure the rest of the file is still looking as expected
- Write protect the file

Or I guess we could also make the file hybrid like `package.json`, but ideally I'd prefer not to,


"Establish if plugin data has changed" - I suspect this will be the trickiest part. We could check the contents of `.yarn/plugins` and compare it with the expected plugins.

PR #3091, opened 2023-11-06
fix(yarn): npmrc and yarnrc cannot be set to the same value

blimmer:
Fixes issue introduced with #3048.

We should allow people to set potentially conflicting values to the same value. Previously, I was just checking to see if they were both set at the same time.

---


src/javascript/node-package.ts
```diff
@@ -1519,15 +1519,23 @@ export class NodePackage extends Component {
   }
 
   private checkForConflictingYarnOptions(yarnRcOptions: YarnrcOptions) {
```
blimmer, line 1521: We should allow them both to be set, we just care if they're in conflict with each other.

mrgrain (approved):

PR #3134, opened 2023-11-20
fix(eslint): use recommended configuration from eslint-plugin-prettier

gmeligio:
## Motivation

This is an *small* change to reduce prettier configuration in projen. Use [the recommended configuration from `eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) and remove it from `projen`.

> Then you need to add `plugin:prettier/recommended` as the _last_ extension in your `.eslintrc.json`:

   ```json
   {
     "extends": ["plugin:prettier/recommended"]
   }
   ```
> Exactly what does `plugin:prettier/recommended` do? Well, this is what it expands to:

```json
{
  "extends": ["prettier"],
  "plugins": ["prettier"],
  "rules": {
    "prettier/prettier": "error",
    "arrow-body-style": "off",
    "prefer-arrow-callback": "off"
  }
}
```

## Changes

1. Don't re-add `prettier` to "plugins" .
1. Don't re-add `prettier`  to "extends".
1. Don't re-add `"prettier/prettier": ["error"]` to "rules".

## Test

1. Modified the test for checking that prettier is configured in eslint with `"plugin:prettier/recommended"`
1. Removed the test for the existence of `"prettier/prettier": ["error"]` in the project's eslint config. It could be possible to test the existence in the full resolved configuration with `eslint --print-config` . That would require an integration test using something like `executeTask` from test/tasks/runtime.test.ts that runs the `eslint` in the synthesized project.
1. Comparison of the full resolved configuration in the projen repository before and after the change, showing that, after resolving all the `extends`, the `plugins` are the same, and the changed rule is reordered.
image

---


mrgrain:
This looks good thank you! I'll need a bit more time to review this week.

mrgrain (approved):
Nice work, thank you!

gmeligio:
Thank you @mrgrain for the review!

I'm sorry I missed a `console.log` and it got merged. It will add noise to the test output. I'll remove that in another PR I already have to not create only one for this console.log. Sorry again for what happened.

mrgrain:
> Thank you @mrgrain for the review!
> 
> I'm sorry I missed a `console.log` and it got merged. It will add noise to the test output. I'll remove that in another PR I already have to not create only one for this console.log. Sorry again for what happened.

No worries @gmeligio I missed it during the review! 🙈

PR #3136, opened 2023-11-20
fix(upgrades): upgrades failing with wrap-ansi ESM cannot be loaded

kaizencc:
Backport of https://github.com/cdklabs/cdklabs-projen-project-types/pull/312, but the code is substantially different because I could modify the actual source and not have to use escape hatches.

> Basically npm-check-updates is introducing a lot of packages to the dependency tree and the single yarn upgrade npm-check-updates command appears to be causing the incorrect lockfile.

> This change removes npm-check-updates from the dependency tree, in favor of calling it via npx. Thus solving the above problem.
 
Co-authored-by: @mrgrain [mail@moritzkornher.de](mailto:mail@moritzkornher.de)

Here's the reason this change works: `npm-check-updates` depends on `@isaacs/cliui`. This last package uses an [NPM-specific feature in `package.json`](https://github.com/isaacs/cliui/blob/main/package.json#L51) to install various package dependencies under alternate names. In certain cases (we don't quite understand yet) Yarn 1 builds an invalid dependency tree if this feature is used, putting the ESM module in the place where the CJS module is expected. When it comes time to execute `npm-check-updates`, Node loads the wrong module and errors out. By removing `npm-check-updates` from our dependency tree, and instead using `npx` to run it, the tool and its dependencies get installed by NPM in a node_modules tree somewhere off to the side that is not managed by Yarn. 

---


src/javascript/upgrade-dependencies.ts
```diff
@@ -240,16 +239,8 @@ export class UpgradeDependencies extends Component {
...
-    // in it or one of its dependencies. This will make upgrade workflows
-    // slightly more stable and resilient to upstream changes.
-    const ncuDep = this.project.deps.all.find(
-      (d) => d.name === "npm-check-updates"
-    )!;
-    steps.push({ exec: this.renderUpgradePackagesCommand([ncuDep.name]) });
-
     const ncuCommand = [
-      "npm-check-updates",
+      "npx npm-check-updates@latest",
```
mrgrain, line 243: I think we should to adjust this command for package managers

Afaik it's:
`npx` for npm and yarn_classic.
`pnpx` for pnpm
`yarn dlx` for yarn_berry

mrgrain (commented):
Nice! I think just one small improvement.

mrgrain (dismissed):
See inline

src/javascript/upgrade-dependencies.ts
```diff
@@ -230,6 +229,21 @@ export class UpgradeDependencies extends Component {
   }
 
   private renderTaskSteps(): TaskStep[] {
+    function executeCommand(packageManager: NodePackageManager): string {
```
kaizencc, line 232: `YARN` is deprecated and matches `YARN_CLASSIC`.... `YARN2` is deprecatd and matches `YARN_BERRY`.... I looked up `BUN`'s equivalent of `npx` and it seemed to be `bunx`.

rix0rrr (commented):
Won't at least `npx` throw up a prompt like `Are you sure you want to download package <x> ? (y)` ?

We used to have a version constraint on this package. Are we okay dropping it? (Seems dangerous)

rix0rrr:
Needs a comment in the source code for why we're doing it this way.

rix0rrr:
I added the reason for why this works to the PR, but now I have a follow-up question: do we know for a fact that `pnpx` and `bunx` etc also support this renaming feature?

mrgrain:
> Won't at least `npx` throw up a prompt like `Are you sure you want to download $PACKAGE? (y)` ?

The command is executed with `CI=1` so no confirmation is printed.


> We used to have a version constraint on this package. Are we okay dropping it? (Seems dangerous)

Good point. We probably should use `@16`

rix0rrr (approved):

PR #3148, opened 2023-12-01
feat!: bump `minNodeVersion` to `18.12.0`

vinayak-kukreja:
BREAKING CHANGE: Node16 has been EOL for a while. Projen now requires Node18. Please upgrade your Node version.

---

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "@types/node",
-      "version": "^16",
+      "version": "^18",
       "type": "build"
     },
     {
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -76,7 +76,7 @@ const project = new cdk.JsiiProject({
 
   projenDevDependency: false, // because I am projen
   releaseToNpm: true,
-  minNodeVersion: "16.0.0", // Do not change this before a version has been EOL for a while
+  minNodeVersion: "18.14.0",
   workflowNodeVersion: "18.14.0",
 
   codeCov: true,
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -76,7 +76,7 @@ const project = new cdk.JsiiProject({
 
   projenDevDependency: false, // because I am projen
   releaseToNpm: true,
-  minNodeVersion: "18.14.0",
+  minNodeVersion: "18.12.0",
   workflowNodeVersion: "18.14.0",
 
   codeCov: true,
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@types/glob": "^7.2.0",
     "@types/ini": "^1.3.33",
     "@types/jest": "^27",
-    "@types/node": "^16",
+    "@types/node": "^18",
     "@types/semver": "^7.5.6",
     "@types/yargs": "^16.0.9",
     "@typescript-eslint/eslint-plugin": "^6",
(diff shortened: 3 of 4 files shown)


mrgrain (changes requested):

vinayak-kukreja:
Hey @tinovyatkin , could you tell me more about what issues CDK projects are facing with node upgrade with lambda runtime upgrade?

Node16 is EOL and we should be moving to Node18. I see JSII has also moved to it: https://github.com/aws/jsii/releases/tag/v1.92.0. Is there any blocker for you with this upgrade?

tinovyatkin:
@vinayak-kukreja The JSII PR you linked is perfectly fine from my point of view as it adds warnings of not being tested on that platform but keeps minimum Node version the same.

The change in this CR without changing any code or dependency just prevent `projen` from being installed on Node 16 without any benefits for the customers.

Regarding Node version upgrade in CDK project I meant that before Node 18 Lambda runtime were included `aws-sdk` v2 and many lambdas are using that. Now 18.x includes aws-sdk v3 and, while it's great change, to migrate a significant amount of lambdas from `aws-sdk v2` to `aws-sdk v3` some developers bandwidth needed to allocated and testing performed. Not every team has that immediately available and it's not a change you may just do with "closed eyes". So, in some our teams we have to postpone 18.x migration for lambda (and whole project, as we are unit-testing our lambdas on the project' version of Node) until we have a bandwidth for AWS SDK version upgrade also.

mrgrain:
jsii explicitly doesn't support & test EOL versions of node. The warning is a nice touch though. 
The next version of `jsii-compiler` (which will be `5.3`) will also change the minNodeVersion to 18.

> i believe many projects obusing this field (minimum! Node version) for sake of nothing. 

The sake is that it is a way to indicate which node versions a project is actively tested with. I agree that this skews heavily towards one side of scale on how you can use this field. But to quote the official npm docs: 

"You can specify the version of node that your stuff works on:"
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

Once we stop testing with an older version of node, we _just don't know anymore_ if iur stuff is working with it.

---

It is important to note that this field is advisory only in npm:

> Unless the user has set the [engine-strict config](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

Yarn classic swaps the default to enforcing `engines`, but you can easily disable it: `yarn install --ignore-engines`


--- 


Now, I don't have a particular rush to get of Node 16 and would be happy to leave this for a bit. GitHub Actions will probably the thing that forces us to move on at some point in the near future: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

tinovyatkin:
> Now, I don't have a particular rush to get off Node 16 and would be happy to leave this for a bit.

That's my whole point - if projen is not pushed by any dependency or code feature - what is the rush to disable Node 16? Any benefits for customers from merging this PR?

garysassano:
Node 16 reached EOL. Also GitHub urged devs to just [move to Node20 directly](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/).

garysassano:
Node 16 reached EOL. Also GitHub urged devs to just [move to Node 20 directly](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/).

PR #3151, opened 2023-12-02
feat(gitlab): add gitlab multiple cache support

### LINKED ISSUE #3022: Add support for Gitlab multiple caches
Thank you for the amazing work, I am a big fan of projen.

I have noticed a slight gap in the gitlab model.

Support for multiple caches were introduced in GitLab 13.10, see [here](https://docs.gitlab.com/ee/ci/caching/index.html#use-multiple-caches).

Currently the `Job` model only supports type `Cache` for the value `cache`, see [here](https://github.com/projen/projen/blob/75283194163716080f2c75a40f5b5b934f3c9dee/src/gitlab/configuration-model.ts#L310-L311).

Therefore we cannot support a job like:
```
test-job:
  stage: build
  cache:
    - key:
        files:
          - Gemfile.lock
      paths:
        - vendor/ruby
    - key:
        files:
          - yarn.lock
      paths:
        - .yarn-cache/
```

I plan to contribute and fix this gap myself, I am just opening this issue for visibility.

alejandrolorefice:
### Description of changes

In this PR I'm adding support for Gitlab multiple cache definition with a maximum of 4 as described in [their docs](https://docs.gitlab.com/ee/ci/caching/). Note that I've explicitly limited to 4 cause otherwise Gitlab will throw an error when running the pipeline definition produced by a projen with more than 4 cache specs

Closes #3022 

BREAKING CHANGE: current cache definitions will need to be changed into an array or will get a compile time exception.

---


mrgrain:
Change looks okay. I'll need to check how this ends up in jsii languages. But I'm on holiday, so it will have to wait until next week. (Unless you want to do that!)

If it doesn't come out with a reasonable type for them, will can make it an array and add some runtime validation.

alejandrolorefice:
> Change looks okay. I'll need to check how this ends up in jsii languages. But I'm on holiday, so it will have to wait until next week. (Unless you want to do that!)
> 
> If it doesn't come out with a reasonable type for them, will can make it an array and add some runtime validation.

I've checked and jsii doesn't like tuples. I'm switching to an array with runtime validation.

alejandrolorefice:
WDYT about this implementation? @mrgrain

src/gitlab/configuration-model.ts
```diff
@@ -74,8 +74,8 @@ export interface Default {
   readonly artifacts?: Artifacts;
   /* Defines scripts that should run *before* all jobs. Can be overriden by the job level `afterScript`. */
   readonly beforeScript?: string[];
-  /* A list of files and directories to cache between jobs. You can only use paths that are in the local working copy. */
-  readonly cache?: Cache;
+  /* A list of cache definitions (máx. 4) with the files and directories to cache between jobs. You can only use paths that are in the local working copy. */
```
mrgrain, line 77:
```suggestion
  /* A list of cache definitions (max. 4) with the files and directories to cache between jobs. You can only use paths that are in the local working copy. */
```

src/gitlab/configuration-model.ts
```diff
@@ -307,8 +307,8 @@ export interface Job {
   readonly artifacts?: Artifacts;
   /* Defines scripts that should run *before* the job. */
   readonly beforeScript?: string[];
-  /* A list of files and directories to cache between jobs. You can only use paths that are in the local working copy. */
-  readonly cache?: Cache;
+  /* A list of cache definitions (máx. 4) with the files and directories to cache between jobs. You can only use paths that are in the local working copy. */
```
mrgrain, line 310:
```suggestion
  /* A list of cache definitions (max. 4) with the files and directories to cache between jobs. You can only use paths that are in the local working copy. */
```

src/gitlab/configuration.ts
```diff
@@ -80,9 +80,9 @@ export class CiConfiguration extends Component {
    */
   public readonly defaultBeforeScript: string[] = [];
   /**
-   * A default list of files and directories to cache between jobs. You can only use paths that are in the local working copy.
+   * A default list of cache definitions (máx. 4) with the files and directories to cache between jobs. You can only use paths that are in the local working copy.
    */
-  public readonly defaultCache?: Cache;
+  public defaultCache?: Cache[];
```
mrgrain, line 85: What we normally do here is turn the property into a private, mutable one and then expose it via a getter.
```suggestion
  private _defaultCache?: Cache[];
  
  public get defaultCache: Cache[] | undefined {
    return this._defaultCache;
  }
```

mrgrain (changes requested):
Nice, just one design change

src/gitlab/configuration.ts
```diff
@@ -308,9 +308,32 @@ export class CiConfiguration extends Component {
...
+  }
+
+  private isValidCacheSetup(caches: Cache[]): Boolean {
+    const MAX_CONFIGURABLE_CACHES = 4;
+    return caches.length <= MAX_CONFIGURABLE_CACHES;
+  }
+
+  private assertIsValidCacheSetup(caches: Cache[]) {
+    if (!this.isValidCacheSetup(caches)) {
+      throw new Error(`${this.name}: GitLab CI defines more than 4 caches.`);
```
mrgrain, line 324:
```suggestion
      throw new Error(`${this.name}: GitLab CI can only define up to 4 caches, got: ${caches.length}`);
```

mrgrain (approved):
Thank you!

PR #3156, opened 2023-12-05
fix(node): packaging bundled dependencies fails in pnpm projects

agdimech:
When using `pnpm` as a package manager, bundled dependencies are hard linked and if using `npm pack`, these dependencies are not correctly included in the bundle. `pnpm pack` handles this by correctly including these hard linked packages into the output bundle.

---


agdimech:
@mrgrain - can I please get a review for this?

mrgrain (approved):

Issue #2051, opened 2022-08-20
Task runner contract (pluggable task runners)


moltar:
Following up on #1826 to make a more generic issue to foster further discussion.

Specifically this comment: https://github.com/projen/projen/issues/1826#issuecomment-1211720662

> We need to make sure the contract for task runners is right. Then we can implement a default choice for the batteries-included projects based on that.

---

There are several issues at the moment with projen's task implementation:

- Extra params cannot be included (E.g. `projen test -- my.spec.ts`). This is because the projen task can include many forked tasks and thereby it is not possible to infer which sub-process needs these extra parameters.
- It sets up it's own opinion and hirerachy of tasks calling (e.g. see [build task](https://github.com/projen/projen/blob/main/docs/build.md#build)). Which makes it difficult to set up other task runners and optimize the behaviour. For example, turborepo can also run task hierarchies, but in an optimized way (using cache).

There are many competing task runners emerging. Notable ones:

- [Turborepo](https://turborepo.org/docs)
- [Wireit](https://github.com/google/wireit): #1826
- [Lage](https://github.com/microsoft/lage)

---

As a user, I want to be able to swap a task runner for a custom one. Ideally, it would be another component or package that can easily plug into the projen project instance.

Should be as simple as:

```ts
new typescript.TypeScriptProject({
  // ... projen options

  taskRunner: new CustomTaskRunner({
    // optional task runner options
  }),
})
```

Where `CustomTaskRunner` implements an interface:

```ts
class CustomTaskRunner implements IProjenTaskRunner {
  // ...
}
```

---

### Question: What should this interface look like?

```ts
interface IProjenTaskRunner {
  // ❓❓❓
}
```

moltar:
> How are task dependencies modeled?

This is one of the main reasons why I believe the current implementation is lacking. There is no way to setup a chain of dependencies (aka "pipeline") as it is in Turborepo or Lage.

It looks like Lage and Turborepo kind-a settled on a "pipeline" interface.

- https://turborepo.org/docs/core-concepts/pipelines
- https://microsoft.github.io/lage/docs/Guide/pipeline

Maybe we can model the same way?

> How should we passing on extra parameters?

Don't have any good ideas here yet! Certainly a challenging problem.

moltar:
> How should we passing on extra parameters?

I think we just have to accept that it is not possible to pass args to a "pipeline" task (i.e. the root task that calls other tasks).

But at the same time, projen should not block the ability to pass args to a leaf task.

[...]

PR #3162, opened 2023-12-10
feat: turn task runner into dependencies vs explicit execution

rix0rrr:
This PR adds a dependency system to tasks; it also adds a reverse dependency system (an "implies" system). This means that tasks can now require other tasks to run before and after themselves.

The motivation for this was that when you have a `pre-compile` task, you *always* want this task to run when you run `npx projen compile`; but in the old system, the only way to run `pre-compile` was to run the `npx projen build` task instead, which does a lot more work which is not always desired.

When the same task is being requested in a single run of `projen` (for example, if `pre-compile` is requested to be run both as a dependency of `compile`, as well as implied by the `build` task) duplicate invocations will be skipped.

This also has a task implication system -- it used to be that when `compile` was requested, `post-compile` would automatically be run. Then I tried it in practice on the `projen` repo and I was annoyed at having to wait for the `jsii-docgen` tasks in `post-compile` (when all I wanted was some compiled code!) so I took the implication out again. The feature is still in but not really contributing to any UX improvements at the moment. I'm loath to take it out because of the sunken cost fallacy (😅) so if you want it gone, give me a couple of days to shore up my resolve to destroy my own work 🙈

---

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -108,10 +108,11 @@
           "exec": "jsii --silence-warnings=reserved-word"
         }
       ],
-      "dependsOn": [
+      "runFirst": [
         "pre-compile"
       ],
-      "implies": [
+      "alsoRun": [
+        "pre-compile",
         "post-compile"
       ]
     },
@@ -295,7 +296,7 @@
           "spawn": "default"
         }
       ],
-      "dependsOn": [
+      "runFirst": [
         "compile"
       ]
     },
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -107,6 +107,12 @@
         {
           "exec": "jsii --silence-warnings=reserved-word"
         }
+      ],
+      "dependsOn": [
+        "pre-compile"
+      ],
+      "implies": [
+        "post-compile"
       ]
     },
     "contributors:update": {
@@ -288,6 +294,9 @@
         {
           "spawn": "default"
         }
+      ],
+      "dependsOn": [
+        "compile"
       ]
     },
     "post-upgrade": {
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -112,8 +112,7 @@
         "pre-compile"
       ],
       "alsoRun": [
-        "pre-compile",
-        "post-compile"
+        "pre-compile"
       ]
     },
     "contributors:update": {
@@ -295,9 +294,6 @@
         {
           "spawn": "default"
         }
-      ],
-      "runFirst": [
-        "compile"
       ]
     },
     "post-upgrade": {
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -108,11 +108,11 @@
           "exec": "jsii --silence-warnings=reserved-word"
         }
       ],
-      "dependsOn": [
+      "runFirst": [
         "pre-compile"
       ],
-      "implies": [
-        "post-compile"
+      "alsoRun": [
+        "pre-compile"
       ]
     },
     "contributors:update": {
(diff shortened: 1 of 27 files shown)


mrgrain:
That's great, thank you! I wanna run this bit a couple community members. Can you turn this into something that's vaguely README driven by updating/amending the docs? I don't think we need a full RFC for this.

rix0rrr:
Yup, can do, some time tomorrow.

tinovyatkin:
### General note:
We have a very interesting and long term problem to solve with projen task being also rendered as NPM `scripts`: if you name a projen task like `task1` and then add another task like `pretask1`, and another `posttask1` then you will have in `package.json` scripts like `pretask1`, `task1` and `posttask1`. Now, executing `npm run task1` will be executing all 3 of those tasks as per NPM documentation: https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts

I'm saying it here to say that if we are doing to touch this part then it will be a breaking change and from my point of view it will be good to become in sync with NPM in that sense (we may detect whether a task is running by `npm run` by checking env). So, just naming a task "precompile" will make it run as part of `npm run compile` and it will be good to have consistency with `npx projen compile` (however, there are no such concepts for other JSII targets).

tinovyatkin:
As on SD - I believe this PR approaches just one of current limitations of Tasks (that already can be solved by an existing meaning - see my [comment above](https://github.com/projen/projen/pull/3162#issuecomment-1860544515)) but it's definitely good bias for action. Now, while adding that _patches_ on this stage where we have `projen` based on `Constructs` we may also complicate future introduction of "generic task rendered and runner" API that community is looking for so long. So as a minimum I would love to see in this PR a comparison how dependencies are set in few popular task runners on the projen target languages - NPM, [nx](https://nx.dev/concepts/task-pipeline-configuration), [wireit](https://github.com/google/wireit?tab=readme-ov-file#dependencies), as well as Pythonic [invoke](https://docs.pyinvoke.org/en/stable/concepts/invoking-tasks.html#pre-and-post-tasks) and [pdm](https://pdm-project.org/latest/usage/scripts/#pre-post-scripts) (as a limited set of examples).

For example our team longer term wish here would be a plug-in replacement for task renderer - I may want to render tasks as pure NPM `scripts` in Node or `pdm` scripts in Python -- all these for sake of performance and "compatibility". Somebody else for sure will want task to be rendered into `nx` or `wireit`. So when we will add such abstract interface it may shape the API for dependencies in some different way - one particular thing I'm seeing very useful here is cross package tasks dependencies as in `wireit` (we use monorepos for most of projects and in projen currently it means different projects) - so why limit dependencies only to another task and not have spawned shell commands there also? Another pattern I see across task runners mentioned above - `pre` and `post` hook, not just the `pre` implemented in this PR.

moltar:
@tinovyatkin 

> For example our team longer term wish here would be a plug-in replacement for task renderer

FYI the discussion for this is happening in #2051

moltar:
@tinovyatkin Just added entries from your list to the bullet list too

cameroncf:
I love this work and always dislike making comments like this, but my opinion is that tasks should be atomic for reasons I've outlined in [this comment](https://github.com/projen/projen/issues/2051#issuecomment-1862999358) over in the RFC for Task Runners. Meaning I think auto-running compile and pre-compile together is a bad idea.

I do understand the reasoning behind mimicking NPM but Projen is not a package manager and the needs are different.

One alternate strategy is to build your own task of `compile:all` that runs them all together.

rix0rrr:
Closing this as this will never get merged.

PR #3176, opened 2023-12-14
fix(javascript): added bun setup for javascript workflows

abdsahin:
[Bun](https://bun.sh/) added as a package manager but the workflows are still using NodeJS. This is to fix this.

Fixes #

---


mrgrain:
@mergifyio update

mrgrain (approved):

Issue #3203, opened 2023-12-27
bug: In some cases, GitHub Workflow steps include an empty object for "with"


dkershner6:
Example: https://github.com/projen/projen/blob/2b0b032153c1fb826fdb19b97282a1eb21affbb6/test/web/__snapshots__/nextjs-ts-project.test.ts.snap#L182

I am not sure this causes an issue, but seems easy to fix.  I will pop in a PR.

PR #3204, opened 2023-12-27
fix: WorkflowSteps enables completely consistent actions/checkout and setGitIdentity in Workflows

dkershner6:
Fixes https://github.com/projen/projen/issues/3203

Placed a bit of an opinion on WorkflowSteps with a common interface, please weigh in on how you like the pattern.  I moved an unrelated Step to the new interface in order to demonstrate.

This is non-breaking PR, but does include a deprecation, since I moved something.

---

### DIFF
--- a/src/github/task-workflow.ts
+++ b/src/github/task-workflow.ts
@@ -1,6 +1,7 @@
 import { DEFAULT_GITHUB_ACTIONS_USER } from "./constants";
 import { GitHub } from "./github";
-import { CheckoutOptions, WorkflowActions } from "./workflow-actions";
+import { WorkflowActions } from "./workflow-actions";
+import { CheckoutWith, WorkflowSteps } from "./workflow-steps";
 import { GithubWorkflow } from "./workflows";
 import {
   ContainerOptions,
@@ -65,7 +66,7 @@ export interface TaskWorkflowOptions {
    *
    * @default - not set
    */
-  readonly checkoutWith?: CheckoutOptions;
+  readonly checkoutWith?: CheckoutWith;
 
   /**
    * Steps to run before the main build step.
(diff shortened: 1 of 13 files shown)


src/github/task-workflow.ts
```diff
@@ -199,16 +199,10 @@ export class TaskWorkflow extends GithubWorkflow {
         ...preCheckoutSteps,
 
         // check out sources.
-        {
-          name: "Checkout",
-          uses: "actions/checkout@v3",
-          ...(Object.keys(checkoutWith).length > 0
```
mrgrain, line 205: Huh, so this didn't work then?

mrgrain (commented):
Looks good. I think my initial concern was that a single `JobStep` as return type is quite risky, in case it ever needs to be more than just one step. And that led to the invention of `WorkflowActions` which is a admittedly a bit confusing.

dkershner6:
> Looks good. I think my initial concern was that a single JobStep as return type is quite risky, in case it ever needs to be more than just one step. And that led to the invention of WorkflowActions which is a admittedly a bit confusing.

@mrgrain , I actually started this PR with the same opinion, and I understand the inclination entirely.  I changed my opinion abruptly the moment I tried to implement a common interface for a WorkflowStep's "other" fields (id, etc) on something that allowed for more than one step.

I now believe there should be both, and that WorkflowActions can be more opinionated, and WorkflowSteps should be quite generic, so everyone has the building blocks for consistency no matter what they do Workflow wise.

mrgrain (approved):

mrgrain:
Makes sense. I think long-term we probably want to get away from the struct interfaces and implement these as classes that can be either a single step or multiple steps.

Issue #3172, opened 2023-12-13
Support subproject releases


Sparkdaddy:
I took a look through the issues and source code and didn't find a reason this wasn't implemented other than ["By default, GitHub workflows will not be created for subprojects since config files in .github/ only work if they are in the root directory."](https://github.com/projen/projen/blob/main/docs/subproject.md).

I have a use-case where I am defining an SDK for a library as a subproject and would like that specific SDK to be released to npm. It would also be quite nice to have a release for the main project as well. If the workflows reason above is the gist for why it hasn't been implemented then I thought the subproject directory would provide most of the distinction from the other workflows used. For example, `release.yml` for the root and then `release-sdk.yml` for the SDK subproject based on the `sdk` directory. Then you could have a default behavior of using the directory in the `releaseTagPrefix` as well to distinguish the SDK release from the other ones.

Overall, I'm curious if there's a reason besides the root directory needing to have the workflows in Github that this isn't implemented. And if not, what would you think of the above behavior?

mrgrain:
Even that link of yours is just a description of the current state and not a reason.

There is no reason this hasn't been implemented yet other than the it's hasn't been implemented yet and the it's maybe a bit more complicated to generalize as a single project release.

Sparkdaddy:
Totally understand - I'll look into implementing it today and see if I'm worthy enough to make the addition

mrgrain:
I have a half-baked solution ready, but it's highly coupled to the monorepo implementation we use. Hopefully still useful.
- https://github.com/cdklabs/awscdk-service-spec/blob/main/projenrc/monorepo-release.ts
- https://github.com/cdklabs/awscdk-service-spec/blob/main/projenrc/workspace-release.ts

dkershner6:
I half-baked something totally differently than mrgrain and CDKlabs, but i'll take a stab at this since I don't see any in progress PRs on it, maybe I can make all our lives easier.

dkershner6:
Just tried out the new functionality on a 12 library publishing monorepo where I was hacking it together before.

I just deleted my hack, added a compile command for inner-monorepo dependencies to `releaseWorkflowSetupSteps`, and it deployed all 12 like butter.

PR #3208, opened 2023-12-28
feat(release): subproject releases

dkershner6:
Resolves https://github.com/projen/projen/issues/3172

This is non-breaking, but some opinions made:
- Releases for each project should be a separate workflow, named after the project in the case of a subproject
- For the base release workflow job (creation of artifact), the default working-directory should be set to the relative outdir of the subproject, with an exception made for the install command, which should be run from the topLevelParent of the subproject outdir.
- The folder that the artifact is published from, for a subproject, should be the subproject's outdir + artifactDirectory.
- Perhaps we should break TaskWorkflow, but I just decomposed it.
- Settings - Nothing has changed default-wise, and as such, in order to release subprojects, you must manually set ~~`github` to true~~ (@mrgrain comments) and `release` to true.

Snapshots have changed, but it has been verified that the default release behavior for NON-subprojects has remained identical.

I attempted to test this on a monorepo, but the package linking process combined with projen's automated nature proved too difficult, but based on the unit tests, I believe it would satisfy.

Open to more testing ideas, but wanted to ensure the opinions formed are the right ones.

I considered writing a recursive `releaseTagPrefix` conflict checker, but it seemed outside the scope, and should be another request if desired.

---


dkershner6:
In order to test this in a more realistic way, I was able to add a project to projen with the following code.

```javascript
new javascript.NodeProject({
  parent: project,
  name: "my-subproject",
  defaultReleaseBranch: "main",
  outdir: "packages/my-subproject",
  release: true,
  github: true,
  minNodeVersion: "16.0.0",
  workflowNodeVersion: "18.14.0",
  releaseTagPrefix: "my-subproject@",
  releaseToNpm: true,
});
```

This produced the following workflow:

```yaml
# ~~ Generated by projen. To modify, edit .projenrc.js and run "/bin/bash ./projen.bash".

name: release_my-subproject
on:
  push:
    branches:
      - main
  workflow_dispatch: {}
jobs:
  release:
    runs-on: ubuntu-latest
    permissions:
      contents: write
    outputs:
      latest_commit: ${{ steps.git_remote.outputs.latest_commit }}
    env:
      CI: "true"
    defaults:
      run:
        working-directory: ./packages/my-subproject
    steps:
      - name: Checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
      - name: Set git identity
        run: |-
          git config user.name "github-actions"
          git config user.email "github-actions@github.com"
      - name: Setup Node.js
        uses: actions/setup-node@v3
        with:
          node-version: 18.14.0
      - name: Install dependencies
        run: yarn install --check-files --frozen-lockfile
        working-directory: .
      - name: release
        run: npx projen release
      - name: Check for new commits
        id: git_remote
        run: echo "latest_commit=$(git ls-remote origin -h ${{ github.ref }} | cut -f1)" >> $GITHUB_OUTPUT
      - name: Backup artifact permissions
        if: ${{ steps.git_remote.outputs.latest_commit == github.sha }}
        run: cd dist && getfacl -R . > permissions-backup.acl
        continue-on-error: true
      - name: Upload artifact
        if: ${{ steps.git_remote.outputs.latest_commit == github.sha }}
        uses: actions/upload-artifact@v3
        with:
          name: build-artifact
          path: packages/my-subproject/dist
  release_github:
    name: Publish to GitHub Releases
    needs: release
    runs-on: ubuntu-latest
    permissions:
      contents: write
    if: needs.release.outputs.latest_commit == github.sha
    steps:
      - uses: actions/setup-node@v3
        with:
          node-version: 18.14.0
      - name: Download build artifacts
        uses: actions/download-artifact@v3
        with:
          name: build-artifact
          path: dist
      - name: Restore build artifact permissions
        run: cd dist && setfacl --restore=permissions-backup.acl
        continue-on-error: true
      - name: Release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_REPOSITORY: ${{ github.repository }}
          GITHUB_REF: ${{ github.ref }}
        run: errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt) --target $GITHUB_REF 2> $errout && true; exitcode=$?; if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then cat $errout; exit $exitcode; fi
  release_npm:
    name: Publish to npm
    needs: release
    runs-on: ubuntu-latest
    permissions:
      contents: read
    if: needs.release.outputs.latest_commit == github.sha
    steps:
      - uses: actions/setup-node@v3
        with:
          node-version: 18.14.0
      - name: Download build artifacts
        uses: actions/download-artifact@v3
        with:
          name: build-artifact
          path: dist
      - name: Restore build artifact permissions
        run: cd dist && setfacl --restore=permissions-backup.acl
        continue-on-error: true
      - name: Release
        env:
          NPM_DIST_TAG: latest
          NPM_REGISTRY: registry.npmjs.org
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
        run: npx -p publib@latest publib-npm
```

When I compare this against my own monorepo and manual code, it looks identical other than naming and that I use PNPM.  The monorepo publish is confirmed working.

test/release/release.test.ts
```diff
@@ -8,855 +10,987 @@ import {
...
+    expect(
+      subprojectReleaseWorkflow.jobs.release.defaults.run["working-directory"]
+    ).toEqual(`./${subprojectOutdir}`);
+    expect(
+      subprojectReleaseWorkflow.jobs.release.steps.find(
+        (step: any) => step.name === "Upload artifact"
+      ).with.path
+    ).toEqual(`${subprojectOutdir}/${artifactsDirectory}`);
+
+    expect(outdir).toMatchSnapshot();
```
mrgrain, line 966: Let's just use `outdir[".github/workflows/release_my-project.yml"]` as the snapshot that should be enough here.

test/release/release.test.ts
```diff
@@ -8,855 +10,987 @@ import {
...
+      branch: "main",
+      publishTasks: true, // to increase coverage
+      artifactsDirectory: "dist",
+      releaseTagPrefix: "my-project@", // to avoid conflicts with the root project
+    });
+
+    const outdir = synthSnapshot(project);
+    expect(outdir[".github/workflows/pull-request-lint.yml"]).toBeUndefined();
+    expect(outdir[".github/workflows/release my-project.yml"]).toBeUndefined();
+    expect(outdir).toMatchSnapshot();
```
mrgrain, line 994: Not required here.

src/project.ts
```diff
@@ -269,6 +278,7 @@ export class Project extends Construct {
     }
 
     this.outdir = outdir;
+    this.relativeOutdir = options.outdir ?? Project.DEFAULT_OUTDIR;
```
mrgrain, line 281: Is `options.outdir` always relative?

src/project.ts
```diff
@@ -373,14 +383,24 @@ export class Project extends Construct {
...
+  /**
+   * Returns the highest parent of this project, if one exists.
+   */
+  public get topLevelParent(): Project | undefined {
+    let parent = this.parent;
+    while (parent?.parent) {
+      parent = parent.parent;
+    }
+    return parent;
+  }
```
mrgrain, line 402: How is this different than `project.root`?

src/project.ts
```diff
@@ -184,6 +188,11 @@ export class Project extends Construct {
    */
   public readonly outdir: string;
 
+  /**
+   * Relative output directory of this project.
```
mrgrain, line 192: Relative to what? Root or parent?

src/release/release.ts
```diff
@@ -680,12 +700,46 @@ export class Release extends Component {
...
+
+      workflow.addJob(BUILD_JOBID, taskjob);
+
+      return workflow;
     } else {
       return undefined;
     }
   }
+
+  private _findTargetGitHubForWorkflow(): GitHub | undefined {
```
mrgrain, line 718: Trying to get away from this convention. I know it's not consistently enforced.

```suggestion
  private findTargetGitHubForWorkflow(): GitHub | undefined {
```

src/release/release.ts
```diff
@@ -680,12 +700,46 @@ export class Release extends Component {
...
+      }
+
+      if (!topLevelParentProject.github) {
+        throw new Error(
+          `Subproject ${this.project.name} cannot create a release workflow to its top-level parent ${topLevelParentProject.name} because it does not have GitHub activated. Please set "github" to true.`
+        );
+      }
+
+      return topLevelParentProject.github;
+    }
```
mrgrain, line 739: Thanks for adding validations! This feels a bit too prescriptive with the class having to be an `instanceof GitHubProject` and the `github` property. 

How about we use `GitHub.of(this.project.root)`?

--- 

Side note, there is probably a missing enforcement of only allowing a single `GitHub` component in the tree.

src/release/release.ts
```diff
@@ -680,12 +700,46 @@ export class Release extends Component {
...
+      return workflow;
     } else {
       return undefined;
     }
   }
+
+  private _findTargetGitHubForWorkflow(): GitHub | undefined {
+    if (!this.github) {
+      return;
+    }
```
mrgrain, line 721: Not sure I understand this early exit.

src/release/release.ts
dkershner6: I know you are still reviewing, but I actually implemented this as their MUST be more than one `GitHub` set to true in the tree (as in you control each piece of the tree with this option).  This also would be why you are confused one comment below on the early exit.

test/release/release.test.ts
```diff
@@ -8,855 +10,987 @@ import {
...
-});
+  test("childProject should contain no github workflows", () => {
+    // GIVEN
+    const rootProject = new NodeProject({
+      defaultReleaseBranch: "main",
+      name: "parent",
+    });
+    const project = new TestProject({
+      parent: rootProject,
+      github: true,
```
mrgrain, line 977: I wonder if we want this. It's not used in the code, is it?

dkershner6: https://github.com/projen/projen/pull/3208#discussion_r1438362070

It is, at least currently.

dkershner6: My thought on this was that it is `true`, just "hoisted".

src/github/workflow-steps.ts
```diff
@@ -49,35 +49,39 @@ export class WorkflowSteps {
...
+      ...this._adaptCommonJobStep({
+        ...options,
+        name: options.name ?? "Upload artifact",
+      }),
+      uses: "actions/upload-artifact@v3",
+      with: uploadArtifactWith,
+    };
+  }
+
+  private static _adaptCommonJobStep(options: CommonJobStep): CommonJobStep {
```
mrgrain, line 72:
```suggestion
  private static adaptCommonJobStep(options: CommonJobStep): CommonJobStep {
```

src/github/workflows-model.ts
mrgrain, line 1: This doesnt' make sense to me. What's unique what's common and why. I'm guessing you are trying to avoid some duplication here, but I think we've made this more complicated than before. 

Also these end up as public interfaces, so we it's tricky to go back on this.

src/javascript/node-project.ts
```diff
@@ -793,6 +795,16 @@ export class NodeProject extends GitHubProject {
...
 
+  private determineInstallWorkingDirectory(): string | undefined {
+    if (this.topLevelParent) {
+      if (this.topLevelParent.relativeOutdir.startsWith(".")) {
+        return this.topLevelParent.relativeOutdir;
+      }
+      return `./${this.topLevelParent.relativeOutdir}`;
+    }
+    return;
+  }
```
mrgrain, line 806: Feels like we are doing this kind of check quite often. And both `topLevelParent` and `relativeOutdir` are new additions. Are they maybe not quite suitable to what we want to use them for? Or maybe we need a helper to convert relative paths between prefixless an with `./` prefix? IDK.

Also I don't think `working-directory` technically needs the prefix in GitHub Actions.

src/javascript/node-project.ts
```diff
@@ -1192,6 +1205,12 @@ export class NodeProject extends GitHubProject {
  * Options for `renderInstallSteps()`.
  */
 export interface RenderWorkflowSetupOptions {
+  /**
+   * We may be running a Workflow with a default working-directory set, and we must set this in order to install from the root.
```
mrgrain, line 1209: Should describe what this does, not why it exists.

src/github/task-workflow.ts
mrgrain, line 1: Not your fault, but the diff makes this very confusing to review. Could you please describe what the refactor is here? I think it's just a refactor...

src/javascript/node-project.ts
```diff
@@ -612,16 +612,18 @@ export class NodeProject extends GitHubProject {
...
-        artifactsDirectory: this.artifactsDirectory,
         ...options,
 
+        artifactsDirectory: this.artifactsDirectory,
         releaseWorkflowSetupSteps: [
-          ...this.renderWorkflowSetup({ mutable: false }),
+          ...this.renderWorkflowSetup({
+            installWorkingDirectory: this.determineInstallWorkingDirectory(),
+            mutable: false,
+          }),
```
mrgrain, line 626: I think there is an implicit question about the workingDir for other steps in renderWorkflowSetup

src/github/task-workflow.ts
dkershner6: TaskWorkflow has been minimally changed, but otherwise just extracted to two static methods.

The only addition is the ability to control `JobDefaults` (which we use to control default `working-directory`)

mrgrain (changes requested):
🚀 This is awesome, thank you! 

> * Releases for each project should be a separate workflow, named after the project in the case of a subproject
> * For the base release workflow job (creation of artifact), the default working-directory should be set to the relative outdir of the subproject, with an exception made for the install command, which should be run from the topLevelParent of the subproject outdir.

Each of these are reasonable by themselves. But together I'm struggling a bit to make sense of them. Are we treating the projects as separate things or not?
For Node Monorepo projects, it makes sense to run this at the top level only because that's how most monorepos work anway. But then why have it in separate workflows?


> * Perhaps we should break TaskWorkflow, but I just decomposed it.

As little change as possible please, that just makes the PR bigger and more complicated. I'd rather have a follow-up `chore` to clean things up.

> * Settings - Nothing has changed default-wise, and as such, in order to release subprojects, you must manually set `github` to true and `release` to true.

This seems like the biggest point of contention to discuss. Can you elaborate on your thinking?

> Open to more testing ideas, but wanted to ensure the opinions formed are the right ones.

I'll test this as well. Not sure either what we can do beside manual testing of the workflows.

> I considered writing a recursive `releaseTagPrefix` conflict checker, but it seemed outside the scope, and should be another request if desired.

Agreed, out of scope.

dkershner6:
Thanks for the thoughtful review, I know it was a beefy one.

> Each of these are reasonable by themselves. But together I'm struggling a bit to make sense of them. Are we treating the projects as separate things or not?

The projects are treated as fully separate, any generation of workflows (release, anyway) are just "hoisted" to the `topLevelParent` project.  My wording was a little confusing, but the code should clearly dileneate this as the case.

> For Node Monorepo projects, it makes sense to run this at the top level only because that's how most monorepos work anway. But then why have it in separate workflows?

I think by "top level only" you may mean "as part of a single release workflow with the root project and subproject all as one workflow".  In general, I agree, people just rig up a nx or turborepo and just `[monorepohandler] run deploy`.  Most people also don't generally do artifact creation per project, then hand that to several publishing jobs, at least in my experience, and doing so would generally create at least some issues since artifact naming is per workflow.  For projen, the complexity level would be a significant uptick for this same reason, but not insurmountable, just some changes to `Publisher`.

src/release/release.ts
mrgrain: Yes, there is a wider discussion to be had. But this comment and the one about early exit are more basic. The validations as implemented are too prescriptive and will move us further away from being able to get rid of `GitHubProject`, which is a goal.

src/release/release.ts
mrgrain: I _think_ you tried to short circuit the helper, because `workflowTargetGitHub` is not used if `!this.github == true`. But that check is also done further up and that's confusing. I'd expect `findTargetGitHubForWorkflow()` to pretty much always find a workflow unless it is set to false everywhere. Might just need splitting up in a validation helper and a helper to get the object.

dkershner6:
> This seems like the biggest point of contention to discuss. Can you elaborate on your thinking?

Agreed.  I believe we can ignore the `release` part and focus on `github`. My thinking was three things:

1. `github` is _technically_ true, as github-y things are being created and happening, they are just "hoisted" to the top level.
2. If you only have a single project in projen, and you want to release that project using GitHub workflows, `github` must be set to `true`.
3. When it comes to creating things that run automatically, opt-in is better than opt-out (and `release` didnt feel like enough of an opt-in to create workflows).

Open to other ideas, of course, there were a lot of possible paths here.

test/release/release.test.ts
mrgrain: When I say it's not used, I mean that the actual component on the subproject is not used. I probably have missed something, but as far as I can tell we only use `subproject.github` (which is a component, not a boolean) as a boolean. But the actual workflows are created against `workflowTargetGitHub`.

dkershner6: I see.  I believe that the `subproject.github` is used in the formation of the job that performs the release.  Things like `runsOn` and such.  The TARGET of the workflow is indeed the root, but essentially we are only using the root for a filepath.

src/release/release.ts
dkershner6: Your second comment makes sense, thanks for circling back.

src/github/workflow-steps.ts
dkershner6: I try to use design pattern names when I am using one, but my first typed-in name was indeed `build`.

Since we are here chatting:
> Adapter is a structural design pattern, which allows incompatible objects to collaborate.

(the forced camelCase of JSII and forced kebab-case of GitHub, in this case)

> Builder is a creational design pattern that lets you construct complex objects step by step. The pattern allows you to produce different types and representations of an object using the same construction code.

Sure, that could work as well, for a simple single step, of course.

I'll switch this to build regardless as I want to stick to convention of this project, but since you asked! ;)

src/github/workflows-model.ts
dkershner6: Open to naming ideas, but the difference is that:

Common: Every single JobStep in existence can take this fields.
Unique: JobStep's generally only use one or two of these, never all, and we, as projen, build these fields using other interfaces, adapting the input to these.

mrgrain:
Thank you, that helps my understanding much better. 

In my mind, the other alternative is:

- Only allow a single GitHub Component in the tree. By default this Component would be attached to the root project, but it could feasibly be created manually. Obviously this is a breaking change, although I'm not sure if there's currently a meaningful way to use more than one of them.
- All features that the GitHub Component provides only exist at the root anyway. 
- This is opt-out at the root project level like before, releases can be opt-in at subproject level. Saying that, we default to releasing root projects, so I'm questioning why we would make a different default for subprojects.


> 1. `github` is _technically_ true, as github-y things are being created and happening, they are just "hoisted" to the top level.
> 2. If you only have a single project in projen, and you want to release that project using GitHub workflows, `github` must be set to `true`.

I think this is the core of my issue. Even currently it feels wrong to be able to set the `github` boolean option to `true` in the root project and `false` in subprojects. Like as if this makes a difference if the repo is hosted on GitHub or if GitHub Workflows are created or not. In this interpretation, the option would be better called something like `useGitHubFeatures`. However the actual currently implementation isn't strictly that either. It is pretty much `shouldCreateGitHubComponent`. And then features are added from the outside. Take the `UpgradeDependencies` component as an example. That creates a Github Workflow by default (if it can), but also has an option to disable that.

(There is this whole fluke that we are still using a `GitHubProject` but you can then disable `github`. But that's fine, we should pretend `GitHubProject` doesn't exist. That's the goal anyway.)

So what I'm struggling with here is if we can separate the root and subprojects enough to have multiple GitHub components. As an example, I can set the `projenCredentials` option on my subproject. Which credentials are we now going to use? If we enable `pullRequestLint` on the subproject, what does that mean?

I'm not really against having multiple GitHub Components, but we will have to sieve through the consequences of that (at least to some extend) and also we should than put it front-and-center that subprojects use their own `github` and not the root one.

---

> 3. When it comes to creating things that run automatically, opt-in is better than opt-out (and `release` didnt feel like enough of an opt-in to create workflows).

This feels like a discussion about sensible defaults that should not have any bearing on the rest.

mrgrain:
> I think by "top level only" you may mean "as part of a single release workflow with the root project and subproject all as one workflow". In general, I agree, people just rig up a nx or turborepo and just `[monorepohandler] run deploy`. Most people also don't generally do artifact creation per project, then hand that to several publishing jobs, at least in my experience, and doing so would generally create at least some issues since artifact naming is per workflow. For projen, the complexity level would be a significant uptick for this same reason, but not insurmountable, just some changes to `Publisher`.

Yes this is fine, we should not confuse this PR with a full Monorepo implemention. But the question remains what the basic algorithm looks like here. Is it something like this?

```
 (0) // starting at the project root
 (1) yarn install --check-files --frozen-lockfile
 (2) cd subproject/
 (3) projen release
 (4) |- projen bump
 (5) |- projen build
 (6)     |- projen default
 (7)     |- projen compile
 (8)     |- projen test
 (9)     |- projen package
(10) |- projen unbump
(11) ./upload.sh dist
```

This seems fine. (1) is required because we need to this to run projen itself
But it think we are assuming that (5) will just run without further work. Is that a correct assumption to make?

dkershner6:
> But it think we are assuming that (5) will just run without further work. Is that a correct assumption to make?

You are definitely more familiar with projen than I...but I think we make the same assumption for normal projects, but provide the ability to add in `pre` and `post` build steps?  That should be true here as well...

Also, your algorithm is close, but instead of `cd`, I instead set default `working-directory` to the subproject and the `working-directory` for the `install` command to root.  This seemed more in the vein of what GitHub intended, but I think the practical difference is near zero...but I dont know that for sure, there could be some kind of nested action logic that reverts to the default `working-directory`.

test/release/release.test.ts
dkershner6: I am incorrect directly above, `github` is in fact a very simple object, and none of the info is used besides the root project.  The options I refer to are on the project itself, not in `github` (which I am not sure makes sense conceptually, but that is the same GitHubProject discussion)

dkershner6:
> I'm not really against having multiple GitHub Components, but we will have to sieve through the consequences of that (at least to some extend) and also we should than put it front-and-center that subprojects use their own github and not the root one.

When I started this out, I assumed all GitHub related info would reside on `github`.  For whatever reason, I kept that constraint in my head.  Having looked at the actual object and how tiny it is, I think I should remove this constraint, and rely fully on `release` in order to determine if a subproject should generate a release workflow (which I think is what you are inclined to as well, please confirm).

Just as a comment for the sake of commenting, I think both of us knew I would run face-first into GitHubProject and likely somehow make the refactor situation worse, but I did do my best to avoid it where I could, "could" just wasn't a lot of places given current implementation.

src/project.ts
dkershner6: Somehow I missed that one, will use.

src/javascript/node-project.ts
dkershner6: I can't find an example without the `./`...maybe keep it around just so I can sleep at night?

I will create the helper though.

src/project.ts
dkershner6: Moving away from this entirely, as requested.

src/github/workflows-model.ts
dkershner6: Alright, I took a stab at these, and renamed my Unique to Settings, which I think helps, but I need some help from you on this one:
What is the difference between a `Step` and a `JobStep`?  It _feels_ like perhaps someone accidentally conflated the two, and now we have two where we could have one.  `JobStep` only contains optional fields outside of what Step contains.  I only see Step used in JSII.

In any case, if we can combine all of these into one, I think my Common/Full additions will make way more sense.  If not, I just need help differentiating these names.

Leaving this unresolved.

src/github/task-workflow.ts
dkershner6: This is still true on my latest commits.

src/javascript/node-project.ts
dkershner6: Leaving unresolved intentionally.

dkershner6:
Alright, @mrgrain , this is ready for another pass when you have (several) moments.

src/github/workflows-model.ts
dkershner6: Alright, I gave it some long thought and came up with the word `Configuration` to describe the JobSteps Who, Where, Why, and How (but not what).

src/github/github-project.ts
```diff
@@ -205,6 +205,7 @@ export class GitHubProject extends Project {
           projenCredentials: options.projenCredentials,
           mergify: options.mergify,
           mergifyOptions: options.mergifyOptions,
+          workflows: this.parent ? false : true, // Don't create workflows for subprojects, in the subproject
```
mrgrain, line 208: This seems unrelated to the PR? I guess it makes sense as a setting, but the again we have the issue that really non the stuff the `GitHub` component does make sense in a subproject. I guess the default can be changed by `githubOptions`, but I'd feel better if we don't touch the wider issue of `GitHub` component in subproject unless we absolutely have to.

dkershner6: This was more important when github was required to be true, but good catch, can now be removed.

src/util/path.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Helper function to format a path as dot notation regardless of how it
+ * was handed in.
+ *
+ * @param path - can be formatted as "path", "./path", or "/path"
+ * @returns "./path"
+ */
+export function formatPathAsDotNotation(path: string): string {
```
mrgrain, line 8: If you have a better name, I'd be up for that. Dot notation seems overloaded. I though it meant `path.to.file` first and was confused. Maybe `ensureRelativePathStartsWithDot` or something.

src/release/release.ts
```diff
@@ -678,14 +707,42 @@ export class Release extends Component {
...
       return undefined;
     }
   }
+
+  private findTargetGitHubForWorkflow(): GitHub | undefined {
+    if (this.project.parent) {
+      const rootGitHub = GitHub.of(this.project.root);
+      if (!rootGitHub) {
+        throw new Error(
+          `Subproject ${this.project.name} cannot create a release workflow to its top-level parent ${this.project.root.name} because it is not a GitHub project or does not have github set to true.`
```
mrgrain, line 737:
```suggestion
          `Subproject ${this.project.name} cannot create a release workflow to its top-level parent ${this.project.root.name} because it does not have GitHub support enabled. Please enable GitHub support for ${this.project.root.name}.`
```

dkershner6:
Ready to roll, @mrgrain .  Thank you, this back and forth was full of learnings.

src/project.ts
```diff
@@ -131,11 +131,15 @@ export interface GitOptions {
...
 }
-
 /**
  * Base project
  */
 export class Project extends Construct {
+  /**
+   * The default output directory for a project if none is specified.
+   */
+  public static readonly DEFAULT_OUTDIR = ".";
```
mrgrain, line 141: Same as on the other PR, let's not have these in the public interface.

src/github/workflows-model.ts
```diff
@@ -399,24 +413,12 @@ export interface Step {
...
-  /**
-   * Specifies a working directory for a step.
-   * Overrides a job's working directory.
-   */
-  readonly workingDirectory?: string;
 }
 
 /**
- * A job step
+ * These settings differentiate a JobStep from a generic Step.
```
mrgrain, line 419: Although this begs the question why is there difference between a JobStep and a generic Step and what is it, like in prose. 😅

src/github/workflows-model.ts
mrgrain, line 1: @dkershner6 Can you give me prose summary why we need to introduce multiple levels between `Step -> JobStep`? Looks like we now have `Step -> StepConfiguration` and `JobStep -> (JobStepConfiguration -> StepConfiguration),  (Step -> StepConfiguration)`.

dkershner6: I have no insight into the existence of `Step` + `JobStep`, they seem to be the same thing to me and could be combined (this was a decision made before my PR).

As to `Configuration` vs the other fields (`Command`):

Configuration: Fields that describe the How, Why, When, and Who of a (Job)Step.  These fields can have none present, but can be present on every single (Job)Step in existence, and have no effect on one another.  This was created for accuracy, we could _technically_ accept a whole (Job)Step into options when we are accepting parameters to create a (Job)Step(s), but it is more accurate to say we can only always accept these.

Command: These fields describe the What, and are generally mutually exclusive and will error if used together (`run` and `uses`, for instance).  This means if we were creating a (Job)Step that uses `uses`, and someone provided `run`, it would error (in github, once the workflow ran).

Hopefully this helps, @mrgrain , if you can help me with the difference between `Step` and `JobStep`, perhaps we can get this in order.

src/util/path.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param path - can be formatted as "path", "./path", or "/path"
+ * @returns "./path"
+ */
+export function ensureRelativePathStartsWithDot(path: string): string {
+  if (path.startsWith(".")) {
+    return path;
+  }
+
+  if (path.startsWith("/")) {
```
mrgrain, line 13: This is odd. Normally paths starting with `/` are absolute paths?!

src/github/task-workflow.ts
```diff
@@ -130,41 +111,60 @@ export interface TaskWorkflowOptions {
...
+ */
+export class TaskWorkflow extends GithubWorkflow {
+  /**
+   * Allows for more flexible construction of a Workflow that is similar, but not identical, to the TaskWorkflow.
+   *
+   * @param taskStep The main JobStep to be executed.
+   * @param options TaskWorkflowJobOptions
+   * @returns The job that would be created as part of the TaskWorkflow
+   */
+  public static buildJob(
```
mrgrain, line 160: So this is a `TaskJob` then basically? Can you make it a class instead of a static method that returns a struct?

mrgrain:
> Ready to roll, @mrgrain . Thank you, this back and forth was full of learnings.

Thanks @dkershner6 I'm going through it now. I think the result is what we want and the API for it is there as well. Most of my requests will now be about refactors that might or might not have been needed. But because they are shaping the public API, we'll have to go through it. In future, I'd suggest we split out refactorings and new features into multiple PRs.

src/javascript/node-project.ts
```diff
@@ -1189,9 +1202,18 @@ export class NodeProject extends GitHubProject {
...
 export interface RenderWorkflowSetupOptions {
+  /**
+   * Overrides for the install step in the workflow setup.
+   *
+   * @default - No overrides of the install step
+   *
+   * @example - { workingDirectory: "rootproject-dir" } for subprojects installing from root.
+   * @example - { env: { NPM_TOKEN: "token" }} for installing from private npm registry.
+   */
+  readonly installJobStepOverrides?: JobStep;
```
mrgrain, line 1216: Can we make this the configuration? As in don't allow overwriting `run`?

```suggestion
  /**
   * Configure the install step in the workflow setup.
   *
   * @default - `{ name: "Install dependencies" }`
   *
   * @example - { workingDirectory: "rootproject-dir" } for subprojects installing from root.
   * @example - { env: { NPM_TOKEN: "token" }} for installing from private npm registry.
   */
  readonly installStepConfiguration?: JobStepConfiguration;
```

The `@default` might not work and just require text instead

src/github/task-workflow.ts
dkershner6: I can implement this, but it is effectively just an adapter renaming all of the `TaskWorkflow` fields back to `Job` fields, when the `TaskWorkflow` fields are just renamed `Job` fields in the first place.  It is the reason I was asking if breaking was a good idea (to just make the fields the same names).

Regardless, my half measure was not the correct one, as you identified.

src/github/task-workflow.ts
```diff
@@ -207,14 +207,57 @@ export class TaskWorkflow extends GithubWorkflow {
...
     };
+  }
+
+  /**
+   * Prepares a workflow to be used as a TaskWorkflow.
+   *
+   * @param workflow The workflow to prepare.
+   * @param triggers The triggers to add to the workflow.
+   */
+  public static prepareWorkflow(
```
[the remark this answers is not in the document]
dkershner6: Was an attempt to deduplicate logic, but it isn't a lot, so I will use duplication instead.

dkershner6:
> Thanks @dkershner6 I'm going through it now. I think the result is what we want and the API for it is there as well. Most of my requests will now be about refactors that might or might not have been needed. But because they are shaping the public API, we'll have to go through it. In future, I'd suggest we split out refactorings and new features into multiple PRs.

My message came across as an attempt to push, it was actually just meant to say "I am done with your most recent review".  Apologies.

src/github/workflows-model.ts
dkershner6: Resolving in favor of detailed prose in conversation one below.

src/javascript/node-project.ts
dkershner6: Worked as is.

dkershner6:
Ready for another pass, @mrgrain .

src/github/workflows-model.ts
dkershner6: It's odd, I actually had some of this in a JSDoc, but must have lost them in a rename.  Re-adding now.

src/github/task-workflow.ts
```diff
@@ -161,10 +173,45 @@ export class TaskWorkflow extends GithubWorkflow {
...
+  readonly email: string;
+}
+
+/**
+ * The primary or initial job of a TaskWorkflow.
+ */
+export class TaskWorkflowJob {
+  private readonly job: Job;
+
+  constructor(taskStep: JobStep, options: TaskWorkflowJobOptions) {
```
mrgrain, line 210: I think the interface here would be the task itself. Both usage you introduce are just taking in the same input.

src/github/task-workflow.ts
```diff
@@ -161,10 +173,45 @@ export class TaskWorkflow extends GithubWorkflow {
...
+  /**
+   * The email address of the git user.
+   */
+  readonly email: string;
+}
+
+/**
+ * The primary or initial job of a TaskWorkflow.
+ */
+export class TaskWorkflowJob {
```
mrgrain, line 207: So I think according to our new nomenclature this is now a WorkflowAction. Correct?

Also, let's put this into a new file please.

src/github/task-workflow.ts
dkershner6: The reason why I didn't do this is to allow for a flexible `GitHub.Project` to run `runTaskCommand` from.  The project attached to the `TaskWorkflow` or this job isn't necessarily the one we want to use (we use the subproject, but place the workflow on the rootProject, for instance).

If I am missing something surrounding this, please be more specific, it is entirely possible I am.

src/github/task-workflow.ts
dkershner6: New file heard.

`WorkflowAction` seems to be `group of JobSteps`.  That makes sense to me.

`Job` includes a field that contains not just a group of JobSteps, but all the steps for a given job, plus it has additional configuration associated with the Job itself (outside the steps).

I think it is named appropriately, as it is a `Job`, and more specifically, one that matches the (partial) interface of `TaskWorkflow`.

src/github/workflows-model.ts
```diff
@@ -357,9 +357,12 @@ export interface RunSettings {
...
 
 /**
- * A generic step
+ * Fields that describe the How, Why, When, and Who of a Step.
+ * These fields can have none present, but can be present on every Step, and have no effect on one another.
+ *
+ * This stands in contrast to the Command (non-Configuration) fields, which are mutually exclusive, and describe the What.
  */
-export interface Step {
+export interface StepConfiguration {
```
mrgrain, line 365: Okay, so this is down to PR https://github.com/projen/projen/pull/1787 and the `Step` interface being used for steps in the [action.yaml Metadata file for GitHub Actions ](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions).

Now I don't necessarily agree with this, or the naming, but that's what we have today. So we can keep it the way you have it. But would you mind updating the comment on `Step` and `StepConfiguration` to explain their purpose. Thanks!

dkershner6: Oh...I see, that makes sense why it doesn't make sense here...it isn't for something in this library.

Yeah...I will do my best to explain.

src/github/task-workflow-job.ts
```diff
@@ -0,0 +1,184 @@
...
+   * Default settings for all steps in the TaskWorkflow Job.
+   */
+  readonly jobDefaults?: JobDefaults;
+}
+
+/**
+ * The primary or initial job of a TaskWorkflow.
+ */
+
+export class TaskWorkflowJob {
```
mrgrain, line 120: this should implement `Job` right?

src/release/release.ts
```diff
@@ -678,14 +707,42 @@ export class Release extends Component {
...
+        jobDefaults: this.project.parent
+          ? // If this is a subproject, we need to set the working directory to the outdir of the subproject
+            {
+              run: {
+                workingDirectory: ensureRelativePathStartsWithDot(
+                  path.relative(this.project.root.outdir, this.project.outdir)
+                ),
+              },
+            }
+          : undefined,
```
mrgrain, line 720: I think this always is true, the working directory is always `path.relative(this.project.root.outdir, this.project.outdir)` because in the case of root project, the result is `''` empty string. We can check for that to just omit the whole thing, but don't need to check the parent anymore.

src/release/release.ts
```diff
@@ -628,42 +635,64 @@ export class Release extends Component {
...
+    const workflowTargetGitHub = this.findTargetGitHubForWorkflow();
+
+    if (workflowTargetGitHub && !this.releaseTrigger.isManual) {
+      // Use target (possible parent) GitHub to create the workflow
+      const workflow = new GithubWorkflow(
+        workflowTargetGitHub,
+        this.project.parent
+          ? `${workflowName}_${fileSafeProjectName}`
+          : workflowName
+      );
```
mrgrain, line 672: Let's move this condition into the place where we are creating the release workflow.

src/github/task-workflow.ts
dkershner6: Resolving based on below comment.

src/release/release.ts
```diff
@@ -628,42 +635,64 @@ export class Release extends Component {
...
-        uses: "actions/upload-artifact@v3",
         with: {
           name: BUILD_ARTIFACT_NAME,
-          path: this.artifactsDirectory,
+          path: this.project.parent
+            ? `${path.relative(
+                this.project.root.outdir,
+                this.project.outdir
+              )}/${this.artifactsDirectory}`
+            : this.artifactsDirectory,
```
mrgrain, line 658: We can just always do this, because in the case of root project, the result of `path.relative(this.project.root.outdir, this.project.outdir) ` is `''` the empty string.

src/release/release.ts
dkershner6: I believe `outdir` on the OUTPUT of a Project class is an absolute path.  We are striving to get to relative here, and I removed the relativeOutDir property that I added earlier in the PR.

Basically, what I am trying to accomplish here is:
If subproject
subproject relative directory
else
default is good (because we are the root)

I think this code works, but if you have a cleaner implementation, I would need more help to find it.

src/release/release.ts
```diff
@@ -678,14 +707,42 @@ export class Release extends Component {
...
     } else {
       return undefined;
     }
   }
+
+  private findTargetGitHubForWorkflow(): GitHub | undefined {
+    // this is not a subproject
+    if (!this.project.parent) {
+      return this.github;
+    }
```
mrgrain, line 736:
```suggestion
```

We don't actually need this. If we are not a subproject, we are a root project. And the code below will return the correct project.

Further more, we can lift this up to where we currently say `this.github = project.github;` => `this.github = this. findTargetGitHubForWorkflow()`. And then we can actually change the constructor from the current `project: Project` => `scope: IConstruct`.

dkershner6: I see, there is no third level (subproject of subproject), I didn't know that limit was in there, I can change accordingly.

mrgrain (changes requested):
Okay, so my last round of reviews on the `Release` component has the goal to rid it of needing to know about subprojects and root projects. That's not something it should concern itself with. I left the inline notes that show you how to get rid of all the few remaining references.

Almost there, thanks for bearing with me!

src/github/task-workflow-job.ts
dkershner6: I get a JSII error if I literally `implements`, but I have done the legwork and it matches.

dkershner6:
> Almost there, thanks for bearing with me!

The product we are ending with is much better than the one I started with, no question.

dkershner6:
> Okay, so my last round of reviews on the Release component has the goal to rid it of needing to know about subprojects and root projects.

@mrgrain , inherent in some of your suggestions is that there can never be a subproject of a subproject.  I didn't go in with this constraint, and just want to be sure it, in fact, exists.

src/release/release.ts
mrgrain: We can actually have third levels, but what we are saying is that we always just use the `GitHub` component of the root project.

The code currently does this:

- `if not this.project.parent` => only the root project does not have a parent
  - `then use this.github` => that is the `GitHub` component of the root project
- `otherwise` => any other project is a subproject, no matter how many levels
  -  `then use GitHub.of(this.project.root) or throw` which is also the `GitHub` component of the root project or we fail

mrgrain:
> @mrgrain , inherent in some of your suggestions is that there can never be a subproject of a subproject. I didn't go in with this constraint, and just want to be sure it, in fact, exists.

No, this is in fact not a constraint that exists. But it is also not inherent to my assumptions. At least I don't think so. 😅

dkershner6:
> > @mrgrain , inherent in some of your suggestions is that there can never be a subproject of a subproject. I didn't go in with this constraint, and just want to be sure it, in fact, exists.
> 
> No, this is in fact not a constraint that exists. But it is also not inherent to my assumptions. At least I don't think so. 😅

I think I still am reading some of your words incorrectly, I will see if anything comes up as I implement.

mrgrain:
> > Thanks @dkershner6 I'm going through it now. I think the result is what we want and the API for it is there as well. Most of my requests will now be about refactors that might or might not have been needed. But because they are shaping the public API, we'll have to go through it. In future, I'd suggest we split out refactorings and new features into multiple PRs.
> 
> My message came across as an attempt to push, it was actually just meant to say "I am done with your most recent review". Apologies.

Oh no, didn't read it that way actually. My message was really just telling you that you will get quicker results if you keep changes to the API surface tight. As much I as appreciate good code and refactoring, it inevitably makes everything a bit slower.

mrgrain:
> > > @mrgrain , inherent in some of your suggestions is that there can never be a subproject of a subproject. I didn't go in with this constraint, and just want to be sure it, in fact, exists.
> > 
> > 
> > No, this is in fact not a constraint that exists. But it is also not inherent to my assumptions. At least I don't think so. 😅
> 
> I think I still am reading some of your words incorrectly, I will see if anything comes up as I implement.

Might be my words. I'm not a native English speaker.

src/release/release.ts
mrgrain: By that I mean `NodeProject` should check if it is a subproject, and in that case pass a different default workflow name.

dkershner6:
> > My message came across as an attempt to push, it was actually just meant to say "I am done with your most recent review". Apologies.
> 
> Oh no, didn't read it that way at all. My message was really just letting you know that you will get quicker results if you keep changes to the API surface tight. As much I as appreciate good code and refactoring, it inevitably makes everything a bit slower.

Asking me to turn my head is a big ask!

I hear you, but I also think this was very educational for me in my assessment of whether this lib can serve our needs and to what certain things are and how they should be used, so the time was not wasted (at least for me), and now we have slightly cleaner code!

src/github/task-workflow-job.ts
mrgrain: What's the jsii error? If you push I can have a look at it.

dkershner6:
> Might be my words. I'm not a native English speaker.

It isn't, I would've never known unless you said something.

src/github/task-workflow.ts
mrgrain: That's a good point actually. `GitHub.Project` isn't the right project here either. 🤔
I was thinking task has a reference to it's project, but that's not the case. This needs some thinking.

src/release/release.ts
dkershner6: I don't like the feel of `/dist` in the field (normally dist or something/dist)...

BUT, we can have one variable power all of these, I think, and that should help.

src/release/release.ts
dkershner6: We cannot do this because a single `Release` may create several workflows (because of more than one branch being released), and so we are already editing the name inside Release (the nodeProject option it fills could even be called "prefix").

The `workflowName` right here is:
```
const workflowName = branch.workflowName ?? `release-${branchName}`;
```

src/release/release.ts
dkershner6: The exact change throws errors in EVERY root project that doesn't have `github` enabled...which I am trying to determine if it is what we want...

dkershner6: OK, now I remember, this whole error was just in case someone had enabled github on the subproject, but not root.

This is no longer a thing, we just need to know if github is on the root.

dkershner6:
OK, @mrgrain , I handled or replied to all comments.

Note that I had to handle one differently than either of us expected, but it is resolved: https://github.com/projen/projen/pull/3208#discussion_r1443392255

Essentially, we were erroring for reasons that no longer make sense.  Now we just do nothing, as per normal, single project, behavior.

src/github/task-workflow.ts
mrgrain: Alright I figured that out. `TaskWorkflowJob` needs to be a component, which gives us the right project. And that can also also be any project - so it's flexible.

It took the liberty to push the code.

src/release/release.ts
mrgrain: Okay this make sense. We are still loosing the ability to set the workflow name though. I can refactor this to be optional.

src/github/task-workflow-job.ts
```diff
@@ -141,7 +144,8 @@ export class TaskWorkflowJob {
   public readonly if?: string | undefined;
   public readonly strategy?: JobStrategy | undefined;
 
-  constructor(taskStep: JobStep, options: TaskWorkflowJobOptions) {
+  constructor(scope: IConstruct, task: Task, options: TaskWorkflowJobOptions) {
```
dkershner6, line 147: You may consider adding something regarding:

> `scope` should be part of the project the `Task` belongs to.

This wouldn't have been obvious to me as a new user of `Component` with this interface.

dkershner6 (commented):
👍

src/github/task-workflow-job.ts
mrgrain: Done. Thanks!

mrgrain:
@dkershner6 Out of interest, have you read this yet? https://projen.io/constructs.html
Either way, does that help you understanding what scope is or is it missing something?

mrgrain:
@dkershner6 Pushed a couple more changes. I am happy with this now. Are you okay for me to merge?

src/github/task-workflow-job.ts
mrgrain: Yeah, it's actually quite a recent thing that we have fully adopted the constructs model. But it's awesome for exactly this reason!

dkershner6:
> @dkershner6 Out of interest, have you read this yet? https://projen.io/constructs.html Either way, does that help you understanding what scope is or is it missing something?

I have, and I have used em quite a bit outside this lib, and am generally familiar from CDK, but that almost would make me complacent on the `scope` parameter, figuring it to be mostly about context and not that I specifically needed to target the subproject.

I mean, the chances of something going wrong are pretty small regardless, but it is worth telling folks, imo.

dkershner6:
> @dkershner6 Pushed a couple more changes. I am happy with this now. Are you okay for me to merge?

I am.  Thanks for this, was a great experience and I love this feature!

mrgrain:
> > @dkershner6 Out of interest, have you read this yet? https://projen.io/constructs.html Either way, does that help you understanding what scope is or is it missing something?
> 
> I have, and I have used em quite a bit outside this lib, and am generally familiar from CDK, but that almost would make me complacent on the `scope` parameter, figuring it to be mostly about context and not that I specifically needed to target the subproject.
> 
> I mean, the chances of something going wrong are pretty small regardless, but it is worth telling folks, imo.

Oh yeah, I was just curious. And more docs are always nice. In projen the biggest win is direct access to the project. Which I guess is kind of like context.

mrgrain:
> > @dkershner6 Pushed a couple more changes. I am happy with this now. Are you okay for me to merge?
> 
> I am. Thanks for this, was a great experience and I love this feature!

:shipit:

mrgrain (approved):

dkershner6:
> > > @dkershner6 Out of interest, have you read this yet? https://projen.io/constructs.html Either way, does that help you understanding what scope is or is it missing something?
> > 
> > 
> > I have, and I have used em quite a bit outside this lib, and am generally familiar from CDK, but that almost would make me complacent on the `scope` parameter, figuring it to be mostly about context and not that I specifically needed to target the subproject.
> > I mean, the chances of something going wrong are pretty small regardless, but it is worth telling folks, imo.
> 
> Oh yeah, I was just curious. And more docs are always nice. In projen the biggest win is direct access to the project. Which I guess is kind of like context.

Yes, very similar in benefit and use case of `Cdk.Stack.of`.  I will freely admit that it takes a bit of getting used to, but that might be something about convention, like a Java dev would immediately grasp it.  Either way, after getting used to, very useful.

PR #3209, opened 2023-12-28
feat(task): Add support for passing down cwd to spawn steps

icj217:
The existing documentation around the `cwd` task/step property doesn't explicitly state that it only works for `exec` steps, but that's currently the case. There are use cases where having `spawn` tasks inherit the cwd of the task that spawned them is quite useful. For example:
```ts
const sdkBuildTask = project.addTask('sdk:build', { cwd: 'my_folder' });
sdkBuildTask.spawn(project.poetry.poetryTask, { args: ['version', '$VERSION' });
sdkBuildTask.spawn(project.poetry.poetryTask, { args: ['build'] });
```
---


test/tasks/runtime.test.ts
```diff
@@ -340,6 +340,25 @@ describe("cwd", () => {
...
+    task.spawn(echo, { args: ["step1=$PWD"] });
+    task.spawn(echo, { args: ["step1=$PWD"], cwd: stepcwd });
+
+    const noCwdSpecified = executeTask(p, "echo", {}, ["$PWD"]);
+    expect(noCwdSpecified[0]).toContain(p.outdir);
+
+    const lines = executeTask(p, "testme");
+    expect(lines[0].includes("mypwd")).toBeTruthy();
+    expect(lines[1].includes("yourpwd")).toBeTruthy();
+  });
```
mrgrain, line 361: I've tried hard to understand what this test is doing. Could you please refactor this to be more readable. I'd recommend sticking to the same names for things e.g. not `taskcwd` and `mypwd` and make the symbol names more explicit.

test/tasks/runtime.test.ts
```diff
@@ -340,6 +340,25 @@ describe("cwd", () => {
...
+    const stepcwd = join(p.outdir, "yourpwd");
+    mkdirSync(taskcwd, { recursive: true });
+    mkdirSync(stepcwd, { recursive: true });
+    const echo = p.addTask("echo", { exec: "echo", receiveArgs: true });
+    const task = p.addTask("testme", { cwd: taskcwd });
+    task.spawn(echo, { args: ["step1=$PWD"] });
+    task.spawn(echo, { args: ["step1=$PWD"], cwd: stepcwd });
+
+    const noCwdSpecified = executeTask(p, "echo", {}, ["$PWD"]);
+    expect(noCwdSpecified[0]).toContain(p.outdir);
```
mrgrain, line 356: This seems redundant since very task on this project will use the project's outdir.

src/task-runtime.ts
```diff
@@ -153,7 +162,8 @@ class RunTask {
         this.runtime.runTask(
           step.spawn,
           [...this.parents, this.task.name],
-          argsList
+          argsList,
+          { cwd: step.cwd ?? task.cwd }
```
mrgrain, line 166: This looks like a breaking change to me. We are now giving the `step.cwd` preference over the `task.cwd` and there is no way to revert the original behavior?!

mrgrain (changes requested):
Hi, this seems reasonable but I am not sure I follow the proposed change and the implications it has for existing code.

mrgrain:
There will be bigger changes and some deprecations as part of https://github.com/projen/projen/pull/3162
Closing this one here in favor of those. Lesson learned is certainly we need to be more explicit about how cwd is inherited.

Issue #3217, opened 2024-01-01
bug(typescript): "compile" task does not properly update itself to use a renamed `tsconfig` filename (non-dev)


dkershner6:
`tsc --build ${project.tsconfig?.fileName}` should do the trick.

At the moment, only `tsconfig.json` will work.

PR #3222, opened 2024-01-02
fix(typescript): tsconfig filename added to compile and watch commands

dkershner6:
Fixes https://github.com/projen/projen/issues/3217

Snapshots updated as per expected.
Fixed an unrelated JSDoc that was throwing a warning.

---


mrgrain (changes requested):
Can we avoid passing the file name if it's the default?

dkershner6:
> Can we avoid passing the file name if it's the default?

Changes have been made, @mrgrain .

mrgrain (approved):

Issue #3215, opened 2024-01-01
bug: Prettier files and .projenrc.ts are not ignored by Package (npmignore)


dkershner6:
They get packaged along with src files.  .projenrc.js gets blocked, but .ts does not.

PR #3223, opened 2024-01-02
fix(node): Prettier, Projenrc, and .gitattributes are not ignored in npm package

dkershner6:
Fixes https://github.com/projen/projen/issues/3215

.gitattributes was not part of the ticket, but I noticed it while I was working.  As part of that implementation, I noticed there was a race condition on the `addPackageIgnore` method, where if it was called before the `npmignore` `IgnoreFile` was initialized, the additions were lost. Added a comment to call out that if `IgnoreFile` isn't guaranteed to exists, the method should be called in `pre-synth`.

---


src/gitattributes.ts
```diff
@@ -13,6 +14,9 @@ export class GitAttributesFile extends FileBase {
     super(scope, ".gitattributes", {
       editGitignore: false,
     });
+
+    const project = Project.of(this);
+    project.addPackageIgnore("/.gitattributes");
```
mrgrain, line 19:
```suggestion
    this.project.addPackageIgnore("/.gitattributes");
```

src/typescript/projenrc-ts.ts
```diff
@@ -38,6 +38,8 @@ export interface ProjenrcTsOptions {
  * Requires that `npx` is available.
  */
 export class ProjenrcTs extends ProjenrcFile {
+  public static readonly DEFAULT_FILENAME = ".projenrc.ts";
```
mrgrain, line 41: I don't want this to be a public property on the class. Tbh, the components are purposely separate, so I'm fine with the duplication here.

Let's make it a local constant per component.

src/typescript/typescript.ts
```diff
@@ -506,6 +507,7 @@ export class TypeScriptProject extends NodeProject {
             fileName: tsconfigDevFile,
             include: [
               PROJEN_RC,
+              RealProjenrcTs.DEFAULT_FILENAME,
```
mrgrain, line 510: This seems inaccurate? Seems like we want the projenrc instance and get `filePath` from that.

test/javascript/jest.test.ts
```diff
@@ -124,6 +124,7 @@ test("Typescript Project Jest Defaults Configured", () => {
   );
   expect(jestTypescriptConfig.include).toEqual([
     PROJEN_RC,
+    ProjenrcTs.DEFAULT_FILENAME,
```
mrgrain, line 127: Same as the other one.

mrgrain (changes requested):
Right, regarding the race condition (great find by the way!): Let's turn `project.npmignore` into a getter that can initialized itself. Than we are guaranteed to have an ignore file.

src/typescript/typescript.ts
dkershner6: This is a little tougher than meets the eye, since creating the ProjenrcFile requies the tsconfig to already be built (chicken/egg).  I tried to capture the spirit of what you wanted, though.

dkershner6:
> Right, regarding the race condition (great find by the way!): Let's turn `project.npmignore` into a getter that can initialized itself. Than we are guaranteed to have an ignore file.

I just tried this, and it doesn't appear possible because we require storing ignores before `this` is available in NodeProject (during the run of `super`).  If I am missing something, let me know.

mrgrain:
> Another thing that came up: There appear to be two near-identical classes: Projenrc and ProjenrcTs. In some cases one is even renamed to the other. I used the more explicit, and didn't break anything, including inner library renaming, but it made my naming odd in some cases to do so. If you want me to be more opinionated, let me know.

These are two different components. `Projenrc` is for `TypeScriptProjects` only. `ProjenrcTs` can be used with any project type. Yes they are mostly the same, but the former can make assumptions that the latter can't and thus do more things. You'll probably need to update both.

src/typescript/typescript.ts
```diff
@@ -510,7 +514,10 @@ export class TypeScriptProject extends NodeProject {
           {
             fileName: tsconfigDevFile,
             include: [
-              PROJEN_RC,
+              projenrcTsShouldBeCreated
+                ? options.projenrcTsOptions?.filename ??
+                  DEFAULT_PROJENRC_TS_FILENAME
+                : PROJEN_RC,
```
mrgrain, line 520: This already gets added in `Projenrc`. Nice spotting the incorrect addition of `PROJEN_RC`. This can be added conditional as well.

src/gitattributes.ts
```diff
@@ -13,6 +13,8 @@ export class GitAttributesFile extends FileBase {
     super(scope, ".gitattributes", {
       editGitignore: false,
     });
+
+    this.project.addPackageIgnore("/.gitattributes");
```
mrgrain, line 17: Actually never mind the race condition. You just have to add this in the pre-synth phase:
```suggestion
}
  public preSynthesize(): void {
    this.project.addPackageIgnore("/.gitattributes");
```

mrgrain:
> I just tried this, and it doesn't appear possible because we require storing ignores before `this` is available in NodeProject (during the run of `super`). If I am missing something, let me know.

No you are right. I've missed that as well. I remembered the proper way of doing this though and left a comment.

dkershner6:
Ready for another pass, @mrgrain .

src/javascript/node-project.ts
```diff
@@ -1103,8 +1103,19 @@ export class NodeProject extends GitHubProject {
...
+   *
+   * @param pattern The pattern to ignore.
+   *
+   * @remarks
+   * If you are having trouble getting an ignore to populate, try using your construct or component's preSynthesize method to properly delay calling this method.
+   */
+  public override addPackageIgnore(pattern: string): void {
+    if (this.npmignore) {
+      return this.npmignore?.addPatterns(pattern);
+    }
```
mrgrain, line 1117:
```suggestion
    this.npmignore?.addPatterns(pattern);
```

The if statement is exactly what `?` does here.

src/javascript/node-project.ts
```diff
@@ -1103,8 +1103,19 @@ export class NodeProject extends GitHubProject {
...
+    // packageIgnore added prior to initialization or npmignore disabled.
```
mrgrain, line 1118:
```suggestion
```

src/typescript/projenrc-ts.ts
```diff
@@ -48,9 +50,12 @@ export class ProjenrcTs extends ProjenrcFile {
   constructor(project: Project, options: ProjenrcTsOptions = {}) {
     super(project);
 
-    this.filePath = options.filename ?? ".projenrc.ts";
+    this.filePath = options.filename ?? DEFAULT_FILENAME;
     this._projenCodeDir = options.projenCodeDir ?? "projenrc";
 
+    project.addPackageIgnore(`/${this.filePath}`);
+    project.gitignore.include(`/${this.filePath}`);
```
mrgrain, line 57: Wait why do you want this gitignored?
```suggestion
```

test/javascript/node-project.test.ts
```diff
@@ -1713,3 +1713,27 @@ describe("package manager env", () => {
     });
   });
 });
+
+describe("npmignore", () => {
+  test("npmignore should have NodeProject's default ignore patterns", () => {
```
mrgrain, line 1718:
```suggestion
  test("should have NodeProject's default ignore patterns", () => {
```

test/javascript/node-project.test.ts
```diff
@@ -1713,3 +1713,27 @@ describe("package manager env", () => {
...
+    // GIVEN
+    const project = new TestNodeProject();
+
+    // WHEN
+    const output = synthSnapshot(project);
+
+    // THEN
+    expect(output[".npmignore"]).toMatchSnapshot();
```
mrgrain, line 1726:
```suggestion
    expect(output[".npmignore"]).toMatchSnapshot();
    expect(output[".npmignore"]).toContain("/.gitattributes");
```

src/typescript/typescript.ts
```diff
@@ -574,6 +571,8 @@ export class TypeScriptProject extends NodeProject {
 
     if (!this.parent && options.projenrcTs) {
       new ProjenrcTs(this, options.projenrcTsOptions);
+    } else {
+      this.tsconfigDev.addInclude(PROJEN_RC); // projenrc.js created in NodeProject needs to be added in tsconfigDev
```
mrgrain, line 575: Not required for this PR, but if you are down for it.

```suggestion
    // when this is a root
    if (!this.parent) {
      if (options.projenrcTs) {
        new ProjenrcTs(this, options.projenrcTsOptions);
      } else {
        // projenrc.js created in NodeProject needs to be added in tsconfigDev
        const projenrcJs = Projenrc.of(this);
        if (projenrcJs) {
          this.tsconfigDev.addInclude(projenrcJs.filePath);
        }
      }
```

mrgrain (changes requested):
Almost there. just a few minor changes now (one totally optional).

The big one that's confusing me is adding `.prjojenrc.ts` to gitignore. But I'm guessing that was a mistake?

src/typescript/projenrc-ts.ts
dkershner6: I think Copilot got me!

src/typescript/typescript.ts
dkershner6: I'll track down the rest.

src/typescript/projenrc-ts.ts
dkershner6: WAIT, we both got fooled, this is INCLUDE for gitignore, as in `!`, as in ALWAYS DON'T ignore.

Here is the JS equivalent: https://github.com/projen/projen/blob/92e8b5ae1048d7185fcae5619abb4dae90041e4a/src/javascript/node-project.ts#L553

dkershner6:
> Almost there. just a few minor changes now (one totally optional).
> 
> The big one that's confusing me is adding `.prjojenrc.ts` to gitignore. But I'm guessing that was a mistake?

@mrgrain , reposting here to ensure you see it:

This is INCLUDE for gitignore, as in `!`, as in ALWAYS DON'T ignore.

Here is the JS equivalent: https://github.com/projen/projen/blob/92e8b5ae1048d7185fcae5619abb4dae90041e4a/src/javascript/node-project.ts#L553

mrgrain:
> This is INCLUDE for gitignore, as in `!`, as in ALWAYS DON'T ignore.

Omg, that's embarrassing. I was wondering why the snapshots didn't change. 🙈  
Thanks for catching my stupidity.

dkershner6:
> > This is INCLUDE for gitignore, as in `!`, as in ALWAYS DON'T ignore.
> 
> Omg, that's embarrassing. I was wondering why the snapshots didn't change. 🙈 Thanks for catching my stupidity.

I think this qualifies as an off by one error (number of negatives), so you are off the hook.

> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

dkershner6:
Alright, @mrgrain , ready for your perusal, including the optional "fix all the projens".  You can note the corrected behavior in all the language snapshots.

`PROJEN_RC` is not exported from the root of the lib, and therefore isnt a public interface, so I gave it a bit more descriptive name.

mrgrain (approved):
🥳

Issue #3213, opened 2023-12-30
feat(node): Implement other common testing practice patterns in non-jsii projects


dkershner6:
More of an RFC, but this should be entirely non-invasive to existing projects, I believe.

I propose the implementation of development directories matching other default Jest test matches, namely the `__tests__` folder and the `src/**/*.(spec|test).[jt]sx?` patterns.  Specifically, adding these patterns to the ESLint devDependencies allow patterns and excluding these patterns from the production tsconfig.

This would enable devs to develop their projen projects (that aren't JSII) with best practice testing strategies (placing the tests next to the  associated source file).

dkershner6:
Going to re-word this one.

PR #3225, opened 2024-01-02
feat(tsconfig): removeInclude/Exclude, deduplication, standard Jest testing pattern handling

dkershner6:
Fixes https://github.com/projen/projen/issues/3213

PS - This should be the last PR for a bit, in case you were worried.

---

### DIFF
--- a/src/cdk/jsii-project.ts
+++ b/src/cdk/jsii-project.ts
@@ -433,6 +433,10 @@ export class JsiiProject extends TypeScriptProject {
         );
       }
     }
+
+    // Discourage use of anything besides an explicit test directory for JSII, since they would be potentially packaged.
+    this.tsconfigDev?.removeInclude(`${this.srcdir}/**/*.spec.*`);
+    this.tsconfigDev?.removeInclude(`${this.srcdir}/**/*.test.*`);
   }
 
   /**
--- a/src/javascript/jest.ts
+++ b/src/javascript/jest.ts
@@ -655,7 +655,7 @@ export class JestReporter {
  */
 export class Jest extends Component {
   /**
-   * Returns the singletone Jest component of a project or undefined if there is none.
+   * Returns the singleton Jest component of a project or undefined if there is none.
    */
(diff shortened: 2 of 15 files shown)


mrgrain (changes requested):
The additional helpers are obviously fine, maybe these can be a separate PR?

The additional test setups are a bit controversial, because we already have `testdir` to specify where tests should go and I don't really see why we should allow multiple test locations at the same time.

mrgrain:
> PS - This should be the last PR for a bit, in case you were worried.

I'm worried that this is the last one 😢 Your PRs are most welcome!

dkershner6:
> > PS - This should be the last PR for a bit, in case you were worried.
> 
> I'm worried that this is the last one 😢 Your PRs are most welcome!

Just for a lil bit! I'll be around.

dkershner6:
> The additional helpers are obviously fine, maybe these can be a separate PR?
> 
> The additional test setups are a bit controversial, because we already have `testdir` to specify where tests should go and I don't really see why we should allow multiple test locations at the same time.

We actually already do.

https://github.com/projen/projen/blob/8ca52e8c443d05a0b54a9e851c06a6207a127ac4/package.json#L134

Basically, I am updated tsconfigs to match this, so that if someone DOES put tests there, it won't build them, publish with the package, and such.

This is all pretty easy to control from outside projen source code, so if projen itself does not want to allow more common testing patterns outside of JSII, this doesn't block me at all, but most JS/TS development places tests inside the `src` directory, in my experience, so supporting it makes sense.  I also don't think this should break anything for existing projects, so long as we have implemented correctly.

Your call on this one, if you want me to pare down the PR to only the helpers, happy to do so.

mrgrain:
Ahahaha. _cry sob_

I'll need some time to look at this then.
Would you mind splitting the PR up anyway please? We can get the new methods out at least.

mrgrain:
Closing. Need separate PRs for additional methods and change in defaults.

Issue #3239, opened 2024-01-05
fix(next): Outdated TSConfig Settings


dkershner6:
Some settings have changed since the Next construct was built.

https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/app/ts/tsconfig.json

The minimum Node.js version has been bumped from 12.22.0 to 16.14.0, since 12.x and 14.x have reached end-of-life.

https://nextjs.org/docs/pages/building-your-application/upgrading/version-13

PR #3241, opened 2024-01-05
fix(nextjs)!: updated `NextJsProject` default options

dkershner6:
Fixes https://github.com/projen/projen/issues/3239

I question eslint and jest (in TS) being off, as Next includes both, but I didn't touch those settings as they are easy to override.  Perhaps this is because the eslint rules are not built.

BREAKING CHANGE: Minimum Node version is now 16.14.0 to match Next 13's minimum node version.  

BREAKING CHANGE: Next.js TSConfig `module` is now `esnext` by default, to preserve previous behavior, set `module` to `CommonJS`.

BREAKING CHANGE: Next.js TSConfig `moduleResolution` is now `BUNDLER` by default, to preserve previous behavior, set `moduleResolution` to `NODE`.

BREAKING CHANGE: Next.js TSConfig `strict` is now set to `true` by default, to preserve previous behavior, set `strict` to `false`.

---


mrgrain:
FYI this is probably a breaking change.

mrgrain:
> I question eslint and jest (in TS) being off, as Next includes both, but I didn't touch those settings as they are easy to override. Perhaps this is because the eslint rules are not built.

Probably. I am not familiar with this project type at all.

dkershner6:
> Probably. I am not familiar with this project type at all.

Jest should work out of the box (aside from tests being prescribed to be in `src` / `pages`, a la `tsconfig` issues).
I'll create a ticket for the eslint rules, and handle as I am converting some `next` projects.

mrgrain (changes requested):
Happy to accept that as is. Can you just add a section into the PR description about why this is a breaking change and how to revert to the previous behavior is needed.

Example: https://github.com/projen/projen/pull/2997

dkershner6:
> Happy to accept that as is. Can you just add a section into the PR description about why this is a breaking change and how to revert to the previous behavior is needed.
> 
> Example: #2997

This is complete, @mrgrain .

mrgrain:
> This is complete, @mrgrain .

Perfect execution! I'll get that merged after the subproject release.

mrgrain (approved):

PR #3244, opened 2024-01-05
chore: new docusaurus-based documentation site

mbonig:
This branch introduces the new /docusaurus directory which contains a new set of docs. These docs were written mostly from scratch to cover what we felt was the best starting point in using projen. However, these are recognized by the contributors as not being complete. We could probably spend another couple of months trying to fill them out better but felt we have hit a good point in the basic framework and content that we can merge this to upstream and let the community continue contributions.

A preview of this is available from our github pages link:

https://defiance-digital.github.io/projen/

Additionally, the images used are either manually created by us and we're not claiming any copyrights on them, or they are AI generated.

---


mrgrain:
Making this a `chore` only because it should not trigger a new release.

mrgrain:
image

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,49 @@
...
+
+name: Publish Docs
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "fix/docs" ]
```
mrgrain, line 11:
```suggestion
```

.projenrc.js
```diff
@@ -194,6 +191,13 @@ new javascript.UpgradeDependencies(project, {
   },
 });
 
+project.tasks
+  .tryFind("docgen")
+  .reset(
+    "jsii-docgen .jsii -o docusaurus/docs/api/projen --split-by-submodule"
+  );
+project.gitignore.addPatterns("docusaurus/docs/api/");
```
mrgrain, line 199: I want to keep this in code. Looking at updated to the API docs is the first thing I do on a review 

```suggestion
```

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Generate JSII docs
+        run: |
+          npm install
+          npm run build
+
+      - name: Build docs
+        working-directory: docusaurus
+        run: |
+          npm install 
+          npm run build
```
mrgrain, line 32:
```suggestion
      - name: Generate JSII docs
        run: |
          yarn install --check-files --frozen-lockfile
          /bin/bash ./projen.bash build

      - name: Build docs
        working-directory: docusaurus
        run: |
          yarn install --check-files --frozen-lockfile
          yarn build
```

.projenrc.js
```diff
@@ -194,6 +191,13 @@ new javascript.UpgradeDependencies(project, {
   },
 });
 
+project.tasks
+  .tryFind("docgen")
+  .reset(
+    "jsii-docgen .jsii -o docusaurus/docs/api/projen --split-by-submodule"
```
mrgrain, line 197: I would also like to keep `docs` at the top-level

```suggestion
    "jsii-docgen .jsii -o docs/api/projen --split-by-submodule"
```

I think these couple of changes ought to do it:
```
diff --git a/docusaurus/docusaurus.config.js b/docusaurus/docusaurus.config.js
index f22c6406..9072a22d 100644
--- a/docusaurus/docusaurus.config.js
+++ b/docusaurus/docusaurus.config.js
@@ -52,7 +52,7 @@ const config = {
           sidebarPath: require.resolve('./sidebars.js'),
           editUrl:
             'https://github.com/projen/projen/tree/main/docusaurus/',
-          path: 'docs',
+          path: '../docs',
           async sidebarItemsGenerator({defaultSidebarItemsGenerator, ...args}) {
             const sidebarItems = await defaultSidebarItemsGenerator(args);
             // @ts-ignore
diff --git a/docusaurus/sidebars/api-sidebar.js b/docusaurus/sidebars/api-sidebar.js
index 6a71c074..f69456f0 100644
--- a/docusaurus/sidebars/api-sidebar.js
+++ b/docusaurus/sidebars/api-sidebar.js
@@ -26,7 +26,7 @@ function getSidebarItems(allFilenames) {
 
 
 module.exports = () => {
-  const apiDirs = fs.readdirSync(path.join(__dirname, '..', 'docs', 'api'));
+  const apiDirs = fs.readdirSync(path.join(__dirname, '..', '..', 'docs', 'api'));
   const sidebarItems = getSidebarItems(apiDirs);
   return Object.values(sidebarItems).sort(sortGroupsInFrontOfTLTopics);
 };
```

mrgrain (commented):
`docs-original` can just be removed. We have access to the git history.

docusaurus/static/img/languages/dotnet.png
mrgrain, line 1: Technically projen isn't published for .NET... yet 😂

docusaurus/static/img/android-chrome-512x512.png
mrgrain, line 1: These look of rather poor quality. Not important though.

.projenrc.js
mrgrain, line 199: This one is rather important ;)

```suggestion
project.npmignore.exclude("/docusaurus/");
```

.projenrc.js
mbonig: one problem with this is that everything with docusaurus has to be in the 'docs' directory. So the api docs need to be in 'docs', and with the change you have then we're seeing things in `/docs/docs/api` which I'm fine with, although it's a little weird.

docusaurus/static/img/android-chrome-512x512.png
mbonig: doesn't seem to be used. will delete it. Do you know where the original source came from so we could get one generated?

docusaurus/static/img/languages/dotnet.png
mbonig: removing.

mbonig:
@mrgrain comments addressed. Please re-review.

mrgrain (approved):

mrgrain:
@mergifyio update

mrgrain (approved):

Issue #3248, opened 2024-01-06
bug(typescript): projenrc.ts is not properly packageIgnored (still)


dkershner6:
Despite my best efforts, using the latest projen version, this is still not working, and `projenrc.js` is still getting ignored instead.

PR #3250, opened 2024-01-06
fix(typescript): projenrc.ts for TS projects is not excluded from npm package

dkershner6:
Fixes https://github.com/projen/projen/issues/3248

---


mrgrain (approved):

dkershner6:
@mrgrain IS excluded (pr name)

mrgrain:
> @mrgrain IS excluded (pr name)

So for a bug fix, the text should describe the issue. Which in this case is that the file was NOT excluded. And the type `fix` says that we fixed it.

That way the changelog reads like "We fixed the following bugs: [list]"

Issue #2763, opened 2023-06-27
Release Workflow attempts to release `v0.0.0` if version was not bumped


mrgrain:
On a scheduled release trigger, if no commits have been made since the last release the `bump` task will be skipped due to the `! git log --oneline -1 | grep -q "chore(release):` condition`.

This doesn't cause the Release Workflow to stop, but instead the package is build with version `v0.0.0` and than erroneously than attempted to be released with this version. This might fail if `v0.0.0` already exists, but e.g. if the package did not start out as a projen package it will actually release the wrong version.

PR #3252, opened 2024-01-07
fix(release): release workflow attempts to publish packages for already existing versions

### LINKED ISSUE #2921: Release mechanism tries to publish existing packages
Projen is relying on remote package managers to reject publishing request for an existing package. We should validate this even before we start the publishing.

Recently, this mechanism caused an outage for Maven since it was opening staging repositories and abandoning them when the package existed. This lead to creation of multiple abandoned staging repositories which impacted Maven's cleanup process.

dkershner6:
The workflow now checks if the to-be-published version already exists as a tag on the repository.
If the tag is found, we don't attempt to publish.

This previously worked, because all supported registries treat versions as immutable and refuse to publish over an existing version. With this change, no unnecessary publish attempts are made.

Resolves #2921 
Possible fix for https://github.com/projen/projen/issues/2763

---


mrgrain (changes requested):
Nice. Only request I have is that we don't take these two micro dependencies on, for what are relatively straight forward shell commands.

Here is an example of how to do this in a single command. If you really care you can probably break this up  into two "actions". (Note that `fetch-depth: 0` is important here)

```
jobs:
  release:
    outputs:
      should-publish: ${{ steps.check-should-publish.outputs.publish }}
    steps:
      - name: Checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
      - id: check-should-publish
        run: (git ls-remote -q --exit-code --tags origin $(cat dist/releasetag.txt) && (echo "publish=false" >> $GITHUB_OUTPUT)) || echo "publish=true" >> $GITHUB_OUTPUT
```
Source: https://github.com/cdklabs/awscdk-service-spec/blob/main/.github/workflows/release.yml#L40C1-L41C170

dkershner6:
@mrgrain , refactored into a single step and used the git CLI. Ready for another pass.

mrgrain:
This looks great! It's also very high impact. So I'll take the fix and will apply it on some repos under my control and will thoroughly test it before merging this.

src/release/release.ts
```diff
@@ -294,6 +297,12 @@ export interface ReleaseOptions extends ReleaseProjectOptions {
 export class Release extends Component {
   public static readonly ANTI_TAMPER_CMD =
     "git diff --ignore-space-at-eol --exit-code";
+
+  /**
+   * Conditional (Github Workflow Job `if`) to check if a release job should be run.
+   */
+  public static readonly DEPENDENT_JOB_CONDITIONAL = `needs.${BUILD_JOBID}.outputs.${TAG_EXISTS_OUTPUT} != 'true' && needs.${BUILD_JOBID}.outputs.${LATEST_COMMIT_OUTPUT} == github.sha`;
```
mrgrain, line 304: Actually one thing, like always I don't want these in the public interface. :)

mrgrain (changes requested):

src/release/release.ts
dkershner6: I will (respectfully) still disagree, but with an example this time.

Let's say I am building a job that publishes an AWS CDK App (something not currently in projen, so I must build it outside the lib).

`Steps` are not that big of a deal, even `needs` would be nice to have something to reference, but not the end of the world.

But the `if` on my job is tough.  I can copy and paste from other jobs, but that isn't forward-compatible, and none of `DEPENDENT_JOB_CONDITIONAL`, `BUILD_JOBID` or `TAG_EXISTS_OUTPUT` or `LATEST_COMMIT_OUTPUT` are public, so I am left with no forward-compatible option, left to manually update this if projen ever does.

I will still make the change, but worth mentioning why this was here in the first place.

dkershner6:
> This looks great! It's also very high impact. So I'll take the fix and will apply it on some repos under my control and will thoroughly test it before merging this.

@mrgrain , makes sense to me, you should be good to test this now.

src/github/workflow-steps.ts
```diff
@@ -69,6 +103,22 @@ export class WorkflowSteps {
...
 
+  /**
+   * From a JobStep "run" command, builds a command that sets an output value.
+   *
+   * @param key The output key
+   * @param value The output value. This may also be a valid bash command.
+   * @returns The CLI Command to set provided output key and value
+   *
+   * @see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter
+   */
```
mrgrain, line 114: Can you add an example here - or explain - that shell commands should use `$(echo 'test')` the `$()`

mrgrain (changes requested):
found one more issue

src/github/workflow-steps.ts
```diff
@@ -69,6 +103,22 @@ export class WorkflowSteps {
...
+  public static buildSetOutputCommand(key: string, value: string): string {
+    return `echo 'Setting Output Key "${key}" to "${value}"' && echo "${key}=${value}" >> $GITHUB_OUTPUT`;
```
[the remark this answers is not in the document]
dkershner6: The suggestion didn't change any code, but I think I probably know where you are thinking...

dkershner6:
> found one more issue

Let me know if I inferred correctly, but should be good for your eyes now, @mrgrain

mrgrain:
Something doesn't work, it can't find the disttag file 🤔 
Also as good as the intention of having a nice echo for the output assignment was, it makes the code a right mess to read. That will have to change.

Nothing for you to do yet, I'm just reporting back.

mrgrain:
> Something doesn't work, it can't find the disttag file 🤔 Also as good as the intention of having a nice echo for the output assignment was, it makes the code a right mess to read. That will have to change.
> 
> Nothing for you to do yet, I'm just reporting back.

- In the CI system the whole repo is copied into another level of `dist`, so the `releasetag.txt` file ends up in `dist/dist/releasetag.txt`
- Added a check to make sure the file was read 
- Use `cat $GITHUB_OUTPUT` to show output

https://github.com/cdklabs/cdklabs-projen-project-types/pull/356/files#diff-37d43b0a9d1d9f2ecf8bc9c4d98cb6f472717d55101fd48a1dea28cde214a612R34-R38

mrgrain:
Pushed the fixes, let me know what you think.

I've reverted the addition `buildSetOutputCommand` because I'm not convinced it's the right abstraction. `tagExists` is to complex to be able to use it anyway.

mrgrain:
> I am surprised about the double nested dist...I have never experienced this occurring, nor have I seen a complete repo (or subproject) in dist. I also don't see the CI doing anything differently than locally... I am wondering if that is something specific to the repo you selected to test on. You also mention `publish task` in your comment, but you use the variable in the `release` task, might just be a conflation, but I noticed the disconnect.

This also surprised me, and technically knew about this before. It's not specific to the repo I was testing on unfortunately, it is a silly quirk in the current implementation of the release system that causes some commands to behave differently when run locally or in CI. This line is the culprit:
https://github.com/projen/projen/blob/2f1a364416c68a93f59bd173c45ff84bd270f27c/.projen/tasks.json#L227

You can reproduce this locally by running `CI=true projen build`. I misspoke in my comment. It's not `publish` but `package`.

> (or subproject)

This is an interesting point you raise. We need to check it for subprojects as well. Though the task will be the same, so I expect my change to still work.

Now, I am sure you are going to ask yourself

> This is silly, should we refactor the code so it is not doing this?

And my answer would be: Yes, let's definitely do this. But on a separate PR.

dkershner6:
> This is an interesting point you raise. We need to check it for subprojects as well. Though the task will be the same, so I expect my change to still work.

Since I have used the `working-directory`, and the command uses `run`, it should in fact spawn out of the nested folder and do the same nesting drill.  AKA I agree.

> This is silly, should we refactor the code so it is not doing this?

You HAVE been paying attention! ;)

> It's not publish but package.

Let's be sure we are accurate here, since it is an oddity, but I am fine with waiting on the full fix, and am good with all of the changes now.

mrgrain:
Found an issue with my new implementation. When the `cat` command fails this now fails the whole job. Not sure yet why the file doesn't exists. Might as well be related to the double `dist`. Looking into it.

In the previous code I think this would have defaulted to "tag does not exists" and continue to attempt publishing which might be what we want either.

```
Run TAG=$(cat dist/dist/releasetag.txt)
  TAG=$(cat dist/dist/releasetag.txt)
  ([ ! -z "$TAG" ] && git ls-remote -q --exit-code --tags origin $TAG && (echo "exists=true" >> $GITHUB_OUTPUT)) || echo "exists=false" >> $GITHUB_OUTPUT
  cat $GITHUB_OUTPUT
  shell: /usr/bin/bash -e {0}
  env:
    CI: true
cat: dist/dist/releasetag.txt: No such file or directory
Error: Process completed with exit code 1.
```
See: https://github.com/cdklabs/json2jsii/actions/runs/7496097383

mrgrain:
Yeah, my bad. This extra copy step only happens in `JsiiProject`s. 🤔 
https://github.com/projen/projen/blob/c072f02a9bbc7c51c778567c45412c33c6f7dd23/src/cdk/jsii-project.ts#L262

dkershner6:
> Yeah, my bad. This extra copy step only happens in `JsiiProject`s. 🤔
> 
> https://github.com/projen/projen/blob/c072f02a9bbc7c51c778567c45412c33c6f7dd23/src/cdk/jsii-project.ts#L262

This makes sense why I have never experienced it.

mrgrain:
I thought a bunch about this, but couldn't find a good solution. So I'm proposing to use a hack until we have fixed the shoddy `JsiiProject` release stuff. I'm not happy with this, but this fix is more important to me.

What do you think? @dkershner6

dkershner6:
> I thought a bunch about this, but couldn't find a good solution. So I'm proposing to use a hack until we have fixed the shoddy `JsiiProject` release stuff. I'm not happy with this, but this fix is more important to me.
> 
> What do you think? @dkershner6

This would save a ton of time for us in big monorepos, and we don't use JsiiProject, so I am OK with whatever your decision is @mrgrain .

mrgrain (approved):

PR #3256, opened 2024-01-09
fix(release): changelog update fails when creating a new releasing when a pre-release with the same version is present

### LINKED ISSUE #3235: bug: changelog update fails when creating a new releasing when a pre-release with the same version is present
## Overview

This is what I'm seeing when trying to create a new release using `npm run release`:

```
👾 release » publish:git | '...node/18.17.1/bin/node' '...node_modules/projen/lib/release/update-changelog.task.js'
👾 Changelog already contains an entry for 0.0.1. Skipping changelog update.
👾 release » publish:git | ...
```

In my case, I have other releases that look like this:

```
v0.0.1-beta.0
v0.0.1-beta.1
...
```

## Potential Cause

I believe this is because of the following test:

https://github.com/projen/projen/blob/314fd0c78e994f1cd0cffbdc6f3db4e2291c4e27/src/release/update-changelog.ts#L74

That checks for the existence of the version (`v0.0.1` in this case) in the existing changelog file. Because `v0.0.1` exists in the tag `v0.0.1-beta.1`, I think that this incorrectly assumes that an entry already exists for 0.0.1 in the changelog and skips that step.

Any thoughts on this?

suhussai:
Fixes #3235 

---


mrgrain (approved):

mrgrain (approved):
Congrats to your first contributions 🎉

mrgrain (approved):

PR #3201, opened 2023-12-26
feat(typescript): ts-jest now uses a non-deprecated implementation

dkershner6:
Resolves https://github.com/projen/projen/issues/2361

Continuing the good work of https://github.com/projen/projen/pull/2636 to get it over the finish line.

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3276, opened 2024-01-16
fix(jest): build fails when running with an old (<29) version of jest that is not explicitly specified as `jestVersion`

rix0rrr:
In #3201, we started generating a new config to deal with an upcoming deprecation of some configuration setting in `ts-jest`.

However, this new configuration now breaks when running with an *old* version of `ts-jest`, showing the following:

```
    Unable to process '***', please make sure that `outDir` in your tsconfig is neither `''` or `'.'`. You can also configure Jest config option `transformIgnorePatterns` to inform `ts-jest` to transform ***
```

The component would deal with versions correctly if a specific version of `jest` was requested via its component properties, but if no `jest` version was requested it would assume it had the most recent version (note that we're using the `jest` version a proxy for the `ts-jest` version).

Add a feature where we check the actual version of `ts-jest` that is installed (by looking at `package.json`), rather than hoping the user tells us what version they're installing. This will make the new feature robust in the face of old dependencies.

---


src/javascript/node-package.ts
```diff
@@ -880,6 +896,68 @@ export class NodePackage extends Component {
...
+   * when `checkRange ⊆ requestedRange`.
+   * false positive in case
+   *
+   * May return `undefined` if the question cannot be answered (for example, if the dependency
+   * is requested via local file dependencies).
+   *
+   * @param dependencyName The name of the dependency
+   * @param checkRange A particular version, or range of versions.
+   */
+  public hasDependencyVersion(
```
mrgrain, line 929: Let's find a way to not make this part of the public interface. This has too many caveats.

mrgrain (changes requested):

package.json
```diff
@@ -93,6 +93,7 @@
     "glob": "^8",
     "ini": "^2.0.0",
     "semver": "^7.5.4",
+    "semver-intersect": "^1.5.0",
```
mrgrain, line 96: Do we really need this?

mrgrain (changes requested):

src/javascript/util.ts
```diff
@@ -217,3 +218,115 @@ export function tryResolveDependencyVersion(
...
+ * the dependency is requested via local file dependencies).
+ *
+ * This API may eventually be added to the public projen API, but only after
+ * we implement exact version checking.
+ *
+ * @param dependencyName The name of the dependency
+ * @param checkRange A particular version, or range of versions.
+ */
+export function hasDependencyVersion(
+  project: Project,
```
mrgrain, line 255:
```suggestion
  nodePackage: NodePackage,
```

src/javascript/util.ts
```diff
@@ -217,3 +218,115 @@ export function tryResolveDependencyVersion(
...
+  dependencyName: string,
+  checkRange: string
+): boolean | undefined {
+  const file = join(project.outdir, "package.json");
```
mrgrain, line 259:
```suggestion
  const file = nodePackage.file
```

src/javascript/util.ts
mrgrain: Or type as NodeProject and then later do `project.package.file`

mrgrain (changes requested):
Thanks for the changes. Now just a minor thing to avoid the hardcoded file path.

src/javascript/util.ts
```diff
@@ -217,3 +219,123 @@ export function tryResolveDependencyVersion(
...
+ * @param checkRange A particular version, or range of versions.
+ */
+export function hasDependencyVersion(
+  project: Project,
+  dependencyName: string,
+  checkRange: string
+): boolean | undefined {
+  const file = NodePackage.of(project)?.file;
+  if (!file) {
+    throw new Error(`Project does not have a NodePackage component`);
```
mrgrain, line 262: This seems like the same situation as the next one. Let's not treat them differently.

```suggestion
    return undefined; 
```

mrgrain (changes requested):

mrgrain (approved):

PR #3285, opened 2024-01-18
fix(release): workflow is failing because of invalid bash command

mrgrain:
A stray `'` made it's way into the codebase in #3252, causing the release workflow to fail.

This is currently blocking the release of projen. Which is good, because it means the broken code did not get published.
I don't know of a better way to catch these kind of mistakes. =/

---


kaizencc (approved):

dkershner6:
Sorry about that.

mrgrain:
> Sorry about that.

That was me!

Issue #3283, opened 2024-01-18
feat(typescript): tsconfig paths handled automatically by ts-jest pathsToModuleMapper


dkershner6:
Since projen removes the ability to use a .ts or .js file for jest config, using ts-jest's pathsToModuleMapper takes some mind-bending (you must use it outside a project, then use the result as options for the project).

Would be nice if projen just detected paths and did this automatically.

https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/

mrgrain:
Nice! Yeah that seems like a relatively straight forward feature.

dkershner6:
> Just a heads up though, if possible we prefer passing values at construction time vs methods. The latter causes more complexity & uncertainty in terms of understanding when certain values are available and when not.

> In other words, as a user using methods is fine. On a framework level they should be last resort because the force users to drop down another level in the chain to using something like pre-synth.

Heard. It does make my other PR unnecessary, but it didn't hurt anything.

dkershner6:
We decided not to implement this, see https://github.com/projen/projen/pull/3289

PR #3287, opened 2024-01-18
feat(jest): addModuleNameMappers, addModulePaths, addRoots

dkershner6:
Dependency of https://github.com/projen/projen/issues/3283

---


mrgrain (approved):
Nice one!

Just a heads up though, if possible we prefer passing values at construction time vs methods. The latter causes more complexity & uncertainty in terms of understanding when certain values are available and when not.

In other words, as a user using methods is fine. On a framework level they should be last resort because the force users to drop down another level in the chain to using something like `pre-synth`.

PR #3289, opened 2024-01-19
feat(typescript): ts-jest pathsToModuleMapping when tsconfig.paths populated

dkershner6:
Resolves https://github.com/projen/projen/issues/3283

---

### DIFF
--- a/src/typescript/typescript.ts
+++ b/src/typescript/typescript.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
 import * as semver from "semver";
-import { pathsToModuleNameMapper } from "ts-jest";
+import { pathsToModuleNameMapper } from "./ts-jest-utils";
 import { PROJEN_DIR } from "../common";
 import { Component } from "../component";
 import {
--- a/src/typescript/typescript.ts
+++ b/src/typescript/typescript.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
 import * as semver from "semver";
+import { pathsToModuleNameMapper } from "ts-jest";
 import { PROJEN_DIR } from "../common";
 import { Component } from "../component";
 import {
@@ -401,6 +402,35 @@ export class TypeScriptProject extends NodeProject {
   public readonly watchTask: Task;
 
   constructor(options: TypeScriptProjectOptions) {
+    const jestConfig = {
+      ...options.jestOptions?.jestConfig,
+      testMatch: options.jestOptions?.jestConfig?.testMatch ?? [],
+    };
+
+    // ts-jest tsconfig paths to module mapping
+    if (options?.tsconfig?.compilerOptions?.paths) {
+      if (
+        Array.isArray(jestConfig.roots) &&
+        !jestConfig.roots.includes("<rootDir>")
+      ) {
+        jestConfig.roots.push("<rootDir>");
+      }
+
+      if (options.tsconfig.compilerOptions.baseUrl) {
+        jestConfig.modulePaths = [
+          ...(jestConfig?.modulePaths ?? []),
+          options.tsconfig.compilerOptions.baseUrl,
+        ];
+      }
+
+      jestConfig.moduleNameMapper = {
+        ...(jestConfig?.moduleNameMapper ?? {}),
+        ...pathsToModuleNameMapper(options.tsconfig.compilerOptions.paths, {
+          useESM: options?.tsJestOptions?.transformOptions?.useESM,
+        }),
+      };
+    }
+
     super({
       ...options,
 
@@ -409,10 +439,7 @@ export class TypeScriptProject extends NodeProject {
 
       jestOptions: {
         ...options.jestOptions,
-        jestConfig: {
-          ...options.jestOptions?.jestConfig,
-          testMatch: options.jestOptions?.jestConfig?.testMatch ?? [],
-        },
+        jestConfig,
       },
     });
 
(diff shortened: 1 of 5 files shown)


dkershner6:
@mrgrain , not sure what to do about ts-jest as a dep.  If I make it a peerDep, it fails the license-check on `caniuse`.  But, currently, lint is rightly complaining it isnt one.

Code is complete and works, though.

mrgrain:
We can't include the `pathsToModuleNameMapper` in our codebase, sorry.

I would have thought you are going to implement the config without helper approach: https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/#jest-config-without-helper

mrgrain:
In an ideal world we could have this client side and emit a JS/TS based config for Jest. But we are blocked by at least #2405 for that.

dkershner6:
> We can't include the `pathsToModuleNameMapper` in our codebase, sorry.
> 
> I would have thought you are going to implement the config without helper approach: https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/#jest-config-without-helper

No worries.  Implementing without the helper is pretty complex due to all the intricacies surrounding filenames with and without ESM (the example only handles one case), and the proper handling of the disparate wildcard syntax, so I would actually recommend that if we are hampered from using the package directly, we just copy it.

dkershner6:
@mrgrain , you can have another look.  Did my best to adapt ts-jest logic into the projen codebase.  We lose forward compatibility if ts-jest makes changes, but otherwise this should handle everything using the package does, and is tested as such.

mrgrain:
Related: what's the current experience like if someone would want to use the helper directly from `ts-jest` with your new `addXyz` methods. Maybe that's an experience we can improve and document, so we don't have to only rely on keeping our version up to date

dkershner6:
> Related: what's the current experience like if someone would want to use the helper directly from `ts-jest` with your new `addXyz` methods. Maybe that's an experience we can improve and document, so we don't have to only rely on keeping our version up to date 

The code would look approximately the same as I have written, but could be placed after the project and use those methods, saving some array spreads. The user could also use the class property tsconfig instead of options.

It would match the ts-jest docs pretty closely though, the only weird part would be calling the `pathsToModuleMapper` inside a class method parameter.

src/typescript/typescript.ts
```diff
@@ -401,6 +402,35 @@ export class TypeScriptProject extends NodeProject {
...
+          options.tsconfig.compilerOptions.baseUrl,
+        ];
+      }
+
+      jestConfig.moduleNameMapper = {
+        ...(jestConfig?.moduleNameMapper ?? {}),
+        ...pathsToModuleNameMapper(options.tsconfig.compilerOptions.paths, {
+          useESM: options?.tsJestOptions?.transformOptions?.useESM,
+        }),
+      };
```
mrgrain, line 431: Could there be situations where I don't want roots, modulePaths and/or moduleNameMapper set? The current code  is additive (which is good), but forces the config on users (which is bad) without an option to disable the behavior.

src/typescript/typescript.ts
```diff
@@ -401,6 +402,35 @@ export class TypeScriptProject extends NodeProject {
...
+      testMatch: options.jestOptions?.jestConfig?.testMatch ?? [],
+    };
+
+    // ts-jest tsconfig paths to module mapping
+    if (options?.tsconfig?.compilerOptions?.paths) {
+      if (
+        Array.isArray(jestConfig.roots) &&
+        !jestConfig.roots.includes("<rootDir>")
+      ) {
+        jestConfig.roots.push("<rootDir>");
```
mrgrain, line 416: Any idea why this required? And could it ever be something like `<rootDir>/whatever`?

dkershner6: I believe that the `baseUrl` in tsconfig is by definition off the root, so if someone changes jest roots, we need one to be rootDir.

dkershner6:
> Not sure I really like going down this approach. Sorry that I didn't realize this isn't a simple mapping. Adding the configuration directly to a `.projenrc.ts` file seems easy enough. Maybe we are be better off documenting this well and making it even easier to add manually?
> 
> ```ts
> project.jest.mapPathsToModuleNames(pathsToModuleNameMapper(compilerOptions.paths));
> ```

It feels odd to be so close, having all of the information and `ts-jest` installed, and yet so far from being able to be more helpful here, which feels very projen-y to do.

That being said, I can't see a way through without either:
1. Moving the implementation until after the project is instantiated and ts-jest is already around.
2. Copying something from ts-jest.
3. Installing ts-jest as part of projen.

If none of those seem palatable, we are indeed only left with documentation and the user must implement themselves, which I am totally fine with and have already done, I just explained it to another developer and it was a bit mind-bending to do when they were used to the ts-jest documented approach.

mrgrain:
Yes, the intended way to deal with this kind of problem is to move the dependency into userland, and generate the required files for the user. In this case the file would look very much [like the one from the ts-jest docs](https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/#jest-config-with-helper).

We don't have a generic way of creating these files at the moment. Although for TS this is probably much easier than for JS since we don't have to deal with the CJS vs ESM problem. It would still be a bit of a task to merge the current JSON based config into this file though.

---

I've decided I'm really not comfortable with importing this code. I'd be open to either
- a helper function and docs that make it effectively a one-liner 
- a PR that generates an appropriate `jest.config.ts`, even without a generic solution for JS based config files.

PR #3291, opened 2024-01-20
fix(python): poetry & setuptools not using python binary provided by `pythonExec`

garysassano:
Fixes the following issue:

image

---


src/python/poetry.ts
```diff
@@ -20,14 +20,17 @@ export class Poetry
...
+  private readonly pythonExec: string;
+
 
   /**
    * A task that uploads the package to the Test PyPI repository.
    */
   public readonly publishTestTask: Task;
 
-  constructor(project: Project, options: PythonPackagingOptions) {
+  constructor(project: Project, options: PythonPackagingOptions, pythonExec: string = "python") {
```
mrgrain, line 31: This needs to go into `options`. And at that point likely becomes a new `PoetryOptions` interface that extends `PythonPackagingOptions`.

mrgrain (approved):

Issue #1904, opened 2022-06-02
Editing Task Steps (Remove Specific Step Without Reseting)


joyliu-q:
In the `.projenrc` file (I'm using the `react-ts` app), it is not possible to conveniently edit the individual steps of a particular task in the `tasks.json` file.

## Current Method
To remove a specific step in a task, there is one way:
### Reset & Repopulate
```
project.buildTask.reset();
// Add stuff to buildTask afterwards
```
Here, you would look for the pre-existing task, nuke it, and refill it with your custom steps. This method is not very compatible if you're just trying to remove one singular thing (e.g. tests) from your build workflow and keep the rest as the projen standard (and change when projen updates).

### Editing steps field of a task directly
```
const newBuildSteps = project.buildTask.steps.filter(step => step.name !== "test");
project.buildTask.steps = newBuildSteps
```
In above, steps is immutable, so it's not possible to set the steps field. I'm assuming there's some reason for this immutability. 
### A `editTask` method
This doesn't exist, but would be pretty helpful to have.

mrgrain:
`Task.reset` currently allows the user to replace the task with a single new command (TaskStep). This might be confusing, especially given the name. Maybe a fluent interface would be better `task.reset().add()` or adding something like `task.replace(new Task())`

PR #3311, opened 2024-01-30
feat(task): getStep, editStep, removeStep, cwd

dkershner6:
Fixes https://github.com/projen/projen/issues/1904

---


mrgrain (commented):
```
    // If the list of steps is a Lazy value, we can't know what the steps
    // are until synthesis occurs, so just return an empty array.
```

The `steps` getter has this note. How does this PR deal with this?

dkershner6:
> ```
>     // If the list of steps is a Lazy value, we can't know what the steps
>     // are until synthesis occurs, so just return an empty array.
> ```
> 
> The `steps` getter has this note. How does this PR deal with this?

Added a warning for folks to try using the `preSynthesize` phase if they somehow run into that for all methods.  Good to look over once more, @mrgrain .

dkershner6:
Tests failing on what seems like an errant type in a lib:

```
node_modules/@types/node/ts4.8/crypto.d.ts:3495:17 - error TS1110: Type expected.
```

src/task.ts
```diff
@@ -154,6 +154,14 @@ export class Task {
    */
   public exec(command: string, options: TaskStepOptions = {}) {
     this.assertUnlocked();
+
+    if (!Array.isArray(this._steps)) {
+      warn(
+        `Cannot add exec to task "${this.name}" because it is a lazy value, try using the preSynthesize phase.`
+      );
+      return;
+    }
```
mrgrain, line 163: Nice and thanks for adding this everywhere! Can you pull this out into a helper, similar to `assertUnlocked()` so we don't have to repeat ourselves.

mrgrain (changes requested):

dkershner6:
@mrgrain , I realized that `reset()` would also have failed, so I added a check there as well, but all now use a centralized helper.  Feel free to have another look at your convenience.

mrgrain (approved):

Issue #3315, opened 2024-01-31
GitHub Job interface made into or complemented with a class that has helpers to aid building a job


dkershner6:
Would make interacting with workflows and jobs significantly easier.  At the moment often you must violate readonly access to achieve goals for already built workflows.

@mrgrain , happy to stab at this, but if you agree with the direction (I believe this was actually your idea), please provide an opinion on how to go about this, as I am sure this is not the first time a transition from interface to class was undertaken.

mrgrain:
I'm not sure I remember the exact situation. If you can help me recall what issues this is attempting to resolve and how, I can be of more assistance to you.

dkershner6:
Ping @mrgrain for feedback because I must always be bothering you with at least 1 thing.

mrgrain:
Thanks @dkershner6 I'm happy to move ahead with this. It will be relatively low priority on my end, so I'd also recommend you to look for experienced peers to provide reviews to make the cycles quicker.

And please make sure you write some docs around this as part of the PR.

dkershner6:
> Thanks @dkershner6 I'm happy to move ahead with this. It will be relatively low priority on my end, so I'd also recommend you to look for experienced peers to provide reviews to make the cycles quicker.
> 
> And please make sure you write some docs around this as part of the PR.

Appreciate the feedback, Momo.  I will think about this some more versus adding more `preSteps` and `postSteps`.  Really, I just don't like violating readonly access, but that is our current course of action, which works as well.

dkershner6:
Maybe worth something someday, but just going to close for now.

PR #3316, opened 2024-01-31
feat(github): Job Builder Class

dkershner6:
Fixes https://github.com/projen/projen/issues/3315

Just a proposal, once we have a direction I will test and such.

---


mrgrain:
Before you get into development, can you write (or update) a pseudo documentation page illustrating how this new API would be intended to be used? This might be better done and discussed on the issue.

Issue #3304, opened 2024-01-26
Issues with New Child Project Release Feature and Duplicate Projen Package Installation


timothymathison:
I'm happy about the new child project release functionality added by #3208. But while taking advantage of this feature, I noticed a few issues that I would like to discuss here before attempting to PR fixes.

### Install Directory Issues

#3208 forced the `Install dependencies` step to be executed in the parent project directory and didn't provide a configuration to override that behavior. @dkershner6 I'm curious about the reason for that decision and if you have any additional context about the manner you envision that working in application?
When a child node project is created in a sub-directory with `projen`, it initializes the sub-directory with its own `package.json` and dependencies. So before the child project can be compiled, the install command needs to be run in the child project directory - not the parent directory. Or am I missing something about how this is supposed to be setup?

Furthermore, [by default](https://github.com/projen/projen/blob/0da09171ef6ce0d22bf6ceddd24c9fb69c228dc3/src/javascript/node-project.ts#L560) `projen` automatically adds itself as a dependency to the child project (in addition to the parent project) - leading to two different copies of the `projen` package being present. This can lead to some weird behavior such as `instanceof` checks returning `false` when `true` is expected. e.g. #3208 added this line:
https://github.com/projen/projen/blob/0da09171ef6ce0d22bf6ceddd24c9fb69c228dc3/src/release/release.ts#L355
Which grabs the parent project `Github` object using this logic:
https://github.com/projen/projen/blob/0da09171ef6ce0d22bf6ceddd24c9fb69c228dc3/src/github/github.ts#L79
However, the child project defaults to importing `Github` class from its own `node_modules` directory while the parent project `Github` class object will be from the parent's `node_modules` directory, leading to `GitHub.of(this.project.root)` returning `undefined`.
With `github` set to `undefined`, the child project release workflow does not get created unless I force the child project to import its projen classes from the parent `node_modules` directory!
Is there a strong reason why `projen` defaults to installing itself in child project directories? In the event that `release: true` is set on a child project, can we change the default behavior or at least call out that gotcha in the docs somehow?

### Trigger Path Configuration

It seems that `projen` does not provide an option to customize the `push` trigger for the `release` workflow, and just sets the `branches` options. In the case of `child` projects, I think it would generally be preferable to only trigger the `release` workflow for a child project if there are changes to the child project directory using the `push.paths` option.
How do we want to go about providing the option to customize that? Should the default `push` trigger include `paths: [{childProjectOutdir/**}]`?

[...]

PR #3317, opened 2024-01-31
feat: build workflow for sub-projects

timothymathison:
Creates a working `build` workflow specific to a sub-project when that project has `buildWorkflow` set to `true`.

Based on #3208. So most of the same decisions were made. The `install` command is always run in the root directory same as the `release` workflow. But based on discussion in #3304, we may want to make that configurable. I could make that change in this PR for `build`. Or just address it in a followup PR for both `build` and `release`.
I added a `buildWorkflowName` option to `NodeProject` to support customizing each build workflow name; but it still defaults to `build` or `build_{package-name}` for sub-projects.

I've successfully tested this with my own monorepo and added a few tests.

---


src/github/workflow-actions.ts
```diff
@@ -39,11 +40,22 @@ export class WorkflowActions {
...
-        with: { name: GIT_PATCH_FILE, path: GIT_PATCH_FILE },
+        with: {
+          name: GIT_PATCH_FILE,
+          path: `${
+            options.workingDirectory &&
+            !options.workingDirectory.match("^.(/)*$")
+              ? `${options.workingDirectory}/`
+              : ""
+          }${GIT_PATCH_FILE}`,
+        },
```
timothymathison, line 58: Instead of adding this option (which I'm just using to set it to `./`), I'm considering just hard-coding it always to `./`. I can't think of a scenario where it would actually be desirable to set it to anything besides `./` (repo root)

timothymathison:
@mrgrain are you able to approve the build workflow for this PR?

timothymathison:
Thanks @mrgrain. The build is failing with a unclear jsii test failure. But I also get the same failure when running tests with the head of `main` checked out. Any idea what's going on?

mrgrain:
Not sure what's going on. The last CI/CD build on main was still successful: https://github.com/projen/projen/actions/runs/7734258354/job/21087924904 🤔

mrgrain:
@timothymathison See https://github.com/projen/projen/pull/3320#issuecomment-1923644372

I't will probably take me over the weekend to get this sorted. Apologies.

timothymathison:
thanks @mrgrain! Mind giving this a review?

mrgrain:
Yeah yeah, I'll get to it. ,🤣

mrgrain:
Sorry this is taken a bit longer than usual. I should have some more time next week.

timothymathison:
@mrgrain any update?

src/javascript/node-project.ts
```diff
@@ -93,6 +94,12 @@ export interface NodeProjectOptions
    */
   readonly buildWorkflow?: boolean;
 
+  /**
+   * Name of PR build workflow.
+   * @default "build"
+   */
+  readonly buildWorkflowName?: string;
```
mrgrain, line 101:
```suggestion
  readonly buildWorkflowOptions?: BuildWorkflowOptions;
```

And then define BuildWorkflowOptions

mrgrain (changes requested):

mrgrain:
Sorry this took so long. Do you know the feeling when you're going into what you expect to be a quiet week, and then it isn't? 🙈

src/javascript/node-project.ts
timothymathison: yeah I was following the pattern of `releaseWorkflowName`. But I agree its a better pattern to define an object with that option nested inside of it

src/build/build-workflow.ts
```diff
@@ -186,7 +203,17 @@ export class BuildWorkflow extends Component {
         contents: JobPermission.WRITE,
         ...options.permissions,
       },
-      steps: (() => this.renderBuildSteps()) as any,
+      defaults:
+        projectPathRelativeToRoot.length > 0 // is subproject,
```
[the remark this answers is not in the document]
timothymathison: Sure. I copied this approach from #3208. But it can probably be more proper

src/build/build-workflow.ts
```diff
@@ -123,21 +132,25 @@ export interface BuildWorkflowOptions {
...
   private readonly artifactsDirectory: string;
-  private readonly name: string;
 
   private readonly _postBuildJobs: string[] = [];
 
   constructor(project: Project, options: BuildWorkflowOptions) {
     super(project);
 
-    const github = GitHub.of(project);
+    const github = GitHub.of(this.project.root);
```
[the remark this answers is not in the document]
timothymathison: Yes

src/github/workflow-actions.ts
```diff
@@ -39,11 +40,22 @@ export class WorkflowActions {
           "git add .",
           `git diff --staged --patch --exit-code > ${GIT_PATCH_FILE} || echo "${options.outputName}=true" >> $GITHUB_OUTPUT`,
         ].join("\n"),
+        workingDirectory: options.workingDirectory
+          ? ensureRelativePathStartsWithDot(options.workingDirectory)
```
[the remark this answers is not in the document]
timothymathison: I'm not 100% sure. The Github actions syntax docs example uses `./` for a relative path but but I didn't see it say if that's required. So I added it based on its use in #3208

src/github/workflow-actions.ts
timothymathison: > Why is it not just workingDir + patchFileName?

The additional logic is just removing any `./` prefix when the working directory is root.

> I don't think it can just be root, given the in the previous step the patch file is generated in a subdirectory.

What I meant was that I don't think we would ever want to generate the patch file in a subdirectory. So we could just hard-code both the previous step and this step to use the root directory.

src/javascript/node-project.ts
timothymathison: I'm not sure it would be desirable to have two different `BuildWorkFlowOptions` types. So I'm considering just using [the existing `BuildWorkflowOptions` interface](https://github.com/projen/projen/blob/0d9027ac26d93c30784c1e184312189bc9e31cf5/src/build/build-workflow.ts#L33) here which would support a customizing a bunch of other things

timothymathison: `BuildWorkflowOptions` has a few properties that shouldn't be exposed at the `NodeProject` constructor level. So I made an interface called `NodeProjectBuildWorkflowOptions`. I wanted to use `Omit` with `BuildWorkflowOptions` to just exclude a few properties from `BuildWorkflowOptions`. But jsii doesn't support `Omit` 😕 . So I copy pasted properties from `BuildWorkflowOptions`.
Since there are several other `NodeProject` options that  are specific to `BuildWorkflow`, I marked them as deprecated and they can be replaced with properties within `NodeProjectBuildWorkflowOptions`.

src/javascript/node-project.ts
```diff
@@ -328,6 +338,81 @@ export interface NodeProjectOptions
   readonly checkLicenses?: LicenseCheckerOptions;
 }
 
+/**
+ * Build workflow options for NodeProject
+ */
+export interface NodeProjectBuildWorkflowOptions {
```
timothymathison, line 344: Wanted to define this type as
```ts
Omit<BuildWorkflowOptions, 'buildTask' | 'artifactsDirectory' | 'gitIdentity'>
```
But alas jsii does not support `Omit`.

src/github/workflow-actions.ts
mrgrain: Gotcha. Yeah I think this would be clearer & safer actually. Because a sub-project could absolutely cause a file outside itself to change, and if we limit the mutation to the subproject we'd miss that.

So let's change it to run mutation detection at root level only. Please make sure to add comments in, so we don't forget about it.

src/javascript/node-project.ts
mrgrain: Fair, but I really don't think we need all of these options here for now. Can we please limit this to `name`, `mutableBuild` and `workflowTriggers` for now please. We can iterate over the other options later. Particularly pre/post build steps needs some cleaning up I think.

PS: The way to do what you wanted to do, would be to pull out the shared options into a `Common BuildWorkflowOptions` interface or something and than extend implementations from it.

mrgrain (changes requested):
I think this will be the last round of changes.

src/javascript/node-project.ts
timothymathison: > Can we please limit this to `name`, `mutableBuild` and `workflowTriggers` for now please.

I have some specific use cases for setting `permissions` and `preBuildSteps`. And I'd like to avoid using a workaround just to patch the job to add those a different way. Any specific reasons you're against going ahead and adding those options along with `postBuildSteps`?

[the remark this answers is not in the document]
timothymathison: Hmm I forgot about the fact its exported from a submodule

mrgrain: Just to keep this already large PR under control and moving. You've correctly picked up on the redundancy for the other props.

Definitely happy for you to discuss additional props on a follow up issue/PR.

timothymathison:
@mrgrain can I get another review please?

mrgrain (changes requested):
One additional request, and the already open request to split the features out into separate PRs.

src/github/workflow-actions.ts
```diff
@@ -39,11 +39,17 @@ export class WorkflowActions {
           "git add .",
           `git diff --staged --patch --exit-code > ${GIT_PATCH_FILE} || echo "${options.outputName}=true" >> $GITHUB_OUTPUT`,
         ].join("\n"),
+        // always run from root of repository
+        // overrides default working directory which is set by some workflows using this function
+        workingDirectory: "./",
```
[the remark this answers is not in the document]
timothymathison: Can you think of a valid use case for running this step from anything besides the root of a git repository?

src/build/build-workflow.ts
```diff
@@ -30,7 +39,41 @@ const IS_FORK =
...
+   */
+  readonly workflowTriggers?: Triggers;
+
+  /**
+   * Permissions granted to the build job
+   * To limit job permissions for `contents`, the desired permissions have to be explicitly set, e.g.: `{ contents: JobPermission.NONE }`
+   * @default `{ contents: JobPermission.WRITE }`
+   */
+  readonly permissions?: JobPermissions;
+}
```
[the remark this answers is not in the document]
timothymathison: I just ripped out all the `buildWorkfowOptions` changes since they're not strictly required for this feature.

src/github/workflow-actions.ts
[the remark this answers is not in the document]
timothymathison: >Use case would be having different workflows that are looking at subsets of the repo.

The diff would still include changes in other directories though - even if its run from a subdirectory (i.e. the unexpected behavior I was referring to).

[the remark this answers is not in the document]
timothymathison: If the `workingDirectory` option is set to something other then the root directory, we could add `./` to the `git diff` command to make it only report the diff of that working directory since I think that's the behavior that having a `workingDirectory` option implies. Thoughts?

>It's a breaking change otherwise and I don't see why we would not make it an option.

I'm not sure I see how its really a breaking change to _not_ have the `workingDirectory` option and hard code it to root (`./`) since that's the default behavior anyway - except for when the working directory of the whole job happens to have been overridden. But if that were to be the case, without my change, I think `uploadArtifact` would fail anyway because it would expect `GIT_PATCH_FILE` to be at the root while `git diff` would have created it inside the subdirectory.

I think making `workingDirectory` an option could be valid. But its not strictly required for this feature and would add complexity, while hardcoding it to `./` (root) updates a bunch of snapshots but should be a functional noop.
Otherwise, I can add the `workingDirectory` option back in and add the `./` path to the `git diff` command to make it behave as one would expect (only collect diff from subdirectory).

mrgrain (approved):

PR #3321, opened 2024-02-02
fix(poetry): maintain frozen lockfile for install task

### LINKED ISSUE #2913: Poetry `install` task updates project dependencies
# Overview
With the introduction of support for Python projects (https://github.com/projen/projen/pull/514) the `install` task for `Poetry` was defined as:

 - `poetry update` (https://github.com/projen/projen/blame/1020d0ad3f95bb4e1201f8640d440ffb43ccf230/src/python/poetry.ts#L30-34)

This is still the case today: https://github.com/projen/projen/blob/5c7f4202f47ff530f600edf1b8aec5e49e2ad3da/src/python/poetry.ts#L32-L34

The description does mention that this task also upgrades dependencies and it does work, but it also means tasks I expect to use my locked dependencies actually update my dependencies and rewrite `poetry.lock` which was unexpected. For example, running `projen build` or simply `projen` to update my project files also has the side effect of updating my Python dependencies.

Especially for fast changing libraries (e.g., `botocore`, `botocore_stubs`) this frequently results in updates which we may not want to take yet (or would like to take in the PRs generated by the `update` workflow). Our workflow may be partially at fault here, but it's also regularly causing failed builds for us due to mutations in files a developer isn't expecting to need to update. The `self-mutation` process does help out here to at least try to patch things up, but it's still not ideal in my view.

I'm not sure if there was specific reasoning behind this decision or not. If so and you believe it should persist, I'll be satisfied with some explanation and assistance. If the reasoning is unknown and/or you agree with me, I'd be happy to help resolve the issue.

Based on an error I recently encountered, I think there is a possibility this was done to prevent a message like the following which occurs if new dependencies are added to `.projenrc` and `poetry install` alone is used:
image

# Potential Solution
We could redefine the `install` task for `poetry` to be: `poetry lock --no-update && poetry install`.

As I understand it, this will refresh the lockfile _without_ updating dependencies which resolved the error shared in my screenshot above and, I believe, should be safe for the general install workflow.

The `--no-update` functionality was added to `poetry` in October 2020, so it's been around for a while at this point:
image

 - https://github.com/python-poetry/poetry/blob/master/CHANGELOG.md#111---2020-10-05
 - https://github.com/python-poetry/poetry/releases/tag/1.1.1

That said, it was pretty new at the time of the original code here and very possibly didn't exist during the original development of these features.

Given the longstanding nature of this behavior, I could see the need to introduce this behavior conditionally, but will definitely leave such decisions to the maintainers. :)

## Thanks
Thanks to everyone involved with `projen` for the hard work on this project!

cogwirrel:
Change the poetry install task from `poetry update` to `poetry lock --no-update && poetry install` to ensure that the lockfile is honoured in CI/CD pipelines.

Post synthesis, run a "mutable install" (`poetry lock && poetry install`) task which updates the lockfile, if the pyproject.toml file has changed (ie dependencies were changed).

Add a new `upgrade` task which runs `poetry update` to update dependencies.

Fixes #2913

---


mrgrain:
Nice one! Can we swap the naming around here to mirror what we have for Node?

```
mutable install -> install
immutable install -> install:ci
```

cogwirrel:
> Nice one! Can we swap the naming around here to mirror what we have for Node?
> 
> ```
> mutable install -> install
> immutable install -> install:ci
> ```

Thanks! Good call - updated! 😄

mrgrain (approved):

cogwirrel:
Thanks Momo! 😄

Issue #2438, opened 2023-02-11
Support multiple workflow engines: Gitlab


mattiamatrix:
Hello everyone,

looking at [projen vision](https://github.com/projen/projen/blob/main/VISION.md) document, one of the _Ideas_ consist of _supporting multiple workflow engines_. Gitlab, what about it? Is it in the interest of the project to support GitLab with the same level of support given to GitHub?

I am mainly asking this because my personal adoption of projen has been impacted by the lack of full support for GitLab workflows.

I am aware of _Manual release_ (#1938 ?), and I am sure there are many possible ways to work on GitLab. 

However, as a beginner user of projen I would like to have my GitLab workflow autogenerated by the framework. 

What do you think? Please correct me if I am completely wrong!

Cheer

mrgrain:
We should do this. I reckon the tricky bit will be finding an abstraction for workflows that can be applied to multiple providers.

agdimech:
I’ve been thinking of this and am actually in the position that we should allow these types of features to be decorated onto a simple kernel project, building the outer layers additively as component add ons.

`new Project({name: “foo”}).with(new Github({defaultBranch: “main”}).with(new Typescript()).with(new CDKApp(…));`

In this approach - users opt in to various features by stringing them together - allowing more tightly scoped required props and more ways to mix and match.

If a user always wanted to apply EsLint and GitHub they can simply configure them once into some type of middleware/decorator and just apply it to all their projects at creation time. For example:

`
const decorator = (project: Project) => project.with(
new Github({defaultBranch: “main”}),
new Eslint({ignorePatterns: [“bar”]})
);

new Project({name: “foo}).with(decorator);
new Project({name: “bar”}).with(decorator);
`

if other users wanted Gitlab, then they can configure that accordingly. And if there are certain combinations that make sense in lots of cases then we can just create default composition that people can use as is.

mrgrain:
I really like the decorator API. Some initial thoughts:
- Currently all components require a project. That means we can't instantiate them and then pass them to `with()`. They'd always need wrapping in a function that takes a project as first argument.
- Free floating functions don't exist in jsii. So the decorator interface might have to be be an `IDecorator` that has a `create` method.
- Of course we could make components not need a project
- Or we lean into jsii save instantiate components by their jsii name (as a string). But we lose strong typing here.
[...]

PR #3330, opened 2024-02-06
feat(node): release from CI with npm provenance statements

gmeligio:
This PR adds `npmProvenance` to allow increasing supply-chain security for a package.

npm and pnpm support it, but Yarn still needs to, as per [npm guide](https://docs.npmjs.com/generating-provenance-statements). It's supported only for public packages.

It will be false by default.

## References

https://docs.npmjs.com/generating-provenance-statements
https://docs.npmjs.com/cli/v10/using-npm/config#provenance

---


mrgrain:
This is awesome! Now projen's default release workflow uses `publib` which interally uses `npm`, so I don't think we need to forcefully limit this for yarn users either. Maybe just write this in the docs?

mrgrain:
PS: The failing build should fix once this PR is merged: https://github.com/projen/projen/pull/3331

gmeligio:
Thank you, @mrgrain, for fixing the build.

I have removed the Yarn limitation, leaving `provenance`=true  by default for public packages. Node package managers can indeed be mixed in different development stages. I added documentation as you suggested.

As a consequence, `projen` itself will have provenance enabled. I think it's a good default.

mrgrain:
We will also need to add the extra `id-token: write` permission to the release workflow:

https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow

mrgrain:
I also have mild concerns about compatibility. How about we don't turn it on by default just yet. Enable it in projen and I can run it on a bunch of repos I have access to. After a month or so we can make it the new default.

gmeligio:
Thanks for reviewing it so promptly.

Sounds good. I changed the default to false. False is actually npm's default. But I think it won't cause compatibility issues.

As you said, it would be nice to test in real repositories and later change the default to true.

mrgrain:
> Sounds good. I changed the default to false. False is actually npm's default. But I think it won't cause compatibility issues.

I think this is one of my concerns: https://docs.npmjs.com/generating-provenance-statements#example-gitlab-ci-job
Once we enable it by default, presumingly npm would try to sign the release but it would fail unless the user change their gitlab config. This is less of an issue for GitHub because we somewhat control the release workflow. But there's currently not GitLab equivalent.

This made me think a bit. Maybe the main option for this should to be on the [Release](https://projen.io/docs/api/release#releaseoptions-) component. We could then use the environment variable instead of adding it package.json (might just be easier). My next leap is than that other ecosystems can sign releases as well! Most notably [PyPi](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)... tl;dr the best place for this flag is probably in [NpmPublishOptions](https://projen.io/docs/api/release#npmpublishoptions-) and that's where we can change the default eventually.

Of course we can still keep the option on the NodePackage, just need to document it well. And potentially be clever with defaults.

gmeligio:
You were right. It makes more sense to persist the provenance configuration in the CI workflow than in the package because it needs other features from the CI provider, such as permissions.

Your thoughts helped me to realize how to change the implementation. I used a similar approach to how `npmTokenSecret` is handled between `NodePackage and `Release` by setting `npmProvenance` in `NodePackage` and passing it from there to `Release.publishToNpm`.

I tested it in a project I use on GitHub and it worked correctly.

I left it for Github for now, which is better integrated with Projen components. I use more Gitlab in general, but to do this provenance part, I think it should wait until after https://github.com/projen/projen/issues/2438 is solved. 

Regarding other ecosystems, it would be nice to have it in PyPi and others. Both npm and PyPi _signatures_ are new to me. I'd like to dive into PyPi's details later and see how it can be done.

mrgrain:
Thanks, I'll review again soon.

If you're interested, for Python the equivalent would be [PyPI Trusted Publishers](https://docs.pypi.org/trusted-publishers/).

src/javascript/node-package.ts
```diff
@@ -494,6 +509,11 @@ export class NodePackage extends Component {
    */
   public readonly npmAccess: NpmAccess;
 
+  /**
+   * Wether provenance statements should be generated when package is published.
```
mrgrain, line 513:
```suggestion
   * Wether provenance statements should be generated when package is published.
   * @default false
```

mrgrain:
- Made some tweaks to docs
- Changed my mind about the default and it's now enabled by default

mrgrain:
@gmeligio This is ready to go now, but I'd like to give you chance to look at my changes first. Thank you so much for this PR though, this is huge!

gmeligio:
Thank you for all the changes. It reads much better now.  It looks great and ready to go. I just did a minor update to a doc comment.

mrgrain (approved):

gmeligio:
I made a mistake in this PR and created the fix in https://github.com/projen/projen/pull/3378 . I'm sorry for that. Could you review @mrgrain please when you have time?

PR #3335, opened 2024-02-07
fix(github): bump upload/download artifact actions to v4; resolves runtime deprecation warnings

icj217:
Update artifact-related github actions that are generating warnings in projects due to nodejs 16 being deprecated in github actions runners.

v4 of actions/upload-artifact [introduced](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) some breaking changes, so we've set `overwrite` to be true by default to mirror the previous behavior as closely as possible.

---

### DIFF
--- a/src/github/workflow-steps.ts
+++ b/src/github/workflow-steps.ts
@@ -230,8 +230,9 @@ export interface UploadArtifactWith {
    * Whether action should overwrite an existing artifact with the same name (should one exist)
    *
    * Introduced in v4 and represents a breaking change from the behavior of the v3 action.
+   * To maintain backwards compatibility with existing, this should be set the `true` (the default).
    *
-   * @default true (to maintain backwards compatibility)
+   * @default true
    */
   readonly overwrite?: boolean;
 }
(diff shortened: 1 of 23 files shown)


mrgrain (changes requested):
Can you keep this limited to upload/download and split other actions in to separate PRs please?

icj217:
> Can you keep this limited to upload/download and split other actions in to separate PRs please?

Sure thing. [Here's the separate PR](https://github.com/projen/projen/pull/3338)

src/github/workflow-steps.ts
```diff
@@ -224,6 +225,15 @@ export interface UploadArtifactWith {
    * @default 6
    */
   readonly compressionLevel?: number;
+
+  /**
+   * Whether action should overwrite an existing artifact with the same name (should one exist)
+   *
+   * Introduced in v4 and represents a breaking change from the behavior of the v3 action.
+   *
+   * @default true (to maintain backwards compatibility)
```
mrgrain, line 234:
```suggestion
   * Introduced in v4 and represents a breaking change from the behavior of the v3 action.
   * To maintain backwards compatibility with existing, this should be set the `true` (the default).
   *
   * @default true
```

mrgrain (approved):
:shipit:

Issue #3340, opened 2024-02-09
github-app-token action throwing error after yesterday's update


dkershner6:
`File not found: '/tmp/ghrxl-1/_actions/tibdex/github-app-token/3eb77c7243b85c65e84acfa93fdbac02fb6bd532/dist/post/index.js'`

https://github.com/projen/projen/pull/3338

@v2 is working fine for us, just not this specific hash.

mrgrain:
Ahhhhhh 💩 

Thanks for noticing. Fix is incoming. You might have to manually intervene with this one 😬

dkershner6:
> Ahhhhhh 💩
> 
> Thanks for noticing. Fix is incoming. You might have to manually intervene to get out of this one 😬

Shouldnt be an issue for us, only hit one repo so far.  Thanks for the quick action, Momo! (@mrgrain )

mrgrain:
@dkershner6 Can you confirm it's working again for you?

dkershner6:
> @dkershner6 Can you confirm it's working again for you?

Testing now, will let you know and ping you in a few minutes.

dkershner6:
Confirmed as working, @mrgrain .

mrgrain:
Thank you, appreciate it!

PR #3338, opened 2024-02-08
fix(github): bump version of tibdex/github-app-token action; resolves runtime deprecation warnings

icj217:
Bump version of action so it uses node20 instead of node16

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #3111, opened 2023-11-12
Switch from `tibdex/github-app-token` to `create-github-app-token` action 


moltar:
GitHub released an official action with the same scope as `tibdex/github-app-token` (a third-party action).

Given the action's sensitivity and potential "blast radius", I think an official one should be preferred.

The API is almost the same, so I think this could be an easy swap.

https://github.com/actions/create-github-app-token#readme

mrgrain:
Very cool!

doublecompile:
I would say that the only thing holding up this change is for the official action to support custom permissions: actions/create-github-app-token#3

polothy:
Perhaps if you are not using custom permissions, it uses GH's action? I also feel a little uneasy using 3rd party action for credentials.

polothy:
Created https://github.com/projen/projen/pull/3591

mrgrain:
We are going to wait for the official action to catch up.

In the meantime, this can be used to switch over if permissions are not needed:

```typescript
project.github?.actions.set(
  "tibdex/github-app-token",
  "actions/create-github-app-token@v1"
);
```

polothy:
Just FYI, looks like GitHub's now has permissions support: https://github.com/actions/create-github-app-token/pull/168

PR #3341, opened 2024-02-09
fix(github): invalid commit for tibdex/github-app-token action

mrgrain:
In PR #3338 we attempted to update the version of the `tibdex/github-app-token` GitHub Action to a later version.
We incorrectly picked the latest commit from the main branch of this action. However that's not how most GitHub Actions work. Instead they a tag a dangling commit that has been specially created with the required build artifacts.

This PR we are updates the hash to the correct working version as referenced from the release page: https://github.com/tibdex/github-app-token/tree/v2.1.0

Fixes: https://github.com/projen/projen/issues/3340

---

### DIFF
--- a/src/github/github-credentials.ts
+++ b/src/github/github-credentials.ts
@@ -65,7 +65,7 @@ export class GithubCredentials {
         {
           name: "Generate token",
           id: "generate_token",
-          uses: "tibdex/github-app-token@3eb77c7243b85c65e84acfa93fdbac02fb6bd532",
+          uses: "tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a",
           with: {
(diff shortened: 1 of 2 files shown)


kaizencc (approved):

dkershner6:
Thanks again!

dkershner6:
Hey, @mrgrain , since we are both staring at this, can I ask why we reference a commit hash instead of a verison on this action?

mrgrain:
> Hey, @mrgrain , since we are both staring at this, can I ask why we reference a commit hash instead of a verison on this action?

For this highly security relevant action that is published by a more or less random user, we've always felt it would be better to use a specific vetted commit.

mrgrain:
There is this issue to switch to the official GitHub version: https://github.com/projen/projen/issues/3111

PR #3343, opened 2024-02-09
fix(upgrade-dependencies): unnecessary attempts to upgrade versions in package.json

mrgrain:
When `some-dep@^10` is added as a dependency, it is incorrectly included in the list of to-be-upgraded dependencies. This was due to a bug in a boolean expression; with the the code comment stating the clear intention to exclude them.

The reason for this is that ncu would upgrade `some-dep@^10` to a later version. But `projen default` would then revert back to the version defined in code.

While fixing the issue, I realized that while we need to exclude these deps from ncu, we still want them to be included in the package manager upgrade command, so that they receive the latest version for the given version constraint.

---

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -398,13 +398,13 @@
       },
       "steps": [
         {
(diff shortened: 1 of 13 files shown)


kaizencc (approved):
Depedencies :)

PR #3365, opened 2024-02-14
feat(github): download artifact workflow step

vinayak-kukreja:
This PR is adding a `download artifact` helper function to create a download artifact workflow step.

Here is the download artifact action for reference: https://github.com/actions/download-artifact?tab=readme-ov-file

---


test/github/__snapshots__/github-workflow.test.ts.snap
```diff
@@ -1,5 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          path: $GITHUB_WORKSPACE
+          repository: \${{ github.repository }}
+          merge-multiple: false
+          run-id: \${{ github.run_id }}
+"
```
mrgrain, line 21: I would have expected this to not be in the template. I think it's because you have added the defaults in to our code. What we usually do, is that we document them with `@default` but don't actually set them in code and rather let GitHub do its thing.

mrgrain (changes requested):

test/github/__snapshots__/github-workflow.test.ts.snap
vinayak-kukreja: Ok, removing it now.

mrgrain (approved):

PR #3366, opened 2024-02-14
feat(bundler): additional esbuild options and generic escape hatch to pass any options to esbuild

### LINKED ISSUE #3364: There are only a small subset of `esbuild` options exposed
Building on the work from #3102 (merged as PR #3103), I'd like to add more `esbuild` options into the bundler. Looking at [`@aws-cdk-lib/aws-lambda-nodejs`](https://github.com/aws/aws-cdk/blob/c3c771c6f6f6790f2298a85a549bded640d2e35b/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts#L195) for inspiration.

Also, since we won't cover *every* option of `esbuild` possible in the future, we'll want an escape hatch as well.

I have a PR for this that I'll open shortly.

giseburt:
Fixes #3364

---


giseburt:
I realize I have one property I haven't implemented yet, but I feel it's a fairly powerful option, but it has ramifications:

```typescript
  /**
   * Run compilation using tsc before running file through bundling step.
   * This usually is not required unless you are using new experimental features that
   * are only supported by typescript's `tsc` compiler.
   * One example of such feature is `emitDecoratorMetadata`.
   *
   * @default false
   */
  readonly preCompilation?: boolean;
```

This is inherited from [@aws-cdk-lib/aws-lambda-nodejs](https://github.com/aws/aws-cdk/blob/c3c771c6f6f6790f2298a85a549bded640d2e35b/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts#L195) as well. I feel the comment 👆🏻 by @hassanazharkhan (in aws/aws-cdk#16543) describes its usefulness quite well.

The issue here is that the existing option `BundlerOptions.addToPreCompile` (with `@default true`) will put the bundle task *before* compile (in `pre-compile`, actually), but with this new `AddBundleOptions.preCompilation` option (which I see now needs renamed), we want compile *first*, then have `esbuild` bundle the *compiled results*.

Unless anyone objects, I will:
- [ ] Add `BundlerOptions.addToPostCompile`, rename the new and unused `AddBundleOptions.preCompilation` to `AddBundleOptions.bundleCompiledResults` to make it more clear
- [ ] Default `BundlerOptions.addToPostCompile` to `true` and `BundlerOptions.addToPreCompile` to `false` when `AddBundleOptions.bundleCompiledResults` is set to `true` (being careful to not adjust existing default behavior)
- [ ] Throw an error if `AddBundleOptions.bundleCompiledResults` is `true` and `AddBundleOptions.preCompilation` was explicitly set to `true`

giseburt:
Ok, I believe all new options are implemented and test coverage is complete.

src/javascript/bundler.ts
```diff
@@ -63,7 +79,8 @@ export class Bundler extends Component {
   public readonly bundledir: string;
 
   private _task: Task | undefined;
-  private readonly addToPreCompile: boolean;
+  private addToPreCompile?: boolean;
+  private addToPostCompile?: boolean;
```
mrgrain, line 83:
```suggestion
  private readonly addToPreCompile?: boolean;
  private readonly addToPostCompile?: boolean;
```

test/javascript/bundler.test.ts
```diff
@@ -329,3 +362,256 @@ test("format can be set to esm", () => {
...
+    defaultReleaseBranch: "main",
+  });
+
+  // p.bundler.addBundle("./lib/hello1.js", {
+  //   platform: "node",
+  //   target: "node18",
+  //   sourcemap: false,
+  //   format: "esm",
+  // });
+
```
mrgrain, line 526:
```suggestion
```

test/javascript/bundler.test.ts
```diff
@@ -329,3 +362,256 @@ test("format can be set to esm", () => {
...
+      bundleCompiledResults: true,
+    });
+
+    Testing.synth(p);
+  }).toThrow(
+    /Cannot auto-set `addToPostCompile` after a bundle has been configured./
+  );
+});
+
+test("bundleCompiledResults  fails if both addToPreCompile and addToPostCompile are set", () => {
```
mrgrain, line 544:
```suggestion
test("bundleCompiledResults fails if both addToPreCompile and addToPostCompile are set", () => {
```

test/javascript/bundler.test.ts
```diff
@@ -329,3 +362,256 @@ test("format can be set to esm", () => {
...
+  const p = new NodeProject({
+    name: "test",
+    defaultReleaseBranch: "main",
+    bundlerOptions: {
+      addToPreCompile: true,
+      addToPostCompile: true,
```
mrgrain, line 550: Is this ever valid? I feel like this should already throw when NodeProject is instantiated.

test/javascript/bundler.test.ts
```diff
@@ -329,3 +362,256 @@ test("format can be set to esm", () => {
...
+      sourcemap: false,
+      format: "esm",
+      bundleCompiledResults: true,
+    });
+
+    Testing.synth(p);
+  }).toThrow(/Cannot have Bundle option `addToPreCompile` set/);
+});
+
+test("bundleCompiledResults fails if bundleCompiledResults are set (later)", () => {
```
mrgrain, line 514: I don't understand what this test is trying to test. Can you help me out by adding `//GIVEN` `//WHEN` & `//THEN` comments and explain in comments what should happen and why.

test/javascript/bundler.test.ts
```diff
@@ -329,3 +362,256 @@ test("format can be set to esm", () => {
...
+test("addToPreCompile set to false works (2)", () => {
+  const p = new NodeProject({
+    name: "test",
+    defaultReleaseBranch: "main",
+    bundlerOptions: {
+      addToPreCompile: false,
+    },
+  });
+
+  p.bundler.bundleTask;
```
mrgrain, line 482: I've seen this a few times. What's up with it? Is this unsuspecting property access call actually mutating something?

mrgrain (changes requested):
This looks great, thank you!  I've left feedback on a few minor things.

The biggest change would be how we are handling `addToPre/PostCompile`. I appreciate the effort to make this just work, but I think it unintentional ended up being a bit complicated and opaque.

I think what we have here is a three-way decision:

```ts
enum RunBundleTask {
  /**
   * Don't bundle automatically as part of the build.
   */
  MANUAL: "manual",
  /**
   * Bundle automatically before compilation.
   */
  PRE_COMPILE: "pre_compile",
  /**
   * Bundle automatically after compilation. Set this when using....
   */
  POST_COMPILE: "post_compile",
}


interface BundlerOptions {
  // ...

  /**
   * 
   * @default RunBundleTask.MANUAL
   */
  runBundleTask?: RunBundleTask;
}
```

And then throw and error when`options.runBundleTask != RunBundleTask.POST_COMPILE` and `bundleCompiledResults` is set.

We can keep backwards compatibility by setting something like `this.runBundleTask = options.runBundleTask ?? (options.addToPreCompile ? RunBundleTask.PRE_COMPILE : RunBundleTask.MANUAL)`.

What do you think?

src/javascript/bundler.ts
mrgrain: Ah I see now why you did it. Sorry this breaking your PR. I'll revert. 😂

src/javascript/bundler.ts
```diff
@@ -63,7 +79,8 @@ export class Bundler extends Component {
   public readonly bundledir: string;
 
   private _task: Task | undefined;
-  private readonly addToPreCompile: boolean;
+  private readonly addToPreCompile?: boolean;
+  private readonly addToPostCompile?: boolean;
```
mrgrain, line 83:
```suggestion
  private addToPreCompile?: boolean;
  private addToPostCompile?: boolean;
```

test/javascript/bundler.test.ts
giseburt: Yeah, this will be cleaned up by your main suggestion.

giseburt:
Thank you for the review!

Looking over your comments, I realize that `addToPreCompile` is completely redundant, and can be removed when moving to `RunBundleTask` as you suggested.

I'm making two minor adjustments:
- Marking `addToPreCompile` as `@depricated` due to the redundancy
- To ensure not breaking downstream, if neither `runBundleTask` nor `addToPreCompile` are defined, assume `RunBundleTask.PRE_COMPILE` as before

I just pushed the new code.

One other note: I'd like to provide more control over the bundle task name. For example, since I switched to `./lib/hello.js` from `./src/hello.ts` then bundle task name changed to `bundle:lib/hello` and I'd prefer to use `bundle:hello` or in fact I may prefer to call it `bundle:production` or whatever. Should I put that in this PR and make another after this? I think it'll be a few lines of code and twice that many of comments/docs.

mrgrain:
>  Should I put that in this PR and make another after this? I think it'll be a few lines of code and twice that many of comments/docs.

I always prefer separate PRs to keep things moving and independent.

mrgrain (approved):

PR #3375, opened 2024-02-17
fix(python)!: Poetry python project is using deprecated `dev-dependencies` notation in  `pyproject.toml`

garysassano:
[Poetry 1.2.0](https://python-poetry.org/blog/announcing-poetry-1.2.0/) introduced the `group` notation for managing dependencies, which has been the recommended way to organize dependencies inside your `pyproject.toml` file for almost two years. The older `dev-dependencies` notation was deprecated, as detailed [here](https://python-poetry.org/docs/master/managing-dependencies/), and its use is discouraged.

BREAKING CHANGE: Poetry is now required to be at version 1.2.0 or above. Please update Poetry if needed.

---


src/python/poetry.ts
```diff
@@ -371,23 +371,30 @@ export class PoetryPyproject extends Component {
...
+          dependencies: dependencies,
+          group: {
+            dev: {
+              dependencies: devDependencies,
+            },
           },
         },
       },
+      "build-system": {
+        requires: ["poetry-core"],
```
mrgrain, line 390: Given that we now require poetry >= 1.2.0 we should probably keep that in there.

mrgrain (changes requested):
Can you also mark this as a breaking change with a note how to resolve it (ie. upgrade poetry). Thanks!

src/python/poetry.ts
garysassano: No, [poetry-core](https://github.com/python-poetry/poetry-core) is just the build system which is part of [poetry](https://github.com/python-poetry/poetry). It's a component which is kept separate so that other projects can make use of it.

What you want to do is currently not possible, despite being a feature that people have been asking for quite a long time as you can see from [this open issue](https://github.com/python-poetry/poetry/issues/3316).

garysassano: It's a very common misconception, one that I also fell for in the past.

src/toml.ts
```diff
@@ -22,10 +22,20 @@ export class TomlFile extends ObjectFile {
       return undefined;
     }
 
+    // Convert JSON content to TOML string
```
mrgrain, line 25: This seems entirely unrelated to this PR. Please keep it separate

mrgrain (changes requested):

src/toml.ts
garysassano: How is it unrelated? Without modifying `toml.ts`, the formatting results in odd indentation that does not align with a fresh `pyproject.toml` file generated by `poetry init`. The aim is to produce a file through projen that appears as though it was directly created by poetry.

mrgrain: None of that is described in the PR or explicitly tested aside from the snapshot. I'd expect a bit of a write up about that and some investigation/reasoning why there is a discrepancy. 

Additionally I'm not at all convinced that that string manipulation is the right approach here. If the current TOML library is not capable of producing the desired output we will have to look for an alternative one. Swapping out the TOML package will definitely require a separate PR.

tl;dr All this just prevents a perfectly fine fix from being merged.

mrgrain (approved):

Issue #2427, opened 2023-02-09
Bump precondition fails on windows


alimhabib:
Hello All I am building projen and try to test on windows ( my dev machine) but no commands ever work for example 

`yarn build` will gives error 

```
Task "build » package" failed when executing "if [ ! -z ${CI} ]; then rsync -a . .repo --exclude .git --exclude node_modules && rm -rf dist && mv .repo dist; else np
x projen package-all; fi" (cwd: C:\Repository\cdk-infra\aws-monitor-alarms-library)
error Command failed with exit code 1.
```
`npm run bump ` will gives error 

```
� bump | condition: ! git log --oneline -1 | grep -q "chore(release):"
'!' is not recognized as an internal or external command,
operable program or batch file.
� bump | condition exited with non-zero - skipping
```

my projen,js looks like 

```
const { awscdk } = require('projen');
const project = new awscdk.AwsCdkConstructLibrary({
  author: 'xxxxx',
  authorAddress: 'xxxxxxxxxxx',
  cdkVersion: '2.1.0',
  peerDependencies: [
    'aws-cdk-lib',
  ],

  devDeps: [
    'aws-sdk',
  ],
  defaultReleaseBranch: 'master',
  name: 'xxxxxxxxxxxxxxxxx',
  repositoryUrl: 'xxxxxxxxxxxxxxxxxxxxx',

  majorVersion: 1,
  minNodeVersion: '14.17.0',
  sampleCode: false,
  // deps: [],                /* Runtime dependencies of this module. */
  // description: undefined,  /* The description is just a string that helps people understand the purpose of the package. */
  // devDeps: [],             /* Build dependencies for this module. */
  // packageName: undefined,  /* The "name" in package.json. */
});
project.gitignore.exclude('.env', '.idea');
project.gitignore.exclude('example/*.js', 'example/*.d.ts');
project.synth();
```

I am using projen version ^0.67.40

mrgrain:
Thanks @alimhabib 
See https://github.com/projen/projen/labels/windows

Windows support unfortunately is very basic at the moment.

RichiCoder1:
@mrgrain would ya'll take PRs to either de-shellify the code, or at least shim them via [@yarnpkg/shell](https://www.npmjs.com/package/@yarnpkg/shell)?

mrgrain:
What exactly do you mean by "de-shellify"?

[...]

Issue #1529, opened 2022-01-15
Build/Tests working on windows


mmcculloh-dms:
Projen itself can't currently be built on windows for a few reasons. One is due to the usage of bash for bootstrapping, which is pretty easy to get around. One of the biggest reasons is that many tests simply do not work on windows (by design). Typically this is due to the task subsystem.

I think it would be lovely to allow projen to be fully buildable on windows, including the testing portion. This may require segregating tests as it makes sense that you can't run certain commands on windows. Or we could have system-conditional logic within tests, but that seems a little scary.

gmeligio:
I think this can be closed since the projen build and test tasks are working on Windows. Here is the build workflow https://github.com/projen/projen/blob/f212132bd15ab0b608beebdd5795c16f631f13cb/.github/workflows/build.yml

PR #3377, opened 2024-02-18
chore: bootstrap with NodeJS script for projen cross-platform development

gmeligio:
## Motivation

I was taking inspiration from https://github.com/projen/projen/pull/2761. I didn't closely follow what was available there because it's outdated, but I used a similar approach.

There are different issues related to developing `projen` on Windows. The most related ones to bootstrapping are #1529 and #2427.

To break the Windows development problem into pieces, I started with bootstraping because it is command I run the most run when developing projen. The other projen tasks can be reviewed and modified if necessary in future PRs.

I should create an issue to expose this idea, but before doing that, I'm trying to see if breaking it like that is viable and starting with bootstrapping.

## Implementation

I decided to do something similar to #2761 and use a NodeJS script. Then, I translated the existing Bash script, as of today, to NodeJS. Also, `.projenrc.js` will generate that script, similar to how it works with the existing Bash script. Windows has no CI implementation because the `build` and `package` tasks could fail. 

## Change 1 projen.js

1. The bootstrapping script is a translation from Bash to NodeJS.
2. The critical bits are the `bootstrap` and `execCommand` functions.
3. In `execCommand`, the important part is `{ stdio: "inherit" }` to make sure to pass input and get output from the executed command by passing down the parent process streams to the child process.
4. I didn't add tests.

## Change 2 .projenrc.js

1. Replacing `projen.bash` with `projen.js`
2. Adding functions in the entrypoint `.projenrc.js`: `execCommand`  and `bootstrap`. These are only to simplify how `projen.js` is generated. Could it be better to move those two functions to another file so as not to pollute .projenrc.js?
4. I didn't add tests.

## Change 3 all files with marker

1. The marker was autoupdated to use `node ./projen.js`

---


.projenrc.js
```diff
@@ -201,23 +201,26 @@ project.npmignore.exclude("/docusaurus/");
...
-    "set -euo pipefail",
-    "if [ ! -f lib/cli/index.js ]; then",
-    '  echo "bootstrapping..."',
-    "  if [ ! -f node_modules/.bin/jsii ]; then",
-    "    yarn install --frozen-lockfile --check-files --non-interactive",
-    "  fi",
-    "  npx jsii --silence-warnings=reserved-word --no-fix-peer-dependencies",
-    "fi",
-    "exec bin/projen $@",
+    `// ${PROJEN_MARKER}`,
```
mrgrain, line 207: Can we add the [Node Hashbang](https://stackoverflow.com/questions/20638520/appropriate-hashbang-for-node-js-scripts) for convenience on unix and make the file executable?

.projenrc.js
```diff
@@ -201,23 +201,26 @@ project.npmignore.exclude("/docusaurus/");
...
+    "  bootstrap();",
+    "}",
+    "",
+    'const args = process.argv.slice(2).join(" ");',
+    "execCommand(`bin/projen ${args}`);",
+    "",
   ],
-  executable: true,
 });
-project.npmignore.exclude("/projen.bash");
```
mrgrain, line 220: Feels like we should keep this?

mrgrain:
This is awesome! Thank you. We should add a (failing) Windows build soon to observe your progress =)

.projenrc.js
gmeligio: Done. I missed that when translating the code.

.projenrc.js
gmeligio: Totally right. Restored.

.projenrc.js
```diff
@@ -331,6 +334,41 @@ function setupBundleTaskRunner() {
...
+ */
+function execCommand(command) {
+  try {
+    execSync(command, { stdio: "inherit" });
+  } catch (error) {
+    console.error(`Failed to execute command: ${command}`, error);
+    process.exit(1);
+  }
+}
+
```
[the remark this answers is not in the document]
gmeligio: Yes. It makes sense to move it there. I was hesitating to ask about guidance on that.

I think `ProjenrcTs` can be used for that. Right now it doesn't add the `projenrc` folder to npmignore and others. If it's OK, I would prefer to do it in another PR to not make this one so big.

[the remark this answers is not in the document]
mrgrain: 100% okay

I think we don't use .projenrc.ts here because it might make Bootstrapping more complicated. If we can, I'd love to though.

.projenrc.js
```diff
@@ -201,23 +201,30 @@ project.npmignore.exclude("/docusaurus/");
...
+    "}",
+    "",
+    'const args = process.argv.slice(2).join(" ");',
+    "execCommand(`node bin/projen ${args}`);",
+    "",
   ],
-  executable: true,
 });
-project.npmignore.exclude("/projen.bash");
+project.npmignore.exclude(`/${bootstrapScript.path}`);
```
mrgrain, line 227: Ha, we totally should just accept a file here. (Not relevant to this PR)

mrgrain (approved):

Issue #3300, opened 2024-01-25
`awscdk-app-py` project does not fully respect `--app-entrypoint` value


toxygene:
When creating a new app with `projen new awscdk-app-py`, the `--app-entrypoint` value is not fully used. The `cdk.json` `app` value is correctly set, but `app.py` is always created, instead of the file specified in the flag.

I believe the issue is here: https://github.com/projen/projen/blob/main/src/awscdk/awscdk-app-py.ts#L101

Instead of using `this.appEntrypoint`, it's always using `app.py`.

```
➜  test poetry run projen new awscdk-app-py --app-entrypoint=infra/app.py          
👾 Project definition file was created at /Users/justin.hendrickson/temp/test/.projenrc.py
👾 Setting up a virtual environment...
👾 Environment successfully created (located in ./.env).
👾 Installing dependencies...
👾 install | pip install --upgrade pip
Requirement already satisfied: pip in ./.env/lib/python3.11/site-packages (23.2.1)
Collecting pip
  Obtaining dependency information for pip from https://files.pythonhosted.org/packages/15/aa/3f4c7bcee2057a76562a5b33ecbd199be08cdb4443a02e26bd2c3cf6fc39/pip-23.3.2-py3-none-any.whl.metadata
  Using cached pip-23.3.2-py3-none-any.whl.metadata (3.5 kB)
Using cached pip-23.3.2-py3-none-any.whl (2.1 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 23.2.1
    Uninstalling pip-23.2.1:
      Successfully uninstalled pip-23.2.1
Successfully installed pip-23.3.2
👾 install | pip install -r requirements.txt
Collecting aws-cdk-lib<3.0.0,>=2.1.0 (from -r requirements.txt (line 2))
  Downloading aws_cdk_lib-2.123.0-py3-none-any.whl.metadata (59 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 59.3/59.3 kB 983.9 kB/s eta 0:00:00
Collecting constructs<11.0.0,>=10.0.5 (from -r requirements.txt (line 3))
  Using cached constructs-10.3.0-py3-none-any.whl.metadata (2.7 kB)
Collecting aws-cdk.asset-awscli-v1<3.0.0,>=2.2.202 (from aws-cdk-lib<3.0.0,>=2.1.0->-r requirements.txt (line 2))
[...]

PR #3378, opened 2024-02-18
fix(release): provenance statement not being generated because of missing quotes in YAML

gmeligio:
By looking into [projen in NPM](https://www.npmjs.com/package/projen), I noticed that it didn't have the provenance widget in the page. Then, looking into the [Github action run](https://github.com/projen/projen/actions/runs/7942036715/job/21685009089), I verified that it wasn't generated.

I realized, I made a mistake in [#3300](https://github.com/projen/projen/pull/3330) because the value `true` should be quoted in YAML. 

Before, wrong
```yaml
- name: Release
        env:
          NPM_DIST_TAG: latest
          NPM_REGISTRY: registry.npmjs.org
          NPM_CONFIG_PROVENANCE: true
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
After, correct
```yaml
- name: Release
        env:
          NPM_DIST_TAG: latest
          NPM_REGISTRY: registry.npmjs.org
          NPM_CONFIG_PROVENANCE: "true"
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

I tested this fix in one of my projects and it worked correctly
https://github.com/gmeligio/project-gen/actions/runs/7950505218/job/21702908858#step:10:57

Sorry for the mistake.

## Change 1 NPM_CONFIG_PROVENANCE

1. Add quotes to `"true"` so it's passed as an string to resulting YAML workflow.
2. Update `release.test.ts` with the new value where needed.

##  Change 2 workflowNodeVersion

1. Upgrade workflowNodeVersion to v18.14.2. According to [nodejs.org/dist/index.json](https://nodejs.org/dist/index.json), it's the first version in `18.14.X` that supports provenance statements.
```bash
❯ nvm use 18.14.2
Now using node v18.14.2 (npm v9.5.0)
```

## Future improvements

I received a warning in my downstream project but didn't see it in `projen`
```bash
👾 Received non-string value for environment variable NPM_CONFIG_PROVENANCE. Value will be stringified.
```
This helped me to confirm again the issue, but I was surprised of how come it doesn't show up when running `rm -rf lib && yarn projen` in the projen repository. I haven't investigated that yet, but maybe it would be nice if the message is also shown in the projen repository.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 18.14.0
+          node-version: 18.14.2
       - name: Install dependencies
         run: yarn install --check-files
       - name: build
@@ -97,7 +97,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.14.0
+          node-version: 18.14.2
       - name: Download build artifacts
         uses: actions/download-artifact@v4
         with:
@@ -126,7 +126,7 @@ jobs:
           java-version: 11.x
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.14.0
+          node-version: 18.14.2
       - name: Download build artifacts
         uses: actions/download-artifact@v4
         with:
@@ -151,7 +151,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.14.0
+          node-version: 18.14.2
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
@@ -179,7 +179,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.14.0
+          node-version: 18.14.2
       - uses: actions/setup-go@v3
         with:
           go-version: ^1.16.0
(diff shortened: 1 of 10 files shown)


mrgrain:
Ha that's a good one. I think there's an other issue as well though: The available npm version isn't quite up to date. We need npm 9.5.0 as minimum.

gmeligio:
Nice catch! npm 9.5.0 appeared in Node v18.14.2 according to [nodejs.org/dist/index.json](https://nodejs.org/dist/index.json). Currently, `workflowNodeVersion` is `18.14.0`, so it should be a minor change.

gmeligio:
Thank you, @mrgrain, for handling this so fast! Now it works https://www.npmjs.com/package/projen
image

Unfortunately, the npm message doesn't show up in the Github job because it was added in npm [v9.6.2](https://docs.npmjs.com/cli/v9/using-npm/changelog#962-2023-03-14), but that is less important I would say. An example of the message is below.
```bash
npm notice publish Signed provenance statement with source and build information from GitHub Actions
npm notice publish Provenance statement published to transparency log: https://search.sigstore.dev/?logIndex=72248776
```

mrgrain:
Awesome! I think there could be an argument to update the npm version for just this particular job to ensure a recent version is used.

Issue #3303, opened 2024-01-26
Error messages for npx projen new are not helpful


Spaideri:
## Update

- `--from <package-name>` is used to create new projects from external packages
- The user accidentally passed the project type to `--from`, so projen attempt to install use the project type as a package name
- Clearly the error message isn't very good here and should be improved

---

## Original issue 

Tried with multiple different templates and they all failed

## typescript
```npx projen new --from typescript```

```
npx projen new --from typescript
/Users/username/path/to/my/project/dir
└── (empty)


added 2 packages in 1s

10 packages are looking for funding
  run `npm fund` for details
👾 installing external module typescript...

added 1 package, and audited 71 packages in 542ms

10 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

Error: Cannot find module 'typescript/.jsii'
Require stack:
- /Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/lib/cli/cmds/new.js
- /Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/lib/cli/index.js
- /Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/bin/projen
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Function.resolve (node:internal/modules/helpers:187:19)
    at initProjectFromModule (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/lib/cli/cmds/new.js:234:44)
    at Object.handler (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/lib/cli/cmds/new.js:95:26)
    at /Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/node_modules/yargs/build/index.cjs:1:8993
    at j (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/node_modules/yargs/build/index.cjs:1:4956)
    at _.handleValidationAndGetResult (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/node_modules/yargs/build/index.cjs:1:8962)
    at _.applyMiddlewareAndGetResult (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/node_modules/yargs/build/index.cjs:1:9604)
    at _.runCommand (/Users/username/.npm/_npx/80cfb0dc84733c29/node_modules/projen/node_modules/yargs/build/index.cjs:1:7231)
[...]

PR #3385, opened 2024-02-22
fix(cli): unclear error messages when using `projen new` with `--from`

adamelkhayyat:
Improvement to error handling and error messages thrown in cases where `--from` flag is used (ex: `npx projen new --from [SPECIFIED-NPM-PACKAGE]`).

This PR covers the cases:
* where npm package specified does not exist in the registry and gracefully handles error with informative message to user
* where npm package specified is not jsii module and gracefully handles error with informative message to user
* add additional tests

Fixes https://github.com/projen/projen/issues/3303

---


src/cli/util.ts
```diff
@@ -55,6 +55,46 @@ export function renderInstallCommand(dir: string, module: string): string {
...
+        paths: [baseDir],
+      })
+    );
+  } catch (error: unknown) {
+    if (
+      error instanceof Error &&
+      "code" in error &&
+      error.code === "MODULE_NOT_FOUND"
+    ) {
+      return undefined;
```
mrgrain, line 90:
```suggestion
      // the provided module is not a jsii module
      return undefined;
```

src/cli/util.ts
```diff
@@ -55,6 +55,46 @@ export function renderInstallCommand(dir: string, module: string): string {
   return `npm install --save --save-dev -f --no-package-lock --prefix="${dir}" ${module}`;
 }
 
+export function moduleExists(cwd: string, module: string): boolean {
+  try {
+    logging.info(`Validating external module's (${module}) existence...`);
```
mrgrain, line 60: I don't think this needs to be visible in the regular logs.
```suggestion
    logging.debug(`Checking if external module '${module}' exists...`);
```

And maybe also add a debug log for the various result cases.

src/cli/util.ts
```diff
@@ -55,6 +55,46 @@ export function renderInstallCommand(dir: string, module: string): string {
...
+    const externalModule = !module.startsWith("./");
+    if (externalModule) {
+      exec(`npm view ${module}`, { cwd, stdio: "ignore" });
+      return true;
+    } else {
+      return fs.existsSync(module);
```
mrgrain, line 66: What is this case covering? Local file?

src/cli/util.ts
```diff
@@ -55,6 +55,46 @@ export function renderInstallCommand(dir: string, module: string): string {
   return `npm install --save --save-dev -f --no-package-lock --prefix="${dir}" ${module}`;
 }
 
+export function moduleExists(cwd: string, module: string): boolean {
+  try {
+    logging.info(`Validating external module's (${module}) existence...`);
+    const externalModule = !module.startsWith("./");
```
mrgrain, line 61: This feels like a new restriction. We might have to find out how `npm` does the disambiguation.

src/cli/cmds/new.ts
```diff
@@ -292,15 +298,26 @@ async function initProjectFromModule(baseDir: string, spec: string, args: any) {
...
-  const moduleDir = path.dirname(
-    require.resolve(`${moduleName}/.jsii`, {
-      paths: [baseDir],
-    })
-  );
+  const moduleDir = findJsiiFilePath(baseDir, moduleName);
+
+  if (!moduleDir) {
+    throw new CliError(
+      `Cannot find '${moduleName}/.jsii', ensure this is a jsii module first!`
```
mrgrain, line 318: Something like this maybe?
```suggestion
      `Module '${moduleName}' does not look like it is compatible with projen. Reason: Cannot find '${moduleName}/.jsii'. All projen module must be jsii modules!`
```

mrgrain (changes requested):
I'm excited by this! A few minor notes.

src/cli/util.ts
adamelkhayyat: This covers the case when you reference a local tarball (using `npm pack [dep]`).

src/cli/util.ts
adamelkhayyat: I looked into this further and I believe if I extend the function to check if the module starts with `./` or `/` (absolute paths) it should cover all cases where we're referencing a local file to install with `npm`.

Example installing a local package `some-local-package`:
- `npm install some-local-package` - ❌ Errors out as it attempts to find the package in registry
- `npm install /some-local-package` - ❌ Errors out as it expects absolute path, incomplete absolute path
- `npm install ./some-local-package` - ✅ reference a file that is in the root of the project
- `npm install --from /Users/adam.el-khayyat/Documents/personal-projects/testing-npm-pack/some-local-package`  - ✅ reference via absolute path

💡 I'd also need to adjust the code to accommodate Windows when checking if the local package/file exists (`\` path separator).

Please challenge my conclusion above if you see anything I might've missed or any flaws and I'll adjust as needed!

mrgrain: This works for me:

```
npm install some-local-package.tgz
```

mrgrain: I think if we extend this test case: https://github.com/projen/projen/blob/main/test/new.test.ts#L122
to run

- without `./` in front
- resolving to an absolute path

we should have tests for most cases

adamelkhayyat: > I think if we extend this test case: https://github.com/projen/projen/blob/main/test/new.test.ts#L122 to run
> 
> * without `./` in front
> * resolving to an absolute path
> 
> we should have tests for most cases

In that case we can check if the `module` ends in `.tgz` (and ensure the tarball exists) and we'll cover all existing test cases.

However, this function will return `false` if the `module` references a local project (example: `/Users/adam.el-khayyat/Documents/projects/my-projen-project`) that isn't a tar file compressed and cause a misleading error that the module doesn't exist.

adamelkhayyat: I've extended the function in https://github.com/projen/projen/pull/3385/commits/00ddc146dc9d7018e24475b082cf29e4f75d6da4 to check all conditions to determine if the module is a local one and changed one test based on your comment, let me know if this covers it well enough!

mrgrain: @adamelkhayyat I'd prefer if we have dedicated tests for all 3 cases, i.e. the existing one plus two new variations.

adamelkhayyat: Got it, wanted to confirm the approach first and then write the additional tests.

I'll try to get to this sometime this weekend!

adamelkhayyat: @mrgrain, I've added tests in https://github.com/projen/projen/pull/3385/commits/452d9ef840f56db1a970e63914845b4a644c0661, please let me know what you think!

adamelkhayyat: Note: absolute path test is failing on CI, I'll try to address this sometime this coming week.

test/new.test.ts
```diff
@@ -70,56 +70,105 @@ test("projen new --from external", () => {
...
+
     execProjenCLI(projectdir, [
       "new",
       "--from",
-      "@pepperize/projen-awscdk-app-ts@0.0.333",
+      `${resolve(
+        __dirname,
+        "..",
+        "pepperize-projen-awscdk-app-ts-0.0.333.tgz"
+      )}`,
```
mrgrain, line 151: This should do it
```suggestion
      `${resolve(
        projectdir,
        "pepperize-projen-awscdk-app-ts-0.0.333.tgz"
      )}`,
```

mrgrain:
Thanks @adamelkhayyat I think I can take it from here!

mrgrain (approved):

mrgrain (approved):

PR #3392, opened 2024-02-24
fix(python): `PythonProject` with Poetry tries to use virtual environment before it's created

### LINKED ISSUE #3390: Can't run `npx projen install` after cloning a `PythonProject` with Poetry
I executed the following sequence of commands:

- `git clone <repo>`
- `npx projen install`

This resulted in the following error message:

```sh
👾 unable to evaluate environment variable VIRTUAL_ENV=$(poetry env info -p): unknown error
```

garysassano:
Fixes #3390 

---


garysassano:
@mrgrain We could resolve this issue by prefixing `poetry env info -p` with `poetry run`, but I don't believe that's the proper solution. The essence of using Poetry is for it to handle the virtual environment management on your behalf. Manually setting up the variables is essentially an attempt to bypass the automatic handling that Poetry is designed to provide.

mrgrain:
> @mrgrain We could resolve this issue by prefixing `poetry env info -p` with `poetry run`, but I don't believe that's the proper solution. The essence of using Poetry is for it to handle the virtual environment management on your behalf. Manually setting up the variables is essentially an attempt to bypass the automatic handling that Poetry is designed to provide.

That's okay. You don't want to use `projen` as a task runner, you are not alone. This is the issue you want to track: https://github.com/projen/projen/issues/2051

---

In the mean time, you've found a bug with the setup of the projen task runner for python projects.

I think this would be the fix, making sure to replace `python` with `option.pythonExec`:

```console
poetry env info -p || (poetry env use python -q && poetry env info -p)
```

garysassano:
I'm not sure to understand how relevant is that issue you linked.

Anyway, the command you provided wouldn't fix the error since the problem here is that the virtual environment doesn't exist once you've just cloned a repo.

I will then fix the issue while maintaining the existing logic, but I must say that I'm not a fan of manually setting the virtual environment like it's currently being done.

Ideally, we should follow Poetry's guidelines to either spawn a sub-shell with `poetry shell` or activate the virtual environment through the provided script using `source {path_to_venv}/bin/activate`.

garysassano:
>How do you propose we find path_to_venv if not using poetry env info -p?

In my view, this could be effortlessly achieved by setting Poetry's configuration to include `virtualenvs.in-project = true`. Such a configuration aligns with the practice of having `node_modules` at the project's root, suggesting that `.venv` should also be located at the root. Using a setting like `virtualenvs.path = "{cache-dir}/virtualenvs"` (Poetry's default) doesn't seem appropriate for our scenario.

garysassano:
I had to adjust the command and make it more verbose in order to make sure it wouldn't throw a warning on the first run (`VIRTUAL_ENV` gets evaluted to `null` with the solution you proposed).

Now you can clone a new repo and run `npx projen install` immediately without any errors or warnings being displayed.

### New solution

#### tasks.json
```json
"env": {
  "VIRTUAL_ENV": "$(poetry env info -p || poetry env use python | grep -oP 'Using virtualenv: \\K.*')",
  "PATH": "$(echo $(poetry env info -p)/bin:$PATH)"
},
```

#### first run
```sh
❯ npx projen install
Creating virtualenv cdk-aws-apigateway-rest-api in /home/user/github/cdk-aws-apigateway-http-api/.venv
👾 install | poetry update
Updating dependencies
Resolving dependencies... (1.0s)
```

### Proposed solution

#### tasks.json
```json
"env": {
  "VIRTUAL_ENV": "$(poetry env info -p || poetry env use python -q)",
  "PATH": "$(echo $(poetry env info -p)/bin:$PATH)"
},
```

#### first run
```sh
❯ npx projen install
The virtual environment found in . seems to be broken.
Recreating virtualenv test-projen in /home/user/github/test-projen/.venv
Updating dependencies
Resolving dependencies... (1.0s)
```

garysassano:
I had to adjust the command and make it more verbose in order to make sure it wouldn't throw a warning on the first run (`VIRTUAL_ENV` gets evaluted to `null` with the solution you proposed).

Now you can clone a new repo and run `npx projen install` immediately without any errors or warnings being displayed.

### New solution

#### tasks.json
```json
"env": {
  "VIRTUAL_ENV": "$(poetry env info -p || poetry env use python | grep -oP 'Using virtualenv: \\K.*')",
  "PATH": "$(echo $(poetry env info -p)/bin:$PATH)"
},
```

#### first run
```sh
❯ npx projen install
Creating virtualenv test-projen in /home/user/github/test-projen/.venv
👾 install | poetry update
Updating dependencies
Resolving dependencies... (1.0s)
```

### Proposed solution

#### tasks.json
```json
"env": {
  "VIRTUAL_ENV": "$(poetry env info -p || poetry env use python -q)",
  "PATH": "$(echo $(poetry env info -p)/bin:$PATH)"
},
```

#### first run
```sh
❯ npx projen install
The virtual environment found in . seems to be broken.
Recreating virtualenv test-projen in /home/user/github/test-projen/.venv
Updating dependencies
Resolving dependencies... (1.0s)
```

garysassano:
There's a problem with grep existing only in UNIX-based OSs, so I think it would be just simpler to just use `poetry run poetry env info -p` for the first run.

garysassano:
There's a problem with grep existing only in UNIX-based OSs, so I think it would be just simpler to use `poetry run poetry env info -p` for the first run.

mrgrain:
Looks great now! If you update the snapshot and revert the doc change, this should be good to go.

mrgrain (approved):

PR #3395, opened 2024-02-25
feat(python)!: default license to `Apache-2.0` if unset

garysassano:
Adding a default `Apache-2.0` license to `PythonProject` to align it with `NodeProject`.

BREAKING CHANGE: The `PythonProject` now defaults to an `Apache-2.0` license. To use a different license, you must specify the `license` property in your project configuration.

---


mrgrain (changes requested):
Do we really not have any tests for this? 😱 

Also this looks like it will be a breaking change.

test/python/setuptools.test.ts
mrgrain, line 1: What we normally do here is having two tests:
- one for setting optional properties
- one for asserting the defaults if optional properties are unset

mrgrain (changes requested):
Sorry, this is going to be super weird. Because `license` is defined here: https://github.com/projen/projen/blob/main/src/python/python-packaging.ts#L48

We need to add a `@default` but then this also needs to change for all usages of `PythonPackagingOptions`, not just poetry.

mrgrain (changes requested):
Code looks good. Will holding this back to group with at least one other Breaking Change.

I'm also still wondering if we have a better transition path here. This feels high risk to unintentionally license a bunch of packages under Apache-2.0

mrgrain:
@garysassano Been thinking more about this. We will at least have to add a warning about undefined license defaulting to Apache2.0. The warning can be removed with GA (make sure to add a `@todo` with respective comment).

mrgrain (changes requested):
We need a warning for this.

Issue #3384, opened 2024-02-22
windows: running tasks for projen development on Windows


gmeligio:
I want to able to develop `projen` on Windows. Being able to use projects that depend on projen on Windows would be even nicer.

I'm currently focusing on `projen` development and not `projen` usage as a tool/library. It's possible many tasks related to projen usage will be fixed too, but that could be analyzed in other issues (like #2427, for example). I'm contributing to this by creating a PR per projen task. The current tasks I'm paying attention to so far and relevant issues that were raised from those PRs are:
- [x] Windows Github workflow to track progress https://github.com/projen/projen/pull/3520
- [x] `projenrc` root folder for projen configuration logic https://github.com/projen/projen/pull/3520
- [x] bootstrapping https://github.com/projen/projen/pull/3377
- [x] pre-compile 
- [x] compile
- [x] post-compile https://github.com/cdklabs/jsii-docgen/pull/1398
- [x] test https://github.com/projen/projen/pull/3578  https://github.com/projen/projen/pull/3582
- [x] package https://github.com/projen/projen/issues/1933 https://github.com/projen/projen/pull/3723
- [x] package:js
- [x] package:python
- [x] package:java
- [x] package:go

General issues:
- [x] https://github.com/projen/projen/issues/3284

mrgrain:
Note that currently no tests are run on windows because of this: https://github.com/cdklabs/jsii-docgen/pull/1398/files#diff-bda22a3de7550efc76dd29e2a427a8d5abd5863f6e96a339ff47390d9bed7d14R51-R53

gmeligio:
Wonderful! Looking back, I think it would have been tricky for me to find that the issue was in the `jsii-docgen` repository.

I'll look into the glob pattern for tests later this week. Thank you for pointing out the reason.

mrgrain:
> I'll look into the glob pattern for tests later this week. Thank you for pointing out the reason.

I think jest used [micromatch](https://www.npmjs.com/package/micromatch) for these patterns, so we'll have to rewrite the default patterns to work on windows.

gmeligio:
Closing, since all projen tasks can be run on Windows now 🎉 , given that the required system packages used by the tasks are installed.

PR #3399, opened 2024-02-25
chore: change projen entrypoint to typescript

gmeligio:

## Motivation

I wanted to add the GitHub workflow for Windows in `projen`. That would be in a `projenrc` folder to not clutter the `.projenrc.js`

I thought about using the `projenrc` root folder based on this PR's comment: https://github.com/projen/projen/pull/3377#discussion_r1496454238 . The `Projenrc` component already has optionsfor that folder that could be used for what's needed for this CI workflow. This PR is the first step to use that `Projenrc` component and change the entrypoint `.projenrc.ts` to TypeScript.

I analyzed converting the bootstrap script `projen.js` to TypeScript too but didn't find a way to solve the hashbang line. So, for now, that file is kept as JavaScript.

## Changelog in .projenrc.ts

1. Change entrypoint from `.projenrc.js` to `.projenrc.ts`.
1. Add `projenrc: true`.
1. bootstrap functions moved back to `.projenrc.ts` as `TextFile.lines` because `function.toString()` transpiles with TypeScript in an undesirable way. 
1. Delete the field `compileBeforeTest` that doesn't exist anymore.
1. Check if tasks are undefined.
1. Replace `authorUrl` with `authorAddress`. `authorUrl` is mentioned as deprecated in https://github.com/projen/projen/blob/cc06084d2990d91e7dd9f9b14b8caf4b9d38684e/src/cdk/jsii-project.ts#L590-L594
1. Remove `templates` inclusion from gitignore because that directory doesn't exist.

Tests:

1. No tests were added.

## Changelog for marker in other files

1. Marker auto-updated in all needed files.

## Migrations needed.

1. `yarn install` needs to be run before any other task because of the added requirement of `ts-node`.

---


mrgrain:
Looks good. I'll just need some time to play with this myself.

mrgrain:
Thanks @gmeligio This is awesome! I think one change I'd suggest is to extend the bootstrap file to make sure ts-node is installed. It's an edge case, but easy enough to cover.

Something like this maybe:
```ts
function bootstrap() {
  console.info("bootstrapping...");

  if (!existsSync("node_modules/.bin/jsii") || !existsSync("node_modules/.bin/ts-node")) {
    execCommand(
      "yarn install --frozen-lockfile --check-files --non-interactive"
    );
  }

  if (!existsSync("lib/cli/index.js")) {
    execCommand(
      "npx jsii --silence-warnings=reserved-word --no-fix-peer-dependencies"
    );
  }
}

bootstrap();
const args = process.argv.slice(2).join(" ");
execCommand(`node bin/projen ${args}`);
```

mrgrain:
I went ahead and made the change to the boostrap script. Let me know what you think. Otherwise I'm happy to ship this.

gmeligio:
Looks great! Thank you for the additions. I like the changes.

mrgrain (approved):

PR #3405, opened 2024-02-27
fix(python): generated `pyproject.toml` has correct formatting style when using Poetry

garysassano:
This PR fixes the following issues:
- TOML sections indentation (.e.g `[tool.poetry.dependencies]` section)
- Spaces surrounding string values within square brackets (e.g. `[ "poetry-core" ]`)

#### pyproject.toml (before)
```toml
# ~~ Generated by projen. To modify, edit .projenrc.py and run "npx projen".

[tool.poetry]
name = "example"
version = "0.1.0"
description = ""
authors = [
  "First Last <email@example.com>"
]
readme = "README.md"

  [tool.poetry.dependencies]
  aws-cdk-lib = "^2.130.0"
  constructs = "^10.0.5"
  python = "^3.8"

[tool.poetry.group.dev.dependencies]
projen = "0.80.1"
pytest = "7.4.3"

[build-system]
requires = [ "poetry-core" ]
build-backend = "poetry.core.masonry.api"
```

#### pyproject.toml (after)
```toml
# ~~ Generated by projen. To modify, edit .projenrc.py and run "npx projen".

[tool.poetry]
name = "example"
version = "0.1.0"
description = ""
authors = [
  "First Last <email@example.com>"
]
readme = "README.md"

[tool.poetry.dependencies]
aws-cdk-lib = "^2.130.0"
constructs = "^10.0.5"
python = "^3.8"

[tool.poetry.group.dev.dependencies]
projen = "0.80.1"
pytest = "7.4.3"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

---


mrgrain (changes requested):
We are not going to bundle 2 TOML libraries. This will have to get rid of the previous one as well.

garysassano:
@mrgrain The `json2toml` module just does a one-way conversion from JSON to TOML, it's not able to parse TOML and convert it to a JS Object, which is used in the latest test I've added.

garysassano:
The `json2toml` module just does a one-way conversion from JSON to TOML, it's not able to parse TOML and convert it to a JS Object, which is used in the latest test I've added.

garysassano:
The `json2toml` module just does a one-way conversion from JSON to TOML, it's not able to parse TOML and convert it to a JS Object, which is done in the latest test I've added.

garysassano:
The `json2toml` module just does a one-way conversion from JSON to TOML, it's not able to parse TOML and convert it to a JS Object, which is used to check the generated `pyproject.toml` in several tests.

mrgrain:
> The `json2toml` module just does a one-way conversion from JSON to TOML, it's not able to parse TOML and convert it to a JS Object, which is used to check the generated `pyproject.toml` in several tests.

We can move `@iarna/toml` to be a unbundled test dependency then, right?

test/python/poetry.test.ts
```diff
@@ -157,30 +157,32 @@ test("poetry enabled with metadata in dependencies", () => {
...
-  expect(snapshot["pyproject.toml"]).toContain('version = "^3.3.3"');
-  expect(snapshot["pyproject.toml"]).toContain(
-    'extras = [ "mypackage-extra" ]'
-  );
-  // Likewise package2 metadata should be rendered
-  expect(snapshot["pyproject.toml"]).toContain(
-    "[tool.poetry.dependencies.package2]"
-  );
-  expect(snapshot["pyproject.toml"]).toContain('path = "../mypackage/foo"');
-  expect(snapshot["pyproject.toml"]).toMatchSnapshot();
```
mrgrain, line 183: Can we keep the snapshot please?

These help maintainers understanding changes better.

test/python/poetry.test.ts
```diff
@@ -157,30 +157,32 @@ test("poetry enabled with metadata in dependencies", () => {
     classifiers: ["Development Status :: 4 - Beta"],
     deps: [
       "regular-version-package@1.2.3",
-      'package1@{version = "^3.3.3", extras = ["mypackage-extra"]}',
-      'package2@{ path = "../mypackage/foo" }',
+      `package1@{ version = "^3.3.3", extras = ["mypackage-extra"] }`,
+      `package2@{ path = "../mypackage/foo" }`,
```
mrgrain, line 161: I don't mind this, but this seems like an entirely unnecessary change (both the spaces and the change in quotes). In future please try to not make these changes. They add no value and only create noise and work for reviewers.

mrgrain (changes requested):
One last thing: Let's keep the snapshot.

mrgrain (approved):

automartin5000:
This one's been annoying me for a while, thanks @garysassano!

agdimech:
This PR is causing issues as it doesn't cater for all cases (comma splitting is to blame) https://github.com/projen/projen/issues/3426

mrgrain:
Hi all, this was just reverted. Apologies for accepting this PR in the first place. I had missed that the new method `permitTomlInlineTableDeps` is effectively attempting to implement TOML parsing.

Let's be very clear that going forward, we are only going to accept a contribution that (in it's essence) only swaps out the calls to `TOML.stringify()` and `TOML.parse()` with a different version from a well regarded library.

I appreciate that this might be a very high bar, given that this library also has to match our licensing requirements.

Issue #3284, opened 2024-01-18
cross-platform: use forward slash as directory separator for all platforms


gmeligio:
## Feature request

Currently `path.join` is used to compose paths from parts wherever is possible. This is the function in NodeJS commonly used for composing paths. In some places, where path.join doesn't fit the use case,
`path.posix.join` is being used to solve it. This issue appears only in Windows because `path.join` is platform-specific. I propose to always use `path.posix.join` that will use forward slash and is supported in all platforms.

Another example is for a project being used both in Windows and Linux. Some files are updated when changing platforms in `.projen/files.json` because of the use of `path.join`. This specific case can be solved in an MR changing the `JsonFile` path with `path.posix.join`, but it's related to the proposal.
![image](https://github.com/projen/projen/assets/22875166/eba253a9-af15-44c4-8f73-84af9ff23ab2)

We could avoid having to use first `path.join` and then change it to `path.posix.join` when it fails after changing platforms, if we use `path.posix.join` since the beginning. This will make the directory separator cross-platform whenever is used.

Windows supports both back and forward slashes as directory separator, [as seen in .NET docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-8.0#examples)

The changes could be easy or hard depending on figuring out if there are breaking changes. I haven't figured out the breaking changes.

I can contribute to the implementation.

## Proposed Implementation

The change can be done by replacing "path" by "path/posix" in all `src/*/**.ts` like
```typescript
import * as path from "path/posix"
import { join } from "path/posix";
```

Another refactors can be done for DevEx, so `posix` is not imported from "path" but only the import "path/posix" is used. Then the functions will be used like `path.join` instead of `path.posix.join`.  Allowing to use a cross-platform separator with the function  `path.join` developers commonly use.

## Alternative implementation

Continue refactoring on a per use case basis, like https://github.com/projen/projen/issues/2344 and in my case will be to refactor `task-runtime.ts` and `dependencies.ts` to use "path/posix".  

Preventing the usage of "path" and require "path/posix" with some mechanism like an Eslint rule.

## Migration Breaking Changes

This _migration_ could have breaking changes whenever backslash (`\`) is used currently. I haven't figured out how functions aside from `join` are affected by using the POSIX-variant with "path/posix". I tried to collect the numbers of the path functions being used in projen codebase. Roughly this was the usage I gathered:
- join: 74
- dirname: 30
- resolve: 21
- basename: 13
[...]

PR #3406, opened 2024-02-27
fix(node): some directory separators are still inconsistent between platforms in tasks.json and package.json

gmeligio:
Related to https://github.com/projen/projen/issues/3284#issuecomment-1964163891

This is the second attempt at fixing the directory separator issue for lambda entrypoints, which was started in https://github.com/projen/projen/pull/3387. 

I tried adding unit tests for `renderBundleName`. Unfortunately, this can't be unit-tested now because NodeJS exports the `path` package for the Linux platform instead of Windows at import time and not at runtime.
https://github.com/nodejs/node/blob/f28ccd3941e70a825786eff7e28f59014ac76ac9/lib/path.js#L1546

I tested it locally in an example project, and it looks like it's fixed now.
image

---

### DIFF
--- a/src/javascript/util.ts
+++ b/src/javascript/util.ts
@@ -27,7 +27,7 @@ export function renderBundleName(entrypoint: string) {
     parts.shift(); // just remove 'src' if its the first element for ergonomics
   }
 
-  const p = parts.join(sep);
+  const p = parts.join(posix.sep);
   const dir = dirname(p);
   const base = basename(p, extname(p));
   return posix.join(dir, base);


mrgrain:
@gmeligio I assume this would be tested automatically once we start running builds on Windows?

mrgrain:
@gmeligio For some reason this broke the release: https://github.com/projen/projen/actions/runs/8079532064/job/22074128464#step:6:1035

Are you around to fix this forward? Otherwise I'll just revert the PR and we can try again.

gmeligio:
Sorry. Haven't had the time to look at it. I'll look into it in an hour or so. Go ahead and revert if necessary.

mrgrain:
No worries, turns out it got fixed via #3409

PR #3409, opened 2024-02-28
fix(tsconfig): support three missing strict tsconfig `compilerOptions`

giseburt:
This is adding support for `compilerOptions` used [in the wild](https://github.com/tsconfig/bases/blob/main/bases/strictest.json) and escape hatches had to be used.

This was found when applying [@tsconfig/strictest](https://github.com/tsconfig/bases?tab=readme-ov-file#strictest-tsconfigjson) and working around TypeStrong/ts-node#1958 (merged, but not yet released, a year later)

No issue was created, and no tests were added. This is merely a pass-thru to tsconfig (like the existing options above them in the code).

---


giseburt:
@mrgrain No problem. I found the source of the issue and made an issue for it: #3410

Should be a single-line PR to fix it.

Issue #3410, opened 2024-02-28
PR #3397 somehow merged without updated snapshots (this repo only)


giseburt:
Note: This is about _this_ project specifically, not configuring other projects.

## Theory: Default `jestConfig.updateSnapshot` allowing test failures to pass silently

In the `.projenrc.js` of the `projen` project itself, `jestOptions.jestConfig.updateSnapshot` is not specified, and due to historical reasons (it appears, looking at closed issues) this defaults to `ALWAYS`.

What happens is that any tests that **should** fail because the snapshot changed doesn't, instead the snapshot is silently "fixed".

This causes a number of bad things at once:

- The test passed, when it shouldn't have, masking the issue the test it supposed to catch
- We don't see _which_ issue caused the snapshot mismatch (it can be determined, but with more effort than it should be)

For this reason, I'm suggesting we change the `.projenrc.js` file *of this repo* to:

```js
const project = new cdk.JsiiProject({
  // ...
  jestOptions: {
    // ...
    jestConfig: {
      // ...
      updateSnapshot: NEVER, // ← add this
    },
  },
```

There is a mechanism in place to detect if file changes would occur such as when a snapshot changes. This *helps* but recently allowed a code change that should have caused a test failure (in a PR that I submitted, #3397) that ended up causing the build to fail in another PR I submitted ([here](https://github.com/projen/projen/pull/3409/commits/68d67dfda28755b0c3554d6d3c820d026879953e)). I'm not sure why it didn't catch it, but more importantly it should have been caught on the test build of the first PR and adjusted there.

mrgrain:
I think the issue is more subtle than that.

This PR introduced the snapshots: https://github.com/projen/projen/pull/3406
This is the build that got the PR merged: https://github.com/projen/projen/actions/runs/8079427878/job/22073814706#step:5:1014

As you can see, there is no mention of updated snapshots.

---

I suspect we have test cases that install the latest version of projen, rather than using the locally build version.

gmeligio:
I tried to narrow down when the snapshot got outdated by checking out older commits from the main branch, and I could reproduce it in all of them before the fix without finding the first commit. I deleted the repo multiple times to check there were no caching issues and still got the same result. The file `test/__snapshots__/new.test.ts.snap` is modified during all my tests for commits older than https://github.com/projen/projen/commit/75072c1ed699dd0e02ecd0627e72c4ff907c9728.

What I did was:
1. git checkout <git-ref-sha>  
[...]

PR #3425, opened 2024-03-06
chore: document --projen-version in projen new help

gmeligio:
## Motivation

After looking for unrelated issue https://github.com/projen/projen/issues/3410#issuecomment-1980700016
, I found that `projen new` supports `--projen-version` but it's not shown in `projen new --help`.

## Changelog

1. Addes the help description
1. Set default as `latest`.
1. Updates a test snapshot.
1. No new tests were added.

---

### DIFF
--- a/src/cli/cmds/new.ts
+++ b/src/cli/cmds/new.ts
@@ -53,6 +53,11 @@ class Command implements yargs.CommandModule {
       default: true,
       desc: "Run `git init` and create an initial commit (use --no-git to disable)",
     });
+    args.option("projen-version", {
+      type: "string",
+      default: "latest",
+      desc: "The version of projen to install",
+    });
     args.example(
       "projen new awscdk-app-ts",
       'Creates a new project of built-in type "awscdk-app-ts"'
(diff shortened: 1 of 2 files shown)


mrgrain:
> What do you mean `projen new` supports `--projen-version`? Where is this coming from? How is the option currently used?

Ah I see it now. It's already documented, but only on the project level because that's where it's coming from:

```console
pj new typescript --help
```
image

gmeligio:
Thanks for checking. Wow! I didn't know all `projen` projects had CLI flags according to the available options.

In this specific case, I refer to the projenVersion in `projen new` itself in the function `initProjectFromModule`. Maybe it is something from the past, and `projenVersion` doesn't belong anymore to that command but to the specific projects.
https://github.com/projen/projen/blob/2fc2b406b4e6013e60b544491ecdc35ab95ea723/src/cli/cmds/new.ts#L286-L300

Would it make sense to document it in `projen new --help` in case is still used?

gmeligio:
Thank you for your review. It makes sense not to document this special case. It could even make sense to deprecate it if it's not used.

Regarding the `--help`, it's worth documenting that projen has that CLI interface with all the options. It blew my mind that CLI capability.

PR #3431, opened 2024-03-07
chore(python): improve poetry test for complex dependencies

garysassano:
- **Renamed the Test for Clarity**: Updated the test name to more accurately reflect its purpose and the specific functionality it tests, enhancing understandability.

- **Removed Redundant Test Case**: Excluded a basic dependency test, as it was covered elsewhere and did not align with the focus of these tests, streamlining the test suite to its core objectives.

- **Added Comprehensive Tests**: Included a series of detailed tests sourced from the [Poetry docs](https://python-poetry.org/docs/dependency-specification/#using-environment-markers), covering every conceivable scenario to ensure thorough validation of dependency specifications.

- **Refined Assertion Method**: Replaced `toContain` with `toEqual` for assertions to ensure exact matching of the expected and actual values.

---


mrgrain (changes requested):
Like last time, let's keep the snapshot

mrgrain (approved):

PR #3449, opened 2024-03-13
feat(node): support more build workflow options

timothymathison:
Split out from #3317 since this feature is not strictly required for that feature.

Adds a `buildWorkflowOptions` prop to `NodeProject` which exposes most options available for build workflow.

---


mrgrain (commented):
The thing that's concerning me here is that we now have two ways to add `postBuildSteps` and also `preBuildSteps` (with a different name at top-level). Permissions and triggers might also have some overlap.

Also the Release workflow does a build as well. Should these be set for it as well were appropriate? 🤔

This isn't your fault at all, but the current distinction between release and build is a bit of a mess.

timothymathison:
> The thing that's concerning me here is that we now have two ways to add `postBuildSteps` and also `preBuildSteps` (with a different name at top-level). 

Yes this has been bugging me. The way the `NodeProject` props inherit directly from `Release` props means that the root level `postBuildSteps` prop is defined by `Release` but then gets passed to both `Release` and `BuildWorkflow`. Ideally, I think those post build steps options should actually be separate nested prop objects for each workflow type. But that would be a breaking change.

> Also the Release workflow does a build as well. Should these be set for it as well were appropriate? 🤔

Yeah the root level `postBuildSteps` prop is already passed to `Release`. Or is that not what you're referencing?

> Permissions and triggers might also have some overlap.

I don't think there is any overlap for these. The root `releaseTrigger` prop gets passed to `Release` only. And the new `buildWorkflowOptions.workflowTriggers` option is passed to `BuildWorkflow`. I agree its weird that they're at different levels; but I think it does make sense for them to remain separate options.
For `permissions`, `NodeProject` does not currently provide an option to override them for `Release`. I think we should keep the `permissions` option for `Release` separate from `BuildWorkflow`, but should at some point provide a prop for it.

timothymathison:
I want to limit the scope of this change to `BuildWorkflow` and not change anything that effects `Release`. So I think the main question is whether the `buildWorkflowOptions.workflowTriggers` option should be added or not since we can't stop passing the root `postBuildSteps` option to `BuildWorkflow` (without a breaking change); and its an arbitrary decision which array comes first when we merge the two options. The downside of not adding `buildWorkflowOptions.workflowTriggers` is that that leaves no way to configure `BuildWorkflow` with post build steps distinct from `Release`, since the current option gets passed to both.

timothymathison:
For now I removed `postBuildSteps` from `BuildWorkflowCommonOptions` since I realized there is also a `addPostBuildSteps` method exposed by `project.buildWorkflow`.

mrgrain:
Thanks, this should be unproblematic to merge them.

This is definitely a bit of a conundrum, because it seems like we need these cases for a bunch of options:

- (main use case) Set for both `build` and `release`
- set only for build
- set only for release

mrgrain (approved):

src/javascript/node-project.ts
```diff
@@ -606,15 +634,17 @@ export class NodeProject extends GitHubProject {
           installStepConfiguration: {
             workingDirectory: this.determineInstallWorkingDirectory(),
           },
-          mutable: options.mutableBuild ?? true,
-        }),
+          mutable:
+            buildWorkflowOptions.mutableBuild ?? options.mutableBuild ?? true,
+        }).concat(buildWorkflowOptions.preBuildSteps ?? []),
```
VrajKD, line 639: @timothymathison If we do provide `preBuildSteps` as part of `buildWorkflowOptions`, this concat wouldn't have an impact as the same field (preBuildSteps) gets overridden on L647.

timothymathison: Ah good catch @VrajKD. You're right. This line should have been moved after the object spread on L647. I'll raise a PR to fix.

timothymathison: #3510

Issue #3240, opened 2024-01-05
Eslint 9


mrgrain:
Eslint 9 release is looming and it comes with a whole bunch of breaking changes: https://eslint.org/docs/next/use/migrate-to-9.0.0

The biggest ones from projen's perspective are:
- [New default config format](https://eslint.org/docs/next/use/migrate-to-9.0.0#-new-default-config-format-eslintconfigjs)
- [Deprecation of formatting rules](https://eslint.org/blog/2023/10/deprecating-formatting-rules/)

I think we will need a
- A new component for basic JS config files (see https://github.com/projen/projen/pull/2414 and https://github.com/projen/projen/issues/2405)
- A new component `Eslint9` or `EslintModern`or whatever
  - Should use the new js config file component
  - Require a correct version of eslint
  - Make this component the default unless a `useLegacy` option is set
  - Enable prettier by default for formatting. 
- Deprecate and warn in the old component

rjmackay:
Eslint 8 is now EOL. Any updates / ETA on updating?

mrgrain:
> Eslint 8 is now EOL. Any updates / ETA on updating?

Not really. Looks like Eslint 9 still supports the old config format though. So that might be a reasonable stop-gap.

mrgrain:
Basic Eslint 9 support landed via #3935 however this doesn't cover the new config style yet.

mrgrain:
Some thoughts based on previous attempts:
 
- The size of the added surface to the public API should be kept minmal. I understand the urge to provide something foundational and flexible. Instead I suggest we focus on a very narrow public API for a JavaScript data/config file.
- What are the must have requirements for that? Surely we won't need the whole JS feature set to write an eslint config. We can start by converting the existing default configs and see what features we need.
- Also focus on the DX. I feel it should be closed to how other object files are defined. They have a really straight forward API and still support lazy-resolution via arrow functions. 
- Ensure we include a new documentation page covering  JS data/config files. Wwe don't have a section for this yet, there probably should be one covering all sorts of files.
- The current proposals seems to be primarily adding two features:  a/ strings with tokens, and b/ conversion from JS literals to valid JavaScript syntax as string. If my understanding is correct here, both might be needed but can and should be implemented independently of each other. E.g. tokenized strings are something that should exist in its own without an inherent connection to (JS) code. We might not really need tokenized strings if get the API right though.

dsdavis4:
[...]

PR #3454, opened 2024-03-15
Eslint js config2

### LINKED ISSUE #2405: eslint: option to generate a js file for config

giseburt:
Fixes #2405, #3240, and replaces #2414

Building on the great work of @andrestone in #2405, which appears to be abandoned (it happens), I've been using techniques I've used in other projects and inspired by Tokens in CDK to make the code work.

The interface is highlighted by this code from the tests:

```ts
  // make a dependencies object to track imports
  const dependencies = new JavascriptDependencies();

  // add a few imports
  const [jsdoc] = dependencies.addImport("jsdoc", "eslint-plugin-jsdoc");
  const [js] = dependencies.addImport("js", "@eslint/js");

  // make a data object
  const data = [
    {
      files: ["**/*.js"],
      plugins: {
        jsdoc,
      },
      rules: {
        "jsdoc/require-description": "error",

        // insert a spread operator, value doesn't matter
        [`...${js}.blah`]: true,
        "jsdoc/check-values": "error",

        // insert a second spread operator, value doesn't matter
        [`...(${jsdoc}.fakeTest ? {"fakeTest": "warn"} : {})`]: true,
      },
    },
  ];
  const value = new JavascriptRaw(`${dependencies}

export default ${JavascriptDataStructure.value(data)};
`).resolve();

  expect(value).toEqual(/*shown below*/);

  expect(value).not.toMatch(/\$\{Token\d+\}/);
```

Here's the contents of `value` it creates, extracted so it'll get properly syntax highlighted:
```js
import jsdoc from 'eslint-plugin-jsdoc';
import js from '@eslint/js';

export default [
  {
    files: [
      "**/*.js",
    ],
    plugins: {
      jsdoc: jsdoc,
    },
    rules: {
      "jsdoc/require-description": "error",
      ...js.blah,
      "jsdoc/check-values": "error",
      ...(jsdoc.fakeTest ? {"fakeTest": "warn"} : {}),
    },
  },
];
```

Features of the JS-file formatted:

- Take data structures and format them, with indention
- Uses stringification to Tokens, and replaces the tokens later
- Accept functions as values, and data structures can be modified between after being provided and before `resolve()` is called for late-binding data
  - For example, can include an array of values that aren't known yet, but will be known before synth time
- Can be built upon to add new features - the `JavascriptDependencies` import tracking feature for example is added in just such a way
  - `JavascriptDependencies` can make CJS `require` or ESM `import` statements
- Supports making Functions (both named and arrow)
- Supports all fundamental types including Dates, Objects, and Arrays (the latter serialized with indention, unless empty)
- Supports inserting spread operators in objects, quoting keys in objects (as needed), and raw javascript values in most places (providing spread operators in arrays, or calling functions, for example)

This is still a draft, I have it using the new capability to make an old-style eslint JS config that works, but the eslint module doesn't know how to make a new-style ("flat") eslint config as mentioned in #3240, so that parts making valid JavaScript just not a valid flat configuration that eslint accepts.

I wanted to post this as a draft to get feedback on what I have so far.

PS: This could be easily expanded to make TypeScript code as well, BTW.

---


package.json
```diff
@@ -127,6 +127,14 @@
...
   "main": "lib/index.js",
   "license": "Apache-2.0",
+  "typesVersions": {
+    "<=3.9": {
+      "lib/*": [
+        "lib/.types-compat/ts3.9/*",
+        "lib/.types-compat/ts3.9/*/index.d.ts"
+      ]
+    }
+  },
```
mrgrain, line 137: That means some new code is too complex for ts3.9 
I'd prefer to not have that. Can you check the file and see what is incompatible?

src/code-resolvable.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  abstract stringify(level: number, idt: string): string;
+
+  resolve(context?: ICodeResolutionContext): string {
+    const idt: string = context?.idt ?? " ".repeat(context?.indentation ?? 2);
+    let value = this.stringify(0, idt) ?? "";
+    // let count = 0;
+    // while (!CodeToken.isResolved(value) && count++ < 10) {
+    //   value = CodeToken.resolve(value, 0, idt);
+    // }
```
mrgrain, line 49: ?

src/javascript/eslint.ts
```diff
@@ -1,12 +1,60 @@
...
+   * @see https://eslint.org/docs/latest/use/configure/configuration-files-new
+   */
+  JAVASCRIPT_FLAT_CJS = "flat-cjs",
+
+  /**
+   * JavaScript file (new flat format) - using CJS-style require/module.exports
+   *
+   * @see https://eslint.org/docs/latest/use/configure/configuration-files-new
+   */
+  JAVASCRIPT_OLD_CJS = "old-cjs",
```
mrgrain, line 36: `OLD` is always a bad name. It just get's outdated. Eslint seems to call this "eslintrc format".

src/javascript/eslint.ts
```diff
@@ -1,12 +1,60 @@
...
+
+  /**
+   * JSON file
+   *
+   * @deprecated ESLINT project is transitioning away from this format, use `JAVASCRIPT_FLAT` instead
+   * @see https://eslint.org/docs/latest/use/configure/configuration-files
+   */
+  JSON = "json",
```
mrgrain, line 44: prefix with `ESLINTRC_JSON`

src/javascript/eslint.ts
```diff
@@ -1,12 +1,60 @@
...
+
+  /**
+   * YAML file
+   *
+   * @deprecated ESLINT project is transitioning away from this format, use `JAVASCRIPT_FLAT` instead
+   * @see https://eslint.org/docs/latest/use/configure/configuration-files
+   */
+  YAML = "yaml",
```
mrgrain, line 52: prefix with `ESLINTRC_YAML`

src/javascript/eslint.ts
```diff
@@ -1,12 +1,60 @@
...
+ * What format should the eslint file be.
+ *
+ */
+export enum EslintConfigFileFormat {
+  /**
+   * JavaScript file (new flat format) - using ESM-style imports/exports
+   *
+   * @see https://eslint.org/docs/latest/use/configure/configuration-files-new
+   */
+  JAVASCRIPT_FLAT_ESM = "flat-esm",
```
mrgrain, line 22: I wonder if the format type is more important:

Enums should have the same string value as the enum:
```suggestion
  FLAT_JAVASCRIPT_ESM = "FLAT_JAVASCRIPT_ESM", 
```

src/javascript/eslint.ts
```diff
@@ -16,8 +64,12 @@ export interface EslintOptions {
 
   /**
    * Files or glob patterns or directories with source files to lint (e.g. [ "src" ])
+   *
+   * @remarks
+   * This is actually **required**, but marked as optional so upstream projects can accept this interface
+   * and provide this value.
```
mrgrain, line 70: Upstream projects need to solve this themselves, sorry.

src/javascript/eslint.ts
```diff
@@ -16,8 +64,12 @@ export interface EslintOptions {
...
    */
-  readonly dirs: string[];
+  readonly dirs?: string[];
```
mrgrain, line 72:
```suggestion
  readonly dirs: string[];
```

src/javascript/eslint.ts
```diff
@@ -359,7 +432,7 @@ export class Eslint extends Component {
       "*.d.ts",
       "node_modules/",
       "*.generated.ts",
-      "coverage",
+      "./coverage",
```
mrgrain, line 435: not this?
```suggestion
      "coverage/",
```

src/javascript/eslint.ts
giseburt: I believe I saw them calling this the legacy format. I’m good either way. I agree “old” is bad naming.

And that answers my unasked question about deprecation of the others. 😄

package.json
giseburt: This came in after a rebase onto new main. I am unsure where it came from. I’ll look and see what’s up.

src/javascript/eslint.ts
```diff
@@ -405,13 +478,135 @@ export class Eslint extends Component {
...
+          marker: true,
+          allowComments: true,
+          cjs: true,
+        });
+      } else {
+        const ext =
+          format === EslintConfigFileFormat.JAVASCRIPT_FLAT_ESM ? "mjs" : "cjs";
+        configFileName = `eslint.config.${ext}`;
+
+        this._javascript = new JavascriptFile(project, configFileName, {
```
mrgrain, line 521: Can we encapsulate this into a new EslintConfigFile Component?

src/javascript/eslint.ts
```diff
@@ -515,14 +710,175 @@ export class Eslint extends Component {
...
+
   /**
    * Update the task with the current list of lint patterns and file extensions
    */
   private updateTask() {
     const taskExecCommand = "eslint";
     const argsSet = new Set<string>();
     if (this._fileExtensions.size > 0) {
-      argsSet.add(`--ext ${[...this._fileExtensions].join(",")}`);
+      if (
```
mrgrain, line 871: this feels like it's in the wrong place

test/javascript/eslint.test.ts
```diff
@@ -245,3 +253,69 @@ test("allow modification of the eslint task", () => {
...
+
+    // THEN
+    expect(output[eslint.fileName]).toMatchSnapshot();
+  }
+);
+
+test.each([
+  // EslintConfigFileFormat.JSON,
+  // EslintConfigFileFormat.YAML,
+  // EslintConfigFileFormat.JAVASCRIPT_OLD_CJS,
```
mrgrain, line 292: ?

src/javascript-file.ts
mrgrain, line 1: This file is missing docblocks

src/code-resolvable.ts
giseburt: I’ll clean this up.

In a previous version of stringing it was practical for nested tokens to not get processed, and needed reprocessed. This is no longer the case.

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+
+export class JavascriptFunction extends CodeResolvableBase {
+  static named(
+    name: string | undefined,
+    properties: Array<unknown>,
+    body: unknown
+  ) {
+    return new JavascriptFunction(name, properties, body);
+  }
+  static arrow(properties: Array<unknown>, body: unknown) {
```
mrgrain, line 15: Why is body unknown in type?

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+    return new JavascriptFunction(undefined, properties, body);
+  }
+  constructor(
+    private readonly name: string | undefined,
+    private readonly properties: Array<unknown>,
+    private readonly body: unknown
```
mrgrain, line 21: Why are these unknown in type?

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+  ): Array<JavascriptRaw> {
+    if (typeof imports === "string") {
+      if (this.defaultImports.has(from)) {
+        if (this.defaultImports.get(from) === imports) {
+          return [JavascriptRaw.value(imports)];
+        }
+        throw new Error(
+          `Default import already exists for ${from}: ${this.defaultImports.get(
+            from
+          )} !== ${imports}`
```
mrgrain, line 107: Instead of hard error, could we warn and re-write the code so that it still works? Not sure if that's feasible though.

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+    if (level !== 0) {
+      throw new Error("JavascriptDependencies cannot be nested");
+    }
+    const out: Array<string> = [];
+    for (const from of this.froms) {
+      const imports = this.imports.get(from) ?? [];
+      const defaultImport = this.defaultImports.get(from);
+      const value = imports.map((i) => i.toString()).join(", ");
+      if (defaultImport && imports.length > 0) {
+        if (this.cjs) {
```
mrgrain, line 132: I wonder if this class should have two flavours: `JavascriptESMDependencies` and `JavascriptCJSDependencies` instead of inlining a lot of if statements

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+export class JavascriptFile extends JsonFile {
+  public dependencies: JavascriptDependencies;
+  cjs: boolean;
+  constructor(
+    project: Project,
+    filePath: string,
+    options: JavascriptFileOptions
+  ) {
+    super(project, filePath, { ...options, allowComments: false });
+    this.cjs = options.cjs ?? false;
```
mrgrain, line 176: Same here, instead of a boolean flag would it be nicer to have two classes?

package.json
giseburt: I was wrong. I was able to diff the two output folders in lib, and determined it was the imports in `code-token-map.ts` - I thought I'd use `type` imports, but apparently that triggered this, so I remored the `type` part, and now `jsii` has stopped adding that to the file.

src/javascript-file.ts
giseburt: Oops, I'll add that.

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
+import { CodeResolvableBase } from "./code-resolvable";
+import { CodeTokenMap, unresolved } from "./code-token-map";
+import { IResolver } from "./file";
+import { JsonFile, JsonFileOptions } from "./json";
+import { Project } from "./project";
+
+export class JavascriptFunction extends CodeResolvableBase {
```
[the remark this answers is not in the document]
giseburt: :+1:

src/javascript-file.ts
giseburt: `unknown` will give errors if we try to use the value without inspecting the contents first. If we use `any` then is disabled typescript checks altogether. In this case, the value is passed to `javascriptStringify` which also take an `unknown` value, which forced us to validate any time we use it as before using it.

I only use `any` as a last resort, and usually will fail a PR I'm reviewing if I see one without a lengthy explanation of why it's necessary.

src/javascript-file.ts
giseburt: If the user only uses the retuened token and doesn't form their own, it'll work. I'll add a warning to that affect.

src/javascript-file.ts
giseburt: I'll split it to three classes, `JavascriptDependenciesBase` (abstract), `CJSJavascriptDependencies` and `ESMJavascriptDependencies`. All the common logic will be in the base.

src/javascript-file.ts
giseburt: I think it would be a lot of duplicated code for very little gain. All the logic remains the same, just eh little bits of syntax output change.

test/javascript/eslint.test.ts
giseburt: Oops. That was me using the testing for development, and I forgot to uncomment them. These tests are terribly slow, since they build a complete project to test actual eslint running properly.

src/javascript/eslint.ts
giseburt: Sorry, @mrgrain , I tried. I have spent a large amount of time on this particular comment. It's much harder than either of us anticipated, I suspect. At least to do it cleanly.

It boils down to a matter of backward compatibility. Even if I assume that the escape hatches don't have an expectation of backward compatibility, I still have issues where we have to decide which interface to expose, the flat one or the legacy one.

I tried to accept both. There's too many gaps to fill. It's _possible_, I'm sure, but not practical and I'm fairly certain not worth the effort at this point.

I'm suggesting we leave this one, leave the code where it is now. Somewhat aggressively depreciate anything that is clearly a direct control of the old config layout, and introduce new controls to replace them. Once we can safely remove the deprecated calls, we can then easily split out a EslintConfigFile Component.

I have a few more ESLint-related PRs I'd like to open, so this can evolve a little as we go.

Is that acceptable?

mrgrain: Yeah, in principle this would be acceptable. But I'm not sure I understand the situation clearly:
 
> It boils down to a matter of backward compatibility. 

This would be new component. What is the Backwards compatibility concern here? I think I general you might have been trying too hard to add this feature to the existing Eslint component. Where in practice the path that I have outlined in #3240 (new component first, then switch over usages of it) appears to be more straight forward.

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+ */
+export interface JavascriptFileOptions
+  extends JavascriptDependenciesOptions,
+    JsonFileOptions {}
+
+/**
+ * Represents a JS configuratin file (e.g. .eslintrc.js).
+ * Suppor
+ */
+export class JavascriptFile extends JsonFile {
```
mrgrain, line 446: Can we get rid of the dependency on `JsonFile`. It doesn't make much sense semantically. What features we actually gaining from it?

```suggestion
export class JavascriptFile extends FileBase {
```

src/javascript-file.ts
```diff
@@ -0,0 +1,288 @@
...
+      // 1. `key: value,`
+      // 2. `"key": value`
+      // 3. `[key]: "value"`
+      // 4. `...key` (value is ignored)
+
+      const value = javascriptStringify(val, level + 1, idt);
+      let keyString = key;
+      // if the key is a token, resolve it (3,4)
+      if (unresolved(key)) {
+        // console.log("key", key);
```
mrgrain, line 543:
```suggestion
```

mrgrain:
We should probably get the JavaScript Code stuff in first without the eslint part.

cameroncf:
> We should probably get the JavaScript Code stuff in first without the eslint part.

I came here looking for eslint9 support after doing down a similar path. There are other needs that could be solved with a standalone `javascript-file.ts` component as well. For example, vite configs want to be in `vite.config.js` files. Merging in that limited change first would allow the js file component to be vetted against other downstream work in addition to the eslint work.

Issue #3462, opened 2024-03-18
bug(upgrade-depdendencies): some dependencies are still attempted to be upgraded by ncu, then reverted by projen


dkershner6:
The bug is that a dependency `some-dep@10.0.0` is currently still included in the `ncu` upgrade command. This might cause the dependency to be upgraded to version `10.1.0`. However when projen is synth'ed the next time, this dependency version is reverted to `10.0.0` again. Therefore the ncu upgrade was unnecessary. Instead drop this dependency from the `ncu` command and use the package manager upgrade command to ensure the latest version matching the version constraint is installed.

---

Original title: bug(upgrade-depdendencies): filter for pinned packages is backwards (it includes pinned packages rather than excludes)

Example: https://github.com/projen/projen/blob/8c14559c3ebc5bf58035d9a3e811fa9813bf4d24/.projen/tasks.json#L425

`comment-json` should not be in this list, but `yaml` should be (backwards): https://github.com/projen/projen/blob/8c14559c3ebc5bf58035d9a3e811fa9813bf4d24/.projenrc.ts#L62

Here is the commit that broke it:
https://github.com/projen/projen/commit/cb2d8987439925efc81874f964b1bc03810c9c23

> When `some-dep@^10` is added as a dependency, it is incorrectly included in the list of to-be-upgraded dependencies. This was due to a bug in a boolean expression; with the the code comment stating the clear intention to exclude them.
>
> The reason for this is that ncu would upgrade `some-dep@^10` to a later version. But `projen default` would then revert back to the version defined in code.
>
> While fixing the issue, I realized that while we need to exclude these deps from ncu, we still want them to be included in the package manager upgrade command, so that they receive the latest version for the given version constraint.

I think this comment is backwards too.  `some-dep@^10` SHOULD be upgraded, but `some-dep@10.1.5` should not.

On a side note, I am also not sure that a `^` is sufficient...as `@10` feels like it should also be upgraded (as though it had a `^`), but as least that case can be handled by adding one.

dkershner6:
@mrgrain , happy to PR a fix, but I want to be sure I am reading this correctly first and not missing something.

dkershner6:
I did end up creating a PR, since it was relatively simple and I had the context: https://github.com/projen/projen/pull/3463

mrgrain:
Sounds like we still haven't got this nailed down. The general rule is:

[...]

PR #3463, opened 2024-03-18
fix(upgrade-dependencies): some dependencies are still attempted to be upgraded by ncu, then reverted by projen

dkershner6:
Fixes https://github.com/projen/projen/issues/3462

---


test/javascript/upgrade-dependencies.test.ts
```diff
@@ -128,9 +158,15 @@ test("ncu upgrade command does not include dependencies with version constraints
 
 test("ncu upgrade command is not added if no ncu upgrades are needed", () => {
   const project = createProject({
-    deps: ["some-dep@^10"],
```
mrgrain, line 131: I don't understand why this test needs changing

mrgrain (changes requested):

test/javascript/upgrade-dependencies.test.ts
dkershner6: I added a more comprehensive function to handles this determination as well as explanatory comments and JSDoc.

dkershner6:
I gave `allowMinorUpgrades` a shot, @mrgrain , let me know your thoughts.

src/javascript/upgrade-dependencies.ts
```diff
@@ -388,6 +388,23 @@ export class UpgradeDependencies extends Component {
...
+    }
+
+    // Caret means "allow minor upgrades"
+    if (version.startsWith("^")) {
+      return true;
+    }
+
+    // If only one part of semver is specified, it means "allow minor upgrades", and "x" is a wildcard.
+    // 10, 10.x, and 10.x.x would all work. 10.1 would not.
+    const parts = version.split(".").filter((ver) => ver !== "x");
```
mrgrain, line 412:
```suggestion
    const parts = version.split(".").filter((ver) => ver !== "x" && ver !== "*");
```

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -96,6 +96,33 @@ test("upgrades command includes all dependencies", () => {
   `);
 });
 
+test("upgrade command does not include non-standard dependency versions", () => {
```
mrgrain, line 99: Can you make this a bit more descriptive? I don't understand what the test non-standard dependency versions are

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -108,7 +135,37 @@ test("ncu upgrade command does not include dependencies with version constraints
...
+        "exec": "npx projen",
+      },
+      {
+        "spawn": "post-upgrade",
+      },
+    ]
+  `);
+});
+
+test("ncu upgrade command should include dependencies with non-pinned versions, along with package manager upgrade", () => {
```
mrgrain, line 156: This test is the opposite of what we are trying to achieve here.

`some-dep` has a version constraint define. Therefore projen will ALWAYS write the version constraint back to `package.json`, therefor running `ncu` to update the version constraint in `package.json` is pointless because projen will ALWAYS override it again. However `yarn upgrade` et. al. will make sure that the latest version according to the constraint is installed.

mrgrain (changes requested):
Something still isn't quite right. Appreciate the issue has changed a bit and a I'm not sure if you have updated the title and code yet in all places. But I'm really missing a simple test case for the problem you are trying to solve.

dkershner6:
Alright, @mrgrain , I think we are on the same page now. Please confirm.

src/javascript/upgrade-dependencies.ts
```diff
@@ -372,8 +372,11 @@ export class UpgradeDependencies extends Component {
     const dependencies = [];
 
     const deps = this.project.deps.all
-      // remove those that have a pinned version
-      .filter((d) => includeConstraint || !(d.version && d.version[0] === "^"))
+      // remove those that have a pinned version (unless includeConstraint is true)
```
mrgrain, line 375:
```suggestion
      // remove those that have a constraint version (unless includeConstraint is true)
```

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -96,9 +96,9 @@ test("upgrades command includes all dependencies", () => {
   `);
 });
 
-test("ncu upgrade command does not include dependencies with version constraints, but package manager upgrade does", () => {
+test("ncu upgrade command does not include dependencies with any version constraint, but package manager upgrade does", () => {
   const project = createProject({
-    deps: ["some-dep@^10"],
+    deps: ["some-dep@10.0.0"],
```
mrgrain, line 101:
```suggestion
    deps: ["some-dep@^10", "other-dep@10.0.0"],
```

test/javascript/upgrade-dependencies.test.ts
```diff
@@ -126,11 +126,47 @@ test("ncu upgrade command does not include dependencies with version constraints
...
+        "spawn": "post-upgrade",
+      },
+    ]
+  `);
+});
+
 test("ncu upgrade command is not added if no ncu upgrades are needed", () => {
   const project = createProject({
-    deps: ["some-dep@^10"],
+    deps: ["some-dep@10.0.0"],
```
mrgrain, line 161: Same as above

```suggestion
    deps: ["some-dep@^10", "other-dep@10.0.0"],
```

src/javascript/upgrade-dependencies.ts
```diff
@@ -388,6 +391,21 @@ export class UpgradeDependencies extends Component {
...
 
+  /**
+   * Projen only uses npm-check-updates to upgrade dependencies when either the version is omitted, or set to "*".
+   * Otherwise, the exact version selected is placed in the package.json file and upgrading is handled through the package manager
+   * rather than npm-check-updates.
+   *
+   * @param version semver from DependencyCoordinates.version, may be undefined
+   * @returns whether the version is the default versioning behavior
+   */
+  private packageIsUsingDefaultVersioning(
```
mrgrain, line 402: Let's give this a more descriptive name
```suggestion
  private isProjenManagedVersion(
```

src/javascript/upgrade-dependencies.ts
```diff
@@ -388,6 +391,21 @@ export class UpgradeDependencies extends Component {
...
+    version: string | undefined
+  ): boolean {
+    // No version means "latest"
+    return !version || version === "*";
+  }
```
mrgrain, line 407: Let's give this a more descriptive name
```suggestion
  private isProjenManagedVersion(
    version: string | undefined
  ): boolean {
    // No version means "latest"
    return version && version !== "*";
  }
```

src/javascript/upgrade-dependencies.ts
```diff
@@ -372,8 +372,11 @@ export class UpgradeDependencies extends Component {
     const dependencies = [];
 
     const deps = this.project.deps.all
-      // remove those that have a pinned version
-      .filter((d) => includeConstraint || !(d.version && d.version[0] === "^"))
+      // remove those that have a pinned version (unless includeConstraint is true)
+      .filter(
+        (d) =>
+          includeConstraint || this.packageIsUsingDefaultVersioning(d.version)
```
mrgrain, line 378:
```suggestion
          includeConstraint || !this.isProjenManagedVersion(d.version)
```

mrgrain (changes requested):
Yes we are ineeded. @dkershner6 Thanks for working through this! Just some minor comments on the naming/tests. Otherwise good to go.

src/javascript/upgrade-dependencies.ts
dkershner6: I think the name doesn't reverse the logic here... "Projen Managed" would be anything where a version is not provided, or IS star...

I'll change the name, but keep the logic.

src/javascript/upgrade-dependencies.ts
dkershner6: See https://github.com/projen/projen/pull/3463#discussion_r1550003129

dkershner6:
@mrgrain , minor changes made. Note that I did correct one of your suggestions regarding the direction of the conditional, but otherwise it is exactly as requested.

src/javascript/upgrade-dependencies.ts
```diff
@@ -388,6 +390,19 @@ export class UpgradeDependencies extends Component {
...
 
+  /**
+   * Projen fully manages a package's version when either the version is omitted, or set to "*".
+   * Otherwise, the exact version selected is placed in the package.json file and upgrading is handled through the package manager
+   * rather than npm-check-updates.
+   *
+   * @param version semver from DependencyCoordinates.version, may be undefined
+   * @returns whether the version is the default versioning behavior
+   */
+  private isProjenManagedVersion(version: string | undefined): boolean {
```
mrgrain, line 401: Sorry to be pedantic here, but it's exactly the other way around. Projen only manges the version if it's `undefined` and not `*`.

Hence my original suggestion.

mrgrain (changes requested):

src/javascript/upgrade-dependencies.ts
dkershner6: > Projen only manges the version if it's undefined and not *.

Just to be clear, you are saying that `*` is a version constraint like any other (`^1.0.0`), and that projen should just let the package manager handle that as well.  This seems a little odd since they seem equal, but I suppose using actual `*` is pretty rare and perhaps explicit should have meaning.

If so, we don't need the `!==`, we can just only look for falsy, period.  I will also add a test using explicit `*` to indicate this.

@mrgrain

mrgrain: No pretty sure `*` is special and will not end up not being managed by projen.

I think the code is all correct now. Just the way it is the function ought to be named `isNotManagedByProjen`. And because I wanted to avoid this kind of negative naming I inverted the name and Boolean expression.

mrgrain: Your comment makes me think we are still not aligned though.

The `version` value here is the one projen knows about and has nothing to do with the package manager yet.

dkershner6: I think maybe "Managed by Projen" might be the unclear part...

This function was designed to detect when NCU was able to change the package.json version for packages.  I called this "default" behavior because it was instigated when the version was omitted (or `*`, presumably, but the lack of a variable is often called default).  This could also be called "managed by projen", since projen is changing versions on a command (upgrade), thereby removing control from the package manager and taking it for itself.

It seems that you mean that "managed by projen" was when the projen default command sets the version as dictated by the version constraint, then delegates control to the package manager.  This feels less "manage" to me, but I could see how you could get there, especially when the version constraint is pinned.

My thought is that we should name it something clearer than either of us have come up with.  My suggestion is `packageCanBeUpgradedInPackageJson`, leaves nothing to the imagination and we can leave affirmative, easy to understand logic.

Since we live in opposite parts of the world, in the interest of time, i'll implement this in the PR, but leave this comment unresolved for now.

Tagging for visibility: @mrgrain

mrgrain: I love the new name, `packageCanBeUpgradedInPackageJson` is so much better!

mrgrain (approved):

Issue #3379, opened 2024-02-18
Add `minPythonVersion`, `maxPythonVersion`, `black` and `flake8` to `PythonProject`


garysassano:
Add the following `NodeProject` equivalents to `PythonProject`:

- `minNodeVersion` (semver) → `minPythonVersion` (semver)
- `maxNodeVersion` (semver) → `maxPythonVersion` (semver)
- `prettier` (boolean) → `black` (boolean)
- `eslint` (boolean) → `flake8` (boolean)

mrgrain:
Agree re `black` and `flake8`.

Not sure I'm a huge fan of `minPythonVersion` and `maxPythonVersion`. I always felt the python way of defining it as a dependency is better. Is there anything we can't use the dep for? Is this only for consistency?

garysassano:
@mrgrain Actually, this is defined in [PEP 621](https://peps.python.org/pep-0621/) (standardized `pyproject.toml`), which Poetry hasn't adopted yet unfortunately, even though there is intention to migrate to it in the future.

The standard mandates the inclusion of a [requires-python](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) field, which specifies the Python version required for the project.

garysassano:
Actually, this is defined in [PEP 621](https://peps.python.org/pep-0621/) (standardized pyproject.toml), which Poetry hasn't adopted yet unfortunately, even though there is intention to migrate to it in the future.

The standard mandates the inclusion of a [requires-python](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) field, which specifies the Python version required for the project.

garysassano:
Actually, this is defined in [PEP 621](https://peps.python.org/pep-0621/) (standardized `pyproject.toml`), which Poetry hasn't adopted yet unfortunately, even though there is intention to migrate to it in the future.

The standard mandates the inclusion of a [requires-python](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) field, which specifies the Python version required for the project.

garysassano:
Actually, this is defined in [PEP 621](https://peps.python.org/pep-0621/) (standardized `pyproject.toml`), which Poetry hasn't adopted yet unfortunately, even though there is intention to migrate to it eventually.

The standard mandates the inclusion of a [requires-python](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) field, which specifies the Python version required for the project.

[...]

PR #3468, opened 2024-03-20
feat(python)!: introduce `pythonCompatibleVersion` and `pythonSemanticVersion` props

garysassano:
After giving it considerable thought, the concepts of `minPythonVersion` and `maxPythonVersion` didn't look like a good idea. A better approach, I realized, would be to introduce properties for defining compatible Python versions and another for specifying complex version requirements. Currently, the default version is set to `"^3.8"` which is somewhat useless given that Python incompatibilities usually occur between different minor versions. Ideally, the default should be `"^3.8.0"`, though that's a topic for future discussions.

In this PR, my aim is to add the `pythonCompatibleVersion` property for those primarily concerned with specifying a Python version to ensure compatibility, such as ensuring Python `3.11.x` is used without automatically upgrading to Python `3.12.x`. Additionally, the `pythonSemanticVersion` property is introduced for users who require more advanced version control.

Presently, to change the default Python version in a project, one must use `project.add_dependency`, which is not only cumbersome but also obscures how to properly set a specific Python version, as there's no direct property for it and it lacks documentation visibility (automatic docgen doesn't cover this scenario).

---

### DIFF
--- a/src/python/python-project.ts
+++ b/src/python/python-project.ts
@@ -30,6 +30,19 @@ export interface PythonExecutableOptions {
    * @default "python"
    */
   readonly pythonExec?: string;
+
+  /**
+   * Defines the Python version that the project can work with.
+   * This version follows the caret (^) notation from the SemVer standard.
+   * @default "3.8"
+   */
+  readonly pythonCompatibleVersion?: string;
+
+  /**
+   * A specific Python version constraint following the SemVer standard.
+   * This version takes precedence over `pythonCompatibleVersion`.
+   */
+  readonly pythonSemanticVersion?: string;
 }
 
 /**
@@ -317,6 +330,8 @@ export class PythonProject extends GitHubProject {
         homepage: options.homepage,
         classifiers: options.classifiers,
         pythonExec: options.pythonExec,
+        pythonCompatibleVersion: options.pythonCompatibleVersion,
+        pythonSemanticVersion: options.pythonSemanticVersion,
         poetryOptions: {
           readme: options.readme?.filename ?? "README.md",
           ...options.poetryOptions,
--- a/test/python/poetry.test.ts
+++ b/test/python/poetry.test.ts
@@ -3,6 +3,8 @@ import { python } from "../../src";
 import * as logging from "../../src/logging";
 import { synthSnapshot } from "../util";
 
+const DEFAULT_PYTHON_VERSION = "^3.8";
+
 test("poetry enabled", () => {
   const p = new TestPythonProject({
     poetry: true,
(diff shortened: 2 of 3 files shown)


mrgrain:
> Presently, to change the default Python version in a project, one must use `project.add_dependency`, which is not only cumbersome but also obscures how to properly set a specific Python version, as there's no direct property for it and it lacks documentation visibility (automatic docgen doesn't cover this scenario).

I don't see how `project.add_dependency` or using `deps` is any more cumbersome than specifying a separate property, however I agree that something like a `pythonVersion` top-level property would be preferred for discoverability.

Either way, we are not going to have two props that do the same thing. One will be sufficient with good documentation.

src/python/python-project.ts
```diff
@@ -317,6 +330,8 @@ export class PythonProject extends GitHubProject {
         homepage: options.homepage,
         classifiers: options.classifiers,
         pythonExec: options.pythonExec,
+        pythonCompatibleVersion: options.pythonCompatibleVersion,
+        pythonSemanticVersion: options.pythonSemanticVersion,
```
mrgrain, line 334: This will have to go into the other tools as well where supported or warn that it's ignored.

mrgrain, line 334: This will have to go into the other tools as well where supported or warn that it's ignored.

test/python/poetry.test.ts
```diff
@@ -3,6 +3,8 @@ import { python } from "../../src";
 import * as logging from "../../src/logging";
 import { synthSnapshot } from "../util";
 
+const DEFAULT_PYTHON_VERSION = "^3.8";
```
mrgrain, line 6: We prefer explicit tests. This is easy enough to update when it changes.

mrgrain (changes requested):
see comments

tinovyatkin:
> This PR introduces a new way to define the Python version in a project by adding the requiresPython property to PythonProject. This change is in line with the standards established by [PEP 621](https://peps.python.org/pep-0621/#requires-python). 

This is mixing two unrelated things. The PEP you linked has no relation to Poetry, and projen currently (wrongly) believes that `pyproject.toml` is something Poetry specific (we don't have a dedicated component for it in every Python project as we probably should).

So, this fix (as well as support for other PEPs related to pyproject.toml) should probably be two steps:

1. Move `pyproject.toml` construct our of Poetry scope. This also will help with tooling like ruff, black, etc - see #3379
2. Add `requiresPython` property support

garysassano:
> The PEP you linked has no relation to Poetry

The PEP in question isn't directly related to Poetry itself. However, it defines a standard way to manage Python project metadata in `pyproject.toml`, which also Poetry leverages. It also mentions how the `requires-python` field got implemented across different tools.

image

Moreover, as indicated by the cautionary message on the website, this PEP has seen further development over the past few years, with the `pyproject.toml` specification now under the stewardship of [Python Packaging](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/). The Poetry team has plans to transition to the `pyproject.toml` standard eventually, but that requires significant work because the existing implementation within Poetry includes features that surpass those of the PyPa specification.

image

PR #3474, opened 2024-03-22
fix(semver): improved support of the Python Version Identification part of PEP440

icj217:
The semver util's current version logic does not fully implement/adhere to PEP440 in two respects:
* There is no support for "[local version identifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers)", which are basically the same as SemVer's build metadata (e.g. `1.2.3+foobar`), when used in conjunction with a pre-release label.
* The current pre-release logic doesn't reflect the ability for python pre-releases to include [post-release](https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases) and [developmental release](https://packaging.python.org/en/latest/specifications/version-specifiers/#developmental-releases) labels in conjunction with the pre-release itself (e.g. `1.2.3.rc1.post2.dev3`)

This PR addresses these gaps so that the python release supports these features. I've kept support for the `pre` label as a synonym for `dev`.

Projen and JSII's logic diverged ever so slightly re: including a `.` between the version and pre-release labels (e.g. projen was yielding `1.2.3rc0` while JSII would yield `1.2.3.rc0`. I decided to update projen's logic to mirror that from JSII.

E.g. now `1.2.3-rc.1.dev.2.post.3+foobar` will now yield `1.2.3.rc1.post3.dev2+foobar` for python packages.

I have an identical PR open over in jsii: https://github.com/aws/jsii/pull/4462

---


mrgrain:
@icj217 Thanks for this! I suspect this needs a new release of pacmak [with the respective change](https://github.com/aws/jsii/pull/4462) first?

mrgrain (changes requested):
See question

icj217:
> @icj217 Thanks for this! I suspect this needs a new release of pacmak [with the respective change](https://github.com/aws/jsii/pull/4462) first?

@mrgrain Not necessarily. I think this can be merged first because this logic applies to any python package (not just a JSII-based one) that a user registers as a dependency in their project. JSII projects won't be able to publish versions that would exercise the new logic here until the pacmak release is published.

mrgrain (approved):

PR #3500, opened 2024-03-30
fix: add missing build workflow strategy matrix controls

### LINKED ISSUE #3499: Make GitHub pipeline run tests on multiple node versions
This comes from the discussion on #3447, where (in that case) been able to build with different combinations of node and ESM would be useful.

The concept being to use `jobs.<job_id>.strategy.matrix` int he GitHub workflows as outlined here: [Using a matrix for your jobs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)

This is a generic capability projen projects should be able to utilize as well, and there's already partial support for in place.

I'm working on a PR for this, and it turns out it's not as well supported as it at first appears.

Issue's I've encountered (so far, checked means I'm past that in my in-progress PR):

- [x] `BuildWorkflow` doesn't expose a way to set the `strategy` in it's options, and there's no mechanism (outside of an escape-hatch) to set it later
- [x] `NodeProject` doesn't expose a way to set the `strategy` of the build workflow
  - I've added it as `buildWorkflowJobStrategy` to match `buildWorkflowTriggers` etc.
  - This only covers build workflow, what if we want workflow matrices for other jobs such as release? I'd assume they wouldn't be the same matrix
- [x] `JobMatrixValue` doesn't support objects
  - Docs linked to above say:
    > A variable configuration in a matrix can be an array of objects.
  - It's unclear if that means it can be arbitrarily deep objects, so I'm going with a two-level 
- [x] For node version I've added `buildWorkflowNodeVersion` and wired it up to override node version in the workflow setup for the build workflow
- [ ] We build and test in the same step, and then release an artifact from that build. We either need a mechanism to choose which build is the one that's released or we need to separate build and test
  - Closely related, we probably don't need or want "Upload coverage to Codecov" on every node of the matrix

I'm stopped on that last part, and am open to suggestions. For now, I'm going to add a property with node `18.14.2` (current one being used to release) that I'll have it add `if` statements to the final "Backup artifact permissions" and "Upload artifact" steps.

I'll update when I have a branch pushed to review.

giseburt:
Fixes #3499

Feedback extra welcome, as some of this feels extra clumsy

---


giseburt:
Appears to have [worked](https://github.com/projen/projen/actions/runs/8494096794?pr=3500)

src/github/workflows.ts
```diff
@@ -366,7 +366,7 @@ function setupTools(tools: workflows.Tools) {
 
   if (tools.java) {
     steps.push({
-      uses: "actions/setup-java@v3",
+      uses: "actions/setup-java@v4",
```
mrgrain, line 369: We need separate PRs for each of these please

mrgrain (commented):
I'll have a look at the rest later. This feels awfully complicated though. We might need an abstraction here.

giseburt:
I agree completely. It is very hacky at a few points to get this working, partially due to unanticipated needs in the node project object.

I rather liked the interface of this (sister?) project: https://github.com/cdklabs/cdk-pipelines-github/blob/main/src/workflows-model.ts

I don't like that they've put the implementation of the interface in the core of the project: https://github.com/cdklabs/cdk-pipelines-github/blob/main/src/pipeline.ts#L196

But still. I particularly like the handling of inputs and outputs.

Gonna related note, this is another place where Tokens (like in CDK) would come in handy. Was there a historic reason that they weren't put into Projen?

src/github/workflows.ts
giseburt: Adjusted back

giseburt:
Ok, so I was able to simplify a bit, and I think it fits the patterns of `projen` a little better:

```ts
const project = new cdk.JsiiProject({
  //...
  buildWorkflowOptions: {
    jobStrategy: {
      matrix: {
        domain: {
          runsOn: ["ubuntu-latest", "windows-latest"],
          node: [
            { version: "18.14.2" },
            { version: "18.18" },
            { version: "18.20" }, // some tools behave differently in 18.20 than 18.18
            { version: "20" },
          ],
        },
        include: [
          {
            runsOn: "ubuntu-latest",
            node: { version: "18.14.2" },
            release: true,
          },
        ],
      },
    },
    nodeVersion: "${{ matrix.node.version }}",
    uploadArtifactsVariable: "matrix.release",
    runsOnVariable: "matrix.runsOn",
  },
  workflowNodeVersion: "18.14.2", // was existing
  // ...
```

Options to simplify from here:

Assume if `buildWorkflowOptions.jobStrategy.matrix.domain.runsOn` exists, then:
  - use `runsOnVariable: "matrix.runsOn"` unless overridden

Assume if `buildWorkflowOptions.jobStrategy.matrix.domain.node` exists, then:
  - use `nodeVersion: "${{ matrix.node.version }}"` unless overridden
  - further, if `buildWorkflowOptions.jobStrategy.matrix.include` does NOT exist, add it as:
    ```ts
    include: [
      {
        runsOn: "ubuntu-latest", // ← if buildWorkflowOptions.jobStrategy.matrix.domain.runsOn
        node: { version: workflowNodeVersion },
        release: true,
      }
    ]
    ```
  - and `uploadArtifactsVariable: "matrix.release"` unless overridden

With all that, then this would be the same as above:
```ts
const project = new cdk.JsiiProject({
  //...
  buildWorkflowOptions: {
    jobStrategy: {
      matrix: {
        domain: {
          runsOn: ["ubuntu-latest", "windows-latest"],
          node: [
            { version: "18.14.2" },
            { version: "18.18" },
            { version: "18.20" }, // some tools behave differently in 18.20 than 18.18
            { version: "20" },
          ],
        },
      },
    },
  },
  workflowNodeVersion: "18.14.2", // was existing
  // ...
```

That's a lot of convention over configuration, but that's kinda The Projen Way™.

We could also just say "useBuildStrategyMatrix: true" to default all the above (minus the windows but until we're ready) but I feel that's too presumptive unless we also allow it to be fully overridden.

mrgrain (changes requested):
Can we split this PR up in two parts?

The low-level workflows should definitely support job strategy/matrix. That is not controversial.

For the Build Workflow, it think we will need to think much more holistically.

PR #3506, opened 2024-04-02
fix(subprojects)!: conflicting projenDevDependency is added to subprojects by default

timothymathison:
Partially addresses: #3304

BREAKING CHANGE: by default `projen` will not be installed in subproject directories. To install `projen` in a subproject directory, explicitly set `projenDevDependency: true` in the subproject.

Installing projen in sub-projects leads to weird behavior where the sub-project is referencing different source than the parent project - which breaks things like `instanceof` expressions when the two class objects are from different sources. I think it makes more sense to only install a single copy of `projen` for an entire project graph.
I suspect that enabling sub-project installation by default was an oversight originally since it makes sense to have that default for the root project. So I think its safe to consider this a non-breaking change since the default will stay unchanged for a root project.

---

### DIFF
--- a/docs/api/awscdk.md
+++ b/docs/api/awscdk.md
@@ -11984,7 +11984,7 @@ public readonly projenDevDependency: boolean;
 ```
 
 - *Type:* boolean
-- *Default:* true
+- *Default:* true if not a subproject
 
 Indicates of "projen" should be installed as a devDependency.
 
@@ -17214,7 +17214,7 @@ public readonly projenDevDependency: boolean;
 ```
 
 - *Type:* boolean
-- *Default:* true
+- *Default:* true if not a subproject
 
 Indicates of "projen" should be installed as a devDependency.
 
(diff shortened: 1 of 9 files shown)


timothymathison:
@mrgrain mind reviewing this when you have a chance?

timothymathison:
> See here for the required format of your PR:
> https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

In this case, that goes in the PR description (since you generally squash merge) right?

mrgrain:
> > See here for the required format of your PR:
> > https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer
> 
> In this case, that goes in the PR description (since you generally squash merge) right?

Correct PR title and description will make up the commit message.

timothymathison:
Okay, PR description updated, let me know if its satisfactory

mrgrain (approved):

PR #3520, opened 2024-04-11
chore: add windows runner in build job

gmeligio:
Related to #3384 

## Motivation

Enable experimental Windows runner to build job in CI. The job will fail while projen tasks that run during the task `build` are not compatible with Windows. This way we can keep track of the progress.

## Changelog

1. No tests added.
1. Allow to fail the `build` step in the `build` job on Windows runner by using `continue-on-error=true`. I used `JsonPatch` in `.github/workflows/build.yml` instead of doing changes in `BuildWorkflow` at this point.
1. `step.if` condition to ignore on Windows the following steps: `Upload coverage to Codecov`, `Upload patch`, `Backup artifact permissions` and `Upload artifact`.
1. Move postconstruction setup to `projenrc` folder. Add `projenrc` to gitignore. 
1. Change checking for undefined by `?.` conditional operator. where possible in setupXxx functions.

## Possible Future Improvements

1. Refactor the multiple setupXxx functions to a class `projenrc/JsiiProjectPatch.ts` to configure the project during construction time too, not only during postconstruction. I didn't do it in this PR because there are already many changes by moving the setup to the `projenrc` folder.
1. Find a way for the `BuildWorkflow` and other workflows to expose the synthesized file path. 

---


mrgrain (approved):
Nice one!

mrgrain (approved):

mrgrain (changes requested):
This isn't merging yet, because you missed to mark the windows build as continue on error.

gmeligio:
Thank you, @mrgrain, for the review!

As you noticed, I missed the PR check, which is a step instead of a job, and I thought that allowing the job to fail would also allow the step to fail. I corrected it and moved the patch of `continue-on-error` to the `build` step in the `build` job. Now all the checks are ✅

mrgrain:
> Thank you, @mrgrain, for the review!
> 
> As you noticed, I missed the PR check, which is a step instead of a job, and I thought that allowing the job to fail would also allow the step to fail. I corrected it and moved the patch of `continue-on-error` to the `build` step in the `build` job. Now all the checks are ✅

Hmm. @gmeligio Why can the `continue-on-error` not be on the job like documented in the guide I posted? I'm not sure I understand what you mean by you "missed the PR check, which is a step instead of a job".

gmeligio:
Never mind, after reading multiple Github issues, I was confused about why it didn't work at the job level. The PR check is indeed for jobs. Sorry for saying the contrary.

I followed the guide you suggested since the beginning when drafting the PR to learn how to use `continue-on-error`. Thank you. It was helpful, but it didn't solve my problem.

This comment was the one that gave me a hint about the step
https://github.com/orgs/community/discussions/27069#discussioncomment-3254501
What I understand now is that even if `continue-on-error` is allowed at the job level, it will still mark the job as failed but allow the subsequent jobs to run. When the specific step that fails is allowed to fail, then the job will marked as successful. That's why I committed again to put `continue-on-error` on the `build` step that was failing.

Please, let me know if you have another idea on how to do it.

mrgrain:
Hmm. When I first reviewed this PR I didn't see a `continue-on-error` at all. I believe I reviewed [this revision](https://github.com/gmeligio/projen/blob/26db8bf2af0592da31c5f1228e95fbf4d0a89aa7/.github/workflows/build.yml), so I must have just missed it! Not sure why it didn't merge the first time around then. Maybe some repo and/or Mergify settings need to change. 🤔 

The other thing I'm noticing now is that we probably shouldn't run some of the steps following the build step, when we are on  windows. Namely the uploads: `Upload coverage to Codecov`, `Upload patch`, `Backup artifact permissions`, `Upload artifact`. We probably still want `Find mutations` and `Fail build on mutation`.

gmeligio:
Thank you.

After studying again what Github supports for this case, I would say there is not much to do for the GitHub PR checks. My understanding is that Github Actions separate the concepts of final status of the step/job from the conditions that can trigger subsequent steps/jobs. Unfortunately, the final status is what is reported in the PR check. I looked into Github repo settings docs and Mergify docs but couldn't find it. Maybe you'll find a neat trick!

Good catch with the upload jobs that are not helpful on Windows. I skipped them using an `step.if` condition.

mrgrain (changes requested):
One small thing and we should be good to ship this.
Thanks again for putting the work in!

mrgrain (approved):

gmeligio:
Hi again @mrgrain . An unexpected setback that prevents merging.

Currently, in the PR checks, there is a status check named `build` with the message "Expected—Waiting for status to be reported." It seems we need to change either the repository settings or the Mergify configuration to solve this one.

We'll need to change the job names in Mergify rules, according to this [Github Support answer](https://github.com/orgs/community/discussions/30724#discussioncomment-3473663) and [this Mergify issue](https://github.com/Mergifyio/mergify/issues/486), to something like the following:
- build (ubuntu-latest, false)
- build (windows-latest, true)

How do you recommend proceeding, changing repository settings or Mergify config, or both?

mrgrain:
I see. I can change the repo settings, but that doesn't solve Mergify.

I remember this issue from other repos and tl;dr it's quite annoying to manage matrix job. So the solution typically was to have a fake compound job that requires all the Matrix jobs we are actually interested in - and then does nothing else:

https://github.com/aws/jsii/blob/main/.github/workflows/main.yml#L363-L370

projenrc/index.ts
```diff
@@ -0,0 +1,392 @@
...
+  project.npmignore?.exclude("/SECURITY.md");
+  project.npmignore?.exclude("/.gitpod.yml");
+}
+
+/**
+ * Setup the github workflow for windows
+ *
+ * @param project The project to add the configuration to
+ */
+export function setupGithubWorkflowWindows(project: NodeProject) {
```
mrgrain, line 342: Can we please split this into a separate file now that we have a directory. Unfortunately it's, getting a bit bigger than anticipated now. 🙈

projenrc/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    JsonPatch.add(
+      `/jobs/build/steps/${stepIndex}/if`,
+      "${{ !matrix.runner.experimental }}"
+    )
+  );
+
+  // TODO: workflows should expose file path. Use workflow path instead of hardcoded string
+  const buildWorkflow = project.tryFindObjectFile(
+    ".github/workflows/build.yml"
+  );
```
mrgrain, line 364:
```suggestion
  const buildWorkflow = project.github?.tryFindWorkflow('build')?.file;
```

gmeligio:
Thank you so much @mrgrain for taking care of this! I didn't had the time to move forward the past few days. Very nice of you.

mrgrain:
No worries at all. Thank you for doing all the heavy lifting here! 

Now, on to the next one 😅 🙈

PR #3526, opened 2024-04-15
fix(release): retrying the workflow can have no effect

iliapolo:
When a release succeeds to create the GitHub tag, but fails on other publish steps (for example NPM), retrying the workflow has no affect. This is because we only run publishing jobs in case the tag doesn't exist. In case of a partial failure, the tag exists but the workflow didn't actually succeed and we'd like to retry.

This PR makes it so the GitHub tag is only created if all other publishing jobs succeeded. 

---


src/release/publisher.ts
```diff
@@ -29,6 +29,7 @@ const PUBLIB_TOOLCHAIN = {
   go: { go: { version: "^1.16.0" } },
   dotnet: { dotnet: { version: "3.x" } },
 };
+const PUBLISH_JOB_PREFIX = "release_";
```
iliapolo, line 32: I took a quick glance and it seems to require quite a bit of refactoring in order to avoid this and have the `addPublishJob` method return the job name. 

This is a lazy way of at least avoiding duplicate strings and "syncing" job name to `dependsOn`.

src/release/publisher.ts
```diff
@@ -166,6 +167,8 @@ export class Publisher extends Component {
   private readonly workflowNodeVersion: string;
   private readonly workflowContainerImage?: string;
 
+  private packageManagersPublishJobs: string[] = [];
```
iliapolo, line 170: Again, lazy way to keep track of publishing jobs.

mrgrain (changes requested):

mrgrain (approved):

PR #3562, opened 2024-05-07
fix(docker-compose): omit version if schemaVersion is not provided

Tucker-Eric:
The `version` element is deprecated in docker compose configs (https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements)

This change omits `version` from the `docker-compose.yml` when `schemaVersion` isn't set. Not removing it completely to allow for backwards compatibility.

---


Tucker-Eric:
> Good catch! Can you also `@deprecated` the options please

Done!

mrgrain:
```
src/docker-compose/docker-compose.ts:43:3 - error JSII7999: @deprecated tag needs a reason and/or suggested alternatives.
```

and then you'll have to update the generated API docs again

mrgrain (approved):

PR #3571, opened 2024-05-11
fix(gitlab): use WorkflowWhen in WorkflowRule

JohannesKonings:
The WorkflowRule has the type JobWhen for the when field. When in a workflow rule can be only `always` and `never`: https://docs.gitlab.com/ee/ci/yaml/#workflowrules
If used in jobs it could have more inputs: https://docs.gitlab.com/ee/ci/yaml/#when

---


mrgrain:
Thanks for the fix and congrats our first contribution!

I think you will also have to run the build locally to update the auto generated API docs.

JohannesKonings:
@mrgrain I run the build and added the updated gitlab readme.

mrgrain (approved):

Issue #1859, opened 2022-05-17
tasks and Environment variables / $ syntax is not working on windows


Console32:
If I add the following task to my solution:

```
const hello = project.addTask('hello');
hello.env('BAR', 'world');
hello.exec('echo $FOO, $BAR!');
```

running it on windows (pwsh) results in:
```
❯ projen hello  
👾 hello | echo $FOO, $BAR!
$FOO, $BAR!
```

executing the same task on wsl results in as expected:
```
❯ npx projen hello
👾 hello | echo $FOO, $BAR!
, world!
```

how can I create a task that works on both platforms?

kichik:
It's not the most portable, but I used `bash -c "echo $FOO, $BAR"` to get around this issue. It requires WSL, but it does work.

https://github.com/CloudSnorkel/cdk-github-runners/blob/1e575d17bcde6513ed8ef67f1ee70a5da7b81a14/.projenrc.js#L104-L105

mrgrain:
I think the best way would be to write the command in JS/TS/Python/etc. There's no easy way to configure projen for that, but something like `project.addTask('node hello-world.js');` should work.

PR #3582, opened 2024-05-16
fix(windows): improve windows compat by successfully running full test suite on windows

gmeligio:

This PR has many Windows-related changes, like newline characters, etc. It also includes many changes in tests to make them work in both Linux and Windows. Please review the tests carefully in case the technique is not the right one.

At this moment, all tests are passing in Windows and Linux. However, the `build` is still failing on the next step, which is the `package`. The `package` step could be better fixed in another PR.

### .gitattributes

1. Autodetect text files and normalize line endings to LF to solve mutation, git shows the message: _warning: in the working copy of 'file.ext', LF will be replaced by CRLF the next time Git touches it_

### .github/workflows/build.yml

1. Use cmd shell when running on the `windows-latest` runner.

### package.json, javascript/jest.ts, test/javascript/jest.test.ts, test/typescript/projenrc.test.ts, test/typescript/typescript.test.ts

1. Correct jest match patterns.

### projenrc/index.ts

1. Add `setupGitattributes` to autodetect text files and normalize line endings to LF.

### projenrc/windows-build.ts

1. Use cmd shell when running on the `windows-latest` runner.
1. Run all projen tasks in `build` excluding the `package` task that doesn't work on Windows.

### awscdk/integrate-test.ts, cdk/integration-test-base.ts, cdk8s/integration-test.ts

1. use POSIX path separator from `join` function in `"path/posix"` package.

### awscdk/lambda-function.ts

1. Normalize the lambda entrypoint path because it persists in the eslint configuration.

### cli/cmds/new.ts

1. Normalize the `--from` argument, which persists in package.json dependencies.

### java/projenrc.ts

1. Normalize the relative path of the java projenrc file.

### javascript/typescript-config.ts

1. Normalize `extends` in `TypescriptConfig` (tsconfig.json).

### project.ts

1. Normalize the `defaultTask.exec.cwd` path.

### release/release.ts

1. Normalize the project path that's used in Github workflows.

### task-runtime.ts

1. Use double quotes for the command process, which both Unix bash and Windows cmd accept. Single quotes fail on Windows cmd. In Powershell needs to be preceded by the call operator `&` like `& "process" "command"` but it's assumed it won't be used right now.

### version.ts

1. Use double quotes in grep regular expression.

### test/cli.test.ts

1. Replaced `echo "text" >> file.txt` with `fs.writeFileSync("file.txt", "text")`.

### test/json.test.ts

1. Change the validated state in the test. `executable` will always be true on Windows, so only `readonly` can be checked. `readonly=true` by default, so changing it to `readonly=false` to be able to detect a change.

### test/new.test.ts

1. Normalize the path of the tarball.

### test/release/bump.test.ts

1. The commit message sometimes contains characters incompatible between Linux and Windows (like newlines). A cross-platform solution is to read the commit message, which is already written in the `filePath` file.

### test/tasks/runtime.test.ts

The biggest challenge for this file was deciding if some existing NodeJS library could be used as a cross-platform shell interpreter. After researching on the NodeJS ecosystem and following the discussion on the issue https://github.com/projen/projen/issues/1859 and specifically this comment https://github.com/projen/projen/issues/1859#issuecomment-1228881958 , I chose to replace the commands with node CLI equivalents when there was no shell alternative.

1. The environment variables in Windows cmd can be called like `%VAR%` while in bash is `${VAR}`. Hence, I opted to use `process.env.VAR` instead.
1. Replaced `echo "text" >> file.txt` with `fs.writeFileSync("file.txt", "text")`.
1. Replaced `cat` with `console.log(fs.readFileSync())`
1. Replaced `$PWD` with `process.cwd()`.
1. Left unmodified simple `echo` commands that behave the same in Windows and Linux. Those could be replaced by `node -e "console.log('text')"` as the rest of `echo` for more consistency.
1. Left unmodified some projen tokens like `$@` and `$(` which don't work in Windows cmd but are interpreted by projen before passing the command to the shell.

### test/util.ts

1. When resolving the package path, `path.normalize` needs to be specific to the OS platform, and `path.posix.normalize` doesn't work in this case.

---


mrgrain:
I can't even tell you how excited I am by this! 😍

gmeligio:
@mrgrain, this MR is ready for the first review. Please take into account that there are many changes to ensure all the tests pass in both Windows and Linux. I tried to include all the details in the PR description because it looks tough to review.

mrgrain:
Will do. Thanks for keeping track of the details.

There seem to be still some snapshots the generate a different output. Are you aware of that?

gmeligio:
I wasn't paying attention to the snapshots. Thank you, @mrgrain, for pointing that out. I'll look into those. I'm switching back to draft to fix those.

mrgrain:
@gmeligio Proposal: Shall we just remove the failing package step from the windows build for now? (using `if` in the workflow). That way we can at least start to properly enforce windows tests. What do you think?

gmeligio:
That's a good idea! I can do the changes for that in a couple of weeks after I'm back from holidays.

I also need to remove the jest changes that I put to debug which snapshots were changing.

gmeligio:
@mrgrain I did some changes regarding the package step, but in a different direction than your proposal. It seemed to me that the effort will be the same to make the package step to work both in Windows and Linux for `JsiiProject`. Because the package would need to run conditionally depending on the OS and I was having a hard time finding a simple way to do it.

The package script turned out to work in local both in Windows and Linux but it introduced an error in the CDK tests. It looks like the error is related to a circular dependency, as you can see in `TypeError: Class extends value undefined is not a constructor or null` https://github.com/projen/projen/actions/runs/9929082362/job/27426178617?pr=3582#step:5:2646

I verified that error was introduced when `TextFile` was imported in `src\cdk\jsii-project.ts` for the package script. Is there any chance you can help me fix this circular dependency? I was checking with the **[dpdm](https://github.com/acrazing/dpdm)** tool `dpdm -o circular.json src/cdk/jsii-project.ts`, but I haven't found how to fix it. If you think is too complex, then I can look again into your original proposal of removing the package step.

Unrelated to this, as an aside, since the tests are failing because of the package step, there was a misconfiguration of jest match patterns in https://github.com/projen/projen/pull/3637, which I already fixed in this PR, too. I noticed it when the tests weren't running after rebasing.

mrgrain:
Thanks for fixing up the jest config again. That's kind of why I would like to get this green asap and not wait the additional time on the package step. Plus there's the risk of rollbacks. I think so far your changes are fairly uncontroversial. But any change to packaging carries some risk of requiring a rollback.

mrgrain:
> Because the package would need to run conditionally depending on the OS and I was having a hard time finding a simple way to do it.

What's your current alternative approach?

My ideas in order of less to more effort are: 

- Use `test` instead of `if` which looks to be supported by `shx`
- Write a JS Task to do it 
- Get rid of the different behaviour in CI

gmeligio:
Thank you for your analysis. I went with writing a JS Task for package and hit the issue "TypeError: Class extends value undefined is not a constructor or null".

Seeing that this JS task could create issues and had to rollback, and also hard to debug with so many changes in this PR, I think I have a clearer way to only modify the CI to remove the package step. Basically, it would be to change the windows build in Github workflow. I'm thinking on running every projen task declared in `build` but the `package` one, which is the last one. I missed this easy approach before.

mrgrain:
Awesome, yes that seems like a good intermediate solution. 

I'm happy to look at your js script replacement as well. I think @epolon went on a similar journey at some point. I'm on annual leave this week, so anything more than comments is unlikely to happen this week from me.

gmeligio:
I made the changes and removed the package step. I'll look into the package JS task in another PR. This is ready for another review when you have time.

test/tasks/runtime.test.ts
```diff
@@ -92,11 +92,19 @@ describe("environment variables", () => {
     const t = p.addTask("test:env:stepwise", {
       env: { VALUE: "something" },
       steps: [
-        { exec: "echo ${VALUE}", env: { VALUE: "foo" } },
-        { exec: "echo ${VALUE}", env: { VALUE: "bar" } },
+        {
+          exec: `node -e "console.log(process.env.VALUE)"`,
```
mrgrain, line 96:
```suggestion
          exec: `node -p "process.env.VALUE"`,
```

mrgrain (approved):
:shipit:

mrgrain:
Let's get this in 💪🏻 

@gmeligio If you want another challenge - any idea why the Windows tests are so much slower? 😅

gmeligio:
Thank you for reviewing @mrgrain ! I'll try to investigate and measure the Jest tests. I have no idea from the top of my head.

PR #3584, opened 2024-05-16
fix(release): github release may tag the wrong commit when multiple commits are released at the same time (race condition)

### LINKED ISSUE #3583: Release tags wrong commit
We recently had two PRs merge and the same time.  The one that merged first, ended up tagging the commit of the 2nd.  Then the 2nd cancelled itself because the commit was already tagged.

In the `release_github` job, it has this step to make the tag:

```yaml
      - name: Release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_REPOSITORY: ${{ github.repository }}
          GITHUB_REF: ${{ github.ref }}
        run: errout=$(mktemp); gh release create $(cat dist/releasetag.txt) -R $GITHUB_REPOSITORY -F dist/changelog.md -t $(cat dist/releasetag.txt) --target $GITHUB_REF 2> $errout && true; exitcode=$?; if [ $exitcode -ne 0 ] && ! grep -q "Release.tag_name already exists" $errout; then cat $errout; exit $exitcode; fi
```

The problem, I think, is that `GITHUB_REF: ${{ github.ref }}` should be `GITHUB_REF: ${{ github.sha }}`.  The ref changed during the release, so it ended up tagging the wrong commit.

polothy:
The problem, I think, is that `GITHUB_REF: ${{ github.ref }}` should be `GITHUB_REF: ${{ github.sha }}`. The ref changed during the release, so it ended up tagging the wrong commit.

Fixes #3583

---


polothy:
I just used GitHub's UI to edit this file, it'll probably break all sorts of tests.

mrgrain:
Eeek! That's a good one. Thanks for investigating and patching this!

mrgrain (approved):

polothy:
@mrgrain first, thanks for merging this so quickly, I'll update today!

I was thinking this morning, it might be smart to set `concurrency: "release"` on the `release` job ([docs](https://docs.github.com/en/actions/using-jobs/using-concurrency)) because it really seems like it shouldn't be run in parallel (if I understand the steps correctly).  I was initially trying to do that before I noticed the s/ref/sha thing, but I couldn't find a way to set it (the property is `readonly`).  I feel like just the `release` job needs this due to the tag creation vs the whole workflow.

mrgrain:
Yes that might be a good idea. The issue with `concurrency` is that it doesn't guarantee order (as per the docs), so it might not solve everything either.

The release job is already supposed to stop if another commit has entered.
https://github.com/projen/projen/blob/main/.github/workflows/release.yml#L46-L50

Can you check if/why this hasn't worked in your case?

polothy:
It must have been a perfect storm, we are talking a few second window hah.  But, I was looking at this step:

```yaml
      - name: Check for new commits
        id: git_remote
        run: |-
          echo "latest_commit=$(git ls-remote origin -h ${{ github.ref }} | cut -f1)" >> $GITHUB_OUTPUT
          cat $GITHUB_OUTPUT
```

What's odd is... I don't think it fetches.  Think it would need to be:

```shell
echo "latest_commit=$(git fetch origin && git ls-remote origin -h ${{ github.ref }} | cut -f1)" >> $GITHUB_OUTPUT
```

Perhaps?

mrgrain:
`git ls-remote` operates on the actual remote repository. The docs aren't super clear about this, what that's what it is doing. So no fetch needed.

> EDIT-2: "it doesn't guarantee order" - drat - but maybe that doesn't matter? If a newer commit runs first, then the 2nd release job would cancel itself because of the Check for new commits step?

That's reasonable but also I'm not sure if it's needed? maybe not needed? At this point all the work is already done, crucially the candidate tag has been identified.

polothy:
Sounds good to me, thanks for your help!

PR #3591, opened 2024-05-21
fix: use actions/create-github-app-token when not using permissions option

polothy:

### DIFF
--- a/src/github/github-credentials.ts
+++ b/src/github/github-credentials.ts
@@ -59,23 +59,21 @@ export class GithubCredentials {
     const appIdSecret = options.appIdSecret ?? "PROJEN_APP_ID";
     const privateKeySecret =
       options.privateKeySecret ?? "PROJEN_APP_PRIVATE_KEY";
-    const permissions = options.permissions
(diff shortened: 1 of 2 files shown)


polothy:
Was thinking either this change, or make the `GithubCredentials` constructor public so we can configure out own:

https://github.com/projen/projen/blob/16a38f9cb0e64ed4238c6ceaf9a3726d905ea33e/src/github/github-credentials.ts#L82

polothy:
This PR partially fixes https://github.com/projen/projen/issues/3111 because it only conditionally replaced `tibdex/github-app-token`.

mrgrain:
> Was thinking either this change, or make the `GithubCredentials` constructor public so we can configure out own:

You should already be able to do this with one line.

https://projen.io/docs/integrations/github/#actions-versions

polothy:
Oh nice, that worked!

```typescript
project.github?.actions.set(
  "tibdex/github-app-token",
  "actions/create-github-app-token@v1"
);
```

Luckily, the `tokenRef` is identical.

PR #3604, opened 2024-05-26
chore(docusaurus): add search

JohannesKonings:
It would be helpful if the docs are searchable.
Orama has a plugin for docusaurus, which can be used: https://docs.askorama.ai/open-source/plugins/plugin-docusaurus#docusaurus-plugin
Here is a preview how it works: https://johanneskonings.dev/projen/
Unfortenately, it has some console errors. Hopefully that could be solved the next days.

image

---

### DIFF
--- a/docusaurus/docusaurus.config.js
+++ b/docusaurus/docusaurus.config.js
@@ -147,7 +147,8 @@ const config = {
           return postcssOptions;
         },
       };
-    }
+    },
+    "@orama/plugin-docusaurus-v3",
   ],
 };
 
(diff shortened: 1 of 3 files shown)


mrgrain:
Ha this is awesome! Thanks for getting it started! 

There are also some (rendering?) issues with certain headings:

mrgrain:
@JohannesKonings thanks for demonstrating how simple this actually is! I went with a different plugin in the end that strips html tags better and has some more configuration options.

Issue #3603, opened 2024-05-26
typeRoots is missing in TypescriptConfigOptions


JohannesKonings:
typeRoots is not configurable in TypescriptConfigOptions (https://github.com/projen/projen/blob/main/src/javascript/typescript-config.ts), but listed here: https://www.typescriptlang.org/tsconfig/#typeRoots

PR #3608, opened 2024-05-27
fix(typescript): typeRoots is missing in typescript config

JohannesKonings:
Fixes https://github.com/projen/projen/issues/3603

---

### DIFF
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -549,8 +549,14 @@ export interface TypeScriptCompilerOptions {
   readonly paths?: { [key: string]: string[] };
 
   /**
+   * If typeRoots is specified, only packages under typeRoots will be included
+   * @see https://www.typescriptlang.org/tsconfig/#typeRoots
+   */
+  readonly typeRoots?: string[];
+
+  /**
    * If types is specified, only packages listed will be included in the global scope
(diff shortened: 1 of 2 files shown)


src/javascript/typescript-config.ts
```diff
@@ -548,6 +548,12 @@ export interface TypeScriptCompilerOptions {
    */
   readonly paths?: { [key: string]: string[] };
 
+  /**
+   * If typeRoots is specified, only packages under typeRoots will be included
+   * @see {@link https://www.typescriptlang.org/tsconfig/#typeRoots}
```
mrgrain, line 553:
```suggestion
   * @see https://www.typescriptlang.org/tsconfig/#typeRoots
```

mrgrain: `@link` currently doesn't render correctly in docs

mrgrain (changes requested):

src/javascript/typescript-config.ts
JohannesKonings: @mrgrain I copied from here: https://github.com/projen/projen/blob/main/docs/api/javascript.md#typesoptional-
So I fixed that also.

Should I also: https://github.com/projen/projen/blob/main/docs/api/javascript.md#declarationmapoptional- ?
That should be the last one with `@link`

mrgrain (approved):

Issue #3590, opened 2024-05-21
!reference tag in gitlab job scripts will be rendered in quotes


JohannesKonings:
Using the [!reference](https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#reference-tags) tag will be rendered in quotes.

```typescript
 const job: Job = {
      extends: ['.release'],
      beforeScript: ['!reference [.release, beforeScript]', 'npm install'],
    };
   return {
      release: job,
    };
```
```yaml
release:
  extends:
    - .release
  before_script:
    - "!reference [.setup, script]"
    - npm install
```

Due to the quotes the gitlab pipeline doesn't merge the reference.

Presumply the used YamlFile needs a custom tag to recognise `!reference`: https://github.com/projen/projen/blob/main/src/gitlab/configuration.ts#L156
https://eemeli.org/yaml/#writing-custom-tags

PR #3616, opened 2024-05-30
fix(gitlab): render tag !reference without quotes

JohannesKonings:
Fixes https://github.com/projen/projen/issues/3590

Here is an approach which works.
Unfortenately I didn't managed to get the custom tags to work like described here: https://eemeli.org/yaml/#writing-custom-tags
I also had some struggle with JSII 😄 
Maybe with some help we find a solution that works and looks not so hacky.

---


mrgrain:
Appreciate the effort, but not keen on the string replacements. Certainly not as a generic option on YAML file.

> I also had some struggle with JSII 😄

I suspect we will need to mark custom tags as `@internal` and basically only allow them for built-in features and TS.
I try to come up with a proposal you can take over the line.

mrgrain:
I played around with this a while ago here: https://github.com/projen/projen/pull/3626
Can't remember how far  got though.

mrgrain:
Closing as the direction we would go would be this: https://github.com/projen/projen/pull/3626

Issue #3630, opened 2024-06-04
Add `projenrc` dir to Jest test match list


moltar:
The `projenrc/*` dir is for building local components, which may have tests.

https://github.com/projen/projen/blob/edc3c0e9c624fb28c2844e7ac6d1bc9b1aa673dd/src/javascript/jest.ts#L732

PR #3637, opened 2024-06-06
feat(typescript): adds `projencodedir` to jest testMatch patterns

dkershner6:
Fixes https://github.com/projen/projen/issues/3630

---


src/typescript/typescript.ts
```diff
@@ -673,9 +673,13 @@ export class TypeScriptProject extends NodeProject {
       `ts-jest${jest.jestVersion}`
     );
 
-    jest.addTestMatch(`<rootDir>/${this.srcdir}/**/__tests__/**/*.ts?(x)`);
+    const validTestRootDirs = [this.srcdir, this.testdir, "projenrc"];
```
mrgrain, line 676: This is a variable as well actually: projenCodeDir 
https://projen.io/docs/api/typescript?_highlight=projenrc#projen.typescript.ProjenrcOptions.property.projenCodeDir

mrgrain (changes requested):
Needs to be dynamic

dkershner6:
Had to change the field from `private` to `public` on `ProjenrcTs`, and store `projenrcTs` on `TypeScriptProject`, and change the order of the constructor to get there, but everything seems as it should now, @mrgrain .

mrgrain (changes requested):
Thanks for the change. I think this is the wrong way around though. We would want the `ProjenRc` component to do `Jest.of()` and add a new pattern. This can also totally be a separate entry in the array.

dkershner6:
As I was diving into the new change, I realized there were several very similar `dirTestMatch` pattern creations going on (and I was creating one more), so I centralized the logic to the `Jest` class in a static method, made it use the Jest best practices (from the docs and defaults), then used that everywhere, including the new feature.

So, it's a bit more than I intended to take on, but should now work really well and the responsibilities are in the correct places (and should be easier to maintain that in the future). @mrgrain

dkershner6:
> Still on my list. Should get to it this week. Sorry for the delay.

No worries, just wanna be sure it wasn't forgotten.

test/typescript/__snapshots__/typescript.test.ts.snap
```diff
@@ -1224,8 +1224,8 @@ tsconfig.tsbuildinfo
         ],
       ],
       "testMatch": [
-        "<rootDir>/src/**/__tests__/**/*.ts?(x)",
-        "<rootDir>/@(test|src)/**/*(*.)@(spec|test).ts?(x)",
+        "<rootDir>/@(src|test)/**/?(*.)+(spec|test).[jt]s?(x)",
+        "<rootDir>/@(src|test)/**/__tests__/**/*.[jt]s?(x)",
```
mrgrain, line 1228: I don't think this is a good idea. This used to match ts files only on purpose.

```suggestion
        "<rootDir>/@(src|test)/**/?(*.)+(spec|test).ts?(x)",
        "<rootDir>/@(src|test)/**/__tests__/**/*.ts?(x)",
```

mrgrain (changes requested):
I like the direction. But there are some gnarly little issues we need to sort out.

src/javascript/jest.ts
```diff
@@ -811,12 +826,11 @@ export class Jest extends Component {
 
   /**
    * Adds a test match pattern.
-   * @param pattern glob pattern to match for tests
+   * @param patterns glob pattern to match for tests
    */
-  public addTestMatch(pattern: string) {
-    this.testMatch.push(pattern);
+  public addTestMatch(...patterns: string[]) {
+    this.testMatch.push(...patterns);
```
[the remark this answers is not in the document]
dkershner6: Reverted.

test/typescript/__snapshots__/typescript.test.ts.snap
dkershner6: I made it follow the `allowJs` variable, which seems like a valid compromise.

dkershner6:
@mrgrain , this has been updated.

In order to accommodate the `discover` method, I had to lazily load the testMatches.  I am reasonably sure this actually fixed a bug where all testMatches added after initialization were ignored, but it does require a `ts-expect-error` due to types.

src/javascript/jest.ts
```diff
@@ -817,6 +822,51 @@ export class Jest extends Component {
...
 
+  /**
+   * Build standard test match patterns for a directory.
+   * @param dirs The directories to add test matches for. Matches any folder if not specified or an empty array.
+   * @param fileExtensionPattern The file extension pattern to use. Defaults to "[jt]s?(x)".
+   */
+  public discoverTestMatchPatternsForDirs(
+    dirs?: string[],
+    fileExtensionPattern = "[jt]s?(x)"
+  ): void {
```
mrgrain, line 833: I think we need one more iteration on the API. Something like this:
```suggestion
  public discoverTests(
    directories: string[],
    options?: DiscoverTestsOptions
  ): void {
  const patterns = options.fileExtensionPattern ?? "[jt]s?(x)";
```

with 

```
interface DiscoverTestsOptions {
    fileExtensionPattern?: String
}
```

src/javascript/jest.ts
```diff
@@ -817,6 +822,51 @@ export class Jest extends Component {
...
+        )})/**/__tests__/**/*.${fileExtensionPattern}`,
+      ];
+    }
+
+    // Jest defaults
+    return [
+      `**/__tests__/**/*.${fileExtensionPattern}`,
+      `**/?(*.)+(spec|test).${fileExtensionPattern}`,
+    ];
+  }
```
mrgrain, line 868: I think this is a bit weird to have the defaults and the `buildMatchPatterns` helper wrapped up in the some method.
Let's make `dirs` required instead and put the defaults into a separate private helper and just call them explicitly.

The way it's done right now allows to add multiple default patterns by calling `discoverTestMatchPatternsForDirs` repetitively.

mrgrain (changes requested):
Sorry this took again longer than expected. We'll need an other iteration on the API.

dkershner6:
Changes have been made, @mrgrain .

mrgrain:
Looks good now. I just have to deal with the failing Maven release first.

mrgrain (approved):

Issue #2806, opened 2023-07-19
Prettier and Eslint extends


thoroc:
When using the [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) the official docs advice for the following:

> Then, add "prettier" to the "extends" array in your .eslintrc.* file. Make sure to put it last, so it gets the chance to override other configs.

As this is not the behaviour of the eslint component, I have to resort to do the following hack to push `prettier` as the last item in the eslint's extends list:

```typescript

// hack to push prettier as the last item in extends for eslint
if (project.eslint !== undefined) {
  const eslintExtends = project.eslint['_extends'];
  const prettierIndex = eslintExtends.indexOf('prettier', 0);
  if (prettierIndex > -1) {
    eslintExtends.splice(prettierIndex, 1);
  }
  eslintExtends.sort();
}
project.eslint?.addExtends('prettier');
```

Is this correct? Am I missing something?

mrgrain:
Just to confirm, your report here is that while `prettier` is added to the `extends` array, it is not added as the last item like it should be?

thoroc:
That is correct, at least when adding other item to the `eslint`'s `extends` list.

If I add the following to the `eslint` config:

```typescript
project.eslint?.addExtends(
  'eslint:recommended',
  'plugin:@typescript-eslint/eslint-recommended',
  'plugin:@typescript-eslint/recommended',
  'plugin:import/errors',
  'plugin:import/warnings',
);
```

I end up with this `eslintrc` config:

```json
{
  ...
  "extends": [
    "plugin:import/typescript",
    "prettier",
    "plugin:prettier/recommended",
    "eslint:recommended",
    "plugin:@typescript-eslint/eslint-recommended",
    "plugin:@typescript-eslint/recommended",
    "plugin:import/errors",
    "plugin:import/warnings"
  ],
  ...
}
```

Whereas this section should be:

```json
{
  ...
  "extends": [
    "eslint:recommended",
    "plugin:@typescript-eslint/eslint-recommended",
    "plugin:@typescript-eslint/recommended",
    "plugin:import/errors",
    "plugin:import/typescript",
    "plugin:import/warnings",
    "plugin:prettier/recommended",
    "prettier"
  ],
  ...
}
```

I have not found a way to better manipulate the `eslint` config outside a change to projen itself.

thoroc:
Oh thank you. I'll check that out.

PR #3641, opened 2024-06-07
feat(eslint): ESLint now makes a best effort to sort the extends array

dkershner6:
Fixes https://github.com/projen/projen/issues/2806

---


test/javascript/eslint.test.ts
```diff
@@ -104,18 +104,46 @@ test("if the prettier is configured, eslint is configured accordingly", () => {
...
+  });
+
+  eslint.addExtends("plugin:some-plugin/recommended");
+
+  // THEN
+  const output = synthSnapshot(project);
+
+  const extendsArray = output[".eslintrc.json"].extends;
+  expect(extendsArray.at(-1)).toEqual("plugin:some-plugin/recommended");
+  expect(extendsArray).toContain("plugin:prettier/recommended");
```
mrgrain, line 144: This test case is not super obvious. Instead, let's match the actual value here. Should not be too long are change to often.

src/javascript/eslint.ts
```diff
@@ -62,6 +61,15 @@ export interface EslintOptions {
...
   readonly prettier?: boolean;
 
+  /**
+   * The extends array in eslint is order dependent.
+   * Setting this option true will make a best effort to properly sort the extends array
+   * according to ESLint best practices.
+   *
+   * @default true
+   */
+  readonly sortExtends?: boolean;
```
mrgrain, line 71: I wonder if we should just allow a compare function here:
```suggestion
  readonly sortExtends?: ICompareString;
```

```ts
interface ICompareString {
  compare(a: string, b: string): number;
}
```

And our Eslint component can provide a default:

class Eslint {
  static public SORT_EXTENDS_DEFAULT(): ICompareString {
    return {
      compare(a: string, b: string): number {
          /// to do
      }
    };
  }
}

dkershner6: I like this idea.  I assume the nesting is a JSII requirement, as I see the CDK do that too, and I can't imagine .NET or java having a function parameter...

mrgrain: Yes, unfortunately free-floating functions are not jsii compatible. So we need a protocol interface for it.

dkershner6:
OK, @mrgrain , we now have a sorting function and an extra test.

src/index.ts
```diff
@@ -33,6 +33,7 @@ export * from "./task-runtime";
 export * from "./testing";
 export * from "./textfile";
 export * from "./toml";
+export * from "./util/types";
```
mrgrain, line 36: Let's go with
```suggestion
export * from "./compare;
```

src/javascript/eslint.ts
```diff
@@ -63,12 +64,11 @@ export interface EslintOptions {
 
   /**
    * The extends array in eslint is order dependent.
-   * Setting this option true will make a best effort to properly sort the extends array
-   * according to ESLint best practices.
+   * This option allows to sort the extends array in any way seen fit.
    *
-   * @default true
+   * @default Use known ESLint best practices to place "prettier" plugins at the end of the array
```
mrgrain, line 69:
```suggestion
   * @default - Use known ESLint best practices to place "prettier" plugins at the end of the array
```

dkershner6:
Updated, @mrgrain .

dkershner6:
Another @mrgrain nudge.

mrgrain:
Still on my list. Should get to it this week. Sorry for the delay.

PR #3690, opened 2024-07-03
fix(javascript): correct strict boolean compiler options to `boolean` type

### LINKED ISSUE #3689: bug(javascript): strictly typed boolean values in typescript config limit flexibility
The additional options added to `TypeScriptCompilerOptions` in #3409 are strictly typed as `false` or `true` instead of the appropriate `boolean` type.

As is, the generated dts (as expected) retains these types and spits out:

````
    /**
     * Allow Unused Labels
     *
     * When:
     *
     * - `undefined` (default) provide suggestions as warnings to editors
     * - `true` unused labels are ignored
     * - `false` raises compiler errors about unused labels
     *
     * Labels are very rare in JavaScript and typically indicate an attempt to write an object literal:
     *
     * ```ts
     * function verifyAge(age: number) {
     *   // Forgot 'return' statement
     *   if (age > 18) {
     *     verified: true;
     * //  ^^^^^^^^ Unused label.
     *   }
     * }
     * ```
     *
     * @see https://www.typescriptlang.org/tsconfig#allowUnusedLabels
     */
    readonly allowUnusedLabels?: false;
    /**
     * Allow Unreachable Code
     *
     * When:
     *
     * - `undefined` (default) provide suggestions as warnings to editors
     * - `true` unreachable code is ignored
     * - `false` raises compiler errors about unreachable code
     *
     * These warnings are only about code which is provably unreachable due to the use of JavaScript syntax.
     *
     * @see https://www.typescriptlang.org/tsconfig#allowUnreachableCode
     */
    readonly allowUnreachableCode?: false;
    /**
     * Check JS
     *
     * Works in tandem with [allowJs](https://www.typescriptlang.org/tsconfig#allowJs). When checkJs is enabled then
     * errors are reported in JavaScript files. This is the equivalent of including // @ts-check at the top of all
     * JavaScript files which are included in your project.
     *
     * @see https://www.typescriptlang.org/tsconfig#checkJs
     */
    readonly checkJs?: true;
````

Which a.) does not represent the actual default types (undefined) for these values and b.) does not allow for the actual `boolean` type to be used.

This causes headaches when trying to work with / extend the type for other usages, leading to situations such as:

```
> pnpm exec projen build
👾 build » compile | pnpm exec tsc --build --emitDeclarationOnly
Types of property 'allowUnusedLabels' are incompatible.
 Type 'boolean | undefined' is not assignable to type 'false | undefined'.
 Type 'true' is not assignable to type 'false'.
```

BradenM:
Signed-off-by: Braden Mars <bradenmars@bradenmars.me>

Fixes #3689

---

### DIFF
--- a/src/javascript/typescript-config.ts
+++ b/src/javascript/typescript-config.ts
@@ -636,7 +636,7 @@ export interface TypeScriptCompilerOptions {
    *
    * @see https://www.typescriptlang.org/tsconfig#allowUnusedLabels
    */
-  readonly allowUnusedLabels?: false;
+  readonly allowUnusedLabels?: boolean;
 
   /**
    * Allow Unreachable Code
@@ -651,7 +651,7 @@ export interface TypeScriptCompilerOptions {
    *
    * @see https://www.typescriptlang.org/tsconfig#allowUnreachableCode
    */
-  readonly allowUnreachableCode?: false;
+  readonly allowUnreachableCode?: boolean;
 
   /**
    * Check JS
@@ -662,7 +662,7 @@ export interface TypeScriptCompilerOptions {
    *
    * @see https://www.typescriptlang.org/tsconfig#checkJs
    */
-  readonly checkJs?: true;
+  readonly checkJs?: boolean;
 }
 
 /**


BradenM:
Interesting. Unsure where the windows-specific test failures are coming from, but I am fairly certain they are not originating from my type-only change.

mrgrain (approved):
Good catch! Thanks for the fix.

PR #3711, opened 2024-07-10
feat: CJS/ESMJavascriptFile (for configuration files in js)

dkershner6:
Fixes https://github.com/projen/projen/pull/2414

100% of credit goes to https://github.com/projen/projen/pull/3454 and @giseburt , was just splitting off the JavascriptFile part to save him some time, as everyone was clamoring for it (including us).

---


dkershner6:
Not sure what is going on with the CLI tests, I get the same error from main, even though it deployed fine, somehow...

mrgrain:
> Not sure what is going on with the CLI tests, I get the same error from main, even though it deployed fine, somehow...

This was an upstream dependency issue. Should be resolved now.

I'm on annual leave this week. Will try to get to this afterwards, but the windows tests PR will get the first priority.

giseburt:
@mrgrain Is there anything I can do to help this one along? Once this is merged (or whatever) I'll refactor the eslint #3454 PR.

src/code-token-map.ts
```diff
@@ -0,0 +1,224 @@
...
+
+// Much of this code is borrowed from the AWS CDK project
+// The AWS CDK usage is significantly more complex, as it can handle tokens in
+// data structures that expect numbers or arrays. This is a simplified version
+// that only handles strings.
+// See: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/private/token-map.ts
+
+const glob = global as unknown & { __ProjenCodeTokenMap?: CodeTokenMap };
+
+const STRING_SYMBOL = Symbol.for("@aws-cdk/core.TokenMap.STRING");
```
mrgrain, line 15: Any reason we cannot use projen symbol?

mrgrain:
Hi all, apologies for the long delay. I think the building blocks are here, but I don't think the (public) API is there yet. My main concerns are:
 
- The pure size of the added surface to the public API. I understand the urge to provide something foundational and flexible, however that significantly increases the scope of the PR.
- Instead I suggest we focus on a very narrow public API for a JavaScript config file. What are the must have requirements for that? Surely we won't need the whole JS feature set to write an eslint config. 
- Let's start from the DX. Looking at the examples in the tests, writing these files doesn't feel right. Compare this to other object files that have a really straight forward API and still support lazy-resolution via arrow functions.
- This last will include a new documentation page covering `JavaScriptDataFile`s (or maybe even just `JavaScriptConfigFile`) (we don't have a section for this yet, there probably should be one covering all sorts of files).
- The current proposal seems to be primarily adding two features:  a/ strings with tokens, and b/ conversion from JS literals to valid JavaScript syntax as string. If my understanding is correct here, than these should be implemented independently of each other. E.g. tokenized strings are something that should exist in its own without an inherent connection to (JS) code. (However I'm also not sure if we really need tokenized strings if we focus this on a narrower API for `JavaScriptDataFile`s.

test/javascript-file.test.ts
```diff
@@ -0,0 +1,550 @@
...
+  ESMJavascriptDependencies,
+  ESMJavascriptFile,
+  JavascriptDataStructure,
+  JavascriptFunction,
+  JavascriptRaw,
+} from "../src/javascript-file";
+
+test("JavascriptRaw can make js-type eslint output", () => {
+  // make a dependencies object to track imports
+  const dependencies = ESMJavascriptDependencies.value();
```
mrgrain, line 16: I'm aware I suggested two separate classes. But it doesn't really make sense from a user perspective to declare ESM vs CJS at the dependency level. Shouldn't dependencies be always the same (w/o someone deciding if it's ESM or CJS) and render differently depending on the file type they are added to?

test/javascript-file.test.ts
```diff
@@ -0,0 +1,550 @@
...
+      };
+      "
+    `);
+  });
+});
+
+// There are doing double duty: They serve as mildly redundant and simple tests,
+// but also as a place to copy the documentation examples and paste them in the
+// docs. This means they're formatted properly and working.
+describe("JavascriptFile doc examples", () => {
```
mrgrain, line 444: So this is basically a fully fledged JS code generator?

test/javascript-file.test.ts
```diff
@@ -0,0 +1,550 @@
...
+
+    const configFileName = "testFilename.mjs";
+
+    const file = new ESMJavascriptFile(project, configFileName, {
+      obj: {
+        exportedValue: "value",
+      },
+      marker: true,
+    });
+    const [newValueToken] = file.dependencies.addImport("fs", "fs");
```
mrgrain, line 380: Import names can be automatic unless desired otherwise. For `fs` it's easy. For `@projen/whatever` it will need sanitation.

test/javascript-file.test.ts
```diff
@@ -0,0 +1,550 @@
...
+    });
+
+    const configFileName = "testFilename.js";
+
+    new CJSJavascriptFile(project, configFileName, {
+      obj: {
+        exportedValue: "value",
+      },
+      marker: true,
+    });
```
mrgrain, line 291: I'd like to see a test (and docs) that create a JavaScriptFile with complex data. Code paths are covered by the unit tests, so this is more for understanding and as examples.

test/javascript-file.test.ts
```diff
@@ -0,0 +1,550 @@
...
+      "// ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
+      import fs from 'fs';
+      export default {
+        exportedValue: fs.readFileSync("file.txt"),
+      };
+      "
+    `);
+  });
+
+  test("can override with an import that is computed later", () => {
```
mrgrain, line 400: Is it really the import that is computed later? Or just another part of the override?

dkershner6:
> Instead I suggest we focus on a very narrow public API for a JavaScript config file. What are the must have requirements for that? Surely we won't need the whole JS feature set to write an eslint config.

Based on this comment, I am inclined to close this PR and allow the ESLint config PR handle the conversion from the all-encompassing JS to something specific, since it will essentially at that point be coupled, at least temporarily.  @giseburt , let me know if you disagree.

src/code-token-map.ts
giseburt: I think that one was on me. Yeah, that should change to `projen.TokenMap.STRING`

test/javascript-file.test.ts
giseburt: It's both. The tea there is can we override the value (which is just obj: {exportedValue: "value"},) with a more complex value that includes Tokens that have a value that's changed after creation.

giseburt:
Also sorry for the slow response.

> Instead I suggest we focus on a very narrow public API for a JavaScript config file. What are the must have requirements for that? Surely we won't need the whole JS feature set to write an eslint config.

Well, I both agree and disagree. In the eslint PR I was trying to do (at least) three things:

1. Update ESLint config file generation, which happened to be in javascript
2. Write a general (but limited) JavaScript code generator that honored and worked off of the JSON and YAML generators
  - Many things are NOT covered for a fully-fledged JS generator, as they're not needed here, like comments
  - The focus was on covering all of the examples in the ESLint documentation for the new file format, which mean including `...` spreads and import/require of modules, so it does cover a lot of ground
  - This was easiest done with Tokens as is used in the CDK project
3. Introduce Tokens to projen

This PR would be 2 and 3. My original PR was all three.

I can see a case for breaking that into three PRs, but we'd have a PR that "doesn't do anything," as having a PR to *just* introduce Tokens would seem silly without context. There are a few places that Tokens *could* be used but it would be unnecessarily changing things that currently work fine just to create a use for Tokens, and I didn't expect that to fly either. But we can, if you want.

I have use for Tokens in other areas of the code. I want to use them in GitHub Workflows, for example to make Tokens for outputs from a Step, and when it's used in the same Job it'll make the reference to the outputting Step, and if it's used in another Job it'll refer to the correct Step via `depends.<jobname>` AND add a dependency across Jobs. Also environment variables, steps that depend on other steps and aren't added if not used, etc. I already have internal code doing some of this, but would like it to be more generally available and not collide with upstream code.

Anyway, whichever way you prefer to go, @mrgrain. I'm getting more and more pressure to move to eslint 9, so I'm motivated on this again. :)

mrgrain:
> I have use for Tokens in other areas of the code. I want to use them in GitHub Workflows, for example to make Tokens for outputs from a Step, and when it's used in the same Job it'll make the reference to the outputting Step, and if it's used in another Job it'll refer to the correct Step via `depends.<jobname>` AND add a dependency across Jobs. Also environment variables, steps that depend on other steps and aren't added if not used, etc. I already have internal code doing some of this, but would like it to be more generally available and not collide with upstream code.

I reckon introducing Tokens for/via an existing feature would make it much easier because it provides us with a concrete existing example and use case. We can write tests, docs etc quite independently of this hypothetical new feature (that is 
eslint flat config). Of course the issue/PR should reference other future use cases and ideally this would be an RFC first. But I acknowledge how all that will slow things down.


> Well, I both agree and disagree. In the eslint PR I was trying to do (at least) three things:
> 
> 1. Update ESLint config file generation, which happened to be in javascript
> 2. Write a general (but limited) JavaScript code generator that honored and worked off of the JSON and YAML generators [...]
> 3. Introduce Tokens to projen
> 
> This PR would be 2 and 3. My original PR was all three.

My concerns aren't really around what combination of 1, 2, 3 are shipped in a single PR. My concerns are about the introduced new API surface area and major code changes associated with it. If I take your original PR an break it apart we have:

- **Tokens** - a new major feature that warrants an RFC on its own, small API surface but potentially huge implications everwhere
- **Generic JavaScript code files** (a code generator) - a new major with large public API surface that warrants an RFC on its own
- **JavaScript config/data files** - a small new feature with a limited API surface
- **Refactoring of the existing `Eslint` class** -  a major breaking change to an existing widely used component


Instead what I think we need to do is limit the changes and introduced API surface per PR:

- Tokens - ideally introduced for features that already exist at the time. Will need an RFC.
- A new "`EslintModern`" component - No generic APIs added, introduced APIs are strictly limited to Eslint, no change to existing component
  - With this, early adopters can set `eslint: false` and use the new component.
  - This could also include a "`EslintFlatConfig`" component if it makes sense to break it up like that
  - Limited to the absolute minimum required for automatic config, starting from the current Eslint API.
  - Needs to think about an escape hatch - how can users combine generated config with a hand written one.
  - Of course this only works if we can agree that such a component can work with a limited API compared to what can be done with flat configs.
  - If this API requires Tokens, than we are signing up for that and we need to consider Tokens first.
- Breaking Change to existing Projects to use `EslintModern`, deprecation of old component
- A more generic JavaScript Config/Data File component that offers an extended API compared to `EslintFlatConfig`

---

> Also sorry for the slow response.

Absolutely no need to apologies for anything here. This is mostly on me/us.

mrgrain:
Writing this down, I note that the core issue really is to decide what is needed for a `EslintFlatConfig`. I think if we start from the point that "it's just JS" and one should be able to write anything they like - then we end up with this PR. Instead I propose we should be looking at the existing interface, remove everything that is not needed anymore and introduce strictly only what's required:

```
// Path to `tsconfig.json` which should be used by eslint
// --> for  typescript-eslint
`tsconfigPath?: string;`

// Files or glob patterns or directories with source files to lint (e.g. [ "src" ])
// -> yes but now it's patterns
`dirs: string[];`

// Files or glob patterns or directories with source files that include tests and build toolsories with source files to lint (e.g. [ "src" ])
// -> yes but now it's patterns
`devdirs?: string[];`

// File types that should be linted (e.g. [ ".js", ".ts" ])
// -> not supported in FlatConfig
`fileExtensions?: string[];`

// Options for eslint command executed by eslint task
// -> still needed
`commandOptions?: EslintCommandOptions;`

// List of file patterns that should not be linted, using the same syntax as .gitignore patterns.
// -> still needed
`ignorePatterns?: string[];`

// deprecated
~`lintProjenRcFile?: string;`~

// Should we lint projenrc files
// -> still needed, just extends patterns
`lintProjenRc?: boolean;`

// Enable prettier for code formatting
// -> not needed anymore as eslint does not style anymore
`prettier?: boolean;`

// The extends array in eslint is order dependent.
// This option allows to sort the extends array in any way seen fit.
// -> see addExtends
`sortExtends?: ICompareString;`

// Enable import alias for module paths
// -> for import/resolver rule from eslint-plugin-import
`aliasMap?: { [key: string]: string };`

// Enable import alias for module paths
// -> for import/resolver rule from eslint-plugin-import
`aliasExtensions?: string[];`

// Always try to resolve types under `<root>@types` directory even it doesn't contain any source code.
// This prevents `import/no-unresolved` eslint errors when importing a `@types/*` module that would otherwise remain unresolved.
// -> for import/resolver rule from eslint-plugin-import
`tsAlwaysTryTypes?: boolean;`

// Write eslint configuration as YAML instead of JSON
// -> not needed anymore
`yaml?: boolean;`

// -> imperative add lint patterns
`addLintPattern(pattern: string)`

// -> seem to be mostly the same as before
`addRules(rules: { [rule: string]: any })`

// -> now need to be imports
`addPlugins(...plugins: string[])`

// -> imperative add ignore
`addIgnorePattern(pattern: string)`

// feature to allow dev deps in files that are not devDirs
`allowDevDeps(pattern: string)`

// -> not needed anymore
` enablePrettier()`


# This is where it becomes different

// -> this is now imports
// 3 different styles
// import single config
// import multiple configs as array
// ^ these are always applied to all files

// as part of the projen defined config <- this is closest to what projen does today.
`addExtends(...extendList: string[])`

// this is offering a completely custom config
// In FlatConfig this is essential the EslintConfig primitive
// projen offers a (dynamic) recommended config
`addOverride(override: EslintOverride)`
```

mrgrain:
In summary I think the new `EslintModern` API is very similar to the existing one. The main difference is how extending and existing configs are handled and that we need an API to represent an import  from a different package.

- We could use `tseslint.config` [to keep support for extends](https://typescript-eslint.io/packages/typescript-eslint/#flat-config-extends)
- I think we need a new primitive `EslintConfig` to replace `EslintOverride`
- A `EslintFlatConfigFile` is an array of `EslintConfig`s with imports resolved etc.
- `EslintConfig` It can have a static method `fromPackage()` or something that represents an imported config
- Plugins require the same, so i think there is a token light version that represents something importable
  - However this does not need to be resolved within a string, it's just a value of an object

mrgrain:
Closing in favor of https://github.com/projen/projen/pull/4408
Thanks for the inspiration and ideation here. Lots of the code is coming from this.

PR #3722, opened 2024-07-18
fix(jsii-project): use TypeScript script for package step

gmeligio:

Create a TypeScript task for the package step in the build projen task to make it cross-platform, inspired by https://github.com/projen/projen/pull/2761 . Kudos to @iliapolo for laying the ground for this. It was difficult to get my mind around the different ways of creating the JS script and how the builtin task works until I thoroughly inspected that MR.

## Changelog

1. Support passing `TaskStepOptions` in `Task.builtin()` to be able to pass inputs as environment variables.
1. Use double quotes when rendering the builtin command for Windows CMD shell compatibility.
1. Translate the bash command for package step to TypeScript task.

---


src/cdk/package-jsii-project.task.ts
```diff
@@ -0,0 +1,63 @@
...
+) {
+  fs.readdirSync(source).forEach((file) => {
+    if (!exclude.includes(file)) {
+      const src = path.join(source, file);
+      const dest = path.join(destination, file);
+
+      fs.cpSync(src, dest, {
+        recursive: true,
+        force: true,
+      });
```
mrgrain, line 29: `cpSync` takes a `filter` function. I reckon we can simplify this to only use `cpSync` and save us the extra `readdirSync`

gmeligio: Got it. Thanks

mrgrain:
@gmeligio I'm inclined to close this now. Ideally the task wouldn't even do two different things and the workflow would just "stop" before packaging. Than the `rsync`/`cp` step would be a workflow only concern. https://github.com/projen/projen/pull/3162 would allow us to do that.

What do you think?

gmeligio:
It makes sense. It would be better not to run the `package` step at all during the CI build. I'll close. Thanks for looking into this.

PR #3723, opened 2024-07-18
fix(jsii): running `projen package` on windows fails

mrgrain:
Fixes the failing packaging step in windows builds.

Instead of a complex bash expression, this now leverages  a feature of the projen task runner.
- Firstly, replacing the `if` statement with task step conditions 
- Secondly, use `node` to check if we are in CI. Unlike bash commands, node should always be available
- Thirdly, installing `rsync` on windows runners

Additionally this reverts the special treatment of packaging in Windows builds.

Resolves #2761

---

### DIFF
--- a/projenrc/windows-build.ts
+++ b/projenrc/windows-build.ts
@@ -75,6 +75,13 @@ export class WindowsBuild extends Component {
       ),
       ...skippedStepPatches,
 
+      // Convert original build step to a bash shell matrix step
+      ...convertStepToShellMatrix(buildJobPath("/steps/3"), "bash"),
+
+      // Copy original build step as a cmd shell matrix step
+      JsonPatch.copy(buildJobPath("/steps/3"), buildJobPath("/steps/4")),
+      ...convertStepToShellMatrix(buildJobPath("/steps/4"), "cmd"),
+
       // Install rsync on Windows
       JsonPatch.add(buildJobPath("/steps/0"), {
         name: "Install rsync on Windows",
@@ -111,4 +118,18 @@ export class WindowsBuild extends Component {
       })
     );
   }
+}
+
+/**
+ * Helper function to convert an existing step to a fake shell matrix step.
+ * This is because 'shell' currently does not support expressions and thus matrix inputs.
+ */
+function convertStepToShellMatrix(pathToStep: string, shell: string) {
+  return [
+    JsonPatch.add(
+      `${pathToStep}/if`,
+      `\${{ matrix.runner.shell }} == '${shell}'`
+    ),
+    JsonPatch.add(`${pathToStep}/shell`, shell),
+  ];
 }
--- a/projenrc/windows-build.ts
+++ b/projenrc/windows-build.ts
@@ -131,10 +131,7 @@ function convertStepToShellMatrix(
 ) {
   return [
     JsonPatch.add(`${pathToStep}/name`, `${name} (${shell})`),
-    JsonPatch.add(
-      `${pathToStep}/if`,
-      `\${{ matrix.runner.shell }} == '${shell}'`
-    ),
     JsonPatch.add(`${pathToStep}/shell`, shell),
+    JsonPatch.add(`${pathToStep}/if`, `matrix.runner.shell == '${shell}'`),
   ];
 }
--- a/projenrc/windows-build.ts
+++ b/projenrc/windows-build.ts
@@ -46,13 +46,11 @@ export class WindowsBuild extends Component {
             {
               os: "ubuntu-latest",
               primary_build: true,
-              shell: "bash",
               allow_failure: false,
             },
             {
               os: "windows-latest",
               primary_build: false,
-              shell: "cmd",
               allow_failure: false,
             },
           ],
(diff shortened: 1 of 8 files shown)


mrgrain:
@gmeligio FYI Trying if this small change will be enough.

gmeligio:
You did a good one. I tried test and other shell variants to make an small change with shx but couldn't make it work. I ended up with the JS script. I'll check it out anyways and see if I can reuse something from this PR.

gmeligio:
Looking at [where `rsync` failed ](https://github.com/projen/projen/actions/runs/9996668356/job/27631538201?pr=3723#step:5:2988), it might be enough to install rsync in the step or the job.

mrgrain:
> Looking at [where `rsync` failed ](https://github.com/projen/projen/actions/runs/9996668356/job/27631538201?pr=3723#step:5:2988), it might be enough to install rsync in the step or the job.

Good idea 🤔

gmeligio:
I think the specific windows build step can't be removed, because `shell` doesn't accept variables nor matrix. We are assuming and testing with the same shell as NodeJS default in local which is `cmd`, while the windows runner sets `pwsh` by default, like in this job https://github.com/projen/projen/actions/runs/9998223427/job/27636610342?pr=3723#step:6:3 .

I got curious, and I tried to reproduce the command you created for the package step locally by cloning your branch. But the `test` command is not found in my local Windows 11 in pwsh or cmd. It could be that it's available in the runner, but I haven't found the documentation that mentions that. It could be that I'm missing something about the runner, but it seems that we need to think more about the cross-platform `if` condition.

```bash
❯ yarn package  
yarn run v1.22.19
$ node ./projen.js package
👾 package | test -n "${CI}" && ((rsync -a . .repo --exclude .git --exclude node_modules && rm -rf dist && mv .repo dist) || exit 1) || node ./projen.js package-all
'test' is not recognized as an internal or external command,
operable program or batch file.
👾 package-all » package:js | jsii-pacmak -v --target js
[jsii-pacmak] [INFO] Found 1 modules to package
[jsii-pacmak] [INFO] Packaging NPM bundles
[jsii-pacmak] [INFO] Loading jsii assemblies and translations
[jsii-pacmak] [INFO] Packaging 'js' for projen
[jsii-pacmak] [INFO] js finished
[jsii-pacmak] [INFO] Packaged. npm pack (3.4s) | js (0.0s) | load jsii (0.0s) | cleanup (0.0s)
👾 package-all » package:java | jsii-pacmak -v --target java
[jsii-pacmak] [INFO] Found 1 modules to package
[jsii-pacmak] [INFO] Packaging NPM bundles
```

What do you think?

gmeligio:
Actually, I was amazed when I noticed the shell used was pwsh, because some tests with tricky quoting or shell specifics have failed for me in pwsh but worked in cmd.

mrgrain:
Hmm. I was thinking the `test` command is made available through `shx` which I thought runs all projen commands. 🤔

mrgrain:
Regarding `shell`: This was something you originally added to the matrix in the big compat PR I think!? But looking at it now, it doesn't seem to be used. 

But then Windows tests are passing and running in this PR, so I'm a little confused about the need for it in the first place. 🤔

mrgrain:
> Hmm. I was thinking the `test` command is made available through `shx` which I thought runs all projen commands. And in this PR the package step seems to work without problems 🤔

This was wrong. Shx only runs for a pre-defined list of commands. `test` is not one of them. I guess it is just available in powershell.

mrgrain:
> Regarding `shell`: This was something you originally added to the matrix in the big compat PR I think!? But looking at it now, it doesn't seem to be used.
> 
> But then Windows tests are passing and running in this PR, so I'm a little confused about the need for it in the first place. 🤔

Okay, I think I am following you know @gmeligio . Out of curiosity: What was the reason you preferred testing with `cmd` over `powershell`?

mrgrain:
Alright @gmeligio I think I have addressed all the issues now.

gmeligio:
> Okay, I think I am following you know @gmeligio . Out of curiosity: What was the reason you preferred testing with `cmd` over `powershell`?

My reason was to use the same shell a contributor will use locally. Multiple sources say that NodeJS will use `cmd` by default in Windows. Here is one in the docs https://docs.npmjs.com/cli/v10/using-npm/config#script-shell

Then, the runner default was `push` and wanted to force `cmd` in the job, thinking that would set the shell for NodeJS. But I was wrong. NodeJS uses `cmd` shell in Windows by default, regardless of the shell of the caller command.

So, I understand now that `cmd` doesn't need to be forced, and it's OK to use the runner shell as it is. Thank you for mentioning this.

I'll do a review later today.

projenrc/windows-build.ts
```diff
@@ -70,38 +58,45 @@ export class WindowsBuild extends Component {
...
-        "${{ matrix.runner.primary_build }}"
-      ),
+      // Convert original build step to a bash shell matrix step
+      ...convertStepToShellMatrix("build", buildJobPath("/steps/3"), "bash"),
 
-      // Skip steps that shouldn't run on Windows
-      ...skippedStepPatches,
+      // Copy original build step as a cmd shell matrix step
+      JsonPatch.copy(buildJobPath("/steps/3"), buildJobPath("/steps/4")),
+      ...convertStepToShellMatrix("build", buildJobPath("/steps/4"), "cmd"),
```
gmeligio, line 83: Not needed anymore. As we discussed, NodeJS will use cmd by default which is the expected one, regardless of the shell of the caller command.

projenrc/windows-build.ts
```diff
@@ -70,38 +58,45 @@ export class WindowsBuild extends Component {
...
-        run: windowsBuild,
-        shell: "cmd",
-        if: "${{ !matrix.runner.primary_build }}",
-      }),
+      // Install rsync on Windows
+      JsonPatch.add(buildJobPath("/steps/0"), {
+        name: "Install rsync on Windows",
+        if: `matrix.runner.os == 'windows-latest'`,
+        run: "choco install --no-progress rsync",
+        shell: "pwsh",
```
gmeligio, line 90: No need to specify the shell here. pwsh is already the default in the Windows runner.

gmeligio (changes requested):
Looks great. Only a couple of changes about the shell. Sorry for the back and forth regarding that.

The task condition is a very helpful feature. I didn't know about that.

gmeligio (changes requested):
Looks great. Only a couple of changes about the shell. Sorry for the back and forth regarding that.

The task condition is a very helpful feature. I didn't know about that.

gmeligio (approved):
Perfect! Thanks.

iliapolo (approved):

PR #3745, opened 2024-07-26
fix(bump): exclamation operator `!` is not recognized when running bump on windows

gmeligio:
Fixes https://github.com/projen/projen/issues/2427

The `!` mark is used in bash because is interpreted by the `test` command. Because `test` doesn't exist in Windows cmd, then running the bump task fails. It turns out that `grep` has a flag `-v` to invert the match and works both in bash and cmd, and it solved the problem. Refactored the `publishTask.condition` to use the same expression. Added a test to check the correct handling of a release commit with message `chore(release):`.

---

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -40,7 +40,7 @@
           "builtin": "release/bump-version"
         }
       ],
-      "condition": "git log HEAD --exit-code --oneline --grep \"chore(release):\""
+      "condition": "git log --oneline -1 | grep -qv \"chore(release):\""
     },
     "bundle:task-runner": {
       "name": "bundle:task-runner",
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -40,7 +40,7 @@
           "builtin": "release/bump-version"
         }
       ],
-      "condition": "(git log --oneline -1 | grep -q \"chore(release):\") && exit 1 || exit 0"
+      "condition": "git log HEAD --exit-code --oneline --grep \"chore(release):\""
     },
     "bundle:task-runner": {
       "name": "bundle:task-runner",
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -40,7 +40,7 @@
           "builtin": "release/bump-version"
         }
       ],
-      "condition": "! git log --oneline -1 | grep -q \"chore(release):\""
+      "condition": "(git log --oneline -1 | grep -q \"chore(release):\") && exit 1 || exit 0"
     },
     "bundle:task-runner": {
       "name": "bundle:task-runner",
--- a/src/release/publisher.ts
+++ b/src/release/publisher.ts
@@ -15,6 +15,7 @@ import {
 import { defaultNpmToken } from "../javascript/node-package";
 import { Project } from "../project";
 import { GroupRunnerOptions, filteredRunsOnOptions } from "../runner-options";
+import { CHANGES_SINCE_LAST_RELEASE } from "../version";
 
 const PUBLIB_VERSION = "latest";
 const GITHUB_PACKAGES_REGISTRY = "npm.pkg.github.com";
@@ -261,7 +262,7 @@ export class Publisher extends Component {
         PROJECT_CHANGELOG_FILE: projectChangelogFile ?? "",
         VERSION_FILE: versionFile,
       },
-      condition: '! git log --oneline -1 | grep -q "chore(release):"',
+      condition: CHANGES_SINCE_LAST_RELEASE,
     });
     if (projectChangelogFile) {
       publishTask.builtin("release/update-changelog");
--- a/src/version.ts
+++ b/src/version.ts
@@ -3,6 +3,12 @@ import { Component } from "./component";
 import { Project } from "./project";
 import { Task } from "./task";
 
+// this command determines if there were any changes since the last release
+// (the top-most commit is not a bump). it is used as a condition for both
+// the `bump` and the `release` tasks.
+export const CHANGES_SINCE_LAST_RELEASE =
+  '(git log --oneline -1 | grep -q "chore(release):") && exit 1 || exit 0';
+
 /**
  * Options for `Version`.
  */
@@ -69,12 +75,6 @@ export class Version extends Component {
 
     const versionInputFile = options.versionInputFile;
 
-    // this command determines if there were any changes since the last release
-    // (the top-most commit is not a bump). it is used as a condition for both
-    // the `bump` and the `release` tasks.
-    const changesSinceLastRelease =
-      '! git log --oneline -1 | grep -q "chore(release):"';
-
     const changelogFile = posix.join(
       options.artifactsDirectory,
       this.changelogFileName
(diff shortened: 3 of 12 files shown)


gmeligio:
Thanks for tidying it up. I did a couple of manual tests on Windows and it's not working. I'm reading about --exit-code and trying to understand if it would work in `git log` but so far it seems it doesn't help. I'll replace `HEAD` by `-1` because it was showing a list of commit instead of the last one.

I need to switch to draft anyways because because I put the tests in `test/release/bump.test.ts` but it's not testing these PR's changes. I'll find a way to move the tests to `test/version.test.ts`.

I also noticed that this `condition` might not be needed for `bump` at least, because there is this logic inside the task that checks if there any commits after the latest tag and the convention is the "chore(release):" message is associated to the latest tag. Can you check if the condition is still necessary?
https://github.com/projen/projen/blob/5f88a9c60ac92ba1bf7bfc9ad2f26fc7fff031d5/src/release/bump-version.ts#L156-L161

mrgrain:
> Thanks for tidying it up. I did a couple of manual tests on Windows and it's not working. I'm reading about --exit-code and trying to understand if it would work in `git log` but so far it seems it doesn't help. 

Hmm, that's a pity. `--exit-code` isn't well documented on log. It's supposed to work the same as `diff`, i.e. fail if something is returned.

> I'll replace `HEAD` by `-1` because it was showing a list of commit instead of the last one.

So that's not right either. Before we used `-1` to always return the latest commit (because we didn't add any other filters to the log command) and we would then pipe that last commit to `grep` to check if it matches or not. 

However in my proposal I use the `--grep` flag on log. That works differently, it's a filter. So if we keep using `-1` what this means is return the first commit that matches the filter. Which would at some point just always return a release commit. Instead we need to just consider the recent commit, hence the use of `HEAD`.

At least that's the theory. 😅

> I also noticed that this `condition` might not be needed for `bump` at least, because there is this logic inside the task that checks if there any commits after the latest tag and the convention is the "chore(release):" message is associated to the latest tag. Can you check if the condition is still necessary?

I'll have a look 👀

mrgrain:
Okay, I've found this SO post https://stackoverflow.com/questions/32853972/what-does-git-log-exit-code-mean

And it seems like this doesn't work if the commits are empty. And maybe the behaviour is undefined anyway.

mrgrain:
> However in my proposal I use the --grep flag on log. That works differently, it's a filter. So if we keep using -1 what this means is "return the first commit that matches the filter". Which would always return something commit, as long a commit has happened in the past. So instead we need to just consider the most recent commit, hence the use of HEAD.

Yeah, nah. I was wrong here as well. Not sure how I ended up with this. I tried `HEAD^..HEAD` but annoyingly that fails if there is only a single commit. 🤔

mrgrain:
Some more research. The addition of `-v` to the grep command inverts the query. And therefore should also allow us to achieve the desired outcome without `!`. Looks like [shx/shelljs also support it](https://www.npmjs.com/package/shelljs#grepoptions-regex_filter-file_array).

Basically:

```
git log --oneline -1 | grep -qv "chore(release):"
```

mrgrain:
> I need to switch to draft anyways because because I put the tests in test/release/bump.test.ts but it's not testing these PR's changes. I'll find a way to move the tests to test/version.test.ts.

Hmm that is bad. How come they are not tested? Is it that the condition is bypassed?

mrgrain:
> I also noticed that this `condition` might not be needed for `bump` at least, because there is this logic inside the task that checks if there any commits after the latest tag and the convention is the "chore(release):" message is associated to the latest tag. Can you check if the condition is still necessary?

I think technically you are correct. Further more, these commits are only added by the [update-changelog task](https://github.com/projen/projen/blob/c86bb9a1fd2794b0385d5b9da333fe273a4fd535/src/release/update-changelog.ts#L89) which is only added by the [publishToGit](https://github.com/projen/projen/blob/a702840a190828abbb1e987726d5e8f70cc64aa6/src/release/publisher.ts#L268) method,  which in turn is only [added for manual release triggers](https://github.com/projen/projen/blob/8b225dcdacb3aacebf368b2e06abdbb39d62c0dc/src/release/release.ts#L637).

So I suppose in future we could really limit when this condition is added or even roll it into `releasableCommits` as something that is added when manual release triggers are used.

---

On the other hand, this doesn't seem to hurt. I'd rather keep this for now. It seems like we have a number of good solutions to fix the windows issue.

gmeligio:
> Some more research. The addition of `-v` to the grep command inverts the query. And therefore should also allow us to achieve the desired outcome without `!`. Looks like [shx/shelljs also support it](https://www.npmjs.com/package/shelljs#grepoptions-regex_filter-file_array).
> 
> Basically:
> 
> ```
> git log --oneline -1 | grep -qv "chore(release):"
> ```

Great investigation! Thank you. This is perfect, and it's still simple.

> Hmm that is bad. How come they are not tested? Is it that the condition is bypassed?
> 
> If moving to `test/version.test.ts` is unfeasible, we might get away with running and inspecting the command directly.

I mean the `function bump(cwd: string, options: BumpOptions)` is tested but I couldn't find a test for the bump task **condition**. After further thinking, I will try to clone and modify the function `testBump` to add the task condition tests and keep them in `test/release/bump.test.ts`. It should be something similar to the test *newline at the end of version file > existing version file keeps newline at the end*.

> I think technically you are correct. Further more, these commits are only added by the [update-changelog task](https://github.com/projen/projen/blob/c86bb9a1fd2794b0385d5b9da333fe273a4fd535/src/release/update-changelog.ts#L89) which is only added by the [publishToGit](https://github.com/projen/projen/blob/a702840a190828abbb1e987726d5e8f70cc64aa6/src/release/publisher.ts#L268) method, which in turn is only [added for manual release triggers](https://github.com/projen/projen/blob/8b225dcdacb3aacebf368b2e06abdbb39d62c0dc/src/release/release.ts#L637).
> 
> So I suppose in future we could really limit when this condition is added or even roll it into `releasableCommits` as something that is added when manual release triggers are used.
> 
> On the other hand, this doesn't seem to hurt. I'd rather keep this for now. It seems like we have a number of good solutions to fix the windows issue.

Thank you for analyzing if the condition is necessary. I understand better now and will continue with our current direction. Maybe I'll look later into how to move the condition into `releasableCommits`.

gmeligio:
> Some more research. The addition of `-v` to the grep command inverts the query. And therefore should also allow us to achieve the desired outcome without `!`. Looks like [shx/shelljs also support it](https://www.npmjs.com/package/shelljs#grepoptions-regex_filter-file_array).
> 
> Basically:
> 
> ```
> git log --oneline -1 | grep -qv "chore(release):"
> ```

I tested `shx grep` as well to avoid the grep dependency on Windows, but the exit code doesn't behave the same as `grep.` So, I discarded shx in this case.

mrgrain:
@gmeligio Just to confirm, `grep -qv` works on windows?

gmeligio:
> @gmeligio Just to confirm, `grep -qv` works on windows?

Yes, it works on Windows.

mrgrain (approved):

mrgrain:
> For more clarity, grep is not preinstalled on Windows (it is somehow installed in the Github Windows runner). But it can be installed, and then that command will work.

Yeah, I'm good with that for now. I suppose long-term we may not want to use shell commands at all. Everything would need to be built-in. But that's for another day.

PR #3747, opened 2024-07-27
feat(typescript):  speedup jest tests runtime

gmeligio:
The Windows test suite is taking close to 20 min, and I'm investigating some general improvements that will speed up the tests in both Unix and Windows. The slowness in Windows is mainly to the filesystem usage by the tests and by Jest. It's a common topic in NodeJS project and not specific to projen.

## Changelog

Enable isolatedModules in `ts-jest`.

---


src/typescript/typescript.ts
```diff
@@ -720,6 +720,7 @@ export class TypeScriptProject extends NodeProject {
         TypeScriptProject.DEFAULT_TS_JEST_TRANFORM_PATTERN]: new Transform(
           "ts-jest",
           {
+            isolatedModules: true,
```
mrgrain, line 723: I know you are still testing, but in the final PR this should just be set on the config for this project, not as a default for everyone.

gmeligio:
I tried some transpiring improvements but couldn't make them work. Transpilating would be like a 1% improvement only. The root cause of the slowness is the I/O problem in Node in Windows because of how Node uses NTFS disks. On top of that, Jest also has slowness problems in Windows. I'll close for now because I don't have any more ideas. Hopefully, new Node, Jest, and Windows versions will improve this.

mrgrain:
> I tried some transpiring improvements but couldn't make them work. Transpilating would be like a 1% improvement only. The root cause of the slowness is the I/O problem in Node in Windows because of how Node uses NTFS disks. On top of that, Jest also has slowness problems in Windows. I'll close for now because I don't have any more ideas. Hopefully, new Node, Jest, and Windows versions will improve this.

Okay no worries. Thanks for investigating though! 🙏

Issue #3761, opened 2024-08-06
Race condition during release


iliapolo:
Normally, a projen release has the following steps:

> For the sake of the example, assume the latest GitHub tag for a package is `2.3.13`

1. Fetch the latest tag from GitHub to determine the next version number -> 2.3.14
2. Build the source code and populate `package.json` with the version.
3. Publish to NPM -> `https://www.npmjs.com/package/my-package/v/2.3.14`
4. Create GitHub tag --> `https://github.com/cdk8s-team/cdk8s-plus/tree/my-package/v2.3.14`

However, when releases run concurrently, the following can happen:

> For the sake of the example, assume commit B can very soon after commit A.

1. *ReleaseA:* Fetch the latest tag from GitHub to determine the next version number -> 2.3.14
2. *ReleaseB:* Fetch the latest tag from GitHub to determine the next version number -> 2.3.14 
3. *ReleaseA:* Build the source code and populate `package.json` with the version.
4. *ReleaseB:* Build the source code and populate `package.json` with the version.
5. *ReleaseA*:  Publish to NPM -> `https://www.npmjs.com/package/my-package/v/2.3.14`
6. *ReleaseB*: Publish to NPM -> Skipped because the version already exists # Oops, this means commit B isn't released.
7. *ReleaseA*:  Create GitHub tag --> `https://github.com/cdk8s-team/cdk8s-plus/tree/my-package/v2.3.14`
8. *ReleaseB*:  Create GitHub tag --> Not sure what happens, does it override?

Note that steps 7 and 8 can also come in the reverse order, which creates a situation where the GitHub tag differs from the artifact in NPM. 

The race condition is rooted in the fact that GitHub tags are used as a common state. I think this approach should be revisited, but its also a pretty big change. In the meantime, we can limit the release workflow concurrency by [using concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-concurrency).

iliapolo:
Oh I see concurrency is already available on a workflow:

https://github.com/projen/projen/blob/87aa8cd6e5d7326ff8a80d3a69fc249cf2effc8d/src/github/workflows.ts#L27

Its weird its a string though, as it has multiple configuration options. @mrgrain How do you feel about breaking this API to support both `group` and `cancel-in-progress` and also make the release workflow use this by default?

mrgrain:
> Oh I see concurrency is already available on a workflow:
> 
[...]

PR #3762, opened 2024-08-06
fix(github)!: github workflow concurrency api is out of date

iliapolo:
According to [GitHub Docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-concurrency#overview), the `concurrency` setting can accept multiple options, and not just a string. 

Its likely that in the 3 years since this feature was originally introduced, GitHub changed the API, and so should we.

In preparation of https://github.com/projen/projen/issues/3761

BREAKING CHANGE: `concurrency` option on `GithubWorkflow` now accepts multiple configuration options instead of a string.

---


src/github/workflows.ts
```diff
@@ -18,13 +42,21 @@ export interface GithubWorkflowOptions {
    * @default false
    */
   readonly force?: boolean;
+
+  /**
+   * Enable concurrency limitations. Use `concurrencyOptions` to configure specific non default values.
+   *
+   * @default false
+   */
+  readonly limitConcurrency?: boolean;
```
iliapolo, line 51: Since all options have default values, we need this to differentiate between *disable* and *enable with defaults*

test/github/github-workflow.test.ts
```diff
@@ -6,31 +6,42 @@ import { synthSnapshot, TestProject } from "../util";
...
-  test("Default concurrency allowed", () => {
+  test("conccurency is not set by default", () => {
     const project = new TestProject();
 
     new GithubWorkflow(project.github!, workflowName);
 
     const snapshot = synthSnapshot(project);
 
-    expect(snapshot[`.github/workflows/${workflowName}.yml`]).not.toContain(
-      "concurrency"
```
iliapolo, line 17: This is very brittle assertion as this word may appear in the workflow in a different context. Happy to make a specific structural assertion but figured a snapshot would be ok too?

mrgrain: Let's add structural assertions please.

mrgrain (changes requested):

test/github/github-workflow.test.ts
iliapolo: Added

mrgrain (approved):
Thank you!

PR #3763, opened 2024-08-06
fix(release): reduce likelihood for race condition by limiting concurrency

iliapolo:
Without concurrency limits, we can have two runs of the same workflow in parallel:

Screenshot 2024-08-06 at 15 47 01

After adding:

```yaml
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: false
```

The later one will be pending until the first one finishes:

Screenshot 2024-08-06 at 16 09 22

Fixes https://github.com/projen/projen/issues/3761

---


src/github/workflows.ts
```diff
@@ -116,6 +116,7 @@ export class GithubWorkflow extends Component {
 
     const defaultConcurrency: ConcurrencyOptions = {
       cancelInProgress: false,
+      group: "${{ github.workflow }}",
```
iliapolo, line 119: Apparently `group` is required...

Screenshot 2024-08-06 at 15 51 48

iliapolo:
@mrgrain Retries are also put into pending if there another workflow running:

Screenshot 2024-08-07 at 10 58 33
Screenshot 2024-08-07 at 10 58 41

So say maven publishing for *ReleaseA* failed, and *ReleaseB* is now running...retrying the maven publishing for *ReleaseA* will wait until *ReleaseB* finishes. When the retry starts up again, our condition:

```console
if: needs.release.outputs.tag_exists != 'true' && needs.release.outputs.latest_commit == github.sha
```

Should make the retry no-op.

> Question: Will the `release` job run as part of the retry in this case? if not, outputs are now stale and this protection won't work...?

Regardless though there is still a problem. ReleaseB and ReleaseA have the same target version, so it might be the commitA published to maven, but commitB was published to pypi, both under the same version... :\

Ideally we should not retry old releases if there was a later release already triggered. When we manually run workflow, we need to take this into account. And we also should make sure our retry automation doesn't retry old releases. 

I think moving the tag based triggers might solve this actually...we should discuss it. 

But in any case, I believe this PR can still be merged because it makes these problems less likely, and doesn't introduce new ones.

iliapolo:
We decided to first battle-test this change using JsonPatch on some of our own projects before we push it to everyone.

mrgrain (approved):

Issue #3545, opened 2024-04-29
NpmConfig uses package 'ini' which adds CRLF on Windows


pavlovecjakePFG:
I ran into an issue where developers on Windows generated a different .npmrc file with the NpmConfig class then developers on Ubuntu. The cause is the use of the 'ini' package for the .npmrc file: 
https://github.com/projen/projen/blob/6ee24aaded54c9ceaaae8ee5c7472e17c89b87cb/src/javascript/npm-config.ts#L38. 
That package adds CRLF on Windows, but just LF on Ubuntu. 

This results in the mutation check constantly failing as different developers commit slightly different versions of the .npmrc file. 

I believe this file could just as easily be created with TextFile, which would avoid this problem.

mrgrain:
We can upgrade to the latest version of the ini library and hardcode the platform option to unix.

https://www.npmjs.com/package/ini?activeTab=readme

gmeligio:
Hardcoding the platform in `INI.stringify` would work.

But this could happen to multiple file formats managed by multiple packages and tools. A more general solution could be to hardcode the Linux line ending when `.gitattributes` is generated, moving the responsibility to git instead of the particular package. I faced a similar problem when doing changes in projen itself, and `gitattributes` worked well in:

https://github.com/projen/projen/blob/f34f8a020c6a7cb41688b867c69abcb95b921924/projenrc/index.ts#L346

What do you think of using gitattributes?

mrgrain:
Yep I didn't know gitattributes can do that until you did it! 

It's probably a more generic and therefore better solution. I guess we can make it a setting and default to using Unix linebreaks.

PR #3784, opened 2024-08-20
feat(gitattributes)!: support endOfLine with Linux default

gmeligio:
Fixes https://github.com/projen/projen/issues/3545

Adds endOfLine option with LF default to avoid git mutations. This default can be overridden by later lines set by the user. It can be disabled by setting it to `{ endOfLine: EndOfLine.NONE}`, in case of a breaking change or other reason. Added tests for this new option and for other logic of `GitAttributesFile` to improve coverage. 

BREAKING CHANGE: A new gitattribute is added by default to normalize line endings (eol) of text files to LF. This prevents mutations when Windows and Linux/macOS users are working on the same project. To revert back to the previous behavior (line endings behavior is defined by user config), add the option `endOfLine: EndOfLine.NONE` to the project.

---


mrgrain:
Initial thoughts:
- `endOfLine` is better than `eol`. Maybe `lineEndings` is even better?
- This is probably a breaking change. What are the instructions for someone to revert to the previous behavior?
- This currently always adds `* text=auto [eol=(lf|crlf)]` to gitattributes.  To my understanding this will prevent a user configuration from being used, even if it was the intention to not decide for the user.

gmeligio:
Thank you for the review!

> Initial thoughts:
> 
> * `endOfLine` is better than `eol`. Maybe `lineEndings` is even better?

Thank you for the advice. I think I'll settle on `endOfLine` to be consistent with the abbreviation eol and because it's called like that in other tools (Prettier, NodeJS, etc).

> * This is probably a breaking change. What are the instructions for someone to revert to the previous behavior?

To prevent and revert in the case of a breaking change, I just added the option `EndOfLine.NONE`, which disables this, and the line is not added. I added instructions in the PR description and also updated the docblock. I took inspiration for NONE from `JobPermissions` and a few other places:
https://github.com/projen/projen/blob/3d70e77b9b65788994106cc5c0c70f572cacc5ca/src/build/build-workflow.ts#L62-L67

> * This currently always adds `* text=auto [eol=(lf|crlf)]` to gitattributes.  To my understanding this will prevent a user configuration from being used, even if it was the intention to not decide for the user.

I understand you refer to not allowing to override when you say "preventing user configuration". If my interpretation is correct, then the user can still override this configuration. The rules in .gitattributes work like in .gitignore, from top to bottom. Then this first line is only the default but the user config that would be below this line, can override it. It's important to keep this line as the first one to allow overriding it later.

In other words, this line will set the default for all the _detected_ text files but it can be overridden by any user configuration that is below this line.

src/project.ts
```diff
@@ -135,6 +135,15 @@ export interface GitOptions {
    * @default - No files stored in LFS
    */
   readonly lfsPatterns?: string[];
+
+  /**
+   * The default end of line character for text files.
+   *
+   * endOfLine it's useful to keep the same end of line between Windows and Unix operative systems for git checking/checkout operations. Hence, it can avoid simple repository mutations consisting only of changes in the end of line characters. It will be set in the first line of the .gitattributes file to make it the first match with high priority but it can be overriden in a later line. Can be disabled by setting explicitly: `{ endOfLine: EndOfLine.NONE }`.
```
mrgrain, line 142:
```suggestion
   * endOfLine it's useful to keep the same end of line between Windows and Unix operative systems for git checking/checkout operations.
   * Hence, it can avoid simple repository mutations consisting only of changes in the end of line characters.
   * It will be set in the first line of the .gitattributes file to make it the first match with high priority but it can be overriden in a later line.
   * Can be disabled by setting: `endOfLine: EndOfLine.NONE`.
```

mrgrain (changes requested):

mrgrain:
Looks great! One small change please. 

---

I actually meant how can you configure projen to not make a decision at all, and use the local setting of a user on a machine. And a different setting for a different user on a different machine. 

But `EndOfLine.None` does that! So I'm happy 😄

mrgrain (approved):

PR #3791, opened 2024-08-22
fix(bump): cannot release multiple prereleases for 0.0.0

### LINKED ISSUE #3788: Commits are not updating version
I've made an initial commit that built version 0.0.0-alpha.0 (using prerelease `alpha`) and github automatically ran the package distribution workflows. I made some changes to my source and committed a change with a message like "fix: new update" and pushed. The release workflow runs, but the version isn't updated and the package distribution jobs are skipped. The result of the "Check if new version has been tagged" step results in: 

```
Run TAG=$(cat dist/dist/releasetag.txt)
57a03007c6e4c8bc788984be2cf3788a97b4ad29	refs/tags/my-package/v0.0.0-alpha.0
exists=true
```

What am I doing wrong? I thought every conventional commit would increase the version?

mrgrain:
Fixes #3788

---

### DIFF
--- a/test/release/bump.test.ts
+++ b/test/release/bump.test.ts
@@ -80,6 +80,22 @@ test("first release, with prerelease", async () => {
   expect(result.bumpfile).toStrictEqual("0.0.0-beta.0");
 });
 
+test("two initial releases, with prerelease", async () => {
+  const resultA = await testBump({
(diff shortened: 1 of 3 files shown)


mrgrain:
closing stale draft

PR #3792, opened 2024-08-22
fix(python): dynamic environment variables not working on Windows

### LINKED ISSUE #1003: new python project creation crashes on windows
Can't create a python project. `pip` executable is not found by projen, despite it's installed and available.

**Given:**
- OS Windows:
``` 
Edition	Windows 10 Pro
Version	21H1
Installed on	‎26/‎07/‎2021
OS build	19043.1165
Experience	Windows Feature Experience Pack 120.2212.3530.0
```
- projen is installed globally,
- cmd.exe with administator privilege,
- node version: v16.6.2,
- npm version: 7.20.3,
- npx version: 7.20.3,
- python version: 3.9.6,
- pip version: 21.1.3,

When executing the following:
```shell
mkdir newpython
cd newpython
git init
npx projen new python
```

Then, the process ends with an error:
```shell
'pip' is not recognized as an internal or external command,
operable program or batch file.
```

Full log:
```shell
C:\Users\Ivan\PycharmProjects\newpython>git init
Initialized empty Git repository in C:/Users/Ivan/PycharmProjects/newpython/.git/

C:\Users\Ivan\PycharmProjects\newpython>node -v
v16.6.2

C:\Users\Ivan\PycharmProjects\newpython>npm -v
7.20.3

C:\Users\Ivan\PycharmProjects\newpython>npx -v
7.20.3

C:\Users\Ivan\PycharmProjects\newpython>python --version
Python 3.9.6

C:\Users\Ivan\PycharmProjects\newpython>pip --version
pip 21.1.3 from c:\python39\lib\site-packages\pip (python 3.9)

C:\Users\Ivan\PycharmProjects\newpython>npx projen new python
✨ Project definition file was created at C:\Users\Ivan\PycharmProjects\newpython\.projenrc.js
✨ Synthesizing project...
✨ Setting up a virtual environment...
🤖 python -m venv .env
✨ Environment successfully created (located in ./.env).
✨ Installing dependencies...
🤖 install | env: VIRTUAL_ENV=$PWD/.env PATH=$PWD/.env/bin:$PATH
🤖 install | pip install --upgrade pip
'pip' is not recognized as an internal or external command,
operable program or batch file.
C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\tasks\runtime.js:143
                    throw new Error(`Task "${this.fullname}" failed when executing "${command}" (cwd: ${path_1.resolve(cwd !== null && cwd !== void 0 ? cwd : this.workdir)})`);
                    ^

Error: Task "install" failed when executing "pip install --upgrade pip" (cwd: C:\Users\Ivan\PycharmProjects\newpython)
    at new RunTask (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\tasks\runtime.js:143:27)
    at TaskRuntime.runTask (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\tasks\runtime.js:62:9)
    at Pip.installDependencies (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\python\pip.js:56:17)
    at PythonProject.postSynthesize (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\python\python-project.js:152:26)
    at PythonProject.synth (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\project.js:293:18)
    at evalmachine.<anonymous>:19:9
    at Script.runInContext (node:vm:139:12)
    at Object.runInContext (node:vm:289:6)
    at createProject (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\cli\cmds\new.js:120:8)
    at newProject (C:\Users\Ivan\AppData\Roaming\npm\node_modules\projen\lib\cli\cmds\new.js:248:5)

C:\Users\Ivan\PycharmProjects\newpython>
```

gmeligio:
Fixes #1003

Use Python CLI to resolve the dynamic environment variables. Also, the upgrade command for pip was fixed because it needs to be different on Windows.

## Test

1. I tested manually both in Windows and Linux. I didn't add an integration test. I guess the tests for `projen new` cover it.
1. Extracted the duplicated class `TestPythonProject` into test/python/util.ts.

---


src/python/venv.ts
```diff
@@ -44,13 +44,18 @@ export class Venv extends Component implements IPythonEnv {
...
     this.project.addGitIgnore(`/${this.envdir}`);
+
+    const pythonDir = process.platform === "win32" ? "Scripts" : "bin";
+    const venvPathExpression = `\${process.cwd()}/${this.envdir}`;
+    const pythonDirExpression = `${venvPathExpression}/${pythonDir}`;
+
     this.project.tasks.addEnvironment(
       "VIRTUAL_ENV",
-      `$(echo $PWD/${this.envdir})`
+      `$(node -p "\`${venvPathExpression}\`")`
```
mrgrain, line 54: Given that is python, can we use python?

(Node will be fine since it's required for projen, but you know...)

gmeligio: Good idea! I'll check with Python.

src/python/venv.ts
```diff
@@ -44,13 +44,32 @@ export class Venv extends Component implements IPythonEnv {
     this.pythonExec = options.pythonExec ?? "python";
 
     this.project.addGitIgnore(`/${this.envdir}`);
+
+    const isWindows = process.platform === "win32";
+    const pythonDir = isWindows ? "Scripts" : "bin";
```
mrgrain, line 49: This will still change the task definition, depending on the OS that projen is run on.

mrgrain (changes requested):
Code looks good, but the pythonDir check needs to move into the python script as well.

> This adds the withPlatforms test helper to save snapshots for both Linux and Windows with minimal boilerplate. In the case of Python, the virtual environment will be stored in ./.env/Scripts on Windows and ./.env/bin on Linux. It's good to save both snapshots instead of discarding that path difference to have only one snapshot.

I don't think this is the right approach. We want snapshots to be identical on both platforms, otherwise we get the issue with differences between Win and Linux when `projen default` runs.

src/python/venv.ts
gmeligio: Thank you for the review. I didn't notice that it could also be moved inside the script. I'll do that.

src/python/venv.ts
```diff
@@ -44,13 +44,33 @@ export class Venv extends Component implements IPythonEnv {
     this.pythonExec = options.pythonExec ?? "python";
 
     this.project.addGitIgnore(`/${this.envdir}`);
+
+    const pythonDirExpression = `python_dir = 'Scripts' if sys.platform == 'win32' else 'bin'`;
+
+    const isWindows = process.platform === "win32";
+
+    // In Windows CMD the PATH double quote needs to be escaped with another double quotes.
+    const quote = isWindows ? '""' : '\\"';
```
gmeligio, line 53: @mrgrain This quoting part depends on the shell (cmd or bash), which happens before running the Python script, so it can't be moved into the Python script. I haven't found a way to make it cross-platform. As it is right now, it will produce different snapshots for Linux and Windows.

gmeligio: Another place that it comes to my mind is to have a special escape syntax for projen, like `$(` is for expressions, so that evalEnvironment does the escaping depending on the platform.

gmeligio: On the other hand, it's possible that escaping the quotes is not needed after moving the script into a function in `.projenrc.py` or in another generated file.

What's the best option in this case?

mrgrain: To confirm my understanding (using simplified python code):

In bash (or shell) we need to do this:
```bash
python -c "import os; os.environ[\"PATH\"]"
```

But in cmd we need to do this?
```cmd
python -c "import os; os.environ[""PATH""]"
```

mrgrain: I think in this situation we can get around this with adjusting the python syntax:
```bash
python3 -c  "import os; path = os.environ['PATH']; print(f'{path}')"
```
or maybe even this due to better backwards compat:
```bash
"import os; print('{}'.format(os.environ['PATH']));"
```

mrgrain: Now long term, this is obviously super annoying. I think we should write code in one form (likely nix) and not have to worry about the shell it will be executed in. 

Not entirely how that would be implemented, but we can look at what other tools are doing (e.g. npm's [run-script](https://github.com/npm/run-script) or [shelljs](https://www.npmjs.com/package/shelljs) or other tasks runners. This sounds like a hard, but solved problem. It might also just be a case of running commands always throw shelljs.

src/python/venv.ts
```diff
@@ -44,13 +44,33 @@ export class Venv extends Component implements IPythonEnv {
...
+
+    // For example: /my/path/.env
+    const envdirPathExpression = `{os.getcwd()}/${this.envdir}`;
+
+    // For example: /my/path/.env/bin
+    const pythonDirPathExpression = `${envdirPathExpression}/{python_dir}`;
+
+    // For example: /my/path/.env/bin:{os.environ["PATH"]}
+    const pathEnvironmentExpression = `${pythonDirPathExpression}{os.pathsep}{os.environ[${quote}PATH${quote}]}`;
```
mrgrain, line 62: Side note, I think we should accept the duplication here and define the both python code snippets as full strings. The way this is now, it's really hard to argue about.

gmeligio: You're right. It's better to duplicate to read the TS, shell, and Python mix a bit easier.

src/python/venv.ts
gmeligio: > python3 -c  "import os; print('{}'.format(os.environ['PATH']));"

The str.format() actually solved the quote problem. Thank you for providing alternatives. I was trying to use f-string and then working around it instead of taking a step back and find a shell compatible way.

gmeligio: Regarding the long term, I think we can for sure understand how other shell tools do it and design a simple and cross-platform way for projen. I need to investigate more before proposing an RFC or some plan.

From my point of view, projen is already more advanced than the other tools. For example: npm run-script, shelljs don't manage cross-platform environment variables. But there is room for simplification in terms of standardizing the projen syntax for scripts, likely shell/*nix as you said.

mrgrain (approved):

gmeligio:
Thank you again for the review @mrgrain ! Sorry, I just updated the PR description because it was outdated after removing `withPlatforms`. I hope there is still time before the actual commit message is created.

mrgrain:
Ah great! It wasn't merged yet, so you are all good. Thanks for catching that. <3

PR #3814, opened 2024-08-30
feat(github): set a custom file name for GitHub Workflow files

### LINKED ISSUE #3811: Feature request: Allow setting a different file name vs name for github workflows
Following GithubWorkflow component will generate

```ts
new GithubWorkflow(github, 'This is a workflow name');
```
a file named `this is a workflow name.yml` with the following content.

```yaml
name: This is a workflow name
```

Ideally GithubWorkflow component should allow users to set a custom filename or a custom workflow name. For example, I'd like to create a file named `this-is-a-workflow-name.yml` using snake case.

### LINKED ISSUE #3851: Unable to change GitHub workflow file extension
Since the filename of a GitHub workflow must not change between branches for the same repo (difficulties arise with branch protections in particular, but also `workflow_dispatch` has to exist in the primary branch in the same named file before it can be used), and Projen may be added to a project after it's established (our use case) and we have a `build.yaml` (note the **a**), we cannot use Projen to make the build workflow or we have to change the filename in all of the existing branches (not practical).

The fix is simple: add control over choosing `.yml` or `.yaml` as the file extension. I don't see a practical use case for any other options, so an enum should do, with the current `.yml` being the default if not specified.

shinebayar-g:
This PR allows user to set the file name of the workflow explicitly in an option and falls back to the `name` constructor parameter if not provided.

Fixes #3811
Fixes #3851
Closes #3850

---


src/github/workflows.ts
```diff
@@ -57,6 +57,13 @@ export interface GithubWorkflowOptions {
    * @see https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency
    */
   readonly concurrencyOptions?: ConcurrencyOptions;
+  /**
+   * The name of the workflow. GitHub displays the names of your workflows under your repository's
+   * "Actions" tab. If you omit `name`, GitHub displays the workflow file path relative to the
+   * root of the repository.
+   * @see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#name
+   */
+  readonly name?: string;
```
mrgrain, line 66: We will have to improve the API for this. With your (very welcome!) change we now have two things that are called `name`. The argument doesn't even have a description right now 🙈

mrgrain (changes requested):
I like the change, but can you suggest an API that doesn't use `name` everywhere. And make sure to update (and add where needed) existing doc blocks as well.

shinebayar-g:
Changed `name` parameter in the constructor to `filePath`. I think it's fitting since the underlying Yamlfile also uses the same parameter.

mrgrain:
Thanks for the change. In the meantime #3851 was raised which is somewhat similar.

I think what we should do is flip this around:
- Keep the the `name` parameter as it is. It represents the name of the workflow and will be used to derive a file path.
- Add a new `fileName` option (not path because it's going ti be places into the right path automatically) 
- When `fileName` is provided, we use that **exactly as it is** as the filename. So make sure we add an example showing the file extension.
- If `fileName` option is not provided, we default it to `${name}. yml`

That we we can solve both issues with a not too complicated API.

mrgrain (approved):

PR #3850, opened 2024-09-19
fix: add file extension control to GitHub workflows

### LINKED ISSUE #3851: Unable to change GitHub workflow file extension

giseburt:
Fixes #3851

---

### DIFF
--- a/src/github/workflows.ts
+++ b/src/github/workflows.ts
@@ -9,6 +9,14 @@ import { deepMerge, kebabCaseKeys } from "../util";
 import { YamlFile } from "../yaml";
 
 /**
+ * File extension for the workflow file.
+ */
+export enum WorkflowYamlExtension {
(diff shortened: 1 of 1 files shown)


mrgrain:
Wondering if we can combine this with this: https://github.com/projen/projen/pull/3814

PR #3855, opened 2024-09-21
feat(node)!: default to LTS version in GitHub Actions workflows and match `@types/node` to TypeScript version

mrgrain:
Previously the default node version used in GitHub Actions workflows was explicitly defined by projen and updated periodically. A similar change has been made to the version of `@types/node` for typescript projects. With these changes projen won't have to release an update anymore to update the versions. Instead a sensible default is used that will automatically move forward. If users desire more explicit control over versions, they can opt-in to control these versions themselves.

Also clarifies the intended usage of `minNodeVersion` and `maxNodeVersion` which should not be used in most cases.

BREAKING CHANGE: The `workflowNodeVersion` used in GitHub Actions Workflows now defaults to `lts/*`. To revert to the previous behavior, set `workflowNodeVersion: '18.x'`. 

BREAKING CHANGE: The version of the`@types/node` dev dependency now falls back to match the version of TypeScript and then the latest version if `minNodeVersion` is not set. To revert to the previous behavior, add an explicit devDependency on `@types/node` with the desired version.

---


src/javascript/node-package.ts
```diff
@@ -216,16 +216,35 @@ export interface NodePackageOptions {
...
    *
-   * @default - no "engines" specified
+   * The value indicates that the package is incompatible with any older versions of node.
+   * This requirement is enforced via the engines field.
+   *
+   * You will normally not need to set this option, even if your package is incompatible with EOL versions of node.
+   * Consider this option only if your package depends on a specific feature, that is not available in other LTS versions.
+   * Setting this option has very high impact on the consumers of your package,
+   * as package managers will actively prevent usage with node versions you have marked as incompatible.
+   *
```
rix0rrr, line 229: ❤️

src/javascript/node-project.ts
```diff
@@ -170,9 +170,11 @@ export interface NodeProjectOptions
   readonly releaseToNpm?: boolean;
 
   /**
-   * The node version to use in GitHub workflows.
+   * The node version used in GitHub Actions workflows.
    *
-   * @default - same as `minNodeVersion`
+   * Always use this option if your GitHub Actions workflows require a specific to run.
+   *
+   * @default - `lts/*` or `minNodeVersion` if set
```
rix0rrr, line 177:
```suggestion
   * @default - `minNodeVersion` if set, otherwise `lts/*`.
```

src/typescript/typescript.ts
```diff
@@ -611,6 +601,35 @@ export class TypeScriptProject extends NodeProject {
...
+   *
+   * If the user has already added this dependency, do nothing.
+   * Otherwise match the version to the used typescript version.
+   * If not available, fall back to the major version of `minNodeVersion`.
+   * If not available, we use latest and let the user manage the version.
+   */
+  private addNodeTypesVersion(tsVersion?: string, minNodeVersion?: string) {
+    const name = "@types/node";
+
+    if (this.deps.tryGetDependency(name, DependencyType.BUILD)) {
```
rix0rrr, line 615: Yes it should be a BUILD dependency, but what if a user added it in the wrong place?

src/typescript/typescript.ts
```diff
@@ -611,6 +601,35 @@ export class TypeScriptProject extends NodeProject {
...
+      return;
+    }
+
+    // coerce version, since the ts version likely something like ~5.3.0
+    const tsParsed = semver.coerce(tsVersion);
+    if (tsParsed) {
+      return this.addDevDeps(`${name}@ts${tsParsed.major}.${tsParsed.minor}`);
+    }
```
rix0rrr, line 623: I'm not sure exactly what this `ts5.5` tag is for, but this seems a bit dangerous?

Presumably it's the *latest* `@types/node` package that is compatible with the given TS version? 

Shouldn't we do the lower bound check on `minNodeVersion` before defaulting to "the newest @types/node that we can support" ? You are liable to use Node features that aren't supported on a version of Node you'll be deploying to?

src/typescript/typescript.ts
mrgrain: This certainly is a somewhat undefined oddity in projen at the moment. It exists in a few places. I reckon we would want a set of conditional add methods that check, check and add, check and warn, check and warn and add etc.

But I think this is okay to be out of scope here. The issue already exists pre this PR.

src/typescript/typescript.ts
mrgrain: Swapped the order

rix0rrr (approved):

PR #3861, opened 2024-09-25
feat(node-package): allow reset node package manager

floydspace:
Hello,

Following PR makes an ability to reset the node package manager dynamically.
I have the following use case, the component which decorates the project and sets a new package manager:

```ts
import { Component, Project, javascript } from "projen";

export interface OnlyAllowOptions {
  packageManager?: javascript.NodePackageManager;
}

export class OnlyAllow extends Component {
  constructor(project: javascript.NodeProject, options?: OnlyAllowOptions) {
    super(project);

    project.package.setPackageManager(
      options?.packageManager ?? javascript.NodePackageManager.NPM,
    );

    project.addDevDeps("only-allow");

    project.addScripts({
      preinstall: `npx only-allow ${project.package.packageManager}`,
    });
  }
}

//...

new OnlyAllow(project, {
  packageManager: javascript.NodePackageManager.PNPM
});
```

---


src/javascript/node-package.ts
```diff
@@ -809,6 +807,33 @@ export class NodePackage extends Component {
...
 
+  /**
+   * Sets the package manager for the project and configures it accordingly.
+   *
+   * @param packageManager - The Node Package Manager used to execute scripts.
+   * @param yarnBerryOptions - Optional configuration options for Yarn Berry.
+   */
+  public setPackageManager(
+    packageManager: NodePackageManager,
+    yarnBerryOptions?: YarnBerryOptions
```
mrgrain, line 818:
```suggestion
    options?: NodePackageManager
```

mrgrain (changes requested):
Started reviewing, but I don't think this fits the direction projen wants to go in. The PR adds a lot of imperative functionality, way beyond escape hatching, whereas projen strives to be a declarative tool. You need to make a lot of changes because projen is not setup to work like that.

There are some imperative changes in here, that are probably okay to make it in and we can certainly talk about the blockers you encounter to make work what you need to make work.

---

Taking a step back, what is the actually problem you are trying to solve here. The example you are providing feels rather contrived to me. As a user, it would boil down to the following:

```ts
  packageManager: NodePackageManager.NPM,
  // ...
});

new OnlyAllow(project, {
  packageManager: javascript.NodePackageManager.PNPM
});
```

Which seems rather redundant, to first set the package manager to one tool and than use a component to re-set it to another one. Maybe you can provide more insight into the real world scenario you are trying address.

floydspace:
hi @mrgrain 
thanks for the feedback

indeed `OnlyAllow` example could seem silly, but I just used it for simplicity.
so my main motivation first of all, have an ability to revert changes that was made by a Component, in most cases what I need is to overwrite some built-in projen behaviour, however I cannot just overwrite properties, I also need to revert changes made by internal components.

Yes probably the right way to do that is simply declare correct properties in the native project and declare component you need on top, like your example, and this is how I have to do now indeed. however this approach does not let implementing components with full incapsulation, I ahve to make sure that certain properties are disabled or properly set before using the component, there is an attempt to solve this issue here they author made attempt to implement this kind of encapsulation as [PseudoComponent](https://github.com/mountain-pass/cool-bits-for-projen/blob/aef5cc62918c9f73511533223cb6fa09695568c0/src/util/pseudo-component.ts#L44C1-L48C4), that looks good, but still not perfect, becasue you still have to provide the properties to the project.. I hope I described my issue clearly..

So in general it would be usefull to have some kind of mechanism to reset changes made by any kind of component before they were synthesized. Like if you remove a node from a tree it will also cut the whole branch with their leafs.

But still I can get your point that projen components, like cdk constructs designed to be used declaratively and I can accept it, no worries if we close this PR.

mrgrain:
> hi @mrgrain thanks for the feedback

No worries. This is an interesting and important discussion. Things will likely be moving slow, but it's good we have it.
 
>  there is an attempt to solve this issue here, the author made attempt to implement this kind of encapsulation as [PseudoComponent](https://github.com/mountain-pass/cool-bits-for-projen/blob/aef5cc62918c9f73511533223cb6fa09695568c0/src/util/pseudo-component.ts#L44C1-L48C4), that looks good, but still not perfect, becasue you still have to provide the properties to the project.. I hope I described my issue clearly..

This is interesting. Thanks for sharing. Although I would hesitate to call these components. They look like straight-forward classes to me and effectively serve as as Adapters between a preferred interface and a required interface. 

> indeed `OnlyAllow` example could seem silly, but I just used it for simplicity. so my main motivation first of all, have an ability to revert changes that was made by a Component, in most cases what I need is to overwrite some built-in projen behaviour, however I cannot just overwrite properties, I also need to revert changes made by internal components.
>
> I hope I described my issue clearly.

I'm still not quite sure on the why. Which mostly impedes my ability to suggest you a solution that is more inline with how today's projen works. I'll have a go anyway. My assumption here is that you need to simplify the interface of project options, and introduce new settings on top of the existing projects that have a broader effect. The current best-practice for this is to create your own option interface and a thing wrapper class around existing projects. The options & wrappers are responsible to ensure your desired configuration and validate options accordingly. If you want to re-use existing options, you can use a tool like [jsii-struct-builder](https://github.com/mrgrain/jsii-struct-builder) to easily generate an adjusted copy of it.

For validation within components, the best practice is to throw an error and tell the user what to do. E.g. if `OnlyPnpm` component is used, it should check the value of the package manager and throw if it is incorrect with instructions to the user how to fix it. So you don't make sure that certain properties "are disabled or properly set", instead you check and fail if they are not. Your project options should guide the user to easily make the right choices.

> however this approach does not let implementing components with full incapsulation,

Of course full encapsulation is not possible today with projen. I'd say by design `Project`s don't try to encapsulate either. They are the the glue that is connecting (ideally fully encapsulated) components together. The first difference of projen to CDK is that projen has A LOT of singletons to deal with. In CDK every resource is fairly self contained, they provide information to the outside but rarely change other resources. Policies are probably the biggest exception to this, but even then they tend to not operate as Singletons at the App level, but at a certain point in the construct tree. (You don't have just a single Role/Policy per App). In projen, a lot of configuration files are directory level singletons (`.gitignore`). Some of them are even repo-level singletons (github workflow definitions). The top-level projen project currently effectively acts as a stand-in for a repository. So we'll need to find a way to declare interactions with these singletons in a more constructs native way.

> So in general it would be usefull to have some kind of mechanism to reset changes made by any kind of component before they were synthesized. Like if you remove a node from a tree it will also cut the whole branch with their leafs.
>
> But still I can get your point that projen components, like cdk constructs designed to be used declaratively and I can accept it, no worries if we close this PR. but can we at least link all the changes that are done by a component to a component itself like [here](https://github.com/projen/projen/pull/3861/files#diff-a3cb061d3cad4753225993c3bd0b44de38c7e92c95316e090e2068968a1202a6R384), so removing node would also remove child files generation

This would indeed be ideal. I see two pathways for a solution here:

a) **Remove Node Hook** - Every component gets a hook that is responsible for undoing all the things that need undoing when the construct node is removed from the tree. This is probably straight forward to implement, but will require adding many imperative features that don't exist today and each component will have to implement and maintain almost everything twice (add and remove).

b) **Every addition is a construct** - Instead of using imperative methods, we use constructs to extend additional functionality to such singletons. We would have a `GitIgnoreRules` component that can contain the localized rules for a parent component. The `GitIgnoreFile` singleton than needs to change to be able to collect rules from the tree for synthesis. If the `GitIgnoreRules` component is removed from the tree, nothing can be collected. This probably requires a more careful definition of the pre-synth/synth/post-synth phases to ensure escape hatches can still work.

PR #3863, opened 2024-09-26
fix: bundledDependencies are silently ignored by NPM

rix0rrr:
Certain versions of npm (at least my tested version, `10.8.2`) silently refuse to bundle a dependency if it doesn't occur in `dependencies`, and `dependencies` only. Specifically, if it also occurs in `devDependencies`, `npm pack` will just silently not bundle the directory.

Add a check for that to prevent users from falling into this trap.

Also add a clarification on the `peerDependencies` check. It makes sense, but it took me longer than should be necessary to understand, and I like the computer to explain things to my peanut-sized brain 😉.

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -1257,7 +1257,7 @@ export class NodePackage extends Component {
           // `dependencies` and `dependencies` only.
           if (depDecls.some((d) => d.type === DependencyType.BUILD)) {
             throw new Error(
-              `unable to bundle "${name}": it cannot appear as a devDependency (certain versions of NPM will silently fail to bundle it if it is)`
+              `unable to bundle "${name}": it cannot appear as a devDependency (only prod dependencies are bundled, and any dependency appearing as a devDependency is considered to be not a prod dependency)`
             );
           }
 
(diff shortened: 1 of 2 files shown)


src/javascript/node-package.ts
```diff
@@ -1245,13 +1245,19 @@ export class NodePackage extends Component {
...
+              `unable to bundle "${name}": it cannot appear as a peer dependency (bundled would always take precedence over peer)`
+            );
+          }
+
+          // I've observed that at least npm 10.8.2 will silently fail to bundle
+          // a dependency if it is [also] part of `devDependencies`. It must exist in
+          // `dependencies` and `dependencies` only.
+          if (depDecls.some((d) => d.type === DependencyType.BUILD)) {
+            throw new Error(
+              `unable to bundle "${name}": it cannot appear as a devDependency (certain versions of NPM will silently fail to bundle it if it is)`
```
mrgrain, line 1260:
```suggestion
              `unable to bundle "${name}": it cannot appear as a devDependency (only prod dependencies are bundled, and any dependency appearing as a devDependency is considered to be not a prod dependency)`
```

mrgrain (approved):

PR #3914, opened 2024-10-27
feat(pytest): configurable test path

### LINKED ISSUE #1620: (python) pytestOptions.testdir is not used
`pytestOptions` of `PythonProject` accepts a `testdir` property, which should allow specifying the folder pytest should use to find the tests.

However, it appears this property is never actually used: [pytest.ts](https://github.com/projen/projen/blob/main/src/python/pytest.ts#L34-L39).

mikejgray:
chore(docs): clarify pytestoptions.testdir purpose

Fixes #1620

---


mrgrain (changes requested):
I like the direction of this change, but it's getting more confusing with this.

Couple thoughts:

- if `testdir` is not what it claims to be and isn't even used in the component, we should deprecate it (both as options and component property).
- `testPaths` should probably be exposed as a prop on the component
- should the pytest samples be created outside of pytest?

Looking forward to a proposal that will create more consistency!

mikejgray:
> * if `testdir` is not what it claims to be and isn't even used in the component, we should deprecate it (both as options and component property).
`testdir` is used in the component as the location for sample tests. This is consistent with the NodeProject class IIRC, so deprecating it doesn't seem like the right path forward. Ideally it would be renamed, but I'm not sure what the projen maintainers' policy is on renaming. It wouldn't be a big lift to deprecate `testdir` and point it at a newly named property instead.
> * `testPaths` should probably be exposed as a prop on the component
Good callout, done and coming in a followup commit.
> * should the pytest samples be created outside of pytest?
I have no strong opinion on this. What's the behavior in other project types? PythonProject does not support other Python test modules directly like `unittest` or `nose`, nor does it have the class infrastructure in place to handle that currently.

As a general statement I think it would be appropriate to create some kind of testing base class and subclass that for all project types - one base class per type, as I could imagine the basic testing requirements for Python will differ from Node will differ from Java etc. Pytest is probably the most commonly used, largely because of its support for unittest, but nose has a dedicated following too.

mrgrain:
> > * if `testdir` is not what it claims to be and isn't even used in the component, we should deprecate it (both as options and component property).
> 
> `testdir` is used in the component as the location for sample tests. This is consistent with the NodeProject class IIRC, so deprecating it doesn't seem like the right path forward. Ideally it would be renamed, but I'm not sure what the projen maintainers' policy is on renaming. It wouldn't be a big lift to deprecate `testdir` and point it at a newly named property instead.

Hmm. I suppose I don't like having both `testdir` and `testPaths`. That's seems like duplication of the same thing. From a `Pytest` component perspective I think it would make sense to deprecate `testdir` in favor of the more flexible `testPaths`. That's kind of how the Jest component works as well (although it's patterns there). This applies both to the option and class property.

> > * should the pytest samples be created outside of pytest?
> 
> I have no strong opinion on this. What's the behavior in other project types? PythonProject does not support other Python test modules directly like `unittest` or `nose`, nor does it have the class infrastructure in place to handle that currently.

Sample code seems to be generally created by the projects. So I guess the current location is fine. But instead of using (the now deprecated) `testdir` from pytest, we should use the project testdir `this.testdir` here:
https://github.com/projen/projen/blob/9aa183d14a9fc1902082a7c59e83d1dc46b65ed6/src/awscdk/awscdk-app-py.ts#L107

And in [the line above](https://github.com/projen/projen/blob/9aa183d14a9fc1902082a7c59e83d1dc46b65ed6/src/awscdk/awscdk-app-py.ts#L106) we will need to combine `this.testdir` with `options.pytestOptions.testPaths`.

mrgrain:
@mikejgray Are you happy with my proposal? If you're too busy I can also try and pick it up.

mikejgray:
> @mikejgray Are you happy with my proposal? If you're too busy I can also try and pick it up.

Proposal sounds good! Sorry, I did indeed get busy. I will try to carve out some time to finish this one up

mikejgray:
@mrgrain I took a closer look at what is happening with Node/Jest and discovered that `testdir` is not defined or exposed on the project level in Python. I took the opportunity to expose it and make it match Node, and then changed `testPaths` to `testMatch` for consistency with jest. Bringing the Python resources in line with Node makes a lot of sense here and requires less juggling to accomplish the same goals.

Please let me know what you think and if any other changes are made. I'd like to get this right and preferably in a backwards-compatible way.

src/python/pytest.ts
```diff
@@ -33,13 +52,18 @@ export class Pytest extends Component {
 
     this.testdir = options.testdir ?? "tests";
 
+    this.testMatch = options.testMatch ?? [""];
```
mrgrain, line 55: Why not an empty array?

mrgrain:
Thanks @mikejgray I'm pretty opposed to naming something `testdir` that isn't actually used for that. It's also wrong in the Node/Jest components. If we want to have it, I suggest we call it `sampleTestdir`.

`Pytest` component looks good now!

src/python/pytest.ts
```diff
@@ -11,20 +11,39 @@ export interface PytestOptions {
   readonly version?: string;
 
   /**
-   * Directory with tests
+   * Location of sample tests.
+   * Typically the same directory where project tests will be located.
    *
    * @default 'tests'
+   * @deprecated Reference `testdir` on the project instead.
```
mrgrain, line 18:
```suggestion
   * @deprecated Reference `testdir` on the project instead. To change the directory where tests are discovered from, use `testMatch`.
```

src/python/python-project.ts
```diff
@@ -227,6 +234,12 @@ export class PythonProject extends GitHubProject {
    */
   public pytest?: Pytest;
 
+  /**
+   * Directory where sample tests are stored.
+   * @default "tests"
+   */
+  public readonly testdir: string;
```
mrgrain, line 241:
```suggestion
  public readonly sampleTestdir: string;
```

mrgrain (changes requested):

mikejgray:
Understood - I'll make your requested changes and make sure it's backwards-compatible!

mrgrain:
> Understood - I'll make your requested changes and make sure it's backwards-compatible!

Thank you. Just to be clear: Let's not touch Node/Jest in this PR.

src/python/python-project.ts
```diff
@@ -227,6 +234,12 @@ export class PythonProject extends GitHubProject {
    */
   public pytest?: Pytest;
 
+  /**
+   * Directory where sample tests are stored.
```
mrgrain, line 238:
```suggestion
   * Directory where sample tests are located.
```

src/python/pytest.ts
```diff
@@ -14,7 +14,7 @@ export interface PytestOptions {
    * Location of sample tests.
    * Typically the same directory where project tests will be located.
    *
-   * @default 'tests'
+   * @default tests
```
mrgrain, line 17:
```suggestion
   * @default "tests"
```

src/python/python-project.ts
```diff
@@ -144,7 +144,7 @@ export interface PythonProjectOptions
   /**
    * Location of sample tests.
    * Typically the same directory where project tests will be located.
-   * @default 'tests'
+   * @default tests
```
mrgrain, line 147:
```suggestion
   * @default "tests"
```

src/python/python-project.ts
```diff
@@ -236,7 +236,7 @@ export class PythonProject extends GitHubProject {
 
   /**
    * Directory where sample tests are located.
-   * @default "tests"
+   * @default tests
```
mrgrain, line 239:
```suggestion
   * @default "tests"
```

src/awscdk/awscdk-app-py.ts
```diff
@@ -25,12 +25,6 @@ export interface AwsCdkPythonAppOptions
    * @default "app.py"
    */
   readonly appEntrypoint?: string;
-  /**
-   * Python sources directory.
-   *
-   * @default "tests"
-   */
-  readonly testdir?: string;
```
mrgrain, line 33: I think we still needs this as an input, but deprecated.

src/awscdk/awscdk-app-py.ts
```diff
@@ -76,7 +83,7 @@ export class AwsCdkPythonApp extends PythonProject {
       ...options,
     });
     this.appEntrypoint = options.appEntrypoint ?? "app.py";
-    this.testdir = options.testdir ?? "tests";
+    this.testdir = this.sampleTestdir = options.sampleTestdir ?? "tests";
```
mrgrain, line 86:
```suggestion
    this.testdir = this.sampleTestdir = options.sampleTestdir ?? options.testdir ?? "tests";
```

src/awscdk/awscdk-app-py.ts
```diff
@@ -104,7 +111,7 @@ export class AwsCdkPythonApp extends PythonProject {
 
     if (options.pytest ?? true) {
       this.pytest = new Pytest(this, options.pytestOptions);
-      new AwsCdkPytestSample(this, this.pytest.testdir);
+      new AwsCdkPytestSample(this, this.testdir);
```
mrgrain, line 114:
```suggestion
      new AwsCdkPytestSample(this, this.sampleTestdir);
```

src/python/pytest.ts
```diff
@@ -11,20 +11,39 @@ export interface PytestOptions {
   readonly version?: string;
 
   /**
-   * Directory with tests
+   * Location of sample tests.
+   * Typically the same directory where project tests will be located.
    *
-   * @default 'tests'
+   * @default "tests"
+   * @deprecated Reference `testdir` on the project instead. To change the directory where tests are discovered from, use `testMatch`.
```
mrgrain, line 18:
```suggestion
   * @deprecated Reference `sampleTestdir` on the project instead. To change the directory where tests are discovered from, use `testMatch`.
```

src/python/pytest.ts
```diff
@@ -11,20 +11,39 @@ export interface PytestOptions {
...
+   * The array will be concatenated and passed as a single argument to pytest.
+   * @example ["tests/unit", "tests/qa"]
+   * @default []
+   */
+  readonly testMatch?: string[];
 }
 
 export class Pytest extends Component {
+  /**
+   * @deprecated Use `testdir` on the project instead.
```
mrgrain, line 43:
```suggestion
   * @deprecated Use `sampleTestdir` on the project instead.
```

mrgrain (changes requested):
I know this got very confusing with all the name changes. Think this is the last batch of changes.

src/awscdk/awscdk-app-py.ts
mrgrain: When deprecating, we try to keep the functionality around but with less priority.

mrgrain (approved):

PR #3935, opened 2024-11-07
feat(eslint): upgrade eslint to version ^9

### LINKED ISSUE #3907: hardcoded versions lead to outdated eslint for typescript/javascript projects
I am using projen in a CDK Constructs library project (`AwsCdkConstructLibrary`), and I noticed the eslint version had become unsupported, and incompatible with the (meanwhile updated) typescript version.

I will lead with the exact questions/remarks I have.
A more detailed experience follows.

### Questions

- Is this hard-coded versioning desired, perhaps to guarantee compatibility with the configuration props for eslint at the projen project configuration level?
- Can this behavior be documented, because it was somewhat of a frustrating experience setting devDeps, and seeing no effect for some select packages, because of this overriding behavior.
- Can these dependencies be updated?

### My Experience

Recently, I notices that the eslint version that projen installed was no longer supported:

```
❯ pnpx projen
👾 default | ts-node --project tsconfig.dev.json .projenrc.ts
👾 Installing dependencies...
👾 install | pnpm i --no-frozen-lockfile
 WARN  deprecated eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options.
```

Additionally, it is no longer compatible with the default typescript version that is being installed with projen:

```
❯ pnpx projen test
<snip>
👾 test » eslint | eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern  src test build-tools projenrc .projenrc.ts
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.6.3

Please only submit bug reports when using the officially supported version.

=============
```

After some debugging I was able to track the refusal to upgrade eslint (using `pnpx projen upgrade`) to `pnpx projen` insisting on
```
"@typescript-eslint/eslint-plugin": "^7",
"@typescript-eslint/parser": "^7",
```

So I tried updating those using the `devDeps` property for my projen project, which did absolutely nothing.
However, adding the following at the bottom of the `.projenrc.ts` file did help:

```
project.addDevDeps(
  "eslint@^8",
  "@typescript-eslint/eslint-plugin@^8",
  "@typescript-eslint/parser@^8",
  "eslint-import-resolver-typescript",
  "eslint-plugin-import",
);
```

This is because I found out that using eslint in projen will add these hardcoded versions to the devDependencies:

https://github.com/projen/projen/blob/38be291c30f131eba6b58e2b79b9865eebe4ae2e/src/typescript/typescript.ts#L561-L569

https://github.com/projen/projen/blob/38be291c30f131eba6b58e2b79b9865eebe4ae2e/src/javascript/eslint.ts#L205-L211

I kept eslint at version 8, because there are some breaking changes when moving to 9.

nikovirtala:
Fixes #3907

This is a draft of what the `eslint` v9 update could look like — it doesn't seem to break things

---


nikovirtala:
Ok, if done that way... it doesn't work on Windows.

mrgrain:
> Ok, if done that way... it doesn't work on Windows.

Try setting the env on the task, not as part of the exec command.

nikovirtala:
> Try setting the env on the task, not as part of the exec command.

The result should look something like this?

```js
    "eslint": {
      "name": "eslint",
      "description": "Runs eslint against the codebase",
      "env": {
        "ESLINT_USE_FLAT_CONFIG": "false"
      },
      "steps": [
        {
          "exec": "eslint --ext .ts,.tsx --fix --no-error-on-unmatched-pattern $@ src test build-tools projenrc .projenrc.ts",
          "receiveArgs": true
        }
      ]
    },
```

nikovirtala:
> Try setting the env on the task, not as part of the exec command.

That's now done — can you trigger the build?

mrgrain (approved):

PR #3948, opened 2024-11-14
feat: allow control over next version during bump

rix0rrr:
This adds 2 changes to the bump process:
    
- `bumpPackage`: was configurable on `Version`, but not on `Release`. Expose that property.
- `nextVersionCommand`: introduces a shell command that can be used to determine the next version during a bump.

Refactoring:

- Environment variables which control the behavior of the `bump-version` task were set both in `version.ts` and `release.ts`. Centralize all their names in `version.ts`, and make a function `releaseBranchParameters()` that can be called from `release.ts` to achieve the same effect.

---


test/release/release.test.ts
```diff
@@ -71,6 +71,42 @@ describe("Single Project", () => {
...
+      task: project.buildTask,
+      versionFile: "version.json",
+      branch: "10.x",
+      artifactsDirectory: "dist",
+      bumpPackage: "MY-BUMP",
+    });
+
+    // THEN
+    const outdir = synthSnapshot(project);
+    expect(outdir).toMatchSnapshot();
```
mrgrain, line 89: This test doesn't need to match a full project snapshot (I'm aware the existing tests do this. It's equally wrong).

src/release/bump-version.ts
```diff
@@ -186,6 +207,42 @@ export async function bump(cwd: string, options: BumpOptions) {
...
+  // Determine what version to release as
+  let releaseAs: string | undefined;
+  if (minMajorVersion) {
+    const [majorVersion] = latestVersion.split(".");
+    const majorVersionNumber = parseInt(majorVersion, 10);
+    if (majorVersionNumber < minMajorVersion) {
+      releaseAs = `${minMajorVersion}.0.0`;
+    }
+  }
+  if (!releaseAs && options.nextVersionCommand) {
```
mrgrain, line 219: I'm missing this interaction being explained anywhere. The way it is, the `minMajorVersion` setting seems to be taking precedence over `nextVersionCommand`. 

I don't think that's right. If a user defines their own command, shouldn't they take care of handling that? Hadnling `minMajorVersion` feels like what is part of projen's default "nextVersionCommand".

At a minimum this interaction needs to be called out.

mrgrain (changes requested):

src/release/bump-version.ts
mrgrain: My gut feeling is it should be mutex. But I can be convinced otherwise. Making it mutex doesn't feel like changing existing behavior, does it?

In practice it seems you cannot use these two features together. It's either:
(a) When minMajorVersion kicks in, nextVersionCommand is ignored. Or
(b) When nextVersionCommand is defined, minMajorVersion doesn't have an effect when it would have.

The situation is only relevant when the existing minMajorVersion situation kicks in. 

Given that nextVersionCommand can otherwise completely ignore any minMajorVersion or majorVersion instructions, it feels to be that always given nextVersionCommand precedence is less surprising. Especially when clearly called out in the docs.

src/release/bump-version.ts
```diff
@@ -102,6 +102,27 @@ export interface BumpOptions {
...
+   * - `$VERSION`: the current version.
+   *
+   * The command should print one of the following to `stdout`:
+   *
+   * - Nothing: the next version number will be determined based on commit history.
+   * - `x.y.z`: the next version number will be `x.y.z`.
+   * - `major|minor|patch`: the next version number will be the current version number
+   *   with the indicated component bumped.
+   *
+   * @default - The next version will be determined based on the commit history.
```
mrgrain, line 123:
```suggestion
   * @default - The next version will be determined based on the commit history and project settings.
```

src/release/bump-version.ts
```diff
@@ -102,6 +102,27 @@ export interface BumpOptions {
    * @default "commit-and-tag-version@12"
    */
   readonly bumpPackage?: string;
+
+  /**
+   * A shell command to control the next version to release.
+   *
+   * If present, this shell command will be run before the bump is executed, and
+   * it determines what version to release. It will be executed in the following
+   * environment:
```
mrgrain, line 111:
```suggestion
   * If present, this shell command will be used to determine  what version to release.
   * Only the result of this command will be used. Other project settings will have no effect.
   *
   * The command It will be executed in the following
   * environment:
```

src/release/bump-version.ts
rix0rrr: Made it mutex.

test/version.test.ts
```diff
@@ -119,6 +119,56 @@ describe("bump task", () => {
...
+          { message: "chore(release): v0.1.0", tag: "v0.1.0" },
+          { message: "new change" },
+        ],
+      });
+
+      expect(result.version).toEqual("1.0.0");
+    });
+  });
+
+  test("throws an error if the bump command output is not valid", async () => {
```
mrgrain, line 148: This test isn't doing what it claims to do.

src/version.ts
```diff
@@ -176,6 +203,91 @@ export class Version extends Component {
     this.project.addPackageIgnore(`/${changelogFile}`);
     this.project.addPackageIgnore(`/${bumpFile}`);
   }
+
+  /**
+   * Return the environment variables to set modify the bump command for release branches.
```
mrgrain, line 208:
```suggestion
   * Return the environment variables to modify the bump command for release branches.
```

src/version.ts
```diff
@@ -176,6 +203,91 @@ export class Version extends Component {
...
+   *
+   * These options are used to modify the behavior of the version bumping script
+   * for additional branches, by setting environment variables.
+   *
+   * No settings are inherited from the base `Version` object (but any parameters that
+   * control versions do conflict with the use of a `nextVersionCommand`).
+   */
+  public releaseBranchParameters(
```
mrgrain, line 216: Not crazy about the name here.

Maybe this?

```suggestion
  public envForBranch(
```

mrgrain (changes requested):

test/version.test.ts
```diff
@@ -119,6 +119,58 @@ describe("bump task", () => {
...
+    withProjectDir((projectdir) => {
+      const project = new TestProject({
+        outdir: projectdir,
+      });
+      new Version(project, {
+        versionInputFile: "package.json",
+        artifactsDirectory: "dist",
+        nextVersionCommand: "echo banana",
```
mrgrain, line 156: 🍌

src/version.ts
rix0rrr: Done

mrgrain (approved):

PR #3978, opened 2024-12-01
feat(node)!: default pnpm to v9

AminFazlMondo:
This is to upgrade `pnpm` as one of the supported package manager to `v9` as it is the latest released version

BREAKING CHANGE: pnpm now defaults to `v9`. To use a different pnpm version, please set the `pnpmVersion` option on your project.

---


mrgrain (changes requested):
Looks good. Blocking this to group with some other upcoming breaking changes.

src/javascript/node-package.ts
```diff
@@ -485,7 +485,7 @@ export class NodePackage extends Component {
   /**
    * The version of PNPM to use if using PNPM as a package manager.
    *
```
mrgrain, line 487:
```suggestion
```

mrgrain (approved):

PR #4029, opened 2025-01-07
feat: add pr to merge queue on `synchronize`

iliapolo:
If a PR fails status checks *after* is has been added to the merge queue, github will remove it:

image

It will not re-add it when you push new commits though, this doesn't seem desirable since it effectively disables the auto merge.

> Example: https://github.com/cdklabs/jsii-docgen/pull/1687

---

### DIFF
--- a/src/github/auto-queue.ts
+++ b/src/github/auto-queue.ts
@@ -139,7 +139,7 @@ export class AutoQueue extends Component {
       // That way a user can always disable auto-queue if they want to and the workflow will
       // not automatically re-enable it, unless one of the events occurs.
       pullRequestTarget: {
-        types: ["opened", "reopened", "ready_for_review"],
+        types: ["opened", "reopened", "ready_for_review", "synchronize"],
       },
     });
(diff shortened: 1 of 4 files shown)


src/github/auto-queue.ts
```diff
@@ -139,7 +139,7 @@ export class AutoQueue extends Component {
       // That way a user can always disable auto-queue if they want to and the workflow will
       // not automatically re-enable it, unless one of the events occurs.
       pullRequestTarget: {
-        types: ["opened", "reopened", "ready_for_review"],
+        types: ["opened", "reopened", "ready_for_review", "synchronize"],
```
rix0rrr, line 142: This was on purpose, see the docstring above. 

If automerge is disabled by a human, we also want a human to re-enable it.

rix0rrr (changes requested):

mrgrain:
@iliapolo I would support a feature that allows automatic re-enabling of auto-merge based on a label.

PR #4030, opened 2025-01-07
feat(awscdk): default the AWS CDK CLI version to "latest 2.x"

### LINKED ISSUE #4031: Introduce a new `cdkCliVersion` for CDK projects
[CDK CLI versions and CDK library versions will soon diverge](https://github.com/aws/aws-cdk/issues/32775).

At present, specifying the `cdkVersion` in an `AwsCdkTypeScriptApp` project causes Projen to generate a `package.json` file where both the `aws-cdk` and `aws-cdk-lib` dependencies match the defined `cdkVersion`. However, this approach could lead to unintended behaviours once these versions are no longer aligned.

rix0rrr:
projen currently installs the same CDK CLI version as the requested CDK framework version. The CDK requirement used to be that `CLI Version >= Lib Version`, and setting them to the same value was a simple way of satisfying that requirement.

In the near future, the version of the CDK CLI and the CDK Library will not be the same anymore (see
https://github.com/aws/aws-cdk/issues/32775). At that point, this logic will start to fail.

Instead, change the default to `aws-cdk@^2`, i.e. the latest 2.x version that's available. This still satisfies the requirement that `CLI Version >= Lib Version` (for old CDK releases), and more importantly also satisfies the requirement that `CDK Release Date >= Lib Release Date` (for new CDK releases).

A new optional parameter is added to specifically configure the CLI version if desired.

Closes #4031.

---


mrgrain (approved):

rix0rrr:
Failing with this?

```
error Received malformed response from registry for "@typescript-eslint%2ftypescript-estree". The registry may be down.
```

rix0rrr:
Ah no it's this:

```
error lru-cache@11.0.2: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.18.0"
```

rix0rrr:
Depends on https://github.com/projen/projen/pull/4034

PR #4051, opened 2025-01-28
chore(typescript): align default ES version to `cdk init`

garysassano:
Current `tsconfig.json` file generated by `cdk init`:

```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "commonjs",
    "lib": [
      "es2020",
      "dom"
    ],
    "declaration": true,
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "noUnusedLocals": false,
    "noUnusedParameters": false,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": false,
    "inlineSourceMap": true,
    "inlineSources": true,
    "experimentalDecorators": true,
    "strictPropertyInitialization": false,
    "typeRoots": [
      "./node_modules/@types"
    ]
  },
  "exclude": [
    "node_modules",
    "cdk.out"
  ]
}

```

### DIFF
--- a/src/typescript/typescript.ts
+++ b/src/typescript/typescript.ts
@@ -461,7 +461,7 @@ export class TypeScriptProject extends NodeProject {
       experimentalDecorators: true,
       inlineSourceMap: true,
       inlineSources: true,
-      lib: ["es2019"],
+      lib: ["es2020"],
       module: "CommonJS",
       noEmitOnError: false,
       noFallthroughCasesInSwitch: true,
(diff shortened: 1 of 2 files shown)


mrgrain (changes requested):
It sounds like we should have a separate default for cdk apps then, and not changing the default for everyone. 

Also open to changing the default for everyone, but that should probably be closer to something like https://www.npmjs.com/package/@tsconfig/node18

garysassano:
I'd be fine bumping both `lib` and `target` to `es2024` (there are basically no changes compared to `es2023`). I'd also set `module` to `nodenext`.

mrgrain:
> I'd be fine bumping both `lib` and `target` to `es2024` (there are basically no changes compared to `es2023`). I'd also set `module` to `nodenext`.

That's not what the linked config is suggesting. We'd still want separate, different defaults for CDK though. Which makes sense to align with `cdk init`.

garysassano:
@mrgrain Let me know what you think now, or if you had something else in mind.

PR #4071, opened 2025-02-07
feat(task-runtime): support bash-like quoted argument expansion in task arguments 

jrouly:
Custom task definitions do not allow for bash-style variable expansion.

For example, we have a task definition that looks like this:

```json5
{
  exec: 'prettier --write --ignore-unknown --loglevel=warn "$@"',
  receiveArgs: true,
}
```

It expects to invoke `prettier` on a series of filenames to be expanded from `"$@"`. Normally in bash, this quoted variable expansion (`"$@")` would automatically wrap whitespace containing elements in single quotes to preserve whitespace. Without this behavior, filenames including whitespace would be split and interpreted as separate filename tokens.

Currently, if we invoke `npx projen <prettier task> src/foo\ bar.txt` prettier is incorrectly invoked on `src/foo` and `bar.txt` as separate tokens because projen does not preserve whitespace in arguments.

This PR attempts to preserve whitespace in a similar fashion to bash by specifically noting the presence of `"$@"` (as opposed to `$@`) and performing a similar whitespace preserving single quotation wrap of each argument provided to the task.

I've provided an example of the behavior in a unit test. Before the change in this PR, the unit test failed as the whitespace was not being correctly preserved. It passes after the change.

For a demonstration of the behavior I am describing in bash, see the following.

```bash
$ cat script.sh
set -x
echo quoted: "$@" > /dev/null
echo unquoted: $@ > /dev/null
$ ./script.sh "a b" c\ d e f
++ echo quoted: 'a b' 'c d' e f
++ echo unquoted: a b c d e f
```

BREAKING CHANGE: Task definitions that included `"$@"` will now observe individual arguments being wrapped in single quotes, rather than the entire argstring wrapped in double quotes. This may be breaking or incompatible behavior in certain circumstances.

From my perspective, I'm not clear what use case `"<entire arg string>"` legitimately satisfies (compared to `'arg' 'arg' 'arg'`). But I'm open to hearing why this breaking change may not be desired.

---

### DIFF
--- a/docs/api/projen.md
+++ b/docs/api/projen.md
@@ -13064,6 +13064,9 @@ Should this step receive args passed to the task.
 If `true`, args are passed through at the end of the `exec` shell command.\
 The position of the args can be changed by including the marker `$@` inside the command string.
 
+If the marker is explicitly double-quoted ("$@") arguments will be wrapped in single quotes, approximating
+the whitespace preserving behavior of bash variable expansion.
+
 If the step spawns a subtask, args are passed to the subtask.
 The subtask must define steps receiving args for this to have any effect.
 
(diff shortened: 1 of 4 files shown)


mrgrain (approved):
Looks good. I'm not sure if the test case will work on Windows, but let's see.

mrgrain (approved):

PR #4084, opened 2025-02-14
feat(bump-task): allow forcing a version suffix

rix0rrr:
The `yarn bump` task allows forcing a suffix by setting an environment variable. This will be used to force a unique local version for testing by appending `test.0` to all versions before building.

It makes sure that candidate packages have a realistic but definitely unique version during integration tests.

---

### DIFF
--- a/src/release/bump-version.ts
+++ b/src/release/bump-version.ts
@@ -127,6 +127,20 @@ export interface BumpOptions {
    * @default - The next version will be determined based on the commit history and project settings.
    */
   readonly nextVersionCommand?: string;
+
+  /**
+   * Append a version suffix at the end of the version string
+   *
+   * This can be used to generate fake version numbers for pre-release
+   * integration testing.
+   *
+   * Example: if the current version is `1.2.3`, the generated version can be
+   * something like `1.2.4-test.0`, which is guaranteed to never conflict with
+   * any other (real) version, even if `1.2.4` is released in the mean time.
+   *
+   * @default - No suffix
+   */
+  readonly forceSuffix?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/release/bump-version.ts
```diff
@@ -304,6 +319,18 @@ export async function bump(cwd: string, options: BumpOptions) {
     }
   }
 
+  // Apply a forced suffix if given
+  if (options.forceSuffix) {
+    let suffix = options.forceSuffix;
+    if (!suffix.match(/^-/)) suffix = `-${suffix}`;
+    if (!suffix.match(/\d$/)) suffix += '.0';
```
mrgrain: This looks like it also is opinionated?

src/release/bump-version.ts
```diff
@@ -304,6 +319,20 @@ export async function bump(cwd: string, options: BumpOptions) {
...
+    let suffix = options.forceSuffix;
+    if (!suffix.match(/^-/)) suffix = `-${suffix}`;
+    if (!suffix.match(/\d$/)) suffix += ".0";
+    newVersion = `${newVersion}${suffix}`;
+    updatedVersionFile.contents.version = newVersion;
+
+    const updatedContents =
+      JSON.stringify(updatedVersionFile.contents, undefined, 2) +
+      (updatedVersionFile.newline ? "\n" : "");
+    await fs.writeFile(versionFile, updatedContents, "utf-8");
```
mrgrain, line 333: I'm not into manually updating these files. There might be more files that need updating that are passed into the tool via options. If anything you'll need to set the correct versionrc options.

mrgrain (changes requested):

PR #4087, opened 2025-02-15
feat(eslint):  support ESLint FlatConfig [WIP]

ren-yamanashi:
Fixes #

---

### DIFF
--- a/src/javascript/eslint.ts
+++ b/src/javascript/eslint.ts
@@ -165,7 +165,7 @@ export class Eslint extends Component {
   /**
    * eslint rules.
    */
-  public readonly rules: { [rule: string]: any[] };
+  public readonly rules: { [rule: string]: any };
 
   /**
    * eslint overrides.
@@ -180,7 +180,7 @@ export class Eslint extends Component {
   /**
    * Direct access to the eslint configuration (escape hatch)
    */
-  public readonly config: any;
+  public readonly config: string;
 
   /**
    * File patterns that should not be linted
@@ -189,7 +189,7 @@ export class Eslint extends Component {
 
   private _formattingRules: Record<string, any>;
   private readonly _allowDevDeps: Set<string>;
-  private readonly _plugins = new Set<string>();
+  private readonly _plugins = new Map<string, string>();
   private readonly _extends = new Set<string>();
   private readonly _fileExtensions: Set<string>;
   private readonly _flagArgs: Set<string>;
@@ -254,34 +254,33 @@ export class Eslint extends Component {
 
     this._formattingRules = {
       // @see https://github.com/typescript-eslint/typescript-eslint/issues/8072
-      indent: ["off"],
+      indent: "off",
       "@stylistic/indent": ["error", 2],
 
       // Style
-      quotes: ["error", "single", { avoidEscape: true }],
-      "comma-dangle": ["error", "always-multiline"], // ensures clean diffs, see https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8
-      "comma-spacing": ["error", { before: false, after: true }], // space after, no space before
-      "no-multi-spaces": ["error", { ignoreEOLComments: false }], // no multi spaces
-      "array-bracket-spacing": ["error", "never"], // [1, 2, 3]
-      "array-bracket-newline": ["error", "consistent"], // enforce consistent line breaks between brackets
-      "object-curly-spacing": ["error", "always"], // { key: 'value' }
-      "object-curly-newline": ["error", { multiline: true, consistent: true }], // enforce consistent line breaks between braces
-      "object-property-newline": [
+      "@stylistic/quotes": ["error", "single", { avoidEscape: true }],
+      "@stylistic/comma-dangle": ["error", "always-multiline"], // ensures clean diffs, see https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8
+      "@stylistic/comma-spacing": ["error", { before: false, after: true }], // space after, no space before
+      "@stylistic/no-multi-spaces": ["error", { ignoreEOLComments: false }], // no multi spaces
+      "@stylistic/array-bracket-spacing": ["error", "never"], // [1, 2, 3]
+      "@stylistic/array-bracket-newline": ["error", "consistent"], // enforce consistent line breaks between brackets
+      "@stylistic/object-curly-spacing": ["error", "always"], // { key: 'value' }
+      "@stylistic/object-curly-newline": ["error", { multiline: true, consistent: true }], // enforce consistent line breaks between braces
+      "@stylistic/object-property-newline": [
         "error",
         { allowAllPropertiesOnSameLine: true },
       ], // enforce "same line" or "multiple line" on object properties
-      "keyword-spacing": ["error"], // require a space before & after keywords
-      "brace-style": ["error", "1tbs", { allowSingleLine: true }], // enforce one true brace style
-      "space-before-blocks": ["error"], // require space before blocks
+      "@stylistic/keyword-spacing": "error", // require a space before & after keywords
+      "@stylistic/brace-style": ["error", "1tbs", { allowSingleLine: true }], // enforce one true brace style
+      "@stylistic/space-before-blocks": "error", // require space before blocks
       curly: ["error", "multi-line", "consistent"], // require curly braces for multiline control statements
-      // @see https://github.com/typescript-eslint/typescript-eslint/issues/8072
-      "@stylistic/member-delimiter-style": ["error"],
+      "@stylistic/member-delimiter-style": "error",
 
       // Require semicolons
-      semi: ["error", "always"],
+      "@stylistic/semi": ["error", "always"],
 
       // Max line lengths
-      "max-len": [
+      "@stylistic/max-len": [
         "error",
         {
           code: 150,
@@ -294,12 +293,12 @@ export class Eslint extends Component {
       ],
 
       // Don't unnecessarily quote properties
-      "quote-props": ["error", "consistent-as-needed"],
+      "@stylistic/quote-props": ["error", "consistent-as-needed"],
     };
 
     this.rules = {
       // Require use of the `import { foo } from 'bar';` form instead of `import foo = require('bar');`
-      "@typescript-eslint/no-require-imports": ["error"],
+      "@typescript-eslint/no-require-imports": "error",
 
       // Require all imported dependencies are actually declared in package.json
       "import/no-extraneous-dependencies": [
@@ -313,7 +312,7 @@ export class Eslint extends Component {
       ],
 
       // Require all imported libraries actually resolve (!!required for import/no-extraneous-dependencies to work!!)
-      "import/no-unresolved": ["error"],
+      "import/no-unresolved": "error",
 
       // Require an ordering on all imports
       "import/order": [
@@ -325,34 +324,34 @@ export class Eslint extends Component {
       ],
 
       // Cannot import from the same module twice
-      "import/no-duplicates": ["error"],
+      "import/no-duplicates": "error",
 
       // Cannot shadow names
-      "no-shadow": ["off"],
-      "@typescript-eslint/no-shadow": ["error"],
+      "no-shadow": "off",
+      "@typescript-eslint/no-shadow": "error",
 
       // Required spacing in property declarations (copied from TSLint, defaults are good)
-      "key-spacing": ["error"],
+      "@stylistic/key-spacing": "error",
 
       // No multiple empty lines
-      "no-multiple-empty-lines": ["error"],
+      "@stylistic/no-multiple-empty-lines": "error",
 
       // One of the easiest mistakes to make
-      "@typescript-eslint/no-floating-promises": ["error"],
+      "@typescript-eslint/no-floating-promises": "error",
 
       // Make sure that inside try/catch blocks, promises are 'return await'ed
       // (must disable the base rule as it can report incorrect errors)
-      "no-return-await": ["off"],
-      "@typescript-eslint/return-await": ["error"],
+      "no-return-await": "off",
+      "@typescript-eslint/return-await": "error",
 
       // Useless diff results
-      "no-trailing-spaces": ["error"],
+      "@stylistic/no-trailing-spaces": "error",
 
       // Must use foo.bar instead of foo['bar'] if possible
-      "dot-notation": ["error"],
+      "dot-notation": "error",
 
       // Are you sure | is not a typo for || ?
-      "no-bitwise": ["error"],
+      "no-bitwise": "error",
 
       // Member ordering
       "@typescript-eslint/member-ordering": [
@@ -410,46 +409,77 @@ export class Eslint extends Component {
     this.addPlugins("import");
     this.addExtends("plugin:import/typescript");
 
-    this.config = {
-      env: {
-        jest: true,
-        node: true,
+    this.config = `
+import globals from "globals";
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint"
+import importPlugin from 'eslint-plugin-import';
+import stylistic from '@stylistic/eslint-plugin'
+
+/** @type {import('eslint').Linter.Config[]} */
+export default [
+  eslint.configs.recommended,
+  importPlugin.flatConfigs.typescript,
+  {
+    ...${[...this._extends].join(',\n  ...')},
+    files: ['**/*.ts', '**/*.tsx'],
+    languageOptions: { 
+      globals: {
+        ...globals.node,
+        ...globals.jest,
       },
-      root: true,
-      plugins: this._plugins,
-      parser: "@typescript-eslint/parser",
+      parser: tseslint.parser,
       parserOptions: {
         ecmaVersion: 2018,
         sourceType: "module",
-        project: tsconfig,
+        project: ${tsconfig},
       },
-      extends: () =>
-        Array.from(this._extends).sort((a, b) =>
-          this.sortExtends.compare(a, b)
-        ),
-      settings: {
-        "import/parsers": {
-          "@typescript-eslint/parser": [".ts", ".tsx"],
-        },
-        "import/resolver": {
+    },
+    settings: {
+      "import/parsers": {
+        "@typescript-eslint/parser": [".ts", ".tsx"],
+      },
+      "import/resolver": {
+        ${JSON.stringify({
           ...(options.aliasMap && {
             alias: {
               map: Object.entries(options.aliasMap).map(([k, v]) => [k, v]),
               extensions: options.aliasExtensions,
             },
           }),
-          node: {},
-          typescript: {
-            project: tsconfig,
+        }).slice(1, -1)},
+        node: {},
+        typescript: {
+          project: ${tsconfig},
+          ${JSON.stringify({
             ...(options.tsAlwaysTryTypes !== false && { alwaysTryTypes: true }),
-          },
-        },
+          }).slice(1, -1)},
+        }
       },
-      ignorePatterns: this.ignorePatterns,
-      rules: () => ({ ...this._formattingRules, ...this.rules }),
-      overrides: this.overrides,
-    };
-
+    },
+    plugins: ${[...this._plugins].map(([key, value]) => `"${key}": ${value}`).join(',\n      ')},
+    rules: {
+      ${JSON.stringify({...this.rules, ...this._formattingRules}).slice(1, -1)}
+    }
+  },
+  ${this.overrides.map((override) => `
+    {
+      ...${override.extends?.map((extend) => `"${extend}"`).join(',\n  ...')},
+      files: ${override.files},
+      plugins: ${override.plugins},
+      ${override.parser ? `
+      languageOptions: {
+        parser: ${override.parser},
+      }` : ''},
+      ${override.rules ? `rules: ${override.rules}` : ''},
+      ${override.excludedFiles ? `ignores: ${override.excludedFiles}` : ''},
+    }
+    `).join(',\n  ')},
+  {
+    ignores: ${JSON.stringify(this.ignorePatterns).slice(1, -1)}
+  }
+];
+`
     if (options.yaml) {
       new YamlFile(project, ".eslintrc.yml", {
         obj: this.config,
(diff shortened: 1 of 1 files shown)


mrgrain (changes requested):
I like the direction of your PR, but mostly I'd like to see this as a new & separate component first.

Also see some of my research from here:

https://github.com/projen/projen/pull/3711#issuecomment-2488440745

ren-yamanashi:
TODO

- [ ] support cli options
- [ ] add some tests
- [ ]

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1016 @@
...
+  }
+
+  /**
+   * Sync the config file with the current state of the class properties.
+   */
+  public synthesize() {
+    this._config = this.generateConfig();
+    const projectDir = this.project.outdir;
+    const configFile = path.join(projectDir, "eslint.config.mjs");
+    fs.writeFileSync(configFile, this.config);
```
mrgrain, line 395: This needs to use `TextFile` instead.

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1016 @@
...
+    this.initializeIgnorePatterns(options);
+  }
+
+  /**
+   * Sync the config file with the current state of the class properties.
+   */
+  public synthesize() {
+    this._config = this.generateConfig();
+    const projectDir = this.project.outdir;
+    const configFile = path.join(projectDir, "eslint.config.mjs");
```
mrgrain, line 394: Does not need to be in this PR, but we should support both `mjs` and `cjs`. Let's at least add the option so we can implement CJS later.

src/javascript/eslint-flat-config.ts
ren-yamanashi: @mrgrain 

Thanks for comment!

As a prerequisite, this `ESLintFlatConfig` class would need to implement a use case for rewriting `eslint.config.mjs` after the class has been instantiated. (e.g. overriding rules).

Therefore, the current implementation provides methods to update the contents of the configuration file (e.g. `addOverrides`) and methods to overwrite the file contents (`synthesize`), and the concrete flow in the file rewriting is as follows.

```ts
// 1.  `ESLintFlatConfig` class is instanced. (Here `eslint.config.mjs` is generated.)
const eslint = new EslintFlatConfig(project, {
  enablePatterns: ["src"],
});

// 2. Public method is executed (like `addOverrides` method)
eslint.addOverrides({
  enablePatterns: ["projenrc.ts"],
  rules: {
    "@typescript-eslint/no-require-imports": "off",
    "import/no-extraneous-dependencies": "off",
  },
});

// 3. `synthesize` method is executed and regenerated `eslint.config.mjs`(The `synthesize` method contains logic to regenerate `eslint.config.mjs`)
eslint.synthesize();
```

I thought that the `TextFile` implementation would need to be modified to achieve this kind of flow on top of using `TextFile`. What would you think?  
If there is a better way, please like to know.

ren-yamanashi:
TODO 

- [ ] moduleType test
- [ ] rule test (e.g. when enable prettier)

src/javascript/eslint-flat-config.ts
mrgrain: You are right, `TextFile` is unnecessary but `FileBase` an be used. I have pushed two commits to illustrate the usage. The first is just simple, the second also uses the provided `resolver` to add support for lazy functions.

test/javascript/eslint-flat-config.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      enablePatterns: ["**/*.ts", "**/*.tsx"],
+      tsAlwaysTryTypes: true,
+    });
+    eslint.synthesize();
+
+    // THEN
+    const pattern =
+      /settings:\s*{[\s\S]*?"import\/resolver":\s*{(\s*typescript:\s*{[\s\S]*?})\s*}[\s\S]*?}/;
+    const settingsMatch = eslint.config.match(pattern);
+    expect(settingsMatch).not.toBeNull();
```
mrgrain, line 108:
```suggestion
    expect(settingsMatch).toMatch(pattern);
```

test/javascript/eslint-flat-config.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      importPath: "eslint-plugin-foo",
+      moduleName: "fooPlugin",
+      pluginAlias: "foo",
+    });
+    eslint.synthesize();
+
+    // THEN
+    const pattern = /plugins:\s*{[\s\S]*?}/;
+    const pluginMatch = eslint.config.match(pattern);
+    expect(pluginMatch).not.toBeNull();
```
mrgrain, line 219:
```suggestion
    expect(pluginMatch).toMatch(pattern);
```

mrgrain:
I've pushed to commits to illustrate the usage of `FileBase` and `IResolver`. The first commit also includes some fixes to make the build pass.

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+type Rules = { [rule: string]: any };
+
+/**
+ * ESLint plugin configuration information.
+ *
+ * @example
+ * // Configuration
+ * {
+ *   importPath: "typescript-eslint",
+ *   moduleName: "tseslint",
```
mrgrain, line 24: I would say that in 

```ts
import tseslint from "typescript-eslint"
```

- `typescript-eslint` is the moduleName
- `tseslint` is the importName of the module's defaultExport

(the [spec](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-imports) calls them ModuleSpecifier and ImportedBinding)

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+ * // Results in:
+ * import tseslint from "typescript-eslint"
+ *
+ * export default [{
+ *   plugins: {
+ *     "@typescript-eslint": tseslint
+ *   }
+ * }]
+ */
+export interface EslintPlugin {
```
mrgrain, line 37: From what I can tell we only need to make `importPath` required and could automatically generate the other two values, since they don't seem to matter. In fact do we even need `pluginAlias` or could this just always be the same as the importName?

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+   * @example "eslint-plugin-prettier"
+   */
+  readonly importPath: string;
+
+  /**
+   * The local module name to use in the import statement.
+   *
+   * @example "prettierPlugin"
+   */
+  readonly moduleName: string;
```
mrgrain, line 106: Similar to plugin

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+   * When true, the configuration will be included as `...configReference`,
+   * when false, it will be included as `configReference`.
+   *
+   * @example
+   * true  -> ...eslint.configs.recommended
+   * false -> eslint.configs.recommended
+   *
+   * @default false
+   */
+  readonly shouldSpreadConfig?: boolean;
```
mrgrain, line 129:
```suggestion
  readonly spreadConfig?: boolean;
```

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+
+  /**
+   * The configuration reference to be used in the extends section.
+   * This can be a plugin name or a specific configuration path from the plugin.
+   *
+   * @example
+   * - "prettierPlugin" // Direct plugin reference
+   * - "eslint.configs.recommended" // Specific configuration path
+   */
+  readonly configReference: string;
```
mrgrain, line 116: This could default to the direct plugin reference and then be optional, right?

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+   * @example "typescript-eslint"
+   */
+  readonly importPath: string;
+
+  /**
+   * The local module name to use in the import statement.
+   *
+   * @example "tseslint"
+   */
+  readonly moduleName: string;
```
mrgrain, line 165: Similar to plugins

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+   *
+   * @example
+   * // When importing from typescript-eslint as "tseslint"
+   * parserReference: "tseslint.parser"
+   * // Results in:
+   * languageOptions: {
+   *   parser: tseslint.parser
+   * }
+   */
+  readonly parserReference: string;
```
mrgrain, line 179: I wonder if this should only be the part after `tseslint`:

```ts
// When importing from typescript-eslint as "tseslint"
parserReference: "parser"
// Results in:
languageOptions: {
  parser: tseslint.parser
}
```

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+import { TaskStepOptions } from "../task-model";
+import { Prettier } from "./prettier";
+
+const MODULE_TYPE = {
+  COMMONJS: "commonjs",
+  MODULE: "module",
+} as const;
+
+type ModuleType = (typeof MODULE_TYPE)[keyof typeof MODULE_TYPE];
+type Rules = { [rule: string]: any };
```
mrgrain, line 15: Should be a public interface

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+      ${this.convertRulesToString(override.rules ?? {}, "      ")}
+    }
+    ${
+      overrideIgnorePatterns.length
+        ? `ignores: ${this.convertArrayToString(overrideIgnorePatterns)}`
+        : ""
+    }
+  },`;
+      })
+      .join("\n");
```
mrgrain, line 996: Return a list of overrides object-strings and join them together in the main file
```suggestion
```

src/javascript/eslint-flat-config.ts
```diff
@@ -0,0 +1,1091 @@
...
+  }
+
+  /**
+   * Generate overrides configuration for ESLint.
+   * Creates configuration objects for each override, including their specific
+   * extends, parser options, files patterns, plugins, and rules.
+   *
+   * @returns A string containing all override configurations, separated by commas
+   */
+  private generateOverridesConfig(resolver: IResolver): string {
```
mrgrain, line 952: 💅🏻  This might be better as
```suggestion
  private generateOverrideConfig(override: EslintFlatConfigOverride): string {
```
and do the mapping and joining in the main config generator.

mrgrain (changes requested):
This looks mostly great! The main thing I'd like to see is splitting out the functionality into separate "components":

- A component only for the configuration file without any default (I would call this `EslintFlatConfig`)
- A new `Eslint` component that does the actual setup including the task. It would take a configuration 

We can out both into a new `eslint` subdirectory and then they are available for use without us having to change the default just yet.

mrgrain:
PS: I'd also love if we could split the configuration for the two modes (stylistic and prettier) out somehow. That'd probably need something like a `EslintConfiguration` interface that the `Eslint` component would also take. And the configs would implement it but would also have to accept one themselves to allow for overrides?!? All very complicated abd probably out of scope for this PR. But you might have an idea.

src/javascript/eslint-flat-config.ts
ren-yamanashi: Thanks! resolve this thread.

test/javascript/eslint-flat-config.test.ts
ren-yamanashi: Thanks!  
The test will probably fail because `settingsMatch` is already an array of strings taken from a regular expression...  
So, based on your suggestion, I have modified it as follows, what do you think?

(Using the `toMatch` method to verify that the `typescript` property of `import/resolver` is populated, and using the `toContain` method to confirm that `alwaysTryTypes: true` is written in the configuration).

```ts
    // THEN
    const pattern =
      /settings:\s*{[\s\S]*?"import\/resolver":\s*{(\s*typescript:\s*{[\s\S]*?})\s*}[\s\S]*?}/;
    expect(eslint.config).toMatch(pattern);
    expect(eslint.config.match(pattern)![0]).toContain("alwaysTryTypes: true");
```

https://github.com/projen/projen/pull/4087/commits/71fb757d33abd133e7339d1b980fcada16524181

test/javascript/eslint-flat-config.test.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/71fb757d33abd133e7339d1b980fcada16524181

src/javascript/eslint-flat-config.ts
ren-yamanashi: Renamed as follows

- importPath -> moduleSpecifier
- moduleName -> importedBinding

https://github.com/projen/projen/pull/4087/commits/dc2a0644040ffbb5853ef7039e6b4e9dfefdaeaa

src/javascript/eslint-flat-config.ts
ren-yamanashi: #### Assumptions

Assumptions, from a user perspective, when considering use cases where users add rules, I believe it's better for the `pluginAlias` to follow the official naming convention specified by the plugin (as documented in their official documentation).

For example, if we set the pluginAlias for the `typescript-eslint` plugin as `typescript-eslint`, users would need to write rules like `"typescript-eslint/no-require-imports": "error"`.

However, looking at the official documentation, it instructs users to write `"@typescript-eslint/no-require-imports": "error"`. Most users will follow this convention.
In this case, there would be a mismatch where the plugin is specified as `typescript-eslint` but the rules are using `@typescript-eslint`, which would result in an error as the `@typescript-eslint` rules cannot be loaded.

---

#### Conclusion

Based on this reasoning, I don't think automatically generating these values is a good approach.
This is because it would require the system (within the `ESLintFlatConfig` class) to automatically generate the pluginAlias according to the plugin's official naming convention based on the importPath, which isn't practical.

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/dc2a0644040ffbb5853ef7039e6b4e9dfefdaeaa

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/72a5661e19b32059598b2d490e2517c50a31af58

src/javascript/eslint-flat-config.ts
ren-yamanashi: Yes, I have fixed it!

https://github.com/projen/projen/pull/4087/commits/198d7a6adc95bb8886da6eac723bc83861962709

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/dc2a0644040ffbb5853ef7039e6b4e9dfefdaeaa

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/1ca3ddfe3b8530e0c6624575d86e6f820e43a98e

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/254a411956ce646461ca76bac2aad20f0d567773

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/3c40bc2698150eae06a0b9d5ee55f7eb7f05adde

src/javascript/eslint-flat-config.ts
ren-yamanashi: https://github.com/projen/projen/pull/4087/commits/fe860babc6c5d85700ef4cdf5a624d370627ff35

ren-yamanashi:
> This looks mostly great! The main thing I'd like to see is splitting out the functionality into separate "components":
> 
> * A component only for the configuration file without any default (I would call this `EslintFlatConfig`)
> * A new `Eslint` component that does the actual setup including the task. It would take a configuration
> 
> We can out both into a new `eslint` subdirectory and then they are available for use without us having to change the default just yet.

Thanks for review!

I see... I'd like to confirm the approach, is the following idea correct?


### Class Diagram

- Classes extending from `BaseEslintConfig` (e.g. EslintFlatConfig class): purely dedicated to generating configuration file
- `Eslint` class : responsible for task configuration and project initialisation.



```mermaid
classDiagram

    class FileBase {
        <<abstract>>
        +synthesize()
    }

    class Eslint {
        -config: IEslintConfiguration
        -task: Task
        +synthesize()
        +setupTask()
    }

    class IEslintConfiguration {
        <<interface>>
        +generateConfig()
        +getRules()
        +getPlugins()
        +getExtends()
    }

    class BaseEslintConfig {
        <<abstract>>
        #rules
        #plugins
        #extends
        +generateConfig()
        +getRules()
        +getPlugins()
        +getExtends()
    }

    class EslintFlatConfig {
        -moduleType
        -enablePatterns
        -ignorePatterns
        +generateConfig()
    }


    class StylisticConfig {
        +getStylisticRules()
    }

    class PrettierConfig {
        +getPrettierConfig()
    }

    IEslintConfiguration <|.. BaseEslintConfig
    BaseEslintConfig <|-- EslintFlatConfig
    BaseEslintConfig <|-- StylisticConfig
    BaseEslintConfig <|-- PrettierConfig
    Eslint o-- IEslintConfiguration
    FileBase --|> Eslint

```


### TS Code

```ts
const config = new EslintFlatConfig({
  styling: new StylisticConfig(), // or new PrettierConfig()
  // ... and some settings
});

const eslint = new Eslint(project, config);

eslint.synthesize()

eslint.setupTask({
  fix: true,
  // ... and some settings
})
```

ren-yamanashi:
> This looks mostly great! The main thing I'd like to see is splitting out the functionality into separate "components":
> 
> * A component only for the configuration file without any default (I would call this `EslintFlatConfig`)
> * A new `Eslint` component that does the actual setup including the task. It would take a configuration
> 
> We can out both into a new `eslint` subdirectory and then they are available for use without us having to change the default just yet.

@mrgrain 

Thanks for review!

I see... I'd like to confirm the approach, is the following idea correct?


### Class Diagram

- Classes extending from `BaseEslintConfig` (e.g. EslintFlatConfig class): purely dedicated to generating configuration file
- `Eslint` class : responsible for task configuration and project initialisation.



```mermaid
classDiagram

    class FileBase {
        <<abstract>>
        +synthesize()
    }

    class Eslint {
        -config: IEslintConfiguration
        -task: Task
        +synthesize()
        +setupTask()
    }

    class IEslintConfiguration {
        <<interface>>
        +generateConfig()
        +getRules()
        +getPlugins()
        +getExtends()
    }

    class BaseEslintConfig {
        <<abstract>>
        #rules
        #plugins
        #extends
        +generateConfig()
        +getRules()
        +getPlugins()
        +getExtends()
    }

    class EslintFlatConfig {
        -moduleType
        -enablePatterns
        -ignorePatterns
        +generateConfig()
    }


    class StylisticConfig {
        +getStylisticRules()
    }

    class PrettierConfig {
        +getPrettierConfig()
    }

    IEslintConfiguration <|.. BaseEslintConfig
    BaseEslintConfig <|-- EslintFlatConfig
    BaseEslintConfig <|-- StylisticConfig
    BaseEslintConfig <|-- PrettierConfig
    Eslint o-- IEslintConfiguration
    FileBase --|> Eslint

```


### TS Code

```ts
const config = new EslintFlatConfig({
  styling: new StylisticConfig(), // or new PrettierConfig()
  // ... and some settings
});

const eslint = new Eslint(project, config);

eslint.synthesize()

eslint.setupTask({
  fix: true,
  // ... and some settings
})
```

mrgrain:
I'm so sorry, this completely dropped off my radar :( 

> Class diagram

This seems to be proposing on how to integrate with the old code. Forget all about the old code and let's create entirely new classes and relationships. I think it's more like this:


```mermaid
classDiagram

    class FileBase {
        <<abstract>>
        +synthesize()
    }

    class Eslint {
    }

    class IEslintConfig {
    }


    class EslintFlatConfigFile {
        +synthesize()
    }


    class StylisticConfig {
    }

    class PrettierConfig {
    }

    FileBase --|> EslintFlatConfigFile

    IEslintConfig <|-- StylisticConfig
    IEslintConfig <|-- PrettierConfig
    Eslint o-- IEslintConfig
    Eslint o-- EslintFlatConfigFile
    
```

mrgrain:
> TS Code

This works:
```ts
const config = new EslintFlatConfig({
  styleConfig: new StylisticConfig(), // or new PrettierConfig()
  // ... and some settings
});
```

This doesn't, because we need to be able to eventually pass the settings in from a project:
```ts
const eslint = new Eslint(project, config);
eslint.synthesize()

eslint.setupTask({
  fix: true,
  // ... and some settings
})
```

Instead somthing like this
```ts
const eslint = new Eslint(project, {
  styleConfig: new StylisticConfig(), // or new PrettierConfig()
  task: {
    fix: true
  }
});
```

With the `Eslint` implementation using the inputs and passing them through:

```ts
constructor(scope, id, { styleConfig, task }) {
  super(scope, id);
  new EslintFlatConfig(this, 'Config', { styleConfig }); 
  this.setupTask(task);
}
```

Overrides at the config file level maybe?
```ts
eslint.config.addOverrides();
// optionally provide a pass through on the top-level eslint
eslint.addOverrides();
```

ren-yamanashi:
....🤔

WIP

ren-yamanashi:
@mrgrain 

Based on your suggestions, I have made corrections. Please review:

👇 Class Diagram

```mermaid
classDiagram
class FileBase {
  <<abstract>>
  +synthesize()
}

class IESLintConfig {
  <<interface>>
  rules
}

class ESLint {
  +synthesize()
}

class EslintFlatConfigFile {
  +synthesize()
}

class ESLintFlatConfig {
  rules
  +addRules()
  +addPlugins()
  +addOverrides()
}

class StylisticConfig {
  rules
}

class PrettierConfig {
  rules
}

FileBase --|> EslintFlatConfigFile

ESLint o-- IESLintConfig
ESLint o-- EslintFlatConfigFile

IESLintConfig <|.. StylisticConfig
IESLintConfig <|.. PrettierConfig
IESLintConfig <|.. ESLintFlatConfig
```

- ESLint

  - Accepts external configuration

- EslintFlatConfigFile

  - Represents ESLint configuration files

- ESLintFlatConfig
  - Represents ESLint configuration
  - Can add rules and plugins

PR #4091, opened 2025-02-15
feat(typescript): subclasses of `TypeScriptProject` can customize tsc compiler default options

Dzuelu:
Projects inheriting TypeScriptProject may want to modify the default typescript config without overwriting it's projects children from changing the default config. Provides a simple way to allow sub projects to modify compile options while letting sub-sub projects do the same.

Moves `compilerOptionDefaults` from constructor into a protected method that subclasses can overwrite if needed.

---


mrgrain (changes requested):
Good idea, but make it a protected method please.

Having mutable public properties make things more complicated than needed.

src/typescript/typescript.ts
```diff
@@ -363,13 +365,43 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
   readonly tsJestOptions?: TsJestOptions;
 }
 
+// Transforms readonly interface to have modifiable values
+type Writeable<T> = { -readonly [P in keyof T]: T[P] };
+
```
mrgrain, line 370: Related to the other comment, this shouldn't be a writeable prop. Writeable props cause issues w.r.t to state.

```suggestion
```

mrgrain (changes requested):

src/typescript/typescript.ts
```diff
@@ -363,13 +365,43 @@ export interface TypeScriptProjectOptions extends NodeProjectOptions {
...
  * TypeScript project
  * @pjid typescript
  */
 export class TypeScriptProject extends NodeProject {
   public static readonly DEFAULT_TS_JEST_TRANFORM_PATTERN = "^.+\\.[t]sx?$";
 
+  /**
+   * Projen default Typescript compiler options.
+   */
+  protected defaultCompilerOptions: Writeable<TypeScriptCompilerOptions> = {
```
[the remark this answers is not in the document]
Dzuelu: Ah, ok, that makes a lot of sense, thanks!

src/typescript/typescript.ts
```diff
@@ -601,6 +579,35 @@ export class TypeScriptProject extends NodeProject {
     }
   }
 
+  /**
+   * Projen default Typescript compiler options.
+   */
+  protected getDefaultTsCompilerOptions(): TypeScriptCompilerOptions {
```
mrgrain, line 585: Ugh. My bad. Apparently we cannot call this `getFoobar`.... 🤦🏻 

```suggestion
  protected defaultTsCompilerOptions(): TypeScriptCompilerOptions {
```

mrgrain (approved):

mrgrain:
Build is failing because of the package move. That's technically a breaking change I guess. Let's not include that for now.

mrgrain:
(Sorry closed by accident)

mrgrain (approved):

PR #4093, opened 2025-02-16
chore: stop github actions from running on forks

garysassano:
GitHub Actions shouldn't be run automatically on forks.

image

---


mrgrain:
This change needs to be made in projen itself.

garysassano:
@mrgrain Could you point to the piece of code that generates the workflow yaml?

mrgrain:
Probably in this file: https://github.com/projen/projen/blob/main/src/javascript/upgrade-dependencies.ts

Add a new option `runOnForks` here, defaulting to `false`
https://github.com/projen/projen/blob/256e44f1256f244050d093b36e819199150623a8/src/javascript/upgrade-dependencies.ts#L548

And then use it to create the workflow.

---

And now I am realizing this isn't possible because we need to repo name. _big sigh_

mrgrain:
> @mrgrain Why it's not possible?

GH Actions doesn't offer a context value to identify forks. We need the name of the "home" repo.

mrgrain:
I don't see a good way of doing this right now, so will leave this to userland.

PR #4131, opened 2025-03-14
fix(github): allow single string support for matrix domain property

### LINKED ISSUE #4130: Issue setting dynamic matrix strategy values from context
@mrgrain I have a similar issue, when trying to set a value for `strategy -> matrix`, using the domain property as string value.

[Following this example](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-contexts-to-create-matrices):
```
on:
  repository_dispatch:
    types:
      - test
 
jobs:
  example_matrix:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        version: ${{ github.event.client_payload.versions }}
    steps:
      - uses: actions/setup-node@v4
        with:
          node-version: ${{ matrix.version }}
```
Since the `domain` property is of type `Record<string, JobMatrixValue[]>`, I need to define it as 
```
            strategy: {
                matrix: { domain: { version: ['${{ github.event.client_payload.versions }}'] } },
            },
```

Which then becomes:
```
    strategy:
      matrix:
        version:
          - ${{ github.event.client_payload.versions }}
```
Instead of:
```
    strategy:
      matrix:
        version: ${{ github.event.client_payload.versions }}
```
Should the type of domain perhaps become `Record<string, string | JobMatrixValue[]>`, or how else can I achieve this?

janwvjaarsveld:
Fixes #4130 

This PR allows a single string to be defined for GithubWorkflows `/jobs/strategy/matrix/domain`. 

Now you can also pass a dymanically created matrix, rather than a predefined fixed list:
```
 strategy: {
     matrix: { 
        domain: { 
           version: '${{ github.event.client_payload.versions }}' 
        }
     },
 },
```
The output will look like:
```
    strategy:
      matrix:
        version: ${{ github.event.client_payload.versions }}
```
---


src/github/workflows-model.ts
```diff
@@ -497,7 +497,7 @@ export interface JobMatrix {
    * matrix.os property as the value of the runs-on keyword to create a job
    * for each operating system.
    */
-  readonly domain?: Record<string, JobMatrixValue[]>;
+  readonly domain?: Record<string, string | JobMatrixValue[]>;
```
mrgrain, line 500:
```suggestion
  readonly domain?: Record<string, JobMatrixValue | JobMatrixValue[]>;
```

mrgrain (changes requested):
I think we can allow all of these here, right `string | boolean | number`?

janwvjaarsveld:
> I think we can allow all of these here, right `string | boolean | number`?

I was also wondering about this, but don't think that it makes sense? This is specifically to allow for passing a dynamic matrix in. For something like boolean or number, an array with a single entry will have the desired outcome. 
But I can change it if you think it makes more sense?

src/github/workflows-model.ts
janwvjaarsveld: I intially had it like this, but think that only string is really needed for what this PR tries to achieve. See my comment below.

mrgrain:
Gotcha, now I'm with you 👌

mrgrain (approved):

PR #4134, opened 2025-03-15
feat(awscdk): add customizable CDK app synthesis command and improved default for bun users

automartin5000:
Adds a new `app` property to `AwsCdkTypeScriptAppOptions` that allows users to fully customize the command used to synthesize their CDK application. This enables better support for different package managers including Bun, npm, and pnpm.

The implementation includes:
- Bun specific default command
- Path normalization for relative imports
- Validation to prevent conflicting options
- Comprehensive test coverage

---


test/awscdk/awscdk-app.test.ts
```diff
@@ -195,17 +191,58 @@ describe("integ-runner", () => {
...
+    expect(snapshot["package.json"].devDependencies).toMatchObject({
+      "aws-cdk-lib": "^2.0.0-rc.1",
+    });
+  });
+  it("has a constructs dev depdendency", () => {
+    expect(snapshot["package.json"].devDependencies).toMatchObject({
+      constructs: "^10.0.5",
+    });
+  });
+  it("has the correct import for the sample file", () => {
```
automartin5000, line 215: No idea what this test is for, so I took a wild guess

mrgrain:
I'm feeling pretty strongly about `aws-cdk-lib` et. al. being a production aka runtime dependency in an `AwsCdkAppProject`. Why do you think they should be dev deps?

automartin5000:
> I'm feeling pretty strongly about `aws-cdk-lib` et. al. being a production aka runtime dependency in an `AwsCdkAppProject`. Why do you think they should be dev deps?

Dev deps are for packages that are only used locally and should not be able to be accidentally bundled into runtime code. 

"Dependencies" is for code that is run in production.

automartin5000:
Quote from the [npm docs](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file):

> "dependencies": Packages required by your application in production.
"devDependencies": Packages that are only needed for local development and testing.

mrgrain:
My CDK apps definitely run in production and definitely require `aws-cdk-lib`!? What is it that you run in production that's not the CDK app and how come the CDK code isn't required?

---

Either way, we don't want projen to diverge from what `cdk init` does. That's the better place to have this discussion anyway, and not this unrelated PR.

test/awscdk/awscdk-app.test.ts
mrgrain: Yes it's asserting the correct v2 style imports, compared to the old v1 imports

src/awscdk/awscdk-app-ts.ts
```diff
@@ -23,7 +23,11 @@ export interface AwsCdkTypeScriptAppOptions
    * @default "main.ts"
    */
   readonly appEntrypoint?: string;
-
+  /**
+   * Full override for the cdk.json app field.
+   * @example "bunx tsx -P tsconfig.build.json src/main.ts"
```
mrgrain, line 28: Should this not just happen automatically when we use bun?

src/awscdk/awscdk-app-ts.ts
```diff
@@ -23,7 +23,11 @@ export interface AwsCdkTypeScriptAppOptions
...
+   */
+  readonly cdkAppOverride?: string;
```
mrgrain, line 30: Let's just use `app` from `CdkConfigOptions` and not a new option that does the same thing. We'll ned validation to ensure that only one of `appEntrypoint` or `app` is set.

automartin5000:
> My CDK apps definitely run in production and definitely require `aws-cdk-lib`!? What is it that you run in production that's not the CDK app and how come the CDK code isn't required?
> 
> Either way, we don't want projen to diverge from what `cdk init` does (`aws-cdk` are is correctly classified as dev dep). That's the better place to have this discussion anyway, and not this unrelated PR.

Fair I'll remove it. But you're using `aws-cdk-lib` in your _runtime_ code? Or do you mean in the production pipeline? The latter still qualifies as a `devDep` to me (and based on my reading of the npm docs)

src/awscdk/awscdk-app-ts.ts
automartin5000: That's what's fixed by this PR. Right now `npx` is hardcoded

src/awscdk/awscdk-app-ts.ts
automartin5000: I took a pass at this and realized that `app` is required by `CdkConfigOptions`, which means I'd have to use a Partial type which I don't think JSII supports.

mrgrain: I see. I think than we can just make the option here use the same name on maybe copy some of the existing jsdoc.

src/awscdk/awscdk-app-ts.ts
mrgrain: Oh I see! I guess I was wondering if bun should run the app without ts-node involved? But happy to leave it as it is.

automartin5000: I think you're right, I'll update that example.

src/awscdk/awscdk-app-ts.ts
automartin5000: Ok, I've cleaned this up and just copied `app` from `CdkConfigOptions`. Also added the error handling for both `appEntrypoint` and `app` being set.

automartin5000:
@mrgrain ok I think this is ready for another review

mrgrain:
Looks good, but build is failing

test/__snapshots__/inventory.test.ts.snap
```diff
@@ -7038,7 +7038,7 @@ exports[`inventory 1`] = `
         "fullType": {
           "primitive": "string",
         },
-        "initialValue": ""~5.8.0"",
+        "initialValue": ""~5.7.0"",
```
automartin5000, line 7041: @mrgrain any idea why this snapshot changed?

mrgrain: Looks like this was updated on main and the merge didn't pick the right value

automartin5000: hm, I can't seem to figure out how to resolve it, can you assist?

mrgrain: Will do! I'm currently on holiday, but can have a look when I'm back in 2 weeks.

src/awscdk/awscdk-app-ts.ts
```diff
@@ -211,7 +211,7 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
     switch (this.package.packageManager) {
       case NodePackageManager.BUN:
         // bun doesn't support tsconfig overrides yet
-        return `bunx ${appEntrypoint}`;
+        return `bun ${appEntrypoint}`;
```
mrgrain, line 214: I think the explicit `bun run` would be preferred:

```suggestion
        return `bun run ${appEntrypoint}`;
```

automartin5000: updated

mrgrain:
@automartin5000 you need to fix this, otherwise the PR will not get merged:

> Error: Contributor statement missing from PR description. Please include the following text in the PR description: By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

automartin5000:
> @automartin5000 you need to fix this, otherwise the PR will not get merged:
> 
> 
> 
> > Error: Contributor statement missing from PR description. Please include the following text in the PR description: By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

fixed

mrgrain (approved):

mrgrain (approved):

PR #4136, opened 2025-03-15
fix(github): allow setting workflow env at global scope

### LINKED ISSUE #4135: Allow setting environment variables on GithubWorkflow global scope
When defining a GithubWorkflow, I would like to be able to define environment variables at a global scope level, without having to manually add it with the `JsonPatch` class. 

Ideally it should be possible to do this:
```
    new GithubWorkflow(project.github!, 'test-workflow', { env: { FOO: "bar" } });
```
```
    {
      "name": "test-workflow",
      "on": {},
      "env": {
        "FOO": "bar"
      },
      "jobs": {}
    }
```
Is there any specific reason why this is not currently supported, or can I open a PR to introduce this?

janwvjaarsveld:
Fixes #4135 

It is currently not possible to set `env` for GithubWorkflows at a global scope, other than:
```
const workflow = new GithubWorkflow(project.github!, 'test-workflow')
workflow.file?.patch(
    JsonPatch.add('/env', {
        FOO: 'bar',
    })
);
```

This PR makes it possible to pass the `env` object as an option at initialization, to set it on workflow level, which is useful for environment variables that are applicable to all jobs. e.g. `NODE_VERSION`
```
new GithubWorkflow(project.github!, 'test-workflow', { env: { FOO: "bar" } });
``` 
---


mrgrain (approved):

janwvjaarsveld:
@mrgrain I changed the order of where `env` is defined on the interface and class, to force a workflow rerun (I don't seem to have access to do this in the Actions tab). 
Something went wrong after updating with changes from `main`, although everything still works locally.

mrgrain:
> (I don't seem to have access to do this in the Actions tab). 

Yeah. Pushing a new commit is the only way for contributors. Otherwise tag me and wait.

PR #4214, opened 2025-05-10
fix(github): add correct permissions when using GitHub Token

Hi-Fi:
Currently created `upgrade-main` workflow uses by default specific PAT (from secret `PROJEN_GITHUB_TOKEN`) that has rights to write content and pull requests.

While it's possibe to configure secret to be run spefific `GITHUB_TOKEN`, workflow permissions are not modifiable accordingly so that token would get correct rights.

This fix would add needed permissions when using `GITHUB_TOKEN` instead of custom PAT.

---


mrgrain:
Not against this, but please provide a reference why the new permissions is required.

Hi-Fi:
> Not against this, but please provide a reference why the new permissions is required.

Currently instructed permissions for `PROJEN_GITHUB_TOKEN` are https://github.com/projen/projen/blob/9a455a69e69721b2ce40a1986afc66dcc5e5b779/docs/integrations/github/index.md?plain=1#L36-L39

If this is used from secrets, it can create pull requests (and do also other stuff).

When using `GITHUB_TOKEN`, permissions that token gets should be required and limited same time in workflow/job level. And that's what this PR does; `upgrade-main` gets those permissions when using run specific token.

mrgrain (approved):

Issue #4199, opened 2025-05-02
CDK feature flags are not set in cdk.json


ben-buitendijk:
When creating a new CDK app, the context field in the cdk.json is empty and does not have any feature flag set.

I think it is because of this condition that only adds them for CDKv1:
https://github.com/projen/projen/blob/92cbedbac756d03bac399b1d30459b1af1f543c1/src/awscdk/awscdk-app-ts.ts#L147

There is an list of feature flags here (although only applicable for CDKv1):
https://github.com/projen/projen/blob/main/src/awscdk/internal.ts

I would like to add a CDKv2 list that include the latest recommended flags: 
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md

I think we also need to update [this line](https://github.com/projen/projen/blob/92cbedbac756d03bac399b1d30459b1af1f543c1/src/awscdk/cdk-config.ts#L109), because not all feature flags are booleans, for example: 

```
"@aws-cdk/core:target-partitions": [
      "aws",
      "aws-cn"
    ],
```

I may be able to implement/update this

mrgrain:
There is a machine readable file that we attempt to read from:

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/recommended-feature-flags.json

ben-buitendijk:
Oh nice! How should I manage this dependency-wise? 

Is it possible to do that without adding aws-cdk-lib to Projen's dependencies?

mrgrain:
You are right, we cannot add aws-cdk-lib to Projen's dependencies.

From the top of my head, the options could be:

- add a new [macro](https://github.com/projen/projen/blob/838a28e1c131f6238a62906363d30b5bcabc3896/src/cli/macros.ts#L11)
- add them in during the first synth of `cdk.json` (similar to [sample file](https://github.com/projen/projen/blob/838a28e1c131f6238a62906363d30b5bcabc3896/src/sample-file.ts#L82))
- update mechanism in projen repo to import the file and bundle just the file (less ideal, because doesn't match versions)

Either solution will need to allow for the file not existing in an older version.

ben-buitendijk:
I thought maybe we could try to read from `node_modules/aws-cdk-lib`. 
That would match the correct version and keep it up to date, but it would only work for Typescript. 
It seems that json file is not included in the Python library (I didn't check Java, but I assume its also not included)


[...]

PR #4215, opened 2025-05-12
feat: cdkv2 feature flags for awscdk-app-ts

ben-buitendijk:
Fixes # https://github.com/projen/projen/issues/4199

For awscdk-app-ts, this will try to read the recommended-feature-flags.json from the node_modules. 
If it cannot be read/parsed, it will fall back to a default list of feature flags for CDKv2

The defaults are set to CDKv1 so it should be backwards compatible with all existing projects

Something I'm not sure of:  if this changes existing project's their context (add feature flags), can we introduce breaking changes that way?

---

### DIFF
--- a/src/awscdk/awscdk-app-ts.ts
+++ b/src/awscdk/awscdk-app-ts.ts
@@ -1,5 +1,9 @@
 import * as fs from "fs";
 import * as path from "path";
+import { Component } from "../component";
+import { DependencyType } from "../dependencies";
+import { NodePackageManager, RunBundleTask } from "../javascript";
+import { TypeScriptAppProject, TypeScriptProjectOptions } from "../typescript";
 import { AutoDiscover } from "./auto-discover";
 import { AwsCdkDeps, AwsCdkDepsCommonOptions } from "./awscdk-deps";
 import { AwsCdkDepsJs } from "./awscdk-deps-js";
@@ -7,10 +11,6 @@ import { CdkConfig, CdkConfigCommonOptions } from "./cdk-config";
 import { CdkTasks } from "./cdk-tasks";
 import { IntegRunner } from "./integ-runner";
 import { LambdaFunctionCommonOptions } from "./lambda-function";
-import { Component } from "../component";
-import { DependencyType } from "../dependencies";
-import { NodePackageManager, RunBundleTask } from "../javascript";
-import { TypeScriptAppProject, TypeScriptProjectOptions } from "../typescript";
 
 export interface AwsCdkTypeScriptAppOptions
   extends TypeScriptProjectOptions,
@@ -144,7 +144,8 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
     }
 
     this.cdkConfig = new CdkConfig(this, {
-      featureFlags: this.cdkDeps.cdkMajorVersion < 2,
+      // featureFlags: this.cdkDeps.cdkMajorVersion < 2,
+      cdkMajorVersion: this.cdkDeps.cdkMajorVersion,
       buildCommand: this.runTaskCommand(this.bundler.bundleTask),
       watchIncludes: [`${this.srcdir}/**/*.ts`, `${this.testdir}/**/*.ts`],
       watchExcludes: [
--- a/src/awscdk/cdk-config.ts
+++ b/src/awscdk/cdk-config.ts
@@ -1,8 +1,10 @@
+import * as fs from "fs";
+import * as path from "path";
 import { Component } from "../component";
 import { JsonFile } from "../json";
 import { Project } from "../project";
+import { AwsCdkTypeScriptApp } from "./awscdk-app-ts";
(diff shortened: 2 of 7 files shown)


hoegertn:
What about adding the feature flags to the `.projenrc.ts` on generation? So it works for all new projects and does not break existing ones. I think the flags should be visible and not hidden in the projen magic.

ben-buitendijk:
I would like the context to be updated whenever I update my CDK version, could that also break things? (I'm not familiar enough with all feature flags to say for sure)

If it can, then, yeah, it would be better to add then during the projen new command. 

Can I add things to the projenrc.ts from inside a "project" (such as awscdk-app-ts) or do I need to work from `cli/cmds/new.ts`?

hoegertn:
> I would like the context to be updated whenever I update my CDK version, could that also break things? (I'm not familiar enough with all feature flags to say for sure)
> 
> If it can, then, yeah, it would be better to add then during the projen new command.

Feature flags should be chosen on project creation and rarely touch after that. They hide breaking changes so you can update CDK without breaking your app. If you update them too, they would not even be necessary.

mrgrain:
Been pretty busty sorry, but still on my list to give feedback and ideas.

ben-buitendijk:
Oh I just thought of another issue... What if you add projen to an existing CDK project, it would overwrite the cdk.json. You can add them back in via the projenrc.ts, but if I'm reading it from another file, there is no way to know which ones are correct

So the best option would be to write the context to the projenrc.ts file I guess? Are there any examples for that?

src/awscdk/cdk-config.ts
```diff
@@ -158,6 +170,78 @@ export class CdkConfig extends Component {
...
+  private tryLoadFeatureFlags(project: Project) {
+    if (project instanceof AwsCdkTypeScriptApp) {
+      try {
+        const jsonFile = fs.readFileSync(
+          path.join(
+            process.cwd(),
+            "node_modules",
+            "aws-cdk-lib",
+            "recommended-feature-flags.json"
+          ),
```
mrgrain: There might even be helper methods around to do this already.

```suggestion
          require.resolve("aws-cdk-lib/recommended-feature-flags.json"),
```

src/awscdk/awscdk-app-ts.ts
```diff
@@ -205,7 +205,8 @@ export class AwsCdkTypeScriptApp extends TypeScriptAppProject {
     }
 
     this.cdkConfig = new CdkConfig(this, {
-      featureFlags: this.cdkDeps.cdkMajorVersion < 2,
+      // featureFlags: this.cdkDeps.cdkMajorVersion < 2,
```
mrgrain: Guess we should have a new top-level option to confiugure this?
```suggestion
      featureFlags: options.featureFlags ?? this.cdkDeps.cdkMajorVersion < 2,
```

src/awscdk/cdk-config.ts
```diff
@@ -158,6 +170,78 @@ export class CdkConfig extends Component {
...
+   * @param fflagsEnabled Include all feature flags. Defaults to `true`.
+   * @param cdkMajorVersion The major version of the CDK to include the feature flags for.
+   *     Defaults to 1.
+   * @returns The updated context.
+   */
+  private setFeatureFlags(
+    context: Record<string, any>,
+    fflagsEnabled: boolean = true,
+    cdkMajorVersion: number = 1
+  ) {
```
mrgrain: I think customer provided context should take precedence here, i.e. if a user has set a feature flag in the context than we should not override it with the default.

mrgrain:
> The CdkApps are all extending the CdkConfig interface, so featureFlags and context are top-level options
> 
> I think featureflags should be disabled for CDK v2 by default, otherwise its not backwards compatible?

Ah great! We could make a breaking change, but I agree with you for this case. (Also reminds me to remove CDK v1 support 🙃 )

src/awscdk/cdk-config.ts
```diff
@@ -15,12 +16,24 @@ export interface CdkConfigCommonOptions {
...
+   * Include all feature flags in cdk.json. Default is true for CdkVersion 1 and false for CdkVersion 2.
+   *
+   * For CDK version 2:
+   * Enabling this will update the feature flags in the cdk.json when you update the CDK version,
+   * which could introduce breaking changes.
+   *
+   * Make sure to double-check the changes to cdk.json before deploying.
+   * You can overwrite any feature flag by passing it into the context field.
    *
    * @default true
```
mrgrain:
```suggestion
   * Include all feature flags in cdk.json.
   *
   * For CDK version 2:
   * Enabling this will update the feature flags in the cdk.json when you update the CDK version,
   * which could introduce breaking changes.
   *
   * Make sure to double-check the changes to cdk.json before deploying.
   * You can overwrite any feature flag by passing it into the context field.
  *
  * @default - `false` for AWS CDK V2 and `true` for AWS CDK V1
```

PR #4230, opened 2025-05-26
feat: include Biome support to projen

### LINKED ISSUE #3966: Feature request: Add support for biome as an alternative to eslint/prettier
We switched all our projects from eslint and prettier to biome. 

https://biomejs.dev/ is a tool which can be used as a replacement for either prettier or eslint/typescript-eslint or both.
So biome can lint or format or both. Its much faster and the configuration is much easier (at least in my opinion).

It would be great if we can configure projen to use/configure biome.

I could imagine a config like (similar to github/githuboptions):
```jsonc
biome: true // to activate both formatting + linting
biomeOptions: { // optional
  format: false // false=would still use prettier for formatting
  lint: false // false=would still use eslint for linting
  // .. additional options from https://biomejs.dev/reference/configuration/
}
```

<details>
  <summary>Our current `biome.jsonc` looks like this (just for reference):</summary>

```jsonc
{
    "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
    "vcs": {
        "enabled": true,
        "clientKind": "git",
        "useIgnoreFile": true
    },
    "files": {
        "ignore": [
            // projen managed files
            ".projen/",
            ".projenrc.ts",
            "tsconfig*.json",
            "package.json",
            // end of projen managed files
            "*test.ts.snapshot/**" // cdk integTest snapshots
        ]
    },
    "formatter": {
        "enabled": true,
        "formatWithErrors": false,
        "indentStyle": "space",
        "indentWidth": 4,
        "lineEnding": "lf",
        "lineWidth": 120,
        "attributePosition": "auto"
    },
    "organizeImports": { "enabled": true },
    "linter": {
        "enabled": true,
        "rules": {
            "recommended": true,
            "correctness": {
                "noUnusedImports": "warn"
            }
        }
    },
    "overrides": [
        {
            "include": ["**/*.test.ts"],
            "linter": {
                "rules": {
                    "style": {
                        "noNonNullAssertion": "off"
                    },
                    "suspicious": {
                        "noExplicitAny": "off"
                    }
                }
            }
        }
    ]
}
```
</details>

Hi-Fi:
Allows to use [Biome](https://biomejs.dev/) for linting and formatting.

Fixes #3966

---


Hi-Fi:
@mrgrain any comments? Not sure if the will is to have config that produces as close as current ruleset with eslint, or to use biome's recommended rules. 

Of course e.g. quote types, dangling commas etc. are opinionated area but now just coming from defaults.

mrgrain:
@Hi-Fi I'm so sorry. Been super swamped with other things 😞 😢  I _really_ do appreciate the PR though. 

> Not sure if the will is to have config that produces as close as current ruleset with eslint, or to use biome's recommended rules.

We prefer the upstream tool's recommendations and defaults. Eslint doesn't have any, so we had to pick some.

src/javascript/projenrc.ts
```diff
@@ -47,6 +49,11 @@ export class Projenrc extends ProjenrcFile {
         "import/no-extraneous-dependencies": "off",
       },
     });
+
+    if (this.project instanceof NodeProject) {
```
mrgrain, line 53: let the presence of the `Biome` component be enough indication. I.e. remove the if.

src/javascript/projenrc.ts
```diff
@@ -47,6 +49,11 @@ export class Projenrc extends ProjenrcFile {
...
+      const biome = Biome.of(this.project);
+      biome?.addLintPattern(this.filePath);
+    }
```
mrgrain: let the presence of the Biome component be enough indication. I.e. remove the if.
```suggestion
    const biome = Biome.of(this.project);
    biome?.addLintPattern(this.filePath);
```

src/javascript/biome-config.ts
mrgrain, line 1: let's move this into a sub-directory: `src/javascript/biome/biome-config.ts`

src/javascript/biome.ts
mrgrain, line 1: let's move this into a sub-directory: `src/javascript/biome/biome.ts`

src/javascript/index.ts
```diff
@@ -10,3 +10,5 @@ export * from "./projenrc";
 export * from "./typescript-config";
 export * from "./upgrade-dependencies";
 export * from "./yarnrc";
+export * from "./biome";
+export * from "./biome-config";
```
mrgrain: There's a lot of names in this generated file. We should namespace this if possible:

```suggestion
export * as BiomeConfig from "./biome-config";
```

Disclaimer: I don't know if that works with jsii.

src/javascript/biome.ts
```diff
@@ -0,0 +1,313 @@
...
+      overrides: options.overrides,
+    };
+
+    project.addDevDeps(`@biomejs/biome@${this.optionsWithDefaults.version}`);
+
+    this.biomeConfiguration = _createBiomeConfiguration(
+      this.optionsWithDefaults
+    );
+
+    new JsonFile(this, this.configFile, {
```
mrgrain: expose this as a public property
```suggestion
    this.file = new JsonFile(this, this.configFile, {
```

src/javascript/biome.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  private readonly configFile: string;
+  private readonly optionsWithDefaults: BiomeOptions;
+  private readonly biomeConfiguration: IConfiguration;
+  private readonly _lintPatterns: Set<string>;
+  private readonly biomeCommand = "biome check --write";
+  /**
+   * Biome task.
+   */
+  public readonly biomeTask: Task;
```
mrgrain:
```suggestion
  public readonly task: Task;
```

src/javascript/biome.ts
```diff
@@ -0,0 +1,313 @@
...
+      marker: true,
+    });
+
+    this._lintPatterns = new Set([]);
+
+    this.biomeTask = this.createLocalBiomeTask();
+    project.testTask.spawn(this.biomeTask);
+  }
+
+  public override preSynthesize(): void {
```
mrgrain: This feels super strict. You could convince of a warning, but I'm leaning towards not checking this here. Instead we should check in the `NodeProject` and `TypeScriptProject` constructors.

src/javascript/node-project.ts
```diff
@@ -858,6 +873,10 @@ export class NodeProject extends GitHubProject {
       );
     }
 
+    if (options.biome ?? false) {
+      this.biome = new Biome(this, { ...options.biomeOptions });
+    }
+
```
mrgrain, line 883: I think here we want a check that only oneof `biome` and `prettier` can be set.
And similar in `TypeScriptProject` with `eslint`.

There is a `multipleSelected` helper in utils than can be used.


```
    if (multipleSelected([options.biome, options.prettier])) {
      throw new Error("Only one of biome, and prettier can be enabled.");
    }
```

mrgrain (changes requested):
Super close, thanks for this amazing effort!

> This seems to be going every night to conflict with yarn.lock, so maybe there's no point of merging this all the time but only when this is otherwise OK and can be merged before nightly update run.

Yes that's totally fine. When I approve it I will just deal with that myself.

src/javascript/index.ts
Hi-Fi: Namespaces seem to be supported (had to check also self with build), but just a bit different format:
```
src/javascript/index.ts:13:13 - error JSII8004: Submodule namespaces must be camelCased or snake_cased. Rename "Biome" to biome

13 export * as Biome from "./biome/biome";
               ~~~~~
src/javascript/index.ts:14:13 - error JSII8004: Submodule namespaces must be camelCased or snake_cased. Rename "BiomeConfig" to biomeConfig

14 export * as BiomeConfig from "./biome/biome-config";
```

src/javascript/projenrc.ts
Hi-Fi: Changed `of` so that type checking is not needed.

src/javascript/biome.ts
Hi-Fi: Removed check, it came from one other projen components I got idea for this.

src/javascript/node-project.ts
Hi-Fi: Check added.

mrgrain (approved):

mrgrain:
Looking into the python issue

PR #4232, opened 2025-05-27
fix(javascript): remove requiredEnv and dot-exclusion from codeartifact login task

diranged:
This PR has two changes that I think are important to improving the usability of the `yarn run ca:login` task.

#### Remove the `requiredEnv` settings

Requiring that the `ca:login` task sees a `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variable means that the task does not work for local developers who use other methods to log in to AWS. This breaks if you are running in a Kubernetes Pod that is using web identity access files. This breaks if you are operating using AWS SSO logins. This breaks if you simply have your credentials hard-coded in your AWS config file.

I propose removing this requirement because it's just too explicit.

#### Remove `.` exclusion from `codeArtifactRegex`

We have hundreds of CodeArtifact repositories that use a `.` in their name - ie `"myrepo.git"`.. it's completely valid to have a `.` in the name of a CodeArtifact repository, yet the `codeArtifactRegex` was excluding that and then failing to match. I am removing the `.` because it is valid in the naming patterns for codeartifact. I have tested this change locally and it works great.

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -1114,7 +1114,6 @@ export class NodePackage extends Component {
     }
 
     this.project.addTask("ca:login", {
-      requiredEnv: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"],
       steps: [
         { exec: "which aws" }, // check that AWS CLI is installed
         ...this.scopedPackagesOptions.map((scopedPackagesOption) => {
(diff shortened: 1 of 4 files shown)


diranged:
(oops, I will fix the tests)

diranged:
@mrgrain Check out https://github.com/projen/projen/pull/4232/commits/300a7b1d31b73ca1c65c56adb9d6cacd0b6ebe1d and let me know if those tests look reasonable

mrgrain (approved):

PR #4270, opened 2025-06-22
feat(project): Projects can be initialized from env vars or .env files

automartin5000:
# Summary
.projenrc files can end up being extremely lengthy and have a number of constants that are simply copy and pasted between projects. This PR enables them to initialize from a .env file or environment variables which are easily shared from a central location (such as a configuration store or password vault).

## Changes
- Update base project class with a `fromEnv` static method that supports initializing a project from either a .env file or environment variables

## Testing
- Added relevant tests


mrgrain:
> However, I still think the ability to consume a standardized .env file makes this approach more appealing. I was anticipating putting vars other than projen vars in that shared .env file, including local dev runtime vars.

I can sympathize with this. However projen will always require a well-known entry point. This PR is proposing to use the programmatic entrypoint and add convenience helpers to extract project settings from an arbitrary data source. `.projenrc.json` files are going much further by eliminating the need for the programmatic entrypoint. Beyond that, projen even supports _any_ arbitrary command as "synth" command (aka `default` task). So feasibly one could even build a (user-land) helper to synth from an `.env` file or environment variables.


> [...] Also, this PR enabled env var overrides, although I suppose that could also be handled by a custom secondary JSON file.
> 
> I suppose it wouldn't be the end of the world if this PR wasn't accepted, but I also don't see how it would hurt either.

It adds a lot of API surface for a feature that can be implemented very easily as a user land function.

```ts
new Project(projectOptionsFromAnyDataSource());
```

PR #4272, opened 2025-06-23
feat(github): configure `shell` in workflow steps

mrgrain:
Add optional shell property to GitHub Actions workflow steps.

This adds a missing property for configuring workflow steps that require specific shell environments.

---


src/github/workflows-model.ts
```diff
@@ -420,6 +420,15 @@ export interface StepConfiguration {
...
   readonly workingDirectory?: string;
+
+  /**
+   * Overrides the default shell settings in the runner's operating system and the job's default.
+   *
+   * Refer to GitHub documentation for allowed values.
+   *
+   * @see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
+   */
+  readonly shell?: string;
```
otaviomacedo, line 431: Where is this used?

mrgrain: Not currently used in projen directly, but not having it limits our user's ability to configure this setting on their workflows.

rix0rrr (approved):

src/github/workflows-model.ts
mrgrain: Ah no, you were right. Needed to be added to the render step.

PR #4285, opened 2025-07-08
fix(typescript): ts-jest inline compiler options type

### LINKED ISSUE #4284: Inline ts-jest tsconfig options use wrong type
When trying to set custom tsconfig compiler options for ts-jest, the generated output is incorrect:

Example `.projenrc.ts`:
```ts
new awscdk.AwsCdkConstructLibrary({
  // ...
  tsJestOptions: {
    transformOptions: {
      tsconfig: TsJestTsconfig.custom({
        compilerOptions: { isolatedModules: true },
      }),
    },
  },
})
```

This generates the following `package.json`:
```json
{
  "jest": {
    "transform": {
      "^.+\\.[t]sx?$": [
        "ts-jest",
        {
          "tsconfig": {
            "compilerOptions": {
              "isolatedModules": true
            }
          }
        }
      ]
    }
  }
}
```

However, when running `pj test` an error is shown:
```
  ● Test suite failed to run                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                 
    error TS5023: Unknown compiler option 'compilerOptions'.  
```

I'm pretty sure that the `compilerOptions` field in the package.json should not be there. Instead it should look like this, matching the ts-jest docs.

```json
{
  "jest": {
    "transform": {
      "^.+\\.[t]sx?$": [
        "ts-jest",
        {
          "tsconfig": {
            "isolatedModules": true
          }
        }
      ]
    }
  }
}
```

I suppose a possible fix could be to change the [TsJestTsconfig.custom](https://github.com/projen/projen/blob/f55d1b1539b1db289e6c1f1f12a2b6b48043cad4/src/typescript/typescript.ts#L149) function like this:

```diff
-public static custom(config: TypescriptConfigOptions) {
+public static custom(config: TypeScriptCompilerOptions) {
```

jessestricker:
fixes #4284

As outlined in the linked issue, the inline compiler options should not be wrapped in an extra `compilerOptions` property.
See also the [ts-docs for the tsconfig option](https://kulshekhar.github.io/ts-jest/docs/getting-started/options/tsconfig/#inline-compiler-options).

---

### DIFF
--- a/docs/api/typescript.md
+++ b/docs/api/typescript.md
@@ -10493,16 +10493,16 @@ Force` ts-jest` to use its built-in defaults even if there is a `tsconfig.json` 
 ```typescript
 import { typescript } from 'projen'
 
-typescript.TsJestTsconfig.custom(config: TypescriptConfigOptions)
+typescript.TsJestTsconfig.custom(config: TypeScriptCompilerOptions)
 ```
 
 Inline compiler options.
 
-> [TypescriptConfigOptions](TypescriptConfigOptions)
+> [TypeScriptCompilerOptions](TypeScriptCompilerOptions)
 
 ###### `config`<sup>Required</sup> <a name="config" id="projen.typescript.TsJestTsconfig.custom.parameter.config"></a>
 
-- *Type:* projen.javascript.TypescriptConfigOptions
+- *Type:* projen.javascript.TypeScriptCompilerOptions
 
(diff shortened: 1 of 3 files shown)


mrgrain (approved):
Nice! Congrats to your first contribution to projen and thank you!

PR #4326, opened 2025-08-15
feat(cli): framework for interactive prompts in cli

### LINKED ISSUE #4086: Feature Request: Interactive setup for code quality tools and test frameworks
## Describe the feature

Add interactive CLI prompts for setting up code quality tools and test frameworks during project initialization for TypeScript / JavaScript.

⬇️ sample

sample-image

(This issue was once proposed in the [aws-cdk](https://github.com/aws/aws-cdk) repository, so more infomration is available at https://github.com/aws/aws-cdk/issues/33269#issuecomment-2657443881 and can be found at)

### Use Case

Currently, the CLI automatically installs preset tools like Jest for testing. However, developers often have different preferences for their development tooling. An interactive setup would allow users to:

 - Choose their preferred package manager (e.g, npm, yarn, pnpm)
- Choose their preferred code formatter (e.g., Prettier, biome)
- Select linting tools (e.g., ESLint, biome)
- Pick testing frameworks (e.g., Jest, Vitest)

### Proposed Solution

The following are the necessary implementation details as given by @mrgrain on https://github.com/aws/aws-cdk/issues/33269#issuecomment-2636543259

> - Add interactive prompt setup to projen
> - Make the interactive prompts available to createProject function
> - Add missing desired features as projen components (e.g. Biome, Vitest)
> - Create a new "cdk init" like project type (or update the existing one) that matches the desired experience
> - Any tweaks required to allow projen to created projects in ejected mode (projen should be an option, but not forced upon users; this is already possible, but has some rough-edges)

### Acknowledgements

- [x] I may be able to implement this feature request

ren-yamanashi:
Fixes #4086 

## Changes description

Implemented a feature that allows selection of javascript tools (package manager / linter / formatter / test tool) when the `npx projen new` command is executed.

## Confirmation of operation

- [x] Pass unit tests.
- [x] Confirm that the functionality works correctly by actually executing the command.

> Confirm that the functionality works correctly by actually executing the command.

https://github.com/user-attachments/assets/c4607a22-3379-46db-adec-5f55aba49a86

---


ren-yamanashi:
MEMO: 

To make biome etc. available as a linter, you need to modify `. /.jsii` needs to be modified 
(currently biome is not added to .jsii, so biome cannot be selected as a linter)

ren-yamanashi:
### TODO

- [ ] package name
- [ ] package manager
- [ ] create issue
  - add vitest
  - update .jsii (add biome)

ren-yamanashi:
https://github.com/user-attachments/assets/e5e10611-f6aa-4ac9-b3a3-d08fe7363963

src/projects.ts
```diff
@@ -107,13 +108,29 @@ function createProject(opts: CreateProjectOptions) {
...
+      ? {
+          ...opts.projectOptions,
+          name: jsTools.projectName,
+          eslint: opts.projectOptions.eslint || jsTools.linter === "eslint",
+          prettier:
+            opts.projectOptions.prettier || jsTools.formatter === "prettier",
+          jest: opts.projectOptions.jest || jsTools.testTool === "jest",
+          packageManager:
+            opts.projectOptions.packageManager || jsTools.packageManager,
+        }
```
ren-yamanashi: In projen, biome is available as a linter / formatter, but specifying biome in the `args` property does not work.  
(I think this is because biome is not included in `.jsii`.)

---

Since these investigations and solutions are separate issues, this PR has a policy that biome is not selectable.

ren-yamanashi: #4326 assumes that vitest will be selectable as a test tool, but since vitest is currently not available through projen, it is currently not selectable as a test tool.

(I think it would be a good idea to modify this implementation when vitest becomes available in projen in the future.)

package.json
```diff
@@ -90,6 +90,7 @@
     "case": "^1.6.3",
     "chalk": "^4.1.2",
     "comment-json": "4.2.2",
+    "consola": "^3.4.2",
```
ren-yamanashi, line 96: I added [consola](https://github.com/unjs/consola) package because provide an interactive CLI prompt.

test/util.ts
```diff
@@ -99,42 +99,68 @@ export function synthSnapshotWithPost(project: Project) {
...
+
+  return projectdir;
+}
+
+export function withProjectDirSync(
   code: (workdir: string) => void,
   options: { git?: boolean; chdir?: boolean; tmpdir?: string } = {}
 ) {
   const origDir = process.cwd();
   const outdir = options.tmpdir ?? mkdtemp();
```
ren-yamanashi: What was originally a `withProjectDir` function is now defined as `withProjectDirSync`, which handles non-promise methods.

test/util.ts
```diff
@@ -99,42 +99,68 @@ export function synthSnapshotWithPost(project: Project) {
...
 
+export async function withProjectDirAsync(
+  code: (workdir: string) => Promise<void>,
+  options: { git?: boolean; chdir?: boolean; tmpdir?: string } = {}
+) {
+  const origDir = process.cwd();
+  const outdir = options.tmpdir ?? mkdtemp();
+
+  try {
+    const projectdir = setupProjectDir(outdir, options);
```
ren-yamanashi: Defined `withProjectDirAsync` method to execute promise functions as `withProjectDirAsync`.

src/projects.ts
```diff
@@ -69,8 +70,8 @@ export class Projects {
    * so that custom project types can detect whether the current synthesis is the
    * result of a new project creation (and take additional steps accordingly)
    */
-  public static createProject(options: CreateProjectOptions) {
-    createProject(options);
+  public static async createProject(options: CreateProjectOptions) {
```
mrgrain: This is a breaking change. Needs to be called out in the PR.

src/projects.ts
```diff
@@ -107,13 +108,33 @@ function createProject(opts: CreateProjectOptions) {
...
+  const jsTools = await cliPrompts.selectJsTools({
+    projectTypeName: projectType.typename,
+    projectOptions: {
+      packageName: opts.projectOptions.name,
+      linter: opts.projectOptions.eslint && "eslint",
+      formatter: opts.projectOptions.prettier && "prettier",
+      testTool: opts.projectOptions.jest && "jest",
+      packageManager: opts.projectOptions.packageManager,
+    },
+  });
```
mrgrain: I don't think this is the right place to call this. Instead call it in the CLI and pass the result here.

src/projects.ts
mrgrain: That's fine. Could you raise an issue for me to investigate?

mrgrain: Agreed.

src/projects.ts
```diff
@@ -135,8 +156,7 @@ function createProject(opts: CreateProjectOptions) {
     initProjectCode.push(`${varName}.synth();`);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-require-imports
-  const mainModule = require(mod);
+  const mainModule = await import(mod);
```
mrgrain: seems unrelated? pls revert

package.json
mrgrain: Signed-off. Nice find.

src/run-projenrc-json.task.ts
```diff
@@ -24,6 +24,8 @@ if (!type) {
   );
 }
 
+// top-level await is not supported in this file
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
mrgrain: correct but we also shouldn't just assume that this works.

src/cli/prompts/operations/select-js-tools.ts
```diff
@@ -0,0 +1,102 @@
+import { NodePackageManager } from "../../../javascript";
+import { InteractiveCliPrompt } from "../core/interactive-cli-prompt";
+
+type Linter = "eslint";
+
+type Formatter = "prettier";
+
+type TestTool = "jest";
+
+export const selectJsTools =
```
mrgrain, line 10: As an intermediate solution I'm fine with this. However the next step will be to make prompts selectable via the inventory, i.e. building an extensible way to A\ add new prompts and B\ mark properties as part of the prompt.

mrgrain (changes requested):
Thanks for getting this started! Two design issues I have with this approach:

- We should not change `createProject` to include the interaction. The interaction should be a CLI only feature and its result passed on to the existing `createProject`. This will also allow you to revert all the test changes you had to make. **This is a blocking requirement.**
- I'm okay with the current approach of hardcoded, inlined prompting in the CLI. However next step should be to come up with an extensible way to declare prompts as part of the inventory and have the CLI discover the prompts fro the inventory. This should be extensible for external packages as well. **We probably want a small RFC for this, so let's not make this part of this PR.**

Otherwise, we are missing a flag to enable/disable interactive mode, e.g. `--interactive` and `--no-interactive`. We should also auto-detect non TTY and CI environments (the latter using `process.env.CI`) and disable interaction automatically.

src/projects.ts
ren-yamanashi: Reverted this changes so, I resolve conversation.

https://github.com/projen/projen/pull/4326/commits/fadad8e357a9b5d1936aa1f9f4dd5808e8af8e3c

src/run-projenrc-json.task.ts
ren-yamanashi: Reverted this changes so, I resolve conversation.

https://github.com/projen/projen/pull/4326/commits/fb0180d7fecd6b9a759800e0a9f994125f954ce5

src/projects.ts
ren-yamanashi: Reverted this changes so, I resolve conversation.

https://github.com/projen/projen/pull/4326/commits/fb55bbd25544ffcc4424f77b2314736fc19d0e49

src/projects.ts
ren-yamanashi: Thanks!

I've decided to implement it within the `commandLineToProps` method in `src/cli/cmds/new.ts` as follows. What do you think?

https://github.com/projen/projen/blob/b918a53ffb4110b15602ec1b024218a4ca6e4493/src/cli/cmds/new.ts#L294-L314

test/util.ts
ren-yamanashi: Reverted this changes so, I resolve conversation.

test/util.ts
ren-yamanashi: Reverted this changes so, I resolve conversation.

src/projects.ts
ren-yamanashi: Upon checking, it seems this issue has been resolved. Therefore, I have also updated `biome` to be selectable.

see: https://github.com/projen/projen/blob/b918a53ffb4110b15602ec1b024218a4ca6e4493/src/cli/prompts/operations/select-js-tools.ts

ren-yamanashi:
@mrgrain 

> * We should not change `createProject` to include the interaction. The interaction should be a CLI only feature and its result passed on to the existing `createProject`. This will also allow you to revert all the test changes you had to make. **This is a blocking requirement.**


Thanks review!

I added `interactive` cli option, making interactive functionality a cli only feature.([see](https://github.com/projen/projen/blob/b918a53ffb4110b15602ec1b024218a4ca6e4493/src/cli/cmds/new.ts#L62-L66))

And I fixed `createProject` method was concerned with interactive functionality.([see this comment](https://github.com/projen/projen/pull/4326#discussion_r2441586235))

mrgrain:
This lists too many yarn versions. Order alphabetically.
image

mrgrain:
`Ctrl+C` on selection prompt produces an unsightly error.
image

ren-yamanashi:
@mrgrain 

I fixed code so that process exit when `Ctrl+C` is pressed.
And also sorted items alphabetically.

https://github.com/user-attachments/assets/d9dedf6a-95e1-4420-8a25-90a7c78b7c3f

src/cli/prompts/operations/select-js-tools.ts
```diff
@@ -0,0 +1,105 @@
...
+      defaultVal: packageName ?? "my-project",
+      placeholder: packageName ?? "my-project",
+    });
+
+    const packageManager = await interactiveCliPrompt.selectItem({
+      message: "Choose a package manager",
+      items: Object.entries(NodePackageManager).map(([_, val]) => ({
+        label: val,
+        value: val,
+      })),
```
mrgrain, line 47: We somehow need to filter out the `yarn` duplicates here. There should be exactly 2 entries for yarn: `yarn_classic` and `yarn_berry` (or whatever the non deprecated name is).

src/cli/prompts/core/interactive-cli-prompt.ts
```diff
@@ -0,0 +1,115 @@
...
+  async inputText(_args) {
+    return Promise.resolve("sample");
+  },
+  async selectItem<T extends SelectItemValue>(args: {
+    message: string;
+    items: { label: string; value: T }[];
+  }): Promise<T> {
+    return Promise.resolve(args.items[0].value as T);
+  },
+};
```
mrgrain, line 115: move to test directory

src/cli/prompts/operations/select-js-tools.ts
```diff
@@ -0,0 +1,105 @@
...
+
+type Formatter = "prettier" | "biome";
+
+type TestTool = "jest";
+
+export const selectJsTools =
+  (interactiveCliPrompt: InteractiveCliPrompt) =>
+  async (args: {
+    projectTypeName: string;
+    packageName?: string;
```
mrgrain, line 14: I think this is a better name. Also it's not really optional, we always have a value, right?

```suggestion
    defaultPackageName: string;
```

src/cli/prompts/operations/select-js-tools.ts
```diff
@@ -0,0 +1,105 @@
...
+      !projectType.includes("javascript")
+    ) {
+      return undefined;
+    }
+
+    const projectName = await interactiveCliPrompt.inputText({
+      message: "Project name",
+      cancel: "exit",
+      defaultVal: packageName ?? "my-project",
+      placeholder: packageName ?? "my-project",
```
mrgrain, line 39: Is that the fallback we used before? it thought it was the directory name. But also see my other comment: I think we can make `packageName` a required option.
```suggestion
      defaultVal: packageName,
      placeholder: packageName,
```

src/cli/prompts/core/interactive-cli-prompt.ts
```diff
@@ -0,0 +1,115 @@
...
+        type: "text",
+        placeholder: args.placeholder,
+        default: args.defaultVal,
+        cancel: args.cancel === "exit" ? "reject" : args.cancel, // specify how to handle a cancelled prompt (e.g. when user presses `Ctrl+C`)
+      })) as string;
+    } catch {
+      switch (args.cancel) {
+        case "exit": {
+          // when user cancels the prompt, exit the process with code 1
+          process.exit(1);
```
mrgrain, line 53: Let's print a short message here.

mrgrain: You might want to refactor this into a function so it can be reused in the various places.

src/cli/prompts/core/interactive-cli-prompt.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Specify how to handle a cancelled prompt (e.g. by pressing Ctrl+C).
+ *
+ * - `"exit"` - Exit the process with code 1.
+ */
+type CancelOption = "exit";
+
+export interface InteractiveCliPrompt {
+  inputText: (args: {
+    message: string;
+    cancel: CancelOption;
```
mrgrain, line 21: Do we really need this? Looks like it's only different in tests. If so, let's at least default it to the desired behavior.

src/cli/prompts/core/interactive-cli-prompt.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export const interactiveCliPrompt: InteractiveCliPrompt = {
+  async inputText(args): Promise<string> {
+    try {
+      return (await consola.prompt(args.message, {
+        type: "text",
+        placeholder: args.placeholder,
+        default: args.defaultVal,
+        cancel: args.cancel === "exit" ? "reject" : args.cancel, // specify how to handle a cancelled prompt (e.g. when user presses `Ctrl+C`)
```
mrgrain, line 47: If `exit` then `reject`? I'm confused.

mrgrain (changes requested):

PR #4337, opened 2025-08-26
feat(gitattributes): existing attributes can be removed

### LINKED ISSUE #4335: Allow users to choose what files they annotate as generated
Currently, all the Projen generated files are annotated as generated in the `.gitattributes` file with `linguist-generated` tag. This is OK for most files but having a `*.snap` pattern here is problematic for our use case since it effectively renders Jest/Vitest snapshot files hidden by default in GitHub pull requests. Snapshot test diffs are generally important information when reviewing changes and should not be hidden.

However, there seems to be no simple way for modifying these annotations without resorting to dirty hacks such as `fs.writeFileSync()` during the post-synthesis step.

There is already an API method `annotateGenerated` ([link](https://projen.io/docs/api/projen?_highlight=annotategenerated#annotategenerated-)) in the base project, but as far as I know, it can only add annotations. I propose another method that allows removing the said annotations or otherwise modifying them.

Feel free to discuss and suggest possible better solutions.

nikoheikkila:
Fixes #4335

Users can now remove existing attribute mappings as discussed in the above issue.

---


src/gitattributes.ts
```diff
@@ -89,6 +89,20 @@ export class GitAttributesFile extends FileBase {
...
+  /**
+   * Removes an existing mapping from the file.
+   * @param key file pattern to remove.
+   */
+  public removeAttributes(key: string): void {
+    if (!this.attributes.has(key)) {
+      throw new Error(
+        `attribute mapping '${key}' does not exist in .gitattributes`
+      );
+    }
```
mrgrain: No need to fail here.

src/gitattributes.ts
```diff
@@ -89,6 +89,20 @@ export class GitAttributesFile extends FileBase {
     }
   }
 
+  /**
+   * Removes an existing mapping from the file.
+   * @param key file pattern to remove.
+   */
+  public removeAttributes(key: string): void {
```
mrgrain: Similar to `addAttributes`, this should take a `glob` and a list of `attributes` to remove.

I think we can say that if no attributes are provided, all attributes will be removed.

src/gitattributes.ts
nikoheikkila: We can remove it. However, I was thinking if this would prevent dead code in situations where the user doesn't have a particular mapping anymore but has forgotten to remove the "removal" making it noisier.

src/gitattributes.ts
nikoheikkila: Made it so and refactored some tests while at it. 👍🏻

src/gitattributes.ts
mrgrain: Yeah, I think if anything we could add a warning here. But really it's just not something we do elsewhere in projen, so let's just remove it please.

nikoheikkila: Sure thing, removed the check.

src/gitattributes.ts
```diff
@@ -129,4 +143,28 @@ export class GitAttributesFile extends FileBase {
       ),
     ].join("\n");
   }
+
+  private removeSelectedAttributes(glob: string, attributes: string[]): void {
+    const mapping = this.getMapping(glob);
+
+    if (mapping.size === attributes.length) {
+      this.removeMapping(glob);
+    }
```
mrgrain: I don't think that's right.

`mapping` can be `[a, b]` and `attributes` can be `[c, d]` which should result in mapping being unchanged.

mrgrain (changes requested):
Also build is failing

src/gitattributes.ts
nikoheikkila: Thanks! The said test case was indeed missing, so I added it and fixed the logic accordingly.

nikoheikkila: I fixed the build as well. It didn't pass locally since I don't have Maven installed on this computer, but the generated documentation should hopefully mend the mutation check.

nikoheikkila:
@mrgrain, Thanks for your approval. I see you rewrote parts of the pull request, which apparently resulted in a broken build due to lower code coverage. I'll let you finish the work and merge it when the time is right.

mrgrain:
Thanks @nikoheikkila yes I'll take it over the line. 

Pretty sure the coverage wasn't there before either, but the coverage check was kindof broken. Great to see it working again now!

PR #4342, opened 2025-08-30
feat: ignore action version updated with Renovate

### LINKED ISSUE #4331: Actions in projen-generated GitHub Actions workflows should be ignored from renovate config
Components such as `GitHubWorkflow` produce workflow configurations for GitHub Actions. The versions of the used actions are rendered to the workflow code, so projen alone determines which version of the action to use. e.g. `actions/checkout@v4`. However, if renovate is active in the repository, and newer versions are found, renovate tries to suggest updates to these workflows. And hence, the projen generated output differs from the renovate crafted change, and the change cannot be applied. 

Therefore, actions that are used in projen-managed GitHub Actions workflows should be added to renovate configuration as ignored dependencies. Renovate should not update these workflows, as they are managed by projen.

Hi-Fi:
If actions would be updated, self mutation would revert changes resulting empty PRs

Fixes #4331

---


src/github/workflows.ts
```diff
@@ -224,6 +224,17 @@ export class GithubWorkflow extends Component {
     };
   }
 
+  // Can't be getJobs due to JSII5000
+  /**
+   * Get all jobs defined in workflow
+   */
+  public fetchJobs(): Record<
```
mrgrain: suggestion (blocking): Can we make this a getter for `jobs` instead?

src/renovatebot.ts
```diff
@@ -119,6 +121,7 @@ export class Renovatebot extends Component {
 
   private readonly marker?: boolean;
 
+  // biome-ignore lint/suspicious/noExplicitAny: Renovate configuration is not typed
```
mrgrain:
```suggestion
```

src/renovatebot.ts
```diff
@@ -151,6 +154,30 @@ export class Renovatebot extends Component {
       ),
     ];
 
+    // Ignore GitHub Action updates
+    const gitHub: GitHub = (this._project as any).github;
+    if (gitHub?.workflowsEnabled) {
+      const actions: string[] = [];
+      for (const workflow of gitHub.workflows) {
+        this._project.logger.debug(`Handling workflow ${workflow.name}`);
```
mrgrain: todo: we don't normally do this kind of logging. pls remove. also the other ones.

src/renovatebot.ts
```diff
@@ -151,6 +154,30 @@ export class Renovatebot extends Component {
...
+            continue;
+          }
+          for (const step of steps) {
+            this._project.logger.debug(
+              `Handling step ${step.name} that uses action ${step.uses}`
+            );
+            const actionName = (step as JobStep).uses?.split("@")[0];
+            actionName && actions.push(actionName);
+          }
+        }
```
mrgrain: suggestion (strong, but non-blocking): I know functional programming is not for everyone, but this is also hard to read through. I'd typically prefer a filter + map style approach. You could even do a `flatMap` on the  workflows to collect all jobs first. Something like this:

```ts
const actions = gitHub.workflows
  .flatMap((workflow) => workflow.jobs.map(job => job.steps))
  .filter(step => (step as JobStep).uses)
  .map(step =>  step.uses?.split("@")[0]);
```

mrgrain (changes requested):
Nice! Left some feedback.

src/github/workflows.ts
Hi-Fi: Have check if JSII allows that. Tried first that way but got just that JSII5000 error.

src/renovatebot.ts
Hi-Fi: Will do. Especially without logging suggestion is simpler.

src/github/workflows.ts
Hi-Fi: Hmm... I think "getter" in this sense means meaning changing the jobs from private to public.

mrgrain: Got any concerns with that? 

Otherwise, rename the private property to `_jobs` and add `get jobs(): ... { ... }`

Hi-Fi: No, just wondered if it's OK in here. But if not, then that underscore version could be tried.

mrgrain (approved):

mrgrain:
Ended up renaming to `_jobs` and adding a public getter. But otherwise this was great, thank you so much!

mrgrain:
@Hi-Fi FYI not yet released yet as I'm updating the release config atm. Will release asap.

Hi-Fi:
> @Hi-Fi FYI not yet released yet as I'm updating the release config atm. Will release asap.

No worries.

I have to still check if https://docs.renovatebot.com/configuration-options/#ignorepaths would be better way to do this, as then custom files would still get updates and PR in general would be simpler.

PR #4343, opened 2025-08-30
feat: support for Jest 30 with Typescript

### LINKED ISSUE #4308: Feat: support Jest 30
As Jest 30 advertises to be faster to run the tests and newer in general. it would be nice to have support for that.

Currently jest version can be set with `jestOptions.jestVersion`, but as `ts-jest` is no longer needed (nor published for Jest 30) running projen would fail. Ts-jest is currently always added at https://github.com/projen/projen/blob/5d8371ea876ba29909645173be6e79b4cc8f0e93/src/typescript/typescript.ts#L717

Maybe even Projen's own tests could be moved to Jest 30 at same time, but still keep support for older versions so that it's not breaking change (depending how big effort the change of version is).

Hi-Fi:
Install ts-jest 29 when Jest's major is higher than 29

Affects only if jest version is explicitly set. Newly initialized project gets both version correctly as latest which works.

Fixes #4308

---

### DIFF
--- a/docs/api/javascript.md
+++ b/docs/api/javascript.md
@@ -7789,6 +7789,8 @@ The version of jest to use.
 
 Note that same version is used as version of `@types/jest` and `ts-jest` (if Typescript in use), so given version should work also for those.
 
+With Jest 30 ts-jest version 29 is used (if Typescript in use)
+
 ---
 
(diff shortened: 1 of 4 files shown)


mrgrain (approved):

Hi-Fi:
@mrgrain  is there some issue with Codecov as it complains mostly about files that were not edited in PRs?

mrgrain:
Looks like it. Thanks for bringing this to my attention! It's not blocking though.

PR #4349, opened 2025-08-31
feat: move to Jest 30

Hi-Fi:
Jest 30 should improve efficiency and shorten test execution, which are helpfull due to amount of tests that Projen has

---


Hi-Fi:
Jest 30 seems to be taking more time than 29, so currently beneficial update.

PR #4374, opened 2025-09-27
feat: add NuGet trusted publishing support

mrgrain:
Add support for NuGet trusted publishing similar to existing PyPI implementation.

## Changes

- **Add trusted publishing options** to  interface
- **Update  method** to support OIDC authentication with proper permissions
- **Add comprehensive test coverage** for trusted publishing workflow generation
- **Update documentation** with setup instructions and migration guide
- **Improve trusted publishing docs** by consolidating redundant "How It Works" sections

## Features

- Secure authentication using OIDC tokens instead of long-lived API keys
- Automatic permission setup with `id-token: write` when trusted publishing is enabled
- Flexible configuration allowing custom secret names for the NuGet username
- Backward compatibility - existing API key-based publishing continues to work

## Usage

```typescript
publisher.publishToNuget({
  trustedPublishing: true,
  nugetUsernameSecret: "NUGET_USERNAME", // optional, defaults to "NUGET_USERNAME"
});
```

Requires configuring trusted publishing on NuGet.org and setting the `NUGET_USERNAME` secret with your NuGet.org username (profile name, not email).

## References

- [NuGet Trusted Publishing Documentation](https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing)
- [publib NuGet Support](https://github.com/cdklabs/publib#nuget)

---

### DIFF
--- a/docs/publishing/trusted-publishing.md
+++ b/docs/publishing/trusted-publishing.md
@@ -11,12 +11,13 @@ Trusted Publishers ensures packages come from specific CI systems or build pipel
 
 ## How Trusted Publishers Work
 
-Trusted Publishers leverage OpenID Connect (OIDC) to establish trust between your CI/CD system and package repositories:
+When publishing with trusted publishing enabled:
 
-1. **Identity Provider**: Your CI system (GitHub Actions) acts as an OIDC identity provider
-2. **Token Exchange**: The CI system generates a short-lived OIDC token containing metadata about the workflow
-3. **Verification**: The package repository verifies the token and checks it against configured trusted publishers
-4. **API Token**: If verification succeeds, the repository issues a temporary API token for publishing
+1. **Your CI system generates an OIDC token** containing metadata about your workflow and repository (this "ambient identity" is automatically provided by the CI environment)
+2. **The publishing job uses this token** to authenticate with the package repository
+3. **The package repository verifies the token** against your configured trusted publisher settings
+4. **A temporary API token is issued** for publishing your package
+5. **No long-lived secrets** are stored in your repository
 
 This eliminates the need to store long-lived secrets in your repository.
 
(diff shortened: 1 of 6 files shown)


docs/publishing/trusted-publishing.md
```diff
@@ -222,3 +239,30 @@ When publishing with trusted publishing enabled:
...
    - Delete old API tokens in PyPI once you are confident with the new setup
+
+#### For NuGet packages
+
+1. **Update your projen configuration**:
+
+   ```typescript
+   publishToNuget: {
+     trustedPublishing: true,
+     // Remove nugetApiKeySecret if present
```
rix0rrr, line 250: Are we going to throw if both are set?

rix0rrr (approved):

docs/publishing/trusted-publishing.md
mrgrain: We don't currently throw for the other package managers, so no we won't throw. However the key is not set when `trustedPublishing: true`.

rix0rrr: I'm not worried about confusing the tool, I'm worried about protecting users from getting a different config than what they're expecting.

Isn't "this combination of settings doesn't make sense, please be unambiguous so we're sure you haven't made a mistake one way or the other" our usual modus operandi?

> We don't currently throw for the other package managers, so no we won't throw.

"The experience is bad elsewhere so it can be bad here" isn't a good reason to do things a certain way though, is it?

mrgrain: > Isn't "this combination of settings doesn't make sense, please be unambiguous so we're sure you haven't made a mistake one way or the other" our usual modus operandi?

Apparently it's not the usual MO for this particular part of the code. 

> "The experience is bad elsewhere so it can be bad here" isn't a good reason to do things a certain way though, is it?

Of course it's not. I don't think it's really that bad in this case. But if you want to turn this into a blocking request, that's also fine.

rix0rrr: I do not want to turn this into anything blocking. Just seemed like a cheap thing we could do to improve the usability a tiny bit.

mrgrain: I will do a follow up PR to add this to all package managers.

PR #4434, opened 2025-11-06
chore: switch projen to npm

mrgrain:
`yarn` classic has long been end-of-support and we are starting to see cracks in its abilities. For example it's the only package manager that doesn't support a cooldown period for installing dependencies.

We are choosing `npm` as new package manager. `npm` is the default behavior of the ecosystem and will be familiar to most contributors. As an open source project it is important to make it easy to contribute. While other package managers might have more features or are faster, at this time `npm` brings everything that this project needs from a package manager.

The PR also enables the - now unlocked - `cooldown` option for dependency upgrades.

---


hoegertn (approved):
Love it

PR #4016, opened 2024-12-24
feat(python): adds uv packagement support

laithalsaadoon:
Fixes #3497  #3875 

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #3497, opened 2024-03-29
New structure for `PythonProject`


garysassano:
A few things I would like to change for `PythonProject`:

- `pyproject.toml` is the new standard configuration file for all projects, completely replacing the old `requirements.txt`.

 - A new `requiresPython` prop that follows [pyproject.toml PyPA spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#requires-python) is introduced, with the following default value:
   - `>=3.9`

- A new `packageManager` prop is introduced, with the following valid values:
  - `pip` (default)
  - `conda`
  - `poetry`

- A new `buildSystem` prop that follows [PEP 517](https://peps.python.org/pep-0517/)  is introduced, with the following valid values:
  - `setuptools` (default when `packageManager` is `pip` or `conda`)
  - `poetry` (default when `packageManager` is `poetry`)

 - A new `formatAndLinting` array prop is introduced, with the following valid values:
   - `black` (default)
   - `flake8` (default)
    - `ruff`
   - `yapf`
   - `autopep8`
   - `pylint`
   - `pycodestyle`

garysassano:
I think `PythonPackagingOptions` should be renamed to `PyprojectTomlOptions` ; also, the `requiresPython` and `buildSystem` props would need to be added there.

garysassano:
Since we already have `PoetryPyprojectOptions` as an extension, renaming `PythonPackagingOptions` to `PythonPyprojectOptions` would create a more consistent naming scheme, improving overall clarity.

Additionally, the `requiresPython` and `buildSystem` props would need to be added to `PythonPyprojectOptions`.

garysassano:
Possible non-issues when moving `pyproject.toml` to `PythonProject`:

- `pyproject.toml` currently doesn't support "non-package projects" (something that [PEP 735](https://peps.python.org/pep-0735/#motivation) aims to address), so all Python projects would need to be created inside a package directory having the same name as the project (this isn't a big deal and it's actually what we already do for `PoetryProject`).
  - Poetry recently decided to directly address this issue in v1.8.0 by introducing a custom `non-package` option which can be set to `true`. See #3403 for a more detailed explanation.

- Conda [doesn't currently support pyproject.toml](https://github.com/conda/conda/issues/12462), but it will eventually be supported (not a real issue since projen doesn't support it either at the moment).

mrgrain:
This probably needs to be an RFC:
- name it RFC in the title]
- solicit feedback from the community

[...]

PR #4440, opened 2025-11-13
feat(python): add support for uv

themattmorris:
This PR picks up on the good work that @theagenticguy started in #4016.

This PR introduces support for the [`uv` python package manager](https://docs.astral.sh/uv/).

It does not fully deliver the scope that is outlined in #3497, as that issue is large in scope and relates to the design of `PythonProject`. This simply introduces `uv` within the existing `PythonProject` design.

---


src/python/uv.ts
```diff
@@ -0,0 +1,1139 @@
...
+  implements IPythonDeps, IPythonEnv, IPythonPackaging
+{
+  public readonly installTask: Task;
+  public readonly installCiTask: Task;
+  public readonly publishTask: Task;
+  public readonly publishTestTask: Task;
+  private readonly pyProject: UvPyproject;
+  private readonly pythonVersion: string;
+
+  constructor(project: Project, options: UvOptions) {
```
mrgrain:
```suggestion
  constructor(scope: IConstruct, options: UvOptions) {
```

src/python/uv.ts
```diff
@@ -0,0 +1,1139 @@
...
+      name: project.name,
+      dependencies: this.synthDependencies(),
+      devDependencies: this.synthDevDependencies(),
+      pythonVersion: this.pythonVersion,
+      ...options.metadata,
+    });
+  }
+
+  /** Formats dependencies in UV style. */
+  private formatDependency(dep: Dependency): string {
```
mrgrain, line 105: let's add some test cases explicitly asserting all code paths (called via `Uv.addDependency`)

src/python/uv.ts
themattmorris: Thanks for the suggestion. So on this one I largely followed what was done for poetry (which is a pattern used by other modules as well):
https://github.com/projen/projen/blob/2c907c5d8cc942b5b9f28688bf9e5606942c7871/src/python/poetry.ts#L57

If we change the signature to be `IConstruct` (though I understand the signature of `Component` uses `IConstruct`), we run into other problems when compiling:

```bash
src/python/uv.ts:784:32 - error TS2339: Property 'addTask' does not exist on type 'IConstruct'.

784     this.installTask = project.addTask("install", {
                                   ~~~~~~~
src/python/uv.ts:789:34 - error TS2339: Property 'addTask' does not exist on type 'IConstruct'.

789     this.installCiTask = project.addTask("install:ci", {
                                     ~~~~~~~
src/python/uv.ts:797:13 - error TS2339: Property 'packageTask' does not exist on type 'IConstruct'.

797     project.packageTask.exec("uv build");
                ~~~~~~~~~~~
src/python/uv.ts:799:36 - error TS2339: Property 'addTask' does not exist on type 'IConstruct'.

799     this.publishTestTask = project.addTask("publish:test", {
                                       ~~~~~~~
src/python/uv.ts:804:32 - error TS2339: Property 'addTask' does not exist on type 'IConstruct'.

804     this.publishTask = project.addTask("publish", {
                                   ~~~~~~~
src/python/uv.ts:822:38 - error TS2345: Argument of type 'IConstruct' is not assignable to parameter of type 'Project'.
  Type 'IConstruct' is missing the following properties from type 'Project': name, gitignore, gitattributes, outdir, and 35 more.

822     this.pyProject = new UvPyproject(project, {
                                         ~~~~~~~
src/python/uv.ts:823:21 - error TS2339: Property 'name' does not exist on type 'IConstruct'.

823       name: project.name,
                        ~~~~
```

Am I missing something here that I should be considering instead?

mrgrain: Yeah, I know you followed the other components. It's just that they are quite old and best practices evolve.


You will need to replace use of `project` with `this.project`. The `Component` class provides this for you. Except for the `super()` call, there you want to pass `scope`.

Sorry I didn't make this clear.

themattmorris: Gotcha, thanks for the detail!

themattmorris: This [has been addressed](https://github.com/themattmorris/projen/blob/2ed66e1d30f4a865bf5646a3ac4a5419c64c64c1/src/python/uv.ts#L778)

src/python/uv.ts
themattmorris: I have added test cases for both `.addDependency` and `.addDevDependency`. The only public methods that do not have test cases are [`setupEnvironment`](https://github.com/themattmorris/projen/blob/2ed66e1d30f4a865bf5646a3ac4a5419c64c64c1/src/python/uv.ts#L902-L925) and [`installDependencies`](https://github.com/themattmorris/projen/blob/2ed66e1d30f4a865bf5646a3ac4a5419c64c64c1/src/python/uv.ts#L927-L936)

src/python/uv.ts
straygar, line 1: [drive-by comment] I really hope this wasn't written manually - seems fairly difficult to maintain!

If it was auto-generated, could we add a comment about that?
If not - I would suggest using the official [uv json schema](https://www.schemastore.org/uv.json) as the source of truth. That'd make keeping this up-to-date much easier.

mrgrain: That's a fair point. Let me see what we can do about this.

themattmorris: Thanks for the comment! Is there a precedent set for including codegen anywhere for interfaces? Totally agree that we don't want to maintain manual updates when a json schema exists.

When using quicktype to generate these, are you just looking for a comment somewhere in the code that this is what is run to generate?

mrgrain: Yes there is. I took the liberty to generate them for you.

Unfortunately the rest of the code doesn't quite fit the same structure and I'll have to leave the necessary adjustments to you. You can't really change the generated interface, you can only "add" to them by extending them.

To finally write the data to disk, you should use the generated `toJson_*` helpers. For example: https://github.com/projen/projen/pull/4440/files#diff-7f2f5b41b0efdaf06ff88f489c019d93cafa7c76843a1c1e71480a714b3dd649R756
You only need to call them for the top-level, the nesting happens automatically for you.

For `PyProjectToml` I removed the other tools and converted `uv` to an `any`. That's because the tool we are using cannot deal with external references. See: https://github.com/projen/projen/pull/4440/files#diff-e45553937eca65b53352d027639735e785d3c081a9a8a6aeeec4efaf3bc36c6aR326
Given that `PyProjectToml` is a private type for now, that should be fine.

This pseudo code should give you the correct jsonification:

```ts
function toJson(pyProjectToml: PyProjectToml, uvConfig: UvConfiguration) {
  pyProjectToml.tools.uv = toJson_UvConfiguration(uvConfig);
  return toJson_PyProjectToml(pyProjectToml);
}
```

straygar: Thanks @mrgrain !

mrgrain: FYI: This doesn't auto-update the schema files yet. We can do that separately. Or manually.

themattmorris: This is awesome! Thank you. And sorry for mentioning quicktype when projen has `JsiiFromJsonSchema` to do this very thing.

So I went ahead and updated everything to work with `UvConfiguration`. One thing I will mention though regarding a few of these properties is that it seemed to pick one of the allowed types from the json schema.

For example, [this line](https://github.com/themattmorris/projen/blob/84339a53f789c6b8ae7a2a33e87b506dca6422e2/src/python/uv-config.ts#L712) has property `sources` as `{ [key: string]: any[] }`, but according to the [json schema](https://github.com/SchemaStore/schemastore/blob/c4691f3e8bfa6a9bd17d5268812c14a43a981d16/src/schemas/json/uv.json#L1870-L1882) it should be something like this (or at the very least if using `any` could be `{ [key: string]: any | any[] }` to not require an array):

```ts
interface BaseSource {
  extra?: string;
  group?: string;
  marker?: string;
}

/** A remote Git repository, available over HTTPS or SSH */
interface GitSource extends BaseSource {
  git: string;
  branch?: string;
  tag?: string;
  rev?: string;
  subdirectory?: string;
}

/** A remote http(s) URL to a wheel or source distribution */
interface UrlSource extends BaseSource {
  url: string;
  subdirectory?: string;
}

/** A dependency from a local path */
interface PathSource extends BaseSource {
  path: string;
  editable?: boolean;
  package?: boolean;
}

/** A dependency pinned to a specific index */
interface IndexSource extends BaseSource {
  index: string;
}

/** A dependency on another package in the workspace */
interface WorkspaceSource extends BaseSource {
  workspace: boolean;
  editable?: boolean;
}

/** Union type for any kind of source */
type Source = GitSource | UrlSource | PathSource | IndexSource | WorkspaceSource;
type UvSource = Source | Source[];
```

There are others like this. Not a huge deal because the codegen is a huge improvement, but just wanted to point it out.

mrgrain: > One thing I will mention though regarding a few of these properties is that it seemed to pick one of the allowed types from the json schema.
> There are others like this. Not a huge deal because the codegen is a huge improvement, but just wanted to point it out.

Ha, cool that you noticed! This is actually intentional. In jsii (which projen uses) type unions are not well supported. They just turn into an `any` equivalent in same languages - not a great experience. For this particular case, when `Thing` or `Array<Thing>` is allowed. `Array<Thing>` is strictly more powerful than `Thing` but also includes `Thing` very easily. In JavaScript and Python it's as simple as `[....]` around the value. So that's the trade-off. Make it slightly less convenient for some users, but a lot more convenient for users of languages where this union turns into `any`.

I do feel that it should have been able to create a proper type for `Source`, but that's a different story. Maybe a later improvement.

mrgrain:
Hey @themattmorris and @straygar - I ended up making some more changes. Basically in Matt's last version there was still too much duplication for my liking. But not sure if I've struck the right balance for this.

Would love a review!

src/python/uv.ts
```diff
@@ -5,177 +5,35 @@ import { IPythonPackaging } from "./python-packaging";
...
-  readonly dynamic?: PyProjectTomlProjectDynamic[];
-
-  readonly buildSystem?: BuildSystem;
-
-  /**
-   * Package metadata
-   */
-  readonly metadata?: UvConfiguration;
-}
+// type DependencyGroup = string | { "include-group": string };
```
straygar: [nit] Leftover comment?

src/python/uv.ts
```diff
@@ -5,177 +5,35 @@ import { IPythonPackaging } from "./python-packaging";
...
-
+export interface UvOptions extends PythonExecutableOptions {
   /**
-   * List of dev dependencies for this project.
-   * Dependencies use the format: `<module>@<semver>`
-   * @default []
+   * The project's basic metadata configuration.
    */
-  readonly devDeps?: string[];
+  readonly project?: PyprojectTomlProject;
```
straygar, line 22: [nit] Is `PyprojectTomlProject` the best name for this? These days I'd just say that a `pyproject.toml` is the expectation of new Python projects. (unless it's just a folder of scripts)

src/python/uv.ts
mrgrain:
```suggestion
```

src/python/uv.ts
mrgrain: I'm open to suggestions. The name is auto-generated here from two parts:
- `PyprojectToml`, the base type name.  This is a name I choose and we can change.
-  `Project` because it is the type for the the `project` property of the file. This we can't easily change.

`PyprojectToml` is also the name of the full interfaces and used as prefix for other subtypes.

straygar: Yeah, that's fair! I'd probably eventually make this the default `PythonProject`, but that's a breaking change.

I guess it's fine as is and maybe in future we can deprecate `PythonProject` and once it's removed rename `PyprojectTomlProject` to `PythonProject`?

mrgrain: This would absolutely be the plan.

PR #4465, opened 2025-11-27
Support for tag based release triggers

### LINKED ISSUE #4048: Support Tag-first releases
This describes a release process where a commit is first tagged (either manually or through an other workflow), and then we run the release workflow of this tag, while enforcing the version based on the tag.

This comes up frequently as a request and has a few advantages:
- It doesn't require the Workflow to have write permissions
- It's is easier to replicate on other CI/CD platforms, since the GH-specialized parts are fewer
- It allows manual releases
- There is a stronger guarantee between the tagged commit and what's released. (The current system has some edge cases where it can happen that the same version for different platforms uses a different commit)

We have a version of this implemented in jsii-compiler: https://github.com/aws/jsii-compiler/blob/main/projenrc/tag-release.ts

---

### Design

This probably consists of two distinct changes:

- **A new `ReleaseTrigger` for tags**
  This will handle the release workflow being triggered by tags and the required changes to it
- **A new component to auto-tag commits**
  This serves as the scheduling component. It can be based on a time schedule or use releasable commits.

To be resolved:

- If auto-tagging is also using schedule and releasable commits, the API might become a bit awkward. Suddenly we have two places that take "a schedule" and a list of "releasable commits". Could this be one place? Is there a better design?

ringods:
Fixes #4048

This PR only provides the required changes for the first part of the linked issue:
* A new `ReleaseTrigger` for tags

Changes:

* Extended `ReleaseTrigger` to accept tags

---

### DIFF
--- a/src/release/release-trigger.ts
+++ b/src/release/release-trigger.ts
@@ -41,6 +41,13 @@ export interface ContinuousReleaseOptions {
   readonly paths?: string[];
 }
 
+export interface TagReleaseOptions {
+  /**
+   * Tag patterns for which pushes should trigger a release
+   */
+  readonly tags?: string[];
+}
+
 interface ReleaseTriggerOptions {
   /**
    * Project-level changelog file path.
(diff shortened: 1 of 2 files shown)


ringods:
@mrgrain my bad for not reading the contribution guide well enough. My PR title is not compliant to the Conventional Commit style, but at this moment, I can't update it anymore:

Screenshot 2026-02-09 at 20 04 49

May I kindly ask you to update the title to:

```
feat(release): support for tag based release triggers
```

PR #4483, opened 2025-12-30
fix(upgrade-dependencies): `UpgradeDependencies` GitHub workflow fails in a subproject

rix0rrr:
`UpgradeDependencies` creates a GitHub workflow to run a task that it itself creates.

Because the GitHub workflow runs in the repository root by default, implicit in that assumption is that the project that the task is created for lives in the repository root as well:

```
(run at the root of the repo:)
$ npx projen upgrade-dependencies
```

If that is not the case, we get a "no such command" error in the workflow, an [example of which can be seen here](https://github.com/aws/aws-cdk-cli/actions/runs/20355620592).

This PR gives projects a conception of a repository-relative directory and executes the task in the correct directory, while leaving the repository setup tasks to run in the repository root.

So:

* `yarn|npm install` gets run in the repository root
* `npx projen upgrade-dependencies` gets run in the subproject directory

Most likely correct for realistic use cases, but I'm willing to grant that it depends on the exact structure of the repository and if the dependencies are managed on a subproject-by-subproject basis this will still be incorrect.

Not sure how to fix that in projen proper because it does not have enough information to make this distinction accurately.

---


src/project.ts
```diff
@@ -384,6 +388,16 @@ export class Project extends Construct {
...
 
+  /**
+   * The project directory relative to the repository project
+   *
+   * Use this in tasks and workflows to find the working directory of
+   * subprojects in a monorepo.
+   */
+  public get repoRelativeDirectory(): string {
+    return path.relative(this.repoRoot, this.outdir) || ".";
+  }
```
mrgrain, line 399: Hmm. Not entirely sure on the semantics here. Would this make more sense as an internal helper method?

src/project.ts
```diff
@@ -384,6 +388,16 @@ export class Project extends Construct {
...
   }
 
+  /**
+   * The project directory relative to the repository project
+   *
+   * Use this in tasks and workflows to find the working directory of
+   * subprojects in a monorepo.
+   */
+  public get repoRelativeDirectory(): string {
+    return path.relative(this.repoRoot, this.outdir) || ".";
```
mrgrain, line 398:
```suggestion
    return path.relative(this.repoRoot, this.outdir) || DEFAULT_OUTDIR;
```

src/project.ts
```diff
@@ -750,20 +768,35 @@ function determineOutdir(parent?: Project, outdirOption?: string) {
...
+ * Guess at the repository root
+ *
+ * If the root project's `outdir` is relative, it will be evaluated relative to
+ * the current working directory, or if we are running tests, we will create
+ * a temporary directory and use that as the root.
+ *
+ * If the root project's `outdir` is absolute, we have no information on the
+ * repository root and we will just assume it is the same as the root project's
+ * absolute outdir.
+ */
```
mrgrain, line 784: having a bit of a hard time understanding how all of this is different to `this.root.outdir`?

src/project.ts
mrgrain: Or maybe this is an external helper class `RepoPath.of(project).repoRoot` etc.

src/project.ts
```diff
@@ -750,20 +768,35 @@ function determineOutdir(parent?: Project, outdirOption?: string) {
...
+/**
+ * Guess at the repository root
+ *
+ * If the root project's `outdir` is relative, it will be evaluated relative to
+ * the current working directory, or if we are running tests, we will create
+ * a temporary directory and use that as the root.
+ *
+ * If the root project's `outdir` is absolute, we have no information on the
+ * repository root and we will just assume it is the same as the root project's
+ * absolute outdir.
```
mrgrain, line 783: `determineOutdir` enforces that `outdir` is a relative path. With that information, can we simplify this?

mrgrain (changes requested):
I'm down that change. Just a few things I'm not entirely following/have questions on.

src/project.ts
rix0rrr: That only works if `DEFAULT_OUTDIR` is guaranteed to always be equal to `.`. And even then still they don't represent the same thing.

- `DEFAULT_OUTDIR`: used as implied value if a given input parameter is missing.
- `|| "."`: the output of `path.relative()` is the empty string if two paths are in the same directory, and I need to to be `'.'`, not the empty string.

It feels like these values are currently coincidentally the same, but they don't represent the same thing so they shouldn't use the same variable.

I can add a single-use

```ts
const CURRENT_DIRECTORY = '.';
```

And reference that once... is that worth it?

src/project.ts
rix0rrr: It enforces that it's an absolute path. We need a relative path... but relative to what?

src/project.ts
rix0rrr: I had that originally, but it's not guaranteed that the root project is also in the root of the repository. You could put the root project in a subdirectory:

```ts
const parent = new Project({ outdir: 'subdir' });

const child = new Project({ parent, outdir: 'childdir' });
```

Which leads to the following tree:

```
(repo)
├── .projenrc.ts
└── subdir
    ├── childdir
    │   └── package.json
    └── package.json
```

If we took the childdir relative to `root.outdir` we would get `'./childdir'`, but we're looking for `'./subdir/childdir'`.

src/project.ts
rix0rrr: You mean, the semantics might still change so we shouldn't expose this yet?

We can make it `@internal` of course, and it would work for the update to `UpgradeDependencies` but other components might also want this information and they couldn't get it.

Putting the API in a different class -- I suppose we could, but `RepoPath` would need to cast away privateness of certain fields using `as any` to get at that information, and it would be equally as public. So seems like a more roundabout way to get to the same end result?

mrgrain: > and it would be equally as public. So seems like a more roundabout way to get to the same end result?

Sure, but right now `Project` doesn't have a concept of what a Repository and thus a repository root is. This PR is introducing the concept and I want to make sure we are doing this the right way. 

I don't really like the current proposal since it's somewhat implying project and repo are the same thing, with the logical to handle subprojects differently. Previously I thought about adding a explicit wrapper around projects that represents a repo (and can be a github, gitlab, bitbucket version). Similar to what `App` is to `Stack` in CDK.

Keeping everything private is one way to get around by concerns. Externalising might be another way, but for sure needs more thoughts.

src/project.ts
mrgrain: Sure. I think `DEFAULT_OUTDIR` operates in very much the same way and could be renamed to `CURRENT_DIRECTORY` as well. They both have the same goal: If none provided, assume the current working dir. Not a blocker.

src/project.ts
mrgrain: I was talking about the input. 

https://github.com/projen/projen/blob/0e916cf7c633b4a842bd5184b22cfbdddf34e614/src/project.ts#L758-L760

The output is indeed enforced to be an absolute path.


This function takes in `options.outdir`, which must always be a relative path because that's enforced later.

---

I like the direction of the refactor, but it's confusing to me because we now have a strange interdependency between the `repositoryRoot` and `determineOutdir` helpers. Maybe you can refactor this even more to make it more readable.

PR #4496, opened 2026-01-09
chore: add stub integration test workflow for pull requests

mrgrain:
In preparation for #4495, we need a stub integration test workflow file so I can iterate on the PR to build and execute the actual test case. Without it, #4495 will not run the new workflow.

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -8,4 +8,4 @@ jobs:
   integ:
     runs-on: ubuntu-latest
     steps:
-      - run: echo "Success
+      - run: echo "Success"


.github/workflows/integ.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  pull_request: {}
+  workflow_dispatch: {}
+  merge_group: {}
+permissions: {}
+jobs:
+  integ:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Success
```
mrgrain:
```suggestion
      - run: echo "Success"
```

PR #4517, opened 2026-01-22
chore: fix inefficient regular expression to validate emails

mrgrain:
Resolves [https://github.com/projen/projen/security/code-scanning/3](https://github.com/projen/projen/security/code-scanning/3)

The currently used regular expressions to validate that a given string is an email address, is inefficient. For email validation specifically, it’s almost always better to use a pragmatic regex that covers typical addresses and avoids backtracking traps, rather than a fully RFC‑compliant monster expression.

So for projen, the cleanest and safest fix is to replace the current `EMAIL_REGEX` definition with a much simpler, widely used email regex that is efficient and does not contain problematic nested alternations inside a `*` or `+`. It still tests whether a string looks like an email, but with a slightly less strict definition. No other code changes are needed because nothing in the shown snippet depends on internal capture groups or on the precise structure of the pattern.

The pattern now effectively checks that a string looks like `<something>@<something>` and `something` cannot be whitespace or the `@` character. For projen, this is enough.

---


src/cdk/jsii-project.ts
```diff
@@ -22,7 +22,7 @@ import { TypeScriptProject, TypeScriptProjectOptions } from "../typescript";
 import { deepMerge } from "../util";
 
 const EMAIL_REGEX =
-  /(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/;
+  /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
mrgrain:
```suggestion
  /^[^\s@]+@[^\s@]+$/;
```

test/jsii/__snapshots__/jsii.test.ts.snap
mrgrain, line 1: This is correct. from the looks of it, these strings are more likely to be emails than URLs.

PR #4524, opened 2026-01-28
fix(python): fix uv venv exec

### LINKED ISSUE #4487: python: setupEnvironment() with uv breaks with a Python version range
### Steps to reproduce

```sh
mkdir fooapp && cd fooapp
npx projen new python --uv
```

### Expected Behavior

The project is created and synthesized correctly, with python and the project dependencies correctly installed via `uv`.

### Actual Behavior

```
👾 Project definition file was created at /Users/straygar/workspace/fooapp/.projenrc.py
node:child_process:966
    throw err;
    ^

Error: Command failed: uv venv --python >=3.12,<4.0 .venv
/bin/sh: 4.0: No such file or directory
```

### Notes

I suspect the fix is just to [this line](https://github.com/projen/projen/blob/bd5b283a7bc662ba1e931991925d9af30f47bd92/src/python/uv.ts#L188)
- before: `uv venv --python ${this.venvPython} .venv`
- after: `uv venv --python "${this.venvPython}" .venv`

### Environment

OS: macOS 14.4.1 (Sonoma)
Shell: `oh-my-zsh`

### LINKED ISSUE #4523: bug(uv): missing quotes for requiresPython and pythonExec not forwarded for `exec uv venv`
## Context

When using `Uv` in a `PythonProject`, the venv creation command uses `options.project.requiresPython` as the `--python` argument if `pythonExec` is not set. `requiresPython` is often a constraint string (e.g. `>=3.12,<4.0`) which is valid for `uv`, but only when passed as a single quoted argument.
Currently the command is constructed without quotes, so the shell splits the argument at the comma and uv receives an invalid value.

Additionally, `pythonExec` is never forwarded from `PythonProject` to `Uv`, so users cannot override the python selector even when they set `pythonExec`.

Relevant code:
- Missing quotes in uv command: https://github.com/projen/projen/blob/f681a520b33dc0147548fc92342a96dab3a21b73/src/python/uv.ts#L188
- `pythonExec` not forwarded: https://github.com/projen/projen/blob/f681a520b33dc0147548fc92342a96dab3a21b73/src/python/python-project.ts#L348-L365

## Obtained Result

With `uv` enabled and `pythonExec` unset, projen runs:

```
uv venv --python >=3.12,<4.0 .venv
```

This fails because the argument is split by the shell.

## Expected Result

- The constraint should be passed as a single argument, e.g.:
  ```
  uv venv --python ">=3.12,<4.0" .venv
  ```
- If `pythonExec` is provided, it should be used by `Uv`.

## Suggested Fix

1. Quote the `--python` argument in the `uv venv` command.
2. Forward `pythonExec` from `PythonProject` into `Uv`.

v41e:
Fixes #4523
Fixes #4487

---


mrgrain (approved):

mrgrain:
Awesome, thank you!

PR #4564, opened 2026-03-09
feat(awscdk): create `SingletonFunction`s from files

### LINKED ISSUE #4563: Feature request: Support SingletonFunction lambda
Would be good to have support for SingletonFunction in the same way normal and edge functions are supported

Especially good for libraires providing new functionality to CDK with custom recourses

McDoit:
Fixes #4563

---

### DIFF
--- a/docs/api/cdk.md
+++ b/docs/api/cdk.md
@@ -7378,7 +7378,7 @@ GitHub secret name which contains the GPG private key or file that includes it.
 
 This is used to sign your Maven packages. See instructions.
 
-> [https://github.com/aws/publib#maven](https://github.com/aws/publib#maven)
+> [https://github.com/cdklabs/publib#maven](https://github.com/cdklabs/publib#maven)
 
 ---
 
@@ -7396,7 +7396,7 @@ GitHub secret name which contains the GPG private key or file that includes it.
 This is used to sign your Maven
 packages. See instructions.
 
-> [https://github.com/aws/publib#maven](https://github.com/aws/publib#maven)
+> [https://github.com/cdklabs/publib#maven](https://github.com/cdklabs/publib#maven)
 
 ---
 
--- a/docs/api/release.md
+++ b/docs/api/release.md
@@ -6,7 +6,7 @@
 
 Implements GitHub jobs for publishing modules to package managers.
 
-Under the hood, it uses https://github.com/aws/publib
+Under the hood, it uses https://github.com/cdklabs/publib
 
 #### Initializers <a name="Initializers" id="projen.release.Publisher.Initializer"></a>
 
@@ -1880,7 +1880,7 @@ GitHub secret name which contains the GPG private key or file that includes it.
 
 This is used to sign your Maven packages. See instructions.
 
-> [https://github.com/aws/publib#maven](https://github.com/aws/publib#maven)
+> [https://github.com/cdklabs/publib#maven](https://github.com/cdklabs/publib#maven)
 
 ---
 
@@ -1900,7 +1900,7 @@ GitHub secret name which contains the GPG private key or file that includes it.
 This is used to sign your Maven
 packages. See instructions.
 
-> [https://github.com/aws/publib#maven](https://github.com/aws/publib#maven)
+> [https://github.com/cdklabs/publib#maven](https://github.com/cdklabs/publib#maven)
 
 ---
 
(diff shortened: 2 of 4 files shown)


src/awscdk/internal.ts
```diff
@@ -46,3 +52,22 @@ export const TYPESCRIPT_LAMBDA_EXTENSION_EXT = ".lambda-extension.ts";
...
+  const hash = createHash("sha256").update(input).digest("hex").slice(0, 32);
+  const chars = hash.split("");
+
+  chars[12] = "5";
+  const variantNibble = parseInt(chars[16], 16);
+  chars[16] = ["8", "9", "a", "b"][variantNibble % 4];
+
+  const normalized = chars.join("");
+  return `${normalized.slice(0, 8)}-${normalized.slice(8, 12)}-${normalized.slice(12, 16)}-${normalized.slice(16, 20)}-${normalized.slice(20, 32)}`;
+}
```
mrgrain, line 66: can you explain what this magic is and/or where you have copied it from?
code comments will be very much appreciated here to explain the intent and what this is doing.

McDoit: It is a crude implementation of a v5 uuid, something like this:
https://www.rfc-editor.org/rfc/rfc9562.html#name-variant-field
Was a bit hesitant to add a dependency on uuid library https://www.npmjs.com/package/uuid but could def improve the code with that instead

test/awscdk/lambda-function.test.ts
mrgrain, line 1: These new tests should also match on the snapshot of the generated files, in addition to functional assertions. Snapshots help reviewers understand context and are a good way of highlighting accidental (future) changes.

src/awscdk/internal.ts
mrgrain: Make sense. While the `SingletonFunction` field is called `uuid`, it is not currently used as such. 

https://github.com/aws/aws-cdk/blob/5cc30e00e9d13523d1dedd3606456819161796d6/packages/aws-cdk-lib/aws-lambda/lib/singleton-lambda.ts#L82

I propose we just use a stable hash based on the input (pretty much what you already have here) and call it a day:

```
createHash("sha256").update(input).digest("hex").slice(0, 32);
```

McDoit: Good call! Updated the function now!

test/awscdk/lambda-function.test.ts
McDoit: I haven't really worked on these types of testing with snapshots before, do you have any pointers?

mrgrain: Of course! It's very simple:

```ts
generatedSource = snapshot["src/hello-function.ts"];
expect(generatedSource).toMatchSnapshot();
```

And then you need to run the test cases locally to generate an initial snapshot and commit that as well.

test/awscdk/lambda-function.test.ts
```diff
@@ -278,6 +278,98 @@ test("Edge function", () => {
...
+  });
+
+  new awscdk.LambdaFunction(project, {
+    entrypoint: join("src", "hello.lambda.ts"),
+    cdkDeps: cdkDepsForProject(project),
+    singleton: true,
+  });
+
+  const snapshot = Testing.synth(project);
+  const generatedSource = snapshot["src/hello-function.ts"];
```
mrgrain, line 294:
```suggestion
  const generatedSource = snapshot["src/hello-function.ts"];

  expect(generatedSource).toMatchSnapshot();
```

test/awscdk/lambda-function.test.ts
```diff
@@ -278,6 +278,98 @@ test("Edge function", () => {
...
+
+  new awscdk.LambdaFunction(project, {
+    entrypoint: join("src", "hello.lambda.ts"),
+    cdkDeps: cdkDepsForProject(project),
+    singleton: true,
+    singletonUuid: "34426f5a-65f4-44ff-a380-c730f26b27f5",
+  });
+
+  const snapshot = Testing.synth(project);
+  const generatedSource = snapshot["src/hello-function.ts"];
```
mrgrain, line 343:
```suggestion
  const generatedSource = snapshot["src/hello-function.ts"];

  expect(generatedSource).toMatchSnapshot();
```

PR #4599, opened 2026-03-28
chore: [RFC] Introduce Repository as the Construct Tree Root

hoegertn:
This RFC proposes a new Repository class that sits above Project in the projen construct tree - mirroring the App/Stack relationship in AWS CDK.
Today, Project conflates repository-level concerns (CI/CD workflows, .gitignore, PR templates) with project-level concerns (source code, dependencies, build tasks).
This causes real pain in monorepos, where GitHub Actions workflows are repository singletons but the GitHub component is bolted onto the root project, leaving subprojects to walk the parent chain and cast types just to register a workflow.

---


rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+│   ├── PR templates
+│   ├── Dependabot / Mergify
+│   └── CODEOWNERS
+├── Project (root project)
+│   ├── Tasks
+│   ├── Dependencies
+│   ├── ProjectBuild
+│   ├── VSCode / Gitpod / DevContainer
+│   └── Source files
+└── Project (subproject, in monorepo)
```
mrgrain: nit: Construct tree-wise, I think the subproject would be under the root project.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+### Class Hierarchy
+
+```
+Construct
+└── Repository (new)
+    ├── GitRepository (new) — git-specific config
+    │   ├── GitHubRepository (new) — GitHub platform
+    │   └── GitLabRepository (new) — GitLab platform
+    └── Project (existing, modified)
```
mrgrain: nit: I don't think we are proposing `Project` to extend `Repository`

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+  /**
+   * Find the closest Repository by walking up the construct tree.
+   * Results are cached on the construct for performance.
+   */
+  public static of(construct: IConstruct): Repository { ... }
+
+  /**
+   * All projects within this repository.
+   */
+  public get projects(): Project[] { ... }
```
mrgrain, line 105: question: Including subprojects? Or just direct children?

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+export class GitHubRepository extends GitRepository {
+  /**
+   * The GitHub component managing workflows, PR config, etc.
+   */
+  public readonly github: GitHub;
+
+  constructor(options: GitHubRepositoryOptions) {
+    super(options);
+    this.github = new GitHub(this, options.githubOptions);
```
mrgrain: thought: Probably should be an option to not use GHA at all. Saying that, I guess we might need to start splitting this into "mandatory parts" and a "GitHubActions" component.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+export interface GitLabRepositoryOptions extends GitRepositoryOptions {
+  /**
+   * Options for GitLab integration.
+   * @default - default options
+   */
+  readonly gitlabOptions?: GitLabOptions;
+}
+
+export class GitLabRepository extends GitRepository {
+  public readonly gitlab: GitlabConfiguration;
```
mrgrain: thought (non-blocking): should consider renaming to align with `GitHub`

```suggestion
  public readonly gitLab: Gitlab;
```

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+}
+```
+
+### Repository Lookup
+
+Projects can find their parent Repository using `Repository.of(this)`, which walks up the construct tree. The result is cached as a property for performance.
+
+```typescript
+// From any project or component:
+const repo = Repository.of(this);
```
mrgrain, line 217: suggestion: I like `Repository.of(this)` but we could also add a `repo` property to project.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+if (repo instanceof GitHubRepository) {
+  repo.github.addWorkflow(...);
+}
+```
+
+This pattern mirrors `Stack.of(construct)` in the CDK. The tree traversal is the primary mechanism, with the result cached on first access.
+
+### Subproject Access to Repository
+
+In monorepo setups, subprojects can access the Repository directly to register workflows or interact with CI configuration:
```
mrgrain, line 237: Nice!

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+  workflow.on({ push: { branches: ["main"] } });
+  workflow.addJob("deploy", { ... });
+}
+```
+
+This eliminates the current pattern of checking `this.project.parent` and hoping the root project has a `github` property.
+
+### Backward Compatibility
+
+When a user creates a Project without explicitly creating a Repository, a `GitHubRepository` is auto-created as the implicit parent. This preserves current behavior where `GitHubProject` subclasses automatically get GitHub integration.
```
mrgrain, line 255: Perfect!

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+// Current code (continues to work):
+const project = new TypeScriptProject({
+  name: "my-project",
+  defaultReleaseBranch: "main",
+});
+
+// Equivalent to:
+const repo = new GitHubRepository({ name: "my-project" });
+const project = new TypeScriptProject({
+  parent: repo,
```
mrgrain, line 267: Question: reuse `parent` or introduce new `repo` prop?

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+│   └── postSynthesize() [Project + components]
+├── synthesize() [Repository + components]
+└── postSynthesize() [Repository + components]
+```
+
+Repository-level files (workflows, .gitignore) are synthesized at the Repository level. Project-level files (package.json, tsconfig.json) are synthesized at the Project level. This matches the current behavior but makes the ownership explicit.
+
+### Impact on Existing Classes
+
+**`GitHubProject` — removed in next major version.** The class currently serves as the bridge between Project and GitHub integration. In the new model, this bridge is the Repository. `GitHubProject` would be deprecated first, then removed.
```
mrgrain, line 304: Can probably be removed earlier. But yes, scheduled for deprecation.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+**`Project` — simplified.** Git-related properties (`.gitignore`, `.gitattributes`) move to `GitRepository`. The `parent` property for monorepo setups is replaced by the natural construct tree relationship through the Repository.
```
mrgrain: task: This is might be breaking in itself due to some helpers on `Project`. We can do a pass-through, schedule them for deprecation etc.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+- `GitHubRepository` and `GitLabRepository` subclasses
+- `Repository.of()` lookup mechanism
+- Backward compatibility through auto-creation of default Repository
+- Moving git-related and shared CI config from Project to Repository
+- Subproject access to Repository for workflow registration
+- Deprecation plan for `GitHubProject`
+
+### Explicitly Out of Scope (Future Work)
+
+- **Workflow portability** — Building workflows once and running them on any CI platform is a separate, larger effort. This RFC provides the structural foundation but does not address portable workflow definitions.
```
mrgrain, line 327: Agreed!

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+- **Task system revamp / task dependencies** — Introducing formal task dependency graphs is complementary but separate. The Repository restructuring does not depend on it.
```
mrgrain: Seems unnecessary to mention (there's a lot of things this RFC doesn't cover and I don't think it touches on tasks at all)

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+- **Migration tooling** — Detailed migration guides and codemods will be developed alongside the implementation, not as part of this RFC.
```
mrgrain: Will they though? Probably an AI hallucination.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Open Questions
+
+1. **File ownership during transition.** During the deprecation period, both `Project.gitignore` and `GitRepository.gitignore` could exist. How do we handle the overlap? One option: `Project.gitignore` delegates to `Repository.of(this).gitignore` transparently.
```
mrgrain, line 332: If it's the same object, it doesn't really matter. Otherwise it won't work anyway. Delegation + deprecation is probably the way to go.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+2. **CircleCI integration.** The current `Circleci` component is also repository-level. Should there be a `CircleCiRepository` subclass, or is CircleCI different enough to handle separately?
```
mrgrain: CircleCI is not a repository host. This highlights that there's a difference between a repository host and a CICD provider.

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+3. **Projenrc location.** The `.projenrc.ts` file currently lives at the project root. In the new model, should it be a Repository concern? The projenrc instantiates both the Repository and its Projects, so it logically sits outside both.
```
mrgrain, line 336: Good question.... 🤔 What do you/AI think?

rfcs/repository-tree-root.md
```diff
@@ -0,0 +1,323 @@
...
+
+4. **Multi-project CI divergence.** If a monorepo has projects that need different CI behaviors (e.g., one project needs a specialized release workflow), how do we prevent the Repository-level config from being too rigid?
```
mrgrain: Suggestion: Out-of-scope for this RFC. We are not solving the monorepo challenge here. Our "stake in the ground" is that CICD config at the repo level (like GHA) must be owned by the repo. It's up to the monorepo and a future Workflow portability RFC to deal with anything else.

rfcs/repository-tree-root.md
hoegertn: Very good question. Any opinions?

rfcs/repository-tree-root.md
hoegertn: What happens if I do new GitHub but never add a workflow? Does this hurt?

rfcs/repository-tree-root.md
hoegertn: Need to check the current name. afair it is called GitlabConfiguration but will check.

rfcs/repository-tree-root.md
hoegertn: I think of parent as similar to scope in CDK. What happens if I provide a parent (for a sub project) and a repo? With only parent it is clear.

rfcs/repository-tree-root.md
hoegertn: Sure, sounds great.

rfcs/repository-tree-root.md
hoegertn: Sure, previous version had some references

rfcs/repository-tree-root.md
hoegertn: It might help users but I can remove it.

rfcs/repository-tree-root.md
hoegertn: Good catch.

rfcs/repository-tree-root.md
hoegertn: I am wondering what happens if the repo root is not a TS/Py/.. project.

rfcs/repository-tree-root.md
mrgrain: yeah, unrelated to RFC. sorry.

rfcs/repository-tree-root.md
mrgrain: Makes sense, but will limit our type options. With two props we will need to validate.

rfcs/repository-tree-root.md
mrgrain: No that's fine. But we need a signal to other components that just because we are in a `GitHubRepository` doesn't mean we should add GHA workflows.

Again, this is the difference between a repo hoster and a CICD provider.

rfcs/repository-tree-root.md
mrgrain: Keeping it at project level is certainly going to be much easier. Let's stick with that for now.

rfcs/repository-tree-root.md
hoegertn: So to begin with, it could be a Boolean on the githubOptions?

rfcs/repository-tree-root.md
hoegertn: So you propose adding a `repo` and making sure that the `parent.repo` is the same as `repo` if both are provided, correct?

rfcs/repository-tree-root.md
hoegertn: What would that mean? Currently, the `projenrc` is always on the top-most folder, correct?

rfcs/repository-tree-root.md
mrgrain: Probably not. I think eventually it will be something like

```ts
interface GitHubRepositoryOptions {
  githubActions: boolean; // default true
  githubActionsOptions: GitHubActionsOptions;
}
```

rfcs/repository-tree-root.md
mrgrain: I think someone will need to play around with it and see what the better choice is. Our options are:

- use `parent`, find a way to make the types work, possible runtime check allowed values
- introduce `repo`, enforce that only one of `repo` and `parent` can be set

rfcs/repository-tree-root.md
mrgrain: Pretty sure it's in the `outdir` of the root project. Basically subprojects are just skipping adding the `projenrc` component.

rfcs/repository-tree-root.md
hoegertn: Changed it

hoegertn:
Can we merge it? Or do we need to resolve more?

mrgrain (approved):

PR #4606, opened 2026-04-01
feat(github): update GitHub Actions to latest major versions (node24)

ahammond:
Partially addresses #4604

GitHub is deprecating Node.js 20 in GitHub Actions runners starting September 2026. This updates all GitHub Actions emitted by projen to their latest major versions, moving them off Node.js 20 to Node.js 24.

The `actions/setup-node` upgrade to v6 introduces automatic package manager caching by default, which can cause failures in workflows that don't have a lockfile checked in. To prevent this, `package-manager-cache: false` is set explicitly on all `setup-node` steps.

Two actions remain on Node.js 20 because no upstream release with Node.js 22+ support exists yet: `hmarr/auto-approve-action` (v4) and `imjohnbo/issue-bot` (v3).

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #4607, opened 2026-04-01
fix(typescript): pin ts-jest to ^29 when jest version is unspecified

ahammond:

## Summary

- When `jestVersion` is `""` (the default, meaning "latest"), `semver.coerce("")` returns `null`, so `jestMajor` fell back to `0`. The condition `0 > 29` was false, so `tsJestVersion = ""` (latest ts-jest), bypassing the intended `@^29` pin.
- Fix: default `jestMajor` to `30` instead of `0`, so unversioned jest still pins `ts-jest@^29`.
- This prevents breakage from `ts-jest@30` where `injectGlobals` defaults to `false`, causing `TS2593: Cannot find name 'test'` errors.

Fixes the `integ-node` CI failure reported in #4606.

## Root cause

```typescript
// Before — null coercion fell back to 0, bypassing the pin
const jestMajor = semver.coerce(jest.jestVersion)?.major ?? 0;

// After — null coercion falls back to 30, triggering the ^29 pin
const jestMajor = semver.coerce(jest.jestVersion)?.major ?? 30;
```

## Test plan

- [x] `node ./projen.js test test/typescript/typescript.test.ts` — all 37 tests pass, snapshots updated (`ts-jest: "*"` → `ts-jest: "^29"`, removed from upgrade filter)
- [x] `node ./projen.js test test/web/` — all 13 tests pass
- [x] `node ./projen.js eslint` — no linting errors
- [x] Confirmed `node-project.test.ts` failures are pre-existing (same on `main`, unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mrgrain:
integ test is fixed, not needed I think

PR #4613, opened 2026-04-02
feat(typescript): configurable runner for executing .projenrc.ts (tsx, ts-node, node)

### LINKED ISSUE #4585: Migrate away from ts-node
## Problem

ts-node has not published a release since 2023-12-08. A major issue currently is not being able to extend from multiple bases in a  `tsconfig.json` file, as ts-node does not support the array form of `extends`: https://github.com/TypeStrong/ts-node/issues/2000
(Note that the issue is closed, but the fix has not been released yet, aside from a beta version on npm).

In general, ts-node does seem unmaintained.

## Proposed solutions

The most lightweight solution perhaps would be to run the .projenrc.ts files with just Node.js's built-in TypeScript support. A recent enough version is required, and not all TypeScript features are fully supported.

For a near drop-in replacement of ts-node, there are many options:

- https://github.com/poppinss/ts-exec: very small, runs TypeScript exactly as Node.js does, but with full TypeScript language support, not very popular however
- https://github.com/privatenumber/tsx: very popular, but deviates from Node.js import resolution (more permissive)
- https://github.com/swc-project/swc-node: another option, faster than tsx, also used by ts-exec
- possibly more?

mrgrain:
Choose how projen runs your TypeScript `.projenrc.ts` via a new `runner` option
backed by `TypeScriptRunner`:

```ts
new TypeScriptProject({
  // ...
  runner: TypeScriptRunner.tsx(),                  // fast, no type-checking
  // runner: TypeScriptRunner.tsNode({ swc: true }),
  // runner: TypeScriptRunner.nodejs(),             // Node's native TS (Node >= 22.18)
});
```

- Pick the runtime: `tsx()`, `tsNode()`, or `nodejs()`.
- `{ typeCheck: true }` runs `tsc --noEmit` before executing.
- `{ tsconfig: "..." }` runs against a specific tsconfig.
- `runner.copy({ ... })` derives a variant with tweaked options.

Set it project-wide on `TypeScriptProject`, or per file via
`projenrcTsOptions.runner`. A runner you pass is used as-is; when you don't pass
one, projen supplies the projenrc tsconfig and enables type-checking for you.

Defaults are unchanged — projenrc still runs on **ts-node**. Run steps are now
emitted as shell-free `execArgs`, so re-running projen will refresh the
`.projenrc.ts` step in `.projen/tasks.json` (no behavior change).

## Deprecations

- `projenrcTsOptions.tsconfigFileName` → configure the tsconfig on the `runner`.
- `Projenrc`'s `swc` → `runner: TypeScriptRunner.tsNode({ swc: true })`.

---


src/typescript/typescript-runner.ts
```diff
@@ -0,0 +1,135 @@
...
+export interface RunCommandOptions {
+  /**
+   * The tsconfig file path.
+   */
+  readonly tsconfigFile: string;
+
+  /**
+   * The projenrc file path.
+   */
+  readonly rcFile: string;
```
hoegertn: Can we make this a bit more generic. It can be any script not only projenrc

src/typescript/projenrc.ts
```diff
@@ -23,52 +24,57 @@ export interface ProjenrcOptions {
...
+   * @deprecated Use `runner: TypeScriptRunner.tsNode({ swc: true })` instead.
    */
   readonly swc?: boolean;
+
+  /**
+   * The runner to use for executing TypeScript files.
+   *
+   * @default TypeScriptRunner.tsNode()
+   */
+  readonly runner?: TypeScriptRunner;
```
hoegertn, line 38: Can we also make the runner discoverable by other parts of the system to generate all other scripts using the same runner?

src/typescript/typescript-runner.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  /**
+   * Compiler options passed as CLI flags to `tsc` when typechecking.
+   * These are merged with (and override) the built-in defaults.
+   *
+   * Only used when `typeCheck` is true.
+   *
+   * @default - sensible defaults: strict, target ES2022, module nodenext, skipLibCheck
+   */
+  readonly compilerOptions?: TypeScriptCompilerOptions;
```
mrgrain: I think I will revert this in favor of typeChecking using a tsconfig file. With `tsgo` coming up, any speed advantages will eventually be marginal.

src/typescript/typescript-runner.ts
```diff
@@ -0,0 +1,224 @@
...
+
+/**
+ * Options for the ts-node runner.
+ */
+export interface TsNodeRunnerOptions {
+  /**
+   * Whether to use SWC for transpilation.
+   * @default false
+   */
+  readonly swc?: boolean;
```
jessestricker: If I read https://github.com/TypeStrong/ts-node#swc correctly, enabling this will also enable the transpileOnly flag of ts-node and thus effectively disable typechecking.

Is this intended? If so, I feell it should be documented.

src/typescript/typescript-runner.ts
```diff
@@ -0,0 +1,224 @@
...
+   *
+   * @default - discovered from the project
+   */
+  readonly tsconfig?: string;
+}
+
+/**
+ * Options for the native Node.js TypeScript runner.
+ */
+export interface NodeRunnerOptions {
```
jessestricker, line 64: I think it would be nice if the Node.js runner also supported type-checking, the way the tsx runner does.

src/typescript/typescript-runner.ts
mrgrain: Good callout. The option mainly exists to maintain backwards compatibility.

rix0rrr (approved):

PR #4651, opened 2026-04-10
feat(javascript): improve install logging with structured reasons and file diffs

mrgrain:
When projen's default task automatically runs `install`, it can be unclear why. This is especially confusing for new users who don't expect an install step during synthesis.

The install message now hints that more information is available with `--debug`. When running with `--debug`, the specific reason for the install is logged using a structured `InstallReason` enum (`NO_NODE_MODULES`, `PACKAGE_JSON_CHANGED`, or `DEPS_RESOLVED`), each carrying only the data relevant to that trigger.

For `package.json` changes, a unified diff is shown in git-diff style with `@@ ... @@` hunk headers, context lines, and character-level inline highlighting of what exactly changed. This uses a new `diff()` method on `FileBase` that is available for any synthesized file, not just `package.json`.

Resolved dependency versions (e.g. `*` being pinned to `^2.1.3`) are now always printed at info level since this is important information that was previously only visible at verbose level.

---


src/util/diff.ts
mrgrain, line 1: this file is fully vibe coded

src/file.ts
```diff
@@ -177,7 +182,7 @@ export abstract class FileBase extends Component {
       prevReadonly === this.readonly &&
       successfulExecutableAssertion
     ) {
-      this.project.logger.debug(`no change in ${filePath}`);
+      this.project.logger.verbose(`no change in ${filePath}`);
```
mrgrain, line 185: changed on purpose, this was quite noisy and made `--debug` useless.

rix0rrr (approved):

Issue #4654, opened 2026-04-13
`ca:login` step fails in workflow files


AminFazlMondo:
As part of the changes in (https://github.com/projen/projen/pull/4628), with `pnpm` as the package manager, the commands are expecting projen to already be installed when doing the CodeArtifact login.
If the package is using private packages, the install cannot happen prior to the `ca:login` step. Even though, the login step does not require projen to perform aws cli call, it needs it to be able to route to the tasks defined by projen.
One of the possible options for this is to update the script for CodeArtifact login to be `pnpm dlx projen ca:login`

```
sh: 1: projen: not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
Error: Process completed with exit code 1.
```

mrgrain:
This was a deliberate change to move away from unexpected downloads. I can see how this is an issue for the ca:login task. :thinking_face:  

The workflow steps could use a dynamic install `pnpm dlx` or just call the commands directly.

mrgrain:
Resolved by https://github.com/projen/projen/pull/4655

PR #4655, opened 2026-04-14
fix(node): CodeArtifact login step fails in GHA workflows

AminFazlMondo:
Fixes the AWS CodeArtifact login command (#4654 ) to ensure it executes correctly before package installation by updating the command to use the appropriate package manager. 

---

### DIFF
--- a/src/javascript/node-project.ts
+++ b/src/javascript/node-project.ts
@@ -1143,6 +1143,23 @@ export class NodeProject extends GitHubProject {
       authProvider: codeArtifactOptions?.authProvider,
     };
 
+    function executeCommand(packageManager: NodePackageManager): string {
+      switch (packageManager) {
+        case NodePackageManager.NPM:
+        case NodePackageManager.YARN:
+        case NodePackageManager.YARN_CLASSIC:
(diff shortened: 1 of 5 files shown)


src/javascript/node-project.ts
```diff
@@ -1143,6 +1143,23 @@ export class NodeProject extends GitHubProject {
       authProvider: codeArtifactOptions?.authProvider,
     };
 
+    function executeCommand(packageManager: NodePackageManager): string {
```
mrgrain: seems to be the same as here

https://github.com/projen/projen/blob/05bfc4475fb891abae8433dd21184c2b1c7ceaba/src/javascript/upgrade-dependencies.ts#L366

please extract into a shared helper in [javascript/utils.ts](https://github.com/projen/projen/blob/05bfc4475fb891abae8433dd21184c2b1c7ceaba/src/javascript/util.ts)

mrgrain (changes requested):
thanks, small change and we are good to go. You will also need test coverage.

mrgrain (approved):

PR #4669, opened 2026-04-23
feat(github): add addPaths and deleteBranch to CreatePullRequestOptions

### LINKED ISSUE #4660: [WorkflowActions] Expand createPullRequest options available to match Action Params
Hi All,

We periodically use the `WorkflowActions.createPullRequest` action (or an equivalent) in some of our scheduled GH Workflow tasks.
Not needing to hardcode the current action version is pretty convenient for maintenance for us.

Unfortunately, some of our actions use input parameters that are available on the [create-pull-request action](https://github.com/peter-evans/create-pull-request/blob/main/action.yml), but aren't available in the [projen action's params](https://projen.io/docs/api/github#projen.github.CreatePullRequestOptions), like `add-paths` and `delete-branch`.

It would be appreciated if this could be expanded to match the input variables.

Thanks <3

tmchow:
Closes #4660

The issue reporter uses `WorkflowActions.createPullRequest` in scheduled workflows and wants projen to forward two inputs that `peter-evans/create-pull-request@v8` already supports: `add-paths` (restrict which paths get committed) and `delete-branch` (clean up the PR branch when it closes). The wrapper did not expose either, so callers could not use them without dropping back to a hand-written step, which is the convenience tradeoff they were trying to avoid.

This patch adds `addPaths?: string[]` and `deleteBranch?: boolean` to `CreatePullRequestOptions`, forwarded alongside the existing inputs in `src/github/workflow-actions.ts`. `addPaths` is newline-joined to match the action's YAML list shape; `deleteBranch` is forwarded as-is so an explicit `false` survives even if the action's default ever flips. Keys stay in the original order so existing snapshots under `test/__snapshots__`, `test/github/__snapshots__`, and `test/javascript/__snapshots__` do not churn, and `PullRequestFromPatchOptions` inherits both fields automatically because it extends `CreatePullRequestOptions`.

The new `test/github/workflow-actions.test.ts` covers the default (both keys absent), the `addPaths` newline join, and both `deleteBranch: true` / `deleteBranch: false` branches so the opt-out path does not regress. `docs/api/github.md` was regenerated via `node ./projen.js docgen` so jsii docgen stays in sync.

Verified locally with `node ./projen.js test test/github/ test/javascript/upgrade-dependencies.test.ts` (206 tests, 93 snapshots, all passing) and `node ./projen.js compile && node ./projen.js docgen` (only `docs/api/github.md` regenerated).

---


mrgrain (approved):

mrgrain:
Good stuff, thank you!

tmchow:
@mrgrain Thanks! Glad to help.

mrgrain (approved):

PR #4687, opened 2026-05-05
fix(github): synchronize trigger added to auto-queue workflow

kaizencc:
## Problem

PRs that receive force-pushed commits (e.g. from automated daily upgrade workflows) never merge despite passing all checks and being approved.

GitHub automatically disables auto-merge when a PR receives a force-push. The `auto-queue.yml` workflow only triggers on `opened`, `reopened`, and `ready_for_review` — so auto-merge is never re-enabled after the force-push.

## Fix

Add `synchronize` to the `pull_request_target` trigger types so the workflow re-enables auto-merge whenever new commits are pushed to a PR.

---


mrgrain (changes requested):
I still don't think this is the right approach _by itself_.

Adding this event would mean a PR will always be force queued when updated. Regardless if it was previously deliberately dequeued by a human. 

Instead we could probably ensure that auto-approve PRs stay queued. Or we need state management that adds a label when a PR was dequeued by a human.

kaizencc:
Good point — unconditionally re-enabling on every `synchronize` would override a deliberate human dequeue. Here's a revised plan:

## Problem

When the merge queue bot removes a PR (e.g. due to a base branch conflict), GitHub disables auto-merge. The next force-push should re-enable it. But when a *human* disables auto-merge to investigate, we should respect that decision.

## Proposed approach

1. **Add `synchronize` to auto-queue triggers** — re-enable auto-merge on push, *unless* a `do-not-auto-queue` label is present on the PR.

2. **Add `auto_merge_disabled` trigger** — when auto-merge is disabled, check `github.event.sender`:
   - If sender is `github-merge-queue[bot]` → do nothing (this is the conflict/dequeue case; `synchronize` will re-enable on next push)
   - If sender is a human → add a `do-not-auto-queue` label to the PR

3. **Label acts as the override** — as long as the label is present, `synchronize` skips re-enabling. Human removes the label when they're ready to re-queue.

This gives us:
- ✅ Automated force-pushes re-enable auto-merge (the common case)
- ✅ Human dequeue is respected without extra manual steps
- ✅ Human can re-enable by removing the label

Thoughts?

mrgrain:
> PRs that receive force-pushed commits (e.g. from automated daily upgrade workflows) never merge despite passing all checks and being approved.

I can't actually reproduce this. From what I can tell, auto-merge is disabled either
- by a human, on purpose
- by the merge queue because the check failed

The first one, trivially should in my opinion not lead to re-queueing.
The second one is tricky, it could be a genuine failure. It could be transient. We definitely don't want a queue-fail-remove-queue-fail-remove loop. But I guess after a change (e.g. new push) they could be re-queued automatically.

mrgrain:
This is a use case we should be able to safely detect and retry on update.


image

https://github.com/cdklabs/cdk-ecr-deployment/pull/1282#event-26119128034

Issue #4692, opened 2026-05-07
Feature Request(s): Improve GitHub Workflow Patching Ergonomics


Dzuelu:
## Summary

Patching GitHub Actions workflow files in projen is brittle, error-prone, and requires deep knowledge of internal step ordering. This RFC proposes improvements to make workflow modifications safer and more declarative, eliminating the need for fragile positional index-based patching.

### The Problem

Projen generates GitHub Actions workflow YAML files from a structured object model (`GithubWorkflow`, `Job`, `JobStep`). When users need to customize these workflows beyond what the high-level API supports, they are forced into one of three patching strategies, all of which have significant drawbacks:

#### 1. `addOverride()` with hardcoded step indices

```typescript
buildWorkflow?.addOverride("jobs.build.steps.1", modifySetupNode(options));
buildWorkflow?.addOverride("jobs.build.steps.7", backupArtifactPermissions());
```

Step indices are determined by the *current* internal ordering of projen's generated steps. If projen adds, removes, or reorders a step in a future release, these indices silently target the wrong step — producing broken workflows with no compile-time or synth-time error.

#### 2. `JsonPatch` with positional array insertion

```typescript
buildWorkflow?.patch(JsonPatch.add("/jobs/build/steps/2", addSetupJFrogCli()));
// Now the original step 2 shifted to step 3...
buildWorkflow?.addOverride("jobs.build.steps.3.env", addNodeAuthTokenOidc());
```

`JsonPatch.add` on an array index *inserts* at that position, shifting all subsequent elements. This means every subsequent patch must account for all prior insertions. The mental model required is essentially: "track a running offset for each prior insert/remove." Comments like `// Install Dependencies - use original step index 4 before JFrog insertion` are a code smell showing this is unmanageable.

#### 3. `tryFindObjectFile()` + `addOverride()` on the YAML file

```typescript
const deployWorkflow = project.tryFindObjectFile('.github/workflows/deploy.yml');
deployWorkflow?.addOverride('jobs.cdk-deploy.needs', 'dev_approve');
```

This bypasses the typed `GithubWorkflow` API entirely, operating on the raw YAML `ObjectFile`. It works but loses all type safety and discoverability.

### Real-World Evidence

These three patterns are not hypothetical — they represent the actual approaches used in production today:

[...]

PR #4697, opened 2026-05-08
feat: auto-generate stable `id` for all workflow steps

Dzuelu:
Applies Proposed Solution 4 to #4692

## Summary

Automatically assigns a stable, deterministic `id` to every GitHub Actions workflow step that doesn't already have one. This ensures all steps are individually addressable for `needs` references, conditional logic, and output consumption without requiring manual annotation.

## Motivation

GitHub Actions steps without an `id` cannot be referenced by other steps or jobs. Manually assigning IDs is tedious and error-prone, especially as workflows grow. By auto-generating stable IDs at render time, all steps become referenceable by default.

## How it works

A new `assignStepIds()` function runs during workflow rendering with the following priority:

1. **Explicit `id`** — if a step already has an `id`, it is preserved as-is.
2. **Derived from `name`** — the step name is converted to `snake_case` (e.g., `"Install dependencies"` → `install_dependencies`).
3. **Content hash fallback** — for steps with no name, a prefix `auto_` plus an 8-character SHA-256 hash of the `run` or `uses` content is used (e.g., `auto_65b1296e`).

Duplicate IDs within the same job are disambiguated with a `_2`, `_3`, … suffix.

## Changes

- **`src/github/workflows.ts`** — Added `assignStepIds()` function and integrated it into `renderJobs()` so all steps pass through ID assignment before rendering.
- **`test/`** — Updated assertions in `github-workflow.test.ts`, `stale.test.ts`, `node-project.test.ts`, `upgrade-dependencies.test.ts`, and `release.test.ts` to include the new `id` fields. Snapshots updated accordingly.

## Testing

All existing tests pass with updated assertions and regenerated snapshots.

---


test/javascript/node-project.test.ts
```diff
@@ -1968,23 +1985,23 @@ describe("scoped private packages", () => {
...
     const buildWorkflow = yaml.parse(output[".github/workflows/build.yml"]);
     expect(releaseWorkflow.jobs.release.steps).toEqual(
-      expect.arrayContaining(expectedSteps),
+      expect.arrayContaining([
+        { ...expectedStep, id: "aws_codeartifact_login" },
+      ]),
     );
     expect(buildWorkflow.jobs.build.steps).toEqual(
-      expect.arrayContaining(expectedSteps),
+      expect.arrayContaining([expectedStep]),
```
Dzuelu, line 2004: Not real happy about the above test changes but the steps are only populated during synthesis.

Dzuelu (commented):
An undocumented decision I made is to create id's only for steps that can be resolved by projen before synthesis. So you can [see here](https://github.com/projen/projen/pull/4697/changes/b98ec615edbf3dc89fde9a3d752c5dc6c3499032#diff-17eff83e15f034f6102dec1f66f450a454dbdfb4edfb12c2f64668c5c804840aL363) that I took out assigning ids that would be unavailable to be looked up.

Dzuelu:
Locally says I'm up to date with branch and don't get the current failing mutations. Not sure how to resolve that...

Dzuelu:
Going to recreate PR

PR #4726, opened 2026-05-29
feat: auto-generate stable id for workflow steps

Dzuelu:
Applies Proposed Solution 4 from https://github.com/projen/projen/issues/4692 and I believe this should actually close out that issue!

## Summary

Automatically assigns a stable, deterministic `id` to every GitHub Actions workflow step that doesn't already have one. This ensures all steps that are addressable through the projen workflow API are individually referenceable for `needs`, conditional logic (`if: steps.<id>.outputs.*`), and output consumption — without requiring manual annotation.

## Motivation

GitHub Actions steps without an `id` cannot be referenced by other steps or jobs. Manually assigning IDs is tedious and error-prone, especially as workflows grow. By auto-generating stable IDs at job-add time, all addressable steps become referenceable by default.

## How it works

A new `assignStepIds()` function runs when a job is added (`addJobs`) or updated (`updateJobs`), with the following priority:

1. **Explicit `id`** — if a step already has an `id`, it is preserved as-is and is never renamed (it may already be referenced elsewhere).
2. **Derived from `name`** — the step `name` is converted to `snake_case` (e.g., `"Install dependencies"` → `install_dependencies`).
3. **Content hash fallback** — for steps with no `name`, a prefix `auto_` plus an 8-character SHA-256 hash of the `run` or `uses` content is used (e.g., `auto_65b1296e`).

Duplicate explicit IDs in the same job throw an error. Derived/auto IDs that would collide with explicit IDs (or with each other) are disambiguated with a `_2`, `_3`, … suffix.

IDs are assigned eagerly at `addJobs`/`updateJobs` time rather than during render. This keeps a strong invariant: any id that appears in the rendered YAML must also be reachable via `getStep(jobId, stepId)` and the rest of the step-mutation API (`patchStep`, `replaceStep`, `appendStep`, `removeStep`, `insertStepBefore`, `insertStepAfter`).

## Why some rendered steps still don't have an `id`

Steps that come from `job.tools` (e.g. `actions/setup-node`, `actions/setup-java`, `actions/setup-python`, `actions/setup-go`) are intentionally **not** assigned an `id`.

Those steps are produced by `setupTools()` inside `renderJob()` at synth time. They are pushed onto a local steps array and **never written back into `_jobs[jobId].steps`**. Because the public step-mutation API resolves through `_jobs[jobId].steps`, those tool-injected steps are not visible to `getStep`/`patchStep`/`replaceStep`/etc.

If we assigned IDs to them at render time, those IDs would appear in the rendered YAML but would not be addressable through projen — calls like `workflow.getStep("release_maven", "setup_java")` would throw "step not found." That's a misleading API contract: the id would look like a handle, but couldn't actually be used to modify the step.

So the rule is: **only steps that projen can resolve before synthesis get an auto-generated id.** Tool-injected steps stay id-less in the rendered YAML to make it clear they aren't owned by the projen step API. (If we ever want them addressable, the right fix is structural — materialize tool steps into `_jobs[jobId].steps` at `addJobs` time rather than conjuring them inside `renderJob`.)

---


src/github/workflows.ts
```diff
@@ -317,11 +318,16 @@ export class GithubWorkflow extends Component {
   ): void {
     const steps = this.resolveJobSteps(jobId);
     const index = this.findStepIndex(steps, stepId, jobId);
-    const newStep = { ...replacementStep, id: replacementStep.id ?? stepId };
+    const originalId = steps[index].id;
+    const newStep = {
+      ...replacementStep,
+      id: replacementStep.id ?? originalId ?? stepId,
+    };
```
mrgrain: i don't think `originalId` is needed here? `findStepIndex` should not return a step that with an id different to `stepId` anyway...

src/github/workflows.ts
```diff
@@ -317,11 +318,16 @@ export class GithubWorkflow extends Component {
...
 
-    // If the id changed, validate uniqueness of the new id
-    if (newStep.id !== stepId) {
-      this.requireUniqueStepId(steps, newStep.id!, jobId);
+    // If the id changed, validate uniqueness of the new id against the other steps
+    if (newStep.id !== originalId) {
+      const otherSteps = steps.filter((_, i) => i !== index);
+      this.requireUniqueStepId(otherSteps, newStep.id!, jobId);
```
mrgrain: also slightly unexpected. is `requireUniqueStepId` not working correctly?

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+ * Priority:
+ * 1. Explicit `id` on the step — preserved as-is and never renamed.
+ * 2. Derived from `name` — converted to snake_case.
+ * 3. Content hash fallback — `auto_<hash>` based on `run` or `uses`.
+ *
+ * Explicit IDs are reserved in a first pass so that derived/auto IDs never
+ * collide with them. Duplicate explicit IDs in the same job throw an error.
+ * Derived/auto IDs that collide are disambiguated with `_2`, `_3`, etc.
+ */
+function assignStepIds(
```
mrgrain:
```suggestion
function ensureStepsHaveIds(
```

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+  steps: workflows.JobStep[],
+): Array<workflows.JobStep & { id: string }> {
+  const usedIds = new Map<string, number>();
+
+  // First pass: reserve all explicit IDs so derived IDs won't claim them.
```
mrgrain, line 708: can we use `requireUniqueStepId` here?

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+    }
+    const next = count + 1;
+    usedIds.set(base, next);
+    return `${base}_${next}`;
+  }
+
+  return steps.map((step) => {
+    if (step.id) {
+      // Explicit id — already reserved in first pass, use as-is
+      return { ...step, id: step.id };
```
mrgrain, line 734: do we need to create a new object here?

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+    }
+
+    if (step.name) {
+      // Derive from name: snake_case, collapse non-alphanumeric
+      const base = snake(step.name)
+        .replace(/[^a-z0-9_]/g, "_")
+        .replace(/_+/g, "_")
+        .replace(/^_|_$/g, "");
```
mrgrain: We can just remove non allowed chars and then we don't need to collapse or do other complicated stuff.

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+      const id = uniqueId(base);
+      return { ...step, id };
+    }
+
+    // Fallback: content hash
```
mrgrain, line 754: Not really a content hash in the last case...

src/github/workflows.ts
```diff
@@ -663,6 +669,96 @@ function renderJobs(
...
+    const content = step.run ?? step.uses ?? "";
```
mrgrain, line 755: maybe uses is better treated same as `name?`

mrgrain (changes requested):
Some minor feedback. Can you also add documentation to the guide please.

PS: Sorry this took so long to get around to.

src/github/workflows.ts
Dzuelu: Unfortunatly we do need to "convert" it to an object here because the `workflows.JobStep` keys are readonly. So this is kinda a hack. :/

src/github/workflows.ts
Dzuelu: This is the case where there is no `id` or `name` so we have to generate it off of the command or uses hash. Imo it is distinct from has name but no id case because the name is more easily readable.

src/github/workflows.ts
Dzuelu: No unless we want a breaking change. This allows the current workflows to eagerly assign step IDs or generate it. If we wanted to do that we should instead just make the workflow id key required instead imo.

src/github/workflows.ts
Dzuelu: Needed if allowed to change the id. I do agree with your above comment that it shouldn't be allowed tho so removed.

src/github/workflows.ts
Dzuelu: Had ai comment / make it more clear and it ended up making it longer. If we want to remove snake_case it does make it more simple but less readable imo so my preference is snake_case, but can remove if needed.

Dzuelu:
No worries, applied your feedback. I believe I've updated the docs correctly but lmk if there are others / changes. :)

mrgrain (approved):

PR #4748, opened 2026-06-17
feat(tasks): run commands through dax on windows

### LINKED ISSUE #4746: Projen 0.99.73+ ships with broken dependency tree
Due to a CVE https://github.com/projen/projen/blob/v0.99.74/.projenrc.ts#L201-L203 fix the npm package resolution mechanism is broken because of mismatching dependencies.

The bundled `shx@0.4.0` declares `shelljs@^0.9.2`, but projen forces `shelljs@^0.10.0` via addPackageResolutions. The bundle therefore ships `shelljs@0.10.0`, which does not satisfy shx's ^0.9.2 range.

In downstream projects this makes npm ci impossible to satisfy — npm tries to resolve a separate `shelljs@0.9.2` and its legacy subtree (`execa@1.0.0`, `cross-spawn@6`, `semver@5.7.2`, etc.), none of which are in the lockfile, so it fails with:

```
npm error `npm ci` can only install packages when your package.json and package-lock.json are in sync.
npm error Missing: shelljs@0.9.2 from lock file
npm error Missing: execa@1.0.0 from lock file
```

mrgrain:
Fixes #4746

projen tasks are written with POSIX-style commands such as `cat`, `cp`, `mkdir -p`, `rm -rf`, `&&` chains and `$VAR` expansion. Windows' default shell, `cmd.exe`, does not understand any of these, so projen bundled `shx` (a thin CLI over `shelljs`) to provide cross-platform implementations. When a CVE in `shelljs`' transitive `execa@1.0.0` was patched by forcing `shelljs` to a newer version, that version fell outside the range `shx` declared. The result was an internally inconsistent bundled dependency tree: `npm install` accepted the bundle as-is, but a strict `npm ci` against the generated lockfile failed for everyone depending on projen.

This replaces `shx` (and the `shelljs` override) with `dax`, a maintained, cross-platform shell that ships built-in `cat`/`cp`/`mkdir`/`mv`/`rm` commands and carries a lean dependency tree, removing the source of the recurring `shelljs` problems. The task runtime now routes commands through `dax` only on Windows; macOS and Linux continue to run through the system shell exactly as before, so the behavioural change is scoped to the platform that was already relying on a shim. Because `dax` executes commands asynchronously while the task runtime is synchronous, the Windows path drives `dax` from a synchronous child `node` process.

It also adds `Tasks.runTask(name, args?)`, backed by an internal `TaskRuntime`, so that post-synthesis dependency installation runs tasks through a single, consistent entry point instead of each package manager constructing its own runtime.

---


src/task-runtime.ts
```diff
@@ -359,16 +358,41 @@ class RunTask {
...
+
+      return spawnSync(process.execPath, ["-e", runner, options.command], {
+        ...options,
+        cwd,
+        shell: false,
+        stdio: "inherit",
+        env,
+        ...options.spawnOptions,
+      });
+    }
```
mrgrain, line 388: This is ugly, but the minimal fix. I have a better solution in the works but that will be a bigger change that is also technically breaking.

otaviomacedo (approved):

Issue #1563, opened 2022-01-25
Tasks that use "builtin" steps do not work after projen eject


Chriscbr:
The current reason for this is that "builtin" steps depend on TypeScript/JavaScript files that are included in the projen distribution. Once projen is removed from the project and uninstalled, the files in `node_modules/projen/...` can no longer be accessed.

One solution that comes to mind is finding a way to bundle these files and add them to the "scripts" directory after projen is ejected.

Another solution could be to extract the projen task runner into its own tool/dependency.

PR #4752, opened 2026-06-17
fix(tasks)!: make projen eject produce a self-contained runner

### LINKED ISSUE #3679: Projen eject generates run-task which fails for non-node projects
When Projen ejects, it generates a `scripts/run-task` file which is a es bundle of the `src/task-runtime.ts` file. As apart of this, there are two references to package json as follows:

```
// src/common.ts
1598: var packageVersion = require("../package.json").version;

...

// src/task-runtime.ts
...
renderBuiltin(builtin) {
1889:  const moduleRoot = (0, import_path.dirname)(require.resolve("../package.json"));
  const program = require.resolve((0, import_path.join)(moduleRoot, "lib", `${builtin}.task.js`));
  return `'${process.execPath}' '${program}'`;
}
```

As you can see, it assumes the package.json exists one directory above. For non-node projects, this will fail as no package.json will exist.

### LINKED ISSUE #4407: After projen eject npm run build fails
In our typescript project we ran `projen eject` to remove projen and the generated "run-task.cjs" causes `npm run build` to fail.

We get the following output when running the commands:

```
> npx projen eject
👾 eject » default | ts-node --swc --project tsconfig.dev.json .projenrc.ts
👾 Installing dependencies...
👾 install | npm install

removed 93 packages, and audited 435 packages in 608ms

134 packages are looking for funding
  run `npm fund` for details

1 low severity vulnerability

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
```

```
> npm run build                                

> scaffold-test@0.0.0 build
> scripts/run-task.cjs build

/Users/alexandre.carvalho/Desktop/Projects/TreasCloud/scaffold-test/scripts/run-task.cjs:2098
    this.manifest = (0, import_fs.existsSync)(manifestPath) ? parseConflictJSON(
                                                              ^

TypeError: parseConflictJSON is not a function
    at new _TaskRuntime (/Users/xyz/Desktop/Projects/scaffold-test/scripts/run-task.cjs:2098:63)
    at Object.<anonymous> (/Users/xyz/Desktop/Projects/scaffold-test/scripts/run-task.cjs:2361:17)
    at Module._compile (node:internal/modules/cjs/loader:1369:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)
    at Module.load (node:internal/modules/cjs/loader:1206:32)
    at Module._load (node:internal/modules/cjs/loader:1022:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49

Node.js v20.12.2
```

mrgrain:
Fixes #3679
Fixes #4407

`projen eject` produces a bundled task runner (`scripts/run-task.cjs`) so a project can keep running its tasks after projen is removed. Two bugs made that runner fail in real ejected repositories; this PR fixes both and restructures the task runtime so the boundary between the library and the CLI is explicit.

BREAKING CHANGE: `TaskRuntime` is no longer exported from the `projen` package root - it is now an internal part of the CLI. `Task` and `Tasks` are unchanged. To migrate, run tasks via `project.tasks.runTask(name)` during synthesis or `npx projen <task>` at runtime instead of constructing a `TaskRuntime`. The `.projen/tasks.json` manifest path is no longer published as a constant (it was previously `TaskRuntime.MANIFEST_FILE`); it remains `.projen/tasks.json`.

---

### projen eject fixes

The bundled runner assumed a sibling `package.json` existed one directory up (it read projen's version from it at load time). For non-node projects there is no such file, so `npm run build` - and every other task - failed outright (#3679). The bundle also crashed with `parseConflictJSON is not a function` once ejected (#4407).

The version is now inlined into the bundle at build time, so there is no runtime `package.json` lookup; the builtin resolver locates the package root by walking up rather than assuming a fixed depth; and the `parse-conflict-json` import is interop-safe inside the bundle. A new integration test (`scripts/integ-eject.sh`, run as part of the node integ job) ejects a real project, removes projen from `node_modules`, and runs a task purely through `scripts/run-task.cjs`, so this can't regress unnoticed.

### Task runtime restructuring

Task execution now sits behind a small `ITaskRunner` abstraction. The default `ProjenTaskRunner` shells out to the projen CLI to run a task, which keeps the library decoupled from the CLI (enforced by an `import/no-restricted-paths` lint rule: `src/` may not import from `src/cli/`) and leaves room for alternative runners later. As part of this the `TaskRuntime` moved into the CLI and became asynchronous. The runner is intentionally kept internal for now while its shape settles.

The cross-platform Windows behavior (running task commands through [dax](https://github.com/dsherret/dax) instead of `cmd.exe`) is not new - it already exists. Because the runtime is now async, that execution is driven via dax's `$` API directly instead of through a synchronous helper process; the behavior is unchanged.

Builtin task steps still do not work after eject (#1563): they resolve into `node_modules/projen/lib`, which is gone once projen is uninstalled. Bundling builtins into the ejected output is left as a follow-up.

---


src/cli/task-runtime.ts
```diff
@@ -5,18 +5,57 @@ import { dirname, join, resolve } from "path";
...
 const QUOTED_ARGS_MARKER = `"${ARGS_MARKER}"`;
 
+/**
+ * The result of executing a shell command.
+ *
+ * This is a normalized subset of node's `SpawnSyncReturns` so that the
+ * synchronous (`spawnSync`) and asynchronous (`dax`) execution paths can return
+ * a compatible shape.
+ */
+interface ShellResult {
```
kaizencc, line 34: why all the `| nulls` here, even for optional props. shouldn't `undefined` suffice

kaizencc (approved):
broadly ok with this

src/cli/task-runtime.ts
mrgrain: thats what `spawnSync` returns. That's why.

PR #4756, opened 2026-06-18
chore: agent guidance for integration tests

mrgrain:
Adds integration test guidance to the AI instructions so agents are aware of how to properly test end-to-end behavior.

Also, most AI coding agents have converged on supporting `AGENTS.md` (OpenAI Codex, Cursor, Windsurf, etc.) and `CLAUDE.md` (Claude Code) as their standard instruction files. This PR switches to explicitly target only `AGENTS.md` and `CLAUDE.md`, which together cover the vast majority of AI agents. The per-tool files are removed.

---


hoegertn:
Could we take that even further and have the Claude.md just contain an `@AGENTS.md` reference?

mrgrain:
> Could we take that even further and have the Claude.md just contain an `@AGENTS.md` reference?

I think that'd be a regression. Claude will now have to read two files and spend extra tokens. The reason we do this elsewhere is maintenance effort. But with projen we don't have to care, the files are generated anyway.

kaizencc (approved):

PR #4807, opened 2026-07-10
fix(cdk): jsii projects fail to run jest tests with TypeScript v6

### LINKED ISSUE #4806: Build issue with Typescript v6 in JSII projects (like AwsCdkConstructLibrary)
There is a new issue with TypeScript v6. A fresh `AwsCdkConstructLibrary` projen project does not build (you need to add a single test to reproduce it). Only JSII projects seem to be affected as far as I can tell — e.g. `awscdk-app-ts` is not affected.

**Reproducer:** https://github.com/WtfJoke/projen-4595-reproducer/

Steps:

1. `pnpm dlx projen new awscdk-construct`
2. add a dummy test
3. `pnpm run build`

Should I open a new issue or would you prefer to reopen this one?

The error in the pipeline:
```
👾 release » build » test | jest --passWithNoTests --updateSnapshot
ts-jest[config] (WARN) message TS151002: Using hybrid module kind (Node16/18/Next) is only supported in "isolatedModules: true". Please set "isolatedModules: true" in your tsconfig.json. To disable this message, you can set "diagnostics.ignoreCodes" to include 151002 in your ts-jest config. See more at https://kulshekhar.github.io/ts-jest/docs/getting-started/options/diagnostics
FAIL test/dummy.test.ts
  ● Test suite failed to run

    Unable to process '/home/runner/work/projen-4595-reproducer/projen-4595-reproducer/test/dummy.test.ts', please make sure that `outDir` in your tsconfig is neither `''` or `'.'`. You can also configure Jest config option `transformIgnorePatterns` to inform `ts-jest` to transform /home/runner/work/projen-4595-reproducer/projen-4595-reproducer/test/dummy.test.ts

      at TsCompiler.getCompiledOutput (node_modules/.pnpm/ts-jest@29.4.11.../ts-jest/dist/legacy/compiler/ts-compiler.js:334:27)

----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------|---------|----------|---------|---------|-------------------
All files |       0 |        0 |       0 |       0 |
----------|---------|----------|---------|---------|-------------------
Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
```

Here is the result of the analysis created with help from AI:
<details>
  <summary>Root cause (AI-assisted analysis, verified by bisecting one option at a time)</summary>
 
## Root cause (AI-assisted analysis, verified by bisecting one option at a time)

The misleading `outDir` error is not actually about `outDir`. The real trigger is the tsconfig that ts-jest consumes.

For a JSII project, `tsconfigDev` (the tsconfig ts-jest is pointed at via `TypeScriptProject.addJestNoCompileModern`) inherits from the JSII-strict base tsconfig, which forces:

- `module: "node16"` (a "hybrid" module kind), and
- `noEmitOnError: true`

With ts-jest 29, each test goes through TypeScript's `getEmitOutput`. The hybrid module kind attaches a diagnostic (`TS151002`), and because `noEmitOnError: true` is inherited, TypeScript sets `emitSkipped: true`. ts-jest sees `emitSkipped` and throws the confusing "outDir … is '' or '.'" message (in `ts-compiler.ts` `getCompiledOutput`, the `output.emitSkipped` branch).

`awscdk-app-ts` doesn't hit this because the non-JSII base tsconfig uses `noEmitOnError: false` (and `module: NodeNext`).

I confirmed the cause by changing exactly one thing at a time on the generated configs:

| Change (in the ts-jest tsconfig) | Result |
| --- | --- |
| baseline (as generated) | fails with the `outDir` error |
| `isolatedModules: true` | passes, and also clears the `TS151002` warning |
| `noEmitOnError: false` | passes (warning remains) |

## Proposed fix

In `JsiiProject`'s default options (`src/cdk/jsii-project.ts`), make the ts-jest-consumed `tsconfigDev` compatible by adding `isolatedModules`:

```ts
tsconfigDev: {
  compilerOptions: {
    stripInternal: true,
    isolatedModules: true, // <-- add
  },
},
```

Rationale:

- This is exactly what ts-jest recommends for `Node16/NodeNext` modules (the `TS151002` warning), so it also silences that warning.
- It only touches the dev/test tsconfig — JSII's actual build uses the base tsconfig, so the shipped artifact's type-checking and emit are unaffected.

Alternative: set `noEmitOnError: false` on `tsconfigDev` instead (the dev tsconfig is only used for type-checking and jest, never for the emitted build). This keeps test type-checking on but leaves the `TS151002` warning.

More broadly, `addJestNoCompileModern` could set `isolatedModules: true` whenever the effective `module` is `Node16/NodeNext` and `ts-jest >= 29`, which would cover the whole class of problem rather than just JSII.
  
</details>

 _Originally posted by @WtfJoke in [#4595](https://github.com/projen/projen/issues/4595#issuecomment-4926344041)_

mrgrain:
## Summary

Fixes #4806.

`JsiiProject`'s `tsconfigDev` (the config ts-jest actually consumes) inherits `module: "node16"` and `noEmitOnError: true` from the jsii-strict base `tsconfig`. With ts-jest 29+, hybrid module kinds (Node16/18/Next) require `isolatedModules: true`; without it, `noEmitOnError` causes `emitSkipped`, which ts-jest surfaces as a misleading `outDir` error instead of the real diagnostic.

This only adds `isolatedModules: true` to `tsconfigDev`, leaving the jsii build's actual `tsconfig` (and therefore the shipped artifact's type-checking/emit) untouched.

## What was tested

- Added a regression test in `test/jsii/jsii.test.ts` (`supports running jest tests`) that synthesizes a `JsiiProject`, writes a dummy test file, and runs `projen test` end-to-end via the CLI.
- Confirmed the test fails with the exact reported `outDir` error before the fix, and passes after.
- Ran the full `test/jsii/jsii.test.ts` suite (13 tests) — all pass; 3 snapshots updated to reflect `isolatedModules: true` in the generated `test/tsconfig.json`.

### DIFF
--- a/src/cdk/jsii-project.ts
+++ b/src/cdk/jsii-project.ts
@@ -211,6 +211,7 @@ export class JsiiProject extends TypeScriptProject {
       tsconfigDev: {
         compilerOptions: {
           stripInternal: true,
+          isolatedModules: true,
         },
       },
     };
--- a/test/__snapshots__/integ.test.ts.snap
+++ b/test/__snapshots__/integ.test.ts.snap
@@ -1604,7 +1604,8 @@ project.synth();
   "compilerOptions": {
     "noEmit": true,
     "rootDir": "..",
-    "stripInternal": true
+    "stripInternal": true,
+    "isolatedModules": true
   },
   "include": [
     "**/*.ts"
@@ -2889,7 +2890,8 @@ project.synth();
   "compilerOptions": {
     "noEmit": true,
     "rootDir": "..",
-    "stripInternal": true
+    "stripInternal": true,
+    "isolatedModules": true
   },
   "include": [
     "**/*.ts"
--- a/test/cdk/__snapshots__/jsii.test.ts.snap
+++ b/test/cdk/__snapshots__/jsii.test.ts.snap
@@ -1530,6 +1530,7 @@ UNEXPECTED BREAKING CHANGES: add keys such as 'removed:constructs.Node.of' to .c
 }",
   "test/tsconfig.json": {
     "compilerOptions": {
+      "isolatedModules": true,
       "noEmit": true,
       "rootDir": "..",
       "stripInternal": true,
--- a/test/jsii/__snapshots__/jsii.test.ts.snap
+++ b/test/jsii/__snapshots__/jsii.test.ts.snap
@@ -1531,6 +1531,7 @@ UNEXPECTED BREAKING CHANGES: add keys such as 'removed:constructs.Node.of' to .c
 }",
   "test/tsconfig.json": {
     "compilerOptions": {
+      "isolatedModules": true,
       "noEmit": true,
       "rootDir": "..",
       "stripInternal": true,
@@ -3116,6 +3117,7 @@ UNEXPECTED BREAKING CHANGES: add keys such as 'removed:constructs.Node.of' to .c
 }",
   "test/tsconfig.json": {
     "compilerOptions": {
+      "isolatedModules": true,
       "noEmit": true,
       "rootDir": "..",
       "stripInternal": true,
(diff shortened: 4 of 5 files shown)


mrgrain:
Recreating on a branch/PR that follows repo conventions.

PR #4815, opened 2026-07-13
feat(biome): update Biome to v2.5.0+ and new config

ben-buitendijk:
Biome V2.5 deprecated "recommended: true" in the config and replaces it with "preset: recommended".  

This PR updates the Biome component to the config format introduced in v2.5

---


ben-buitendijk:
The build is failing because the package-lock has changed, but this is due to the bump in biome's version.

How should I handle this?

image

mrgrain (approved):
Cool! Thanks for that.

PR #4819, opened 2026-07-15
fix(release): make CHANGES_SINCE_LAST_RELEASE robust to the dax pipe race

alvazjor:
### What broke

`CHANGES_SINCE_LAST_RELEASE` (the condition on the generated `bump` and `release` tasks) is `git log --oneline -1 | grep -qv "chore(release):"`. `grep -q` closes the read end of the pipe as soon as it has an answer. When the condition runs through projen's built-in dax shell, that early close races with `git log` still writing, and dax surfaces the failed write as a spurious non-zero exit (`stdin pipe broken. Invalid state: WritableStream is closed`). The task runtime treats that as "skip", so a `bump`/`release` task is silently skipped even on a normal commit: a release is missed, or (in a monorepo) a dependent publishes a broken `^0.0.0` dependency range.

Real-world impact: `@aws-cdk/service-spec-importers@0.0.133` shipped `@aws-cdk/service-spec-types: "^0.0.0"` (`npm view @aws-cdk/service-spec-importers@0.0.133 dependencies`; the prior `0.0.132` was `^0.0.251`). Release log: https://github.com/cdklabs/awscdk-service-spec/actions/runs/28693317594 shows two `stdin pipe broken` errors matched 1:1 with two skipped bump tasks on a non-release HEAD, versus zero in the clean runs https://github.com/cdklabs/awscdk-service-spec/actions/runs/27250884136 and https://github.com/cdklabs/awscdk-service-spec/actions/runs/28980091973.

### Why fix the condition, not `evalCondition`

On dax 0.49 (bundled by projen), an early-closing pipe under `.noThrow()` returns a plain non-zero exit code rather than throwing (verified: `seq 1 500000 | grep -qv "chore(release):"` returns code 1 with the pipe-broken error on stderr). `evalCondition` inspects only the exit status, so the infrastructure failure is indistinguishable from a legitimate condition-false without brittle stderr string-matching. The robust fix is therefore to remove the trigger in the condition itself.

### The fix

Drop `-q`:

```
git log --oneline -1 | grep -v "chore(release):" > /dev/null
```

`grep -v` reads to EOF instead of short-circuiting, so it never closes the pipe while the writer is active and the race cannot occur. Exit semantics are identical (proceed when the last commit line does not contain `chore(release):`, skip when it does). A comment on the constant explains why `-q` must not be reintroduced.

A pipe-free form was considered and rejected: dax's shell subset does not support `case`, `if`, or `${var#pattern}` (verified on 0.49), and `expr` is not guaranteed on Windows. Dropping `-q` removes the defect with the least change and stays cross-platform.

### Validation

- Same-stress A/B (`seq 1 500000`, 10 runs each): under dax the old `grep -qv` fails 10/10 (code 1 + pipe-broken) while the new form passes 10/10 (code 0, no pipe error); under `/bin/sh` both pass, confirming the defect is specific to dax's in-process pipe.
- New tests pin the exact constant value and execute the condition under both the built-in (dax) and system shells for a normal HEAD (proceeds) and a `chore(release)` HEAD (skips).
- The only piped `grep -q` in the repo is this condition; `src/release/publisher.ts` uses `grep -q` on a file (not a pipe) and is unaffected. Snapshots updated for the new condition string.

### Related PRs

Downstream consumers (cdklabs/cdklabs-projen-project-types) currently carry interim mitigations for this defect; once this fix is released they can be removed:
- Fail-loud guard: https://github.com/cdklabs/cdklabs-projen-project-types/pull/960
- Release-shell hardening + retry: https://github.com/cdklabs/cdklabs-projen-project-types/pull/961

---

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -54,7 +54,7 @@
           "builtin": "release/bump-version"
         }
       ],
-      "condition": "git log --oneline -1 | grep -qv \"chore(release):\""
+      "condition": "git log --oneline -1 | grep -v \"chore(release):\" > /dev/null"
     },
     "bundle:task-runner": {
       "name": "bundle:task-runner",
--- a/src/version.ts
+++ b/src/version.ts
@@ -6,19 +6,35 @@ import { NodePackage } from "./javascript/node-package";
 import type { Task } from "./task";
 
 /**
- * This command determines if there were any changes since the last release in a cross-platform compatible way.
- * It is used as a condition for both the `bump` and the `release` tasks.
+ * This command determines if there were any changes since the last release in a
+ * cross-platform compatible way. It is used as a condition for both the `bump`
+ * and the `release` tasks: it exits 0 (proceed) when the most recent commit is
+ * not a release commit, and non-zero (skip) when the most recent commit is a
+ * `chore(release):` commit.
  *
  * Explanation:
- *  - log commits                                               | git log
- *  - limit log output to a single line per commit              | --oneline
- *  - looks only at the most recent commit                      | -1
- *  - silent grep output                                        | grep -q
- *  - exits with code 0 if a match is found                     | grep -q "chore(release):"
- *  - exits with code 1 if a match is found (reverse-match)     | grep -qv "chore(release):"
+ *  - `git log --oneline -1`          the most recent commit, one line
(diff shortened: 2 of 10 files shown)


mrgrain:
@mergifyio queue

mrgrain (approved):

PR #4822, opened 2026-07-17
feat(javascript): add dedupe task to remove duplicate dependencies

rix0rrr:
Over multiple upgrade cycles, lock files tend to accumulate several versions of the same package, even when a single version could satisfy all requested ranges — think, *as a purely hypothetical example*, 5 copies of the AWS SDK in one dependency tree. Package managers ship a `dedupe` command to clean this up, but nothing was running it.

This PR adds a new `dedupeDeps` option (default: `false`) to Node.js projects. When enabled, projen adds a `dedupe` task that keeps the dependency tree lean:

- **npm** runs `npm dedupe`
- **pnpm** runs `pnpm dedupe`
- **Yarn Berry** runs `yarn dedupe`. If `yarnBerryOptions.dedupePackages` is set, only the listed packages are deduplicated (glob patterns like `@aws-sdk/*` are supported). Setting that list also implies `dedupeDeps: true`, so it works out of the box.
- **Yarn Classic** already dedupes as part of `yarn install`, so the task just prints a note saying the command isn't necessary.
- **Bun** has no dedupe command, so enabling the option throws an error.

The `dedupe` task is spawned as part of the mutable `install` task and during the `upgrade` task. The reasoning: even projects without automatic upgrades should keep the deduping behavior when dependencies are changed by hand, though in practice the upgrade workflow is where it matters most.

Projen itself now dogfoods the feature with `dedupeDeps: true`.

Incidentally, this also refreshes the Yarn Berry `.yarnrc.yml` options with settings added in recent Yarn releases (catalogs, `npmMinimalAgeGate`, `taskPoolMode`, and friends), and deprecates a couple of misspelled options (`workerPoolMode`, `enableConstraintsCheck`) in favor of correctly named ones.

 _PR body rewritten by an AI Agent on behalf of @mrgrain_

---

### DIFF
--- a/src/javascript/node-package.ts
+++ b/src/javascript/node-package.ts
@@ -909,6 +909,11 @@ export class NodePackage extends Component {
       description: "Install project dependencies using frozen lockfile",
       exec: this.installCommand,
     });
+
+    if (isYarnBerry(this.packageManager)) {
+      // Needs to happen after the install tasks have been added
+      this.configureYarnBerryDedupe(options.yarnBerryOptions?.dedupePackages);
+    }
   }
 
   /**
@@ -2137,6 +2142,22 @@ export class NodePackage extends Component {
     });
   }
 
+  /**
+   * Configure package deduping
+   */
+  private configureYarnBerryDedupe(dedupePackages: string[] | undefined) {
+    if (!dedupePackages || dedupePackages.length === 0) {
+      return;
+    }
+
+    const dedupeTask = this.project.addTask("yarn:dedupe", {
+      execArgs: ["yarn", "dedupe", ...dedupePackages],
+    });
+
+    // Add dedupe to non-CI install command
+    this.project.tasks.tryFind("install")?.spawn(dedupeTask);
+  }
+
   private checkForConflictingYarnOptions(yarnRcOptions: YarnrcOptions) {
     if (
       this.npmAccess &&
@@ -2273,6 +2294,20 @@ export interface YarnBerryOptions {
    * @default false
    */
   readonly zeroInstalls?: boolean;
+
+  /**
+   * Packages to deduplicate.
+   *
+   * This will prevent multiple versions of the same package from being
+   * installed in the lock file, if a single version could satisfy all requested
+   * version ranges. This will prevent version proliferation and reduce the size
+   * of the depdendency tree.
+   *
+   * Setting this will run `yarn dedupe` after dependency upgrades.
+   *
+   * Supports glob patterns, e.g. `@aws-sdk/*`.
+   */
+  readonly dedupePackages?: string[];
 }
 
 /**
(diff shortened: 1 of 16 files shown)


src/javascript/node-package.ts
```diff
@@ -2137,6 +2149,25 @@ export class NodePackage extends Component {
...
+    }
+
+    const dedupeCommand = ["yarn", "dedupe", ...dedupePackages];
+
+    // Add dedupe to non-CI install command
+    this.project.tasks.tryFind("install")?.execArgs(dedupeCommand);
+
+    // Also add the action to a post-upgrade step if one is added
+    this.preSynthesizeActions.push(() => {
+      this.project.tasks.tryFind("post-upgrade")?.execArgs(dedupeCommand);
```
mrgrain: yeah this doesn't really do it I think as there  might be more than one `post-upgrade` task.

mrgrain (changes requested):
I'll need to think about this a little more, but could you revert the refactor? I agree that NodeProject/NodePackage are too connected today, but don't think this is the time to change that. You should be able to anchor the tests on NodeProject instead.

src/javascript/upgrade-dependencies.ts
```diff
@@ -589,9 +597,25 @@ export class UpgradeDependencies extends Component {
         : {}),
     };
 
+    const nodeProject =
+      this.project instanceof NodeProject ? this.project : undefined;
```
mrgrain: This shows you that the refactor doesn't really work today. For better or worse we try to anchor GH Workflows to project, not components.

src/javascript/node-package.ts
mrgrain: Typically you do one of:

1. Give `UpgradeDependencies` an explicit option that we just default to if the value is set
2. Expose the option on the `NodePackage` component as a property, have `UpgradeDependencies` read it
3. Both of the above with the explicit option serving as an override

Typically we want (1) and maybe the other two. The main downside of (2) is that it ties `UpgradeDependencies` to `NodePackage` which might or might not be desired.

src/javascript/node-package.ts
```diff
@@ -2137,6 +2149,25 @@ export class NodePackage extends Component {
...
 
+  /**
+   * Configure package deduping
+   */
+  private configureYarnBerryDedupe(dedupePackages: string[] | undefined) {
+    if (!dedupePackages || dedupePackages.length === 0) {
+      return;
+    }
+
+    const dedupeCommand = ["yarn", "dedupe", ...dedupePackages];
```
mrgrain: probably should be a separate task that is than spawned?

rix0rrr:
Simplified the PR a bunch.

mrgrain (approved):

PR #4856, opened 2026-08-06
feat(release): support GitHub App credentials for Go publishing

### LINKED ISSUE #4498: Publisher: Support GH App for GoPublisher
Add support to publish Go targets with GH Apps

https://github.com/projen/projen/blob/8b796c77ecdf3a7bbb26dd8c81904861d552b79f/src/release/publisher.ts#L1306-L1362

mrgrain:
Fixes #4498

Adds a `githubCredentials` option to `publishToGo`, so Go module releases can authenticate with a GitHub App token instead of a personal access token secret or SSH deploy key — avoiding a long-lived credential that has to be rotated by hand. Two projects already worked around this by patching the generated release job after synthesis ([1](https://github.com/mrgrain/cdk-esbuild/pull/1674), [2](https://github.com/cdktn-io/cdktn-provider-project/pull/21)); this makes it a first-class option.

`githubTokenSecret` is now deprecated in favor of `githubCredentials` (a PAT can be modeled via `GithubCredentials.fromPersonalAccessToken()`). `githubUseSsh` and `githubDeployKeySecret` stay as-is for now, since `GithubCredentials` has no SSH/deploy-key equivalent yet. Combining `githubCredentials` with any of the three legacy options throws, as does combining `githubUseSsh` with `githubTokenSecret` (they're incompatible auth methods).

---

### DIFF
--- a/src/release/publisher.ts
+++ b/src/release/publisher.ts
@@ -6,6 +6,7 @@ import {
   DEFAULT_GITHUB_ACTIONS_USER,
   PERMISSION_BACKUP_FILE,
 } from "../github/constants";
+import type { GithubCredentials } from "../github/github-credentials";
 import type {
   Job,
   JobPermissions,
@@ -650,9 +651,30 @@ export class Publisher extends Component {
    * @param options Options
    */
   public publishToGo(options: GoPublishOptions = {}) {
+    if (
+      options.githubCredentials &&
+      (options.githubUseSsh ||
+        options.githubTokenSecret ||
+        options.githubDeployKeySecret)
+    ) {
+      throw new Error(
+        "Only one of 'githubCredentials' or 'githubUseSsh'/'githubTokenSecret'/'githubDeployKeySecret' may be specified, not both.",
+      );
+    }
+
+    if (options.githubUseSsh && options.githubTokenSecret) {
+      throw new Error(
+        "'githubTokenSecret' cannot be used together with 'githubUseSsh'. Use 'githubDeployKeySecret' instead.",
+      );
+    }
+
     const prePublishSteps = options.prePublishSteps ?? [];
     const workflowEnv: { [name: string]: string | undefined } = {};
-    if (options.githubUseSsh) {
+    if (options.githubCredentials) {
+      const credentials = options.githubCredentials;
+      prePublishSteps.push(...credentials.setupSteps);
+      workflowEnv.GITHUB_TOKEN = credentials.tokenRef;
+    } else if (options.githubUseSsh) {
       workflowEnv.GITHUB_USE_SSH = "true";
       workflowEnv.SSH_AUTH_SOCK = "/tmp/ssh_agent.sock";
       prePublishSteps.push({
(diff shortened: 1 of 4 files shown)


src/release/publisher.ts
```diff
@@ -1296,19 +1316,37 @@ export interface GoPublishOptions extends CommonPublishOptions {
    * The name of the secret that includes a GitHub deploy key used to push to the
    * GitHub repository.
    *
-   * Ignored if `githubUseSsh` is `false`.
+   * Ignored if `githubUseSsh` is `false` or `githubCredentials` is set.
    *
+   * @deprecated use `githubCredentials` instead.
```
mrgrain: we cannot deprecate this, githubCredentials doesn't allow for ssh and deploy key

src/release/publisher.ts
```diff
@@ -1296,19 +1316,37 @@ export interface GoPublishOptions extends CommonPublishOptions {
...
+   * Provide API access to the GitHub repository the Go module is pushed to
+   * through a `GithubCredentials` instance, e.g. a GitHub App
+   * (`GithubCredentials.fromApp`).
+   *
+   * This is the recommended way to authenticate. It replaces `githubUseSsh`,
+   * `githubTokenSecret` and `githubDeployKeySecret`, which cannot be combined
+   * with this option. When a GitHub App credential specifies an
+   * `environment`, it is used unless `githubEnvironment` is also set.
+   *
+   * @default - use `githubTokenSecret` or `githubDeployKeySecret`, depending on `githubUseSsh`
```
mrgrain: the default is wrong also no need to mention what it replaces

src/release/publisher.ts
```diff
@@ -650,9 +651,24 @@ export class Publisher extends Component {
...
    */
   public publishToGo(options: GoPublishOptions = {}) {
+    if (
+      options.githubCredentials &&
+      (options.githubUseSsh ||
+        options.githubTokenSecret ||
+        options.githubDeployKeySecret)
+    ) {
+      throw new Error(
+        "Only one of 'githubCredentials' or 'githubUseSsh'/'githubTokenSecret'/'githubDeployKeySecret' may be specified, not both. Remove the deprecated options in favor of 'githubCredentials'.",
```
mrgrain: githubDeployKeySecret should not be deprecated

mrgrain: githubTokenSecret  is also incompatible with githubUseSsh.

src/release/publisher.ts
mrgrain: Fixed: removed the `@deprecated` tag from `githubUseSsh` — it can't be represented via `GithubCredentials` today, so it stays a supported, non-deprecated option.

src/release/publisher.ts
mrgrain: Fixed: removed the `@deprecated` tag from `githubDeployKeySecret` for the same reason.

src/release/publisher.ts
mrgrain: Fixed: `githubCredentials` now has `@default - no GitHub App credentials used` and the doc comment just states the mutual-exclusivity constraint without re-explaining what it replaces.

src/release/publisher.ts
mrgrain: Fixed: added a second check that throws when `githubUseSsh` and `githubTokenSecret` are both set, since they're incompatible auth methods (SSH vs. secret-based token). Added a test for it.

rix0rrr (approved):

PR #4868, opened 2026-08-15
feat(cdktn): add typescript app

eduardomourar:
### Summary

Adds a new `cdktn-app-ts` (`CdktnTypeScriptApp`) project type for scaffolding CDK Terrain (CDKTN) applications in TypeScript, mirroring how `awscdk-app-ts` supports AWS CDK apps.

### Details

- `CdktnTypeScriptApp`: TypeScript CDKTN app project (`src/cdktn/cdktn-app-ts.ts`), including a cdktnConfig (cdktf.json) component, cdktnTasks for synth, and cdktnDeps dependency management.
- `CdktnConfig`: manages cdktf.json generation, shared between apps and constructs (`src/cdktn/cdktn-config.ts`).
- `CdktnDeps`: shared CDKTN dependency-version resolution logic reused by both the app and construct project types.
- `CdktnTasks`: adds synth/synth:silent tasks wired into the build for the new app type.

---


eduardomourar:
@mrgrain, this one is ready for your review.

src/cdktn/cdktn-app-ts.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    this.gitignore.exclude(".terraform/");
+    this.gitignore.exclude(".terraform.lock.hcl");
+    this.gitignore.exclude("*.tfstate*");
+    this.gitignore.exclude("*.tfvars*");
+
+    this.npmignore?.exclude(`${this.cdktnConfig.cdktnOut}/`);
+    this.npmignore?.exclude(".gen/");
+
+    this.addDevDeps("ts-node");
```
mrgrain: do we really want to default this project to ts-node? [It doesn't work with TS7 anymore](https://github.com/TypeStrong/ts-node/issues/2174).

src/cdktn/cdktn-app-ts.ts
```diff
@@ -0,0 +1,204 @@
...
+    if (options.app) {
+      return options.app;
+    }
+
+    const appEntrypoint = path.posix.join(this.srcdir, this.appEntrypoint);
+
+    const tsNodeConfig = this.tsconfig?.fileName
+      ? ` -P ${this.tsconfig?.fileName}`
+      : "";
+    return `npx ts-node${tsNodeConfig} --prefer-ts-exts ${appEntrypoint}`;
```
mrgrain: This should use the new ScriptRunner API.

src/cdktn/cdktn-app-ts.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+}
+
+class SampleCode extends Component {
```
mrgrain, line 144: There are new lifecycle hooks that might be more appropriate: https://projen.io/docs/custom/custom-components#lifecycle-hooks

Specifically `postProjectCreation` as a one off to add the example.

src/cdktn/cdktn-config.ts
```diff
@@ -0,0 +1,133 @@
...
+ * Represents cdktf.json file.
+ *
+ * CDKTN (CDK Terrain) continues to use the `cdktf.json` filename for
+ * backwards compatibility with its CDKTF origins.
+ */
+export class CdktnConfig extends Component {
+  /**
+   * Represents the JSON file.
+   */
+  public readonly json: JsonFile;
```
mrgrain:
```suggestion
  public readonly file: JsonFile;
```

src/cdktn/cdktn-config.ts
```diff
@@ -0,0 +1,133 @@
...
+    super(project);
+
+    this.cdktnOut = options.cdktnOut ?? "cdktf.out";
+    this._terraformProviders = options.terraformProviders ?? [];
+    this._terraformModules = options.terraformModules ?? [];
+
+    this.json = new JsonFile(project, "cdktf.json", {
+      omitEmpty: true,
+      // `cdktn get` and `cdktn provider add` write generated provider/module
+      // entries back into this file, so it can't be read-only.
```
mrgrain, line 109: This currently overrides `terraformProviders` and `terraformModules` and any `cdktn provider add` etc will be lost. We need to make this a hybrid file that reads the current state for `terraformProviders` and `terraformModules` and merges.

src/cdktn/cdktn-tasks.ts
mrgrain, line 1: Why should this be a separate component?

src/cdktn/cdktn-config.ts
```diff
@@ -0,0 +1,133 @@
...
+   * List of Terraform providers to build.
+   */
+  private readonly _terraformProviders: string[];
+
+  /**
+   * List of Terraform modules to build.
+   */
+  private readonly _terraformModules: string[];
+
+  constructor(project: Project, options: CdktnConfigOptions) {
```
mrgrain: Should take any scope. Use `this.project` to access the project.

```suggestion
  constructor(scope: IConstruct, options: CdktnConfigOptions) {
```

src/cdktn/cdktn-config.ts
```diff
@@ -0,0 +1,133 @@
...
+    this._terraformProviders = options.terraformProviders ?? [];
+    this._terraformModules = options.terraformModules ?? [];
+
+    this.json = new JsonFile(project, "cdktf.json", {
+      omitEmpty: true,
+      // `cdktn get` and `cdktn provider add` write generated provider/module
+      // entries back into this file, so it can't be read-only.
+      readonly: false,
+      obj: {
+        language: "typescript",
```
mrgrain: Why not configurable?

src/cdktn/cdktn-app-ts.ts
```diff
@@ -0,0 +1,204 @@
...
+    });
+
+    this.cdktnDeps = new CdktnDepsJs(this, {
+      dependencyType: DependencyType.RUNTIME,
+      ...options,
+    });
+    this.appEntrypoint = options.appEntrypoint ?? "main.ts";
+
+    // CLI
+    this.addDevDeps(`cdktn-cli@${this.cdktnDeps.cdktnCliVersion}`);
```
mrgrain: Why not in `CdktnDepsJs?` or `CdktnDeps`?

mrgrain: Consider using [requestDependency](https://projen.io/docs/api/projen/?_highlight=requestDependency#projen.Dependencies.requestDependency) instead.

src/cdktn/cdktn-deps.ts
```diff
@@ -0,0 +1,166 @@
...
+      requestedVersion ?? defaultConstructsVersion
+    }`;
+
+    // First remove the version added by projen
+    this.project.deps.removeDependency("constructs", DependencyType.BUILD);
+
+    this.project.deps.addDependency(
+      `${this._packageNames.constructs}@${versionRequirement}`,
+      this.dependencyType,
+    );
```
mrgrain, line 149: Consider using [requestDependency](https://projen.io/docs/api/projen/?_highlight=requestDependency#projen.Dependencies.requestDependency) instead.

src/cdktn/cdktn-deps.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Adds a dependency on the CDKTN framework package (e.g. `cdktn`).
+   */
+  private addFrameworkDependency() {
+    this.project.deps.addDependency(
+      `${this._packageNames.core}@${this.cdktnVersion}`,
+      this.dependencyType,
+    );
```
mrgrain: Consider using [requestDependency](https://projen.io/docs/api/projen/?_highlight=requestDependency#projen.Dependencies.requestDependency) instead.

src/cdktn/cdktn-app-ts.ts
eduardomourar: We don't. (I think I also missed this issue here: https://github.com/projen/projen/issues/4585). I will evaluate to use node natively with `--experimental-transform-types` flag as a runner.

src/cdktn/cdktn-deps.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Fully qualified name of the constructs library package.
+   */
+  readonly constructs: string;
+}
+
+/**
+ * Manages dependencies on CDK Terrain (CDKTN).
+ */
+export abstract class CdktnDeps extends Component {
```
mrgrain: Do we need this? what is this giving us? I know this is mimicking CDK App, but there we really only have it because of CDKv1 and how complex that was ....

mrgrain (changes requested):

src/cdktn/cdktn-config.ts
eduardomourar: Perfect. I will make that modification.

src/cdktn/cdktn-deps.ts
eduardomourar: Probably not. I will have a more broader look and will revisit the AWS CDK approach (here and in other parts of the code) to see if matches the needs for CDKTN.

mrgrain: Sounds good. To be clear: Let's not change the existing AWS CDK project. I just want to take the opportunity of a new project (CDKTN APP) to do better.

src/cdktn/cdktn-tasks.ts
eduardomourar: Just following the pattern from before, but I will check if that is relevant still.

src/cdktn/cdktn-deps.ts
eduardomourar: After review (I know I sound like a referee in a football match), I believe the approach used in the AWS CDK is better for the long-term support of multiple languages. Just for you to know, I plan on supporting: Python and Rust. With that in mind, the `cdktn-tasks` and `cdktn-deps` components are important (as they are language-agnostic) so other people can start working on other languages in parallel.

eduardomourar: One thing that I will still do is merge `cdktn-deps-js` into `cdktn-deps` though. Do want me to mark those components you questioned with `@experimental`? That way until other languages are introduced breaking changes can happen.

mrgrain: Thanks for looking into it! We don't need to mark, we can just issue a breaking change if needed.

mrgrain (approved):

PR #4874, opened 2026-08-20
feat(github): global runs-on for workflows

### LINKED ISSUE #2027: Bug: workflowRunsOn property doesn't update all Github Actions workflow files
Adding this setting `workflowRunsOn: ['custom-runner'],` to the `.projenrc.js` only updates the `build.yml` workflow file. The other two, `pull-request-lint.yml` and `upgrade.yml` are not changed and remain as `ubuntu-latest`.

```
const { awscdk } = require('projen');
const project = new awscdk.AwsCdkTypeScriptApp({
  cdkVersion: '2.1.0',
  defaultReleaseBranch: 'main',
  name: 'MyName',
  workflowRunsOn: ['custom-runner'],

  // deps: [],                /* Runtime dependencies of this module. */
  // description: undefined,  /* The description is just a string that helps people understand the purpose of the package. */
  // devDeps: [],             /* Build dependencies for this module. */
  // packageName: undefined,  /* The "name" in package.json. */
});
project.synth();
```

eduardomourar:
Fixes #2027

Adds a global `workflowRunsOn` / `workflowRunsOnGroup` property to `GitHubOptions` that propagates to all workflow-generating components as their default runner configuration.

Previously, users had to manually set `runsOn` on each component individually (auto-approve, auto-queue, stale, pull-request-lint, dependency-review, upgrade-dependencies, backport, release). Now a single project-level setting flows everywhere, while each component still supports its own `runsOn` / `runsOnGroup` to override when needed.

```typescript
  githubOptions: {
    workflowRunsOn: ['self-hosted', 'linux'],
  },
  // all workflows use this runner by default
  autoApproveOptions: {
    runsOn: ['ubuntu-latest'], // override for a specific workflow
  },
});
```

Components updated:
- AutoApprove
- AutoQueue (also added `runsOnGroup` option)
- PullRequestLint
- DependencyReview
- Stale
- PullRequestBackport (added `runsOn` / `runsOnGroup` options, previously hardcoded)
- UpgradeDependencies
- Release (added `runsOn` / `runsOnGroup` options)
- Build workflow (already supported, no change)

`workflowRunsOn` and `workflowRunsOnGroup` have been deprecated from `ReleaseProjectOptions`. The canonical location is now `GitHubOptions`. Use the project-level `workflowRunsOn` or the new `runsOn` / `runsOnGroup` on `ReleaseOptions` for release-specific overrides.

---

### DIFF
--- a/src/cdk/jsii-project.ts
+++ b/src/cdk/jsii-project.ts
@@ -1,5 +1,6 @@
 import type { ValidateTsconfig } from "./jsii-build";
 import { JsiiBuild } from "./jsii-build";
+import { GitHub } from "../github";
 import type {
   GoPublishOptions,
   MavenPublishOptions,
@@ -263,17 +264,17 @@ export class JsiiProject extends TypeScriptProject {
   }
 
   /**
-   * Generates the runs-on config for Jobs.
-   * Throws error if 'runsOn' and 'runsOnGroup' are both set.
-   *
-   * @param options - 'runsOn' or 'runsOnGroup'.
+   * Generates the runs-on config for Jobs, falling back to the project's
+   * global GitHub runner selection (if any).
    */
   private getJobRunsOnConfig(options: JsiiProjectOptions) {
-    return options.workflowRunsOnGroup
-      ? { runsOnGroup: options.workflowRunsOnGroup }
-      : options.workflowRunsOn
-        ? { runsOn: options.workflowRunsOn }
-        : {};
+    const github = this.github ?? GitHub.of(this.root);
+    return (
+      github?.runsOnConfig({
+        runsOn: options.workflowRunsOn,
+        runsOnGroup: options.workflowRunsOnGroup,
+      }) ?? {}
+    );
   }
 }
 
--- a/src/github/auto-queue.ts
+++ b/src/github/auto-queue.ts
@@ -3,6 +3,7 @@ import { GitHub } from "./github";
 import type { GithubCredentials } from "./github-credentials";
 import * as workflows from "./workflows-model";
 import { Component } from "../component";
+import type { RunsOnOptions } from "../runner-options";
 
 /**
  * The merge method used to add the PR to the merge queue
@@ -18,7 +19,7 @@ export enum MergeMethod {
 /**
  * Options for 'AutoQueue'
  */
-export interface AutoQueueOptions {
+export interface AutoQueueOptions extends RunsOnOptions {
   /**
    * Only pull requests authored by these Github usernames will have auto-queue enabled.
    * @default - pull requests from all users are eligible for auto-queuing
@@ -49,12 +50,6 @@ export interface AutoQueueOptions {
    * @default MergeMethod.SQUASH
    */
   readonly mergeMethod?: MergeMethod;
-
-  /**
-   * Github Runner selection labels
-   * @default ["ubuntu-latest"]
-   */
-  readonly runsOn?: string[];
 
   /**
    * The branch names that we should auto-queue for
@@ -166,7 +161,7 @@ export class AutoQueue extends Component {
 
     const autoQueueJob: workflows.Job = {
       name: "Set AutoQueue on PR #${{ github.event.number }}",
-      runsOn: options.runsOn ?? ["ubuntu-latest"],
+      ...workflowEngine.runsOnConfig(options),
       permissions: {
         pullRequests: workflows.JobPermission.WRITE,
         contents: workflows.JobPermission.WRITE,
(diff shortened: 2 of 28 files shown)


mrgrain (changes requested):
We are trying to get rid of `GithubProject`, so no new features for it I am afraid ([see RFC](https://github.com/projen/projen/blob/main/rfcs/repository-tree-root.md))!

But I think you can just re-anchor this on the `Github` component and respective options.

eduardomourar:
Thanks. I think I missed that RFC! I will make the changes accordingly and keep `GithubProject` without any modification.

src/github/github.ts
```diff
@@ -170,6 +188,18 @@ export class GitHub extends Component {
...
+   * @internal
+   */
+  public readonly _workflowRunsOn?: string[];
+
+  /**
+   * The default runner group options for all workflows in this project.
+   * @internal
+   */
+  public readonly _workflowRunsOnGroup?: GroupRunnerOptions;
+
```
mrgrain, line 197: I think instead of that we can have a public helper method to retrieve the run configuration. this method can take an optional object with runsOn/runsOnGroup.

Then all these

```ts
        ...filteredRunsOnOptions(
          options.runsOn ?? github._workflowRunsOn,
          options.runsOnGroup ?? github._workflowRunsOnGroup,
        ),
```

can become something like

```ts
        ...github.runsOnConfig(options),
```

and the internal values don't need to be public/`@internal`.

src/github/auto-queue.ts
```diff
@@ -56,6 +58,13 @@ export interface AutoQueueOptions {
    */
   readonly runsOn?: string[];
 
+  /**
+   * Github Runner Group selection options
+   * @description Defines a target Runner Group by name and/or labels
+   * @throws {Error} if both `runsOn` and `runsOnGroup` are specified
+   */
+  readonly runsOnGroup?: GroupRunnerOptions;
```
mrgrain: Could this be struct that all of these actions just extend on?

src/release/release.ts
```diff
@@ -194,21 +194,6 @@ export interface ReleaseProjectOptions {
...
-   */
-  readonly workflowRunsOn?: string[];
-
-  /**
-   * Github Runner Group selection options
-   * @description Defines a target Runner Group by name and/or labels
-   * @throws {Error} if both `runsOn` and `runsOnGroup` are specified
-   */
-  readonly workflowRunsOnGroup?: GroupRunnerOptions;
-
```
mrgrain, line 209: could we keep this and deprecate to minimize impact and make it non breaking?

mrgrain (changes requested):

src/release/release.ts
eduardomourar: Done. Do you want me to also mark `workflowRunsOn/workflowRunsOnGroup` from `GitHubProject` as deprecated as well?

mrgrain: Oh yes, that'd be great actually! Another one gone from GitHubProject!

src/release/release.ts
```diff
@@ -395,6 +412,18 @@ export class Release extends Component {
...
+      runsOnGroup: options.runsOnGroup ?? options.workflowRunsOnGroup,
+    });
+    this.runsOn =
+      runsOnConfig && "runsOn" in runsOnConfig
+        ? runsOnConfig.runsOn
+        : undefined;
+    this.runsOnGroup =
+      runsOnConfig && "runsOnGroup" in runsOnConfig
+        ? runsOnConfig.runsOnGroup
+        : undefined;
```
mrgrain: Not just this?
```suggestion
    this.runsOn = runsOnConfig?.runsOn;
    this.runsOnGroup = runsOnConfig?.runsOnGroup
```

src/cdk/jsii-project.ts
```diff
@@ -263,17 +264,12 @@ export class JsiiProject extends TypeScriptProject {
...
    */
-  private getJobRunsOnConfig(options: JsiiProjectOptions) {
-    return options.workflowRunsOnGroup
-      ? { runsOnGroup: options.workflowRunsOnGroup }
-      : options.workflowRunsOn
-        ? { runsOn: options.workflowRunsOn }
-        : {};
+  private getJobRunsOnConfig() {
+    const github = this.github ?? GitHub.of(this.root);
+    return github?.runsOnConfig() ?? {};
```
mrgrain: This seems like we are dropping the functionality of `options.workflowRunsOn` and `options.workflowRunsOnGroup` which still exists, they are just deprecated?

mrgrain (changes requested):
Nice one! Almost there.

src/cdk/jsii-project.ts
eduardomourar: My bad. I have put back in order to preserve support for deprecated options.

mrgrain (approved):

PR #4877, opened 2026-08-20
feat: add support for sonarqube properties

eduardomourar:
Adds SonarQube support to projen, generating a `sonar-project.properties` file from typed, strongly-documented options mapped to the `sonar.*` dot-notation keys (scm, coverage, exclusions, log level, region, etc.).

- `PropertiesFile` (`src/properties-file.ts`): a new `ObjectFile` subclass for Java `.properties` files, serialized via the `properties-file` package for spec-compliant escaping. Nested objects flatten into dot-separated keys (e.g. `{ sonar: { sources: "src" } }` → `sonar.sources=src`).
- `SonarqubeProperties` (`src/sonarqube.ts`): base component that renders a `sonar-project.properties` file from `SonarqubePropertiesOptions`, plus `SonarqubeJavascriptProperties`, `SonarqubeTypescriptProperties`, and `SonarqubeRustProperties` subclasses with language-specific defaults.

---


mrgrain (commented):
How about using this? https://www.npmjs.com/package/properties-file

mrgrain (commented):
any chance there's an upstream, maintained, machine parseable source for this?

eduardomourar:
I have only found the parameters documentation [here](https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/analysis-parameters/parameters-not-settable-in-ui) and the sonar-scanner source code [here](https://github.com/SonarSource/sonar-scanner-msbuild/blob/2695104610c44914edb6c3e0a09871f3daa09e38/src/SonarScanner.MSBuild.Common/Telemetry/TelemetryUtils.cs), but nothing parseable that can be easily used.

src/sonarqube/sonarqube.ts
mrgrain, line 1: move this into a submodule please

src/properties-file.ts
```diff
@@ -0,0 +1,133 @@
...
+ * `{ sonar: { sources: "src" } }` produces `sonar.sources=src`.
+ *
+ * This means `addOverride` uses dot notation as a path separator,
+ * consistent with ObjectFile/JsonPatch semantics.
+ *
+ * Serialization leverages the `properties-file` package for full Java
+ * `.properties` specification compliance (proper escaping of special
+ * characters in keys and values).
+ *
+ * @see https://www.npmjs.com/package/properties-file
```
mrgrain:
```suggestion
```

mrgrain (changes requested):

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+   * The key uses dot notation as a nested path separator, matching the
+   * ObjectFile override semantics. For example, `"sonar.coverage.exclusions"`
+   * navigates to `{ sonar: { coverage: { exclusions: value } } }`.
+   *
+   * @param key The property path in dot notation (e.g., `sonar.java.binaries`)
+   * @param value The property value
+   */
+  public addProperty(key: string, value: string) {
+    this.file.addOverride(key, value);
+  }
```
mrgrain: This seems redundant.

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Language-specific presets
+// ---------------------------------------------------------------------------
```
mrgrain: let's split these out into separate files if we can

src/properties-file.ts
```diff
@@ -0,0 +1,127 @@
+import type { IConstruct } from "constructs";
+// eslint-disable-next-line import/no-unresolved
```
mrgrain: why?

mrgrain (changes requested):

src/sonarqube/sonarqube.ts
eduardomourar: I will try to see if makes sense to have one per language. I think it does.

src/properties-file.ts
```diff
@@ -0,0 +1,127 @@
...
+  ): Array<[string, string]> {
+    const entries: Array<[string, string]> = [];
+    for (const [key, value] of Object.entries(obj)) {
+      const fullKey = prefix ? `${prefix}.${key}` : key;
+      if (value === undefined || value === null) {
+        continue;
+      } else if (typeof value === "object" && !Array.isArray(value)) {
+        entries.push(...PropertiesFile.flatten(value, fullKey));
+      } else {
+        entries.push([fullKey, String(value)]);
```
mrgrain: Array handling here is accidental rather than designed: arrays fall through to `String(value)`, so `["a","b"]` → `"a,b"` only because `Array.prototype.toString` happens to comma-join. It also means `[{x:1}]` silently becomes `[object Object]` and `[null,"a"]` becomes `",a"`. The docstring only mentions strings/numbers/booleans as leaves.

Since comma-joined lists are genuinely useful for properties files, I'd make this explicit: `value.join(",")` for arrays (maybe throwing on non-scalar elements) and document it. Or throw on arrays entirely.

src/properties-file.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+export class PropertiesFile extends ObjectFile {
+  /**
+   * Recursively flattens a nested object into dot-separated key/value pairs.
+   *
+   * For example, `{ sonar: { sources: "src" } }` becomes `[["sonar.sources", "src"]]`.
+   * Leaf values (strings, numbers, booleans) are emitted as entries.
+   * Null/undefined values are skipped.
+   */
+  private static flatten(
```
mrgrain: Flattening is lossy, so key collisions are silent: `{ "a.b": "1", a: { b: "2" } }` produces two `a.b` entries and whichever `editor.insert` runs last wins. Not hypothetical — the tests use literal dotted keys while `SonarqubeProperties` uses nesting, so both styles are sanctioned and can collide. Worth detecting duplicates here and throwing (or at least documenting last-wins).

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+      if (Object.keys(qg).length > 0) {
+        sonar.qualitygate = qg;
+      }
+    }
+
+    // Extra properties (user-specified arbitrary properties).
+    if (options.extraProperties) {
+      for (const [key, value] of Object.entries(options.extraProperties)) {
+        const parts = key.split(".");
+        if (parts[0] === "sonar") {
```
mrgrain: I think this whole sonar-prefix branch can go — it hand-rolls what `ObjectFile.addOverride` already does (split on dots, merge into the synthesized object). Both branches collapse to:

```ts
for (const [key, value] of Object.entries(options.extraProperties ?? {})) {
  this.file.addOverride(key, value);
}
```

after the file is created. Less code, one set of path semantics, same output.

The hand-rolled version also has a few bugs the shared code doesn't:
- `target[parts[i]] = {}` silently clobbers existing scalars — e.g. `"sonar.projectKey.foo": "x"` wipes the mandatory `projectKey`.
- A key of exactly `"sonar"` ends up emitting `sonar.sonar=...` (loop doesn't run, then `target["sonar"] = value` on the sonar object itself).
- Inconsistent escaping: sonar keys are split naively on `.`, while non-sonar keys go through `addOverride` which honors `\\.` escapes — same option, two path grammars.

Also worth a line in the `extraProperties` docs that they override the typed options (a leaf like `"sonar.coverage"` replaces the whole coverage subtree).

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+
+    // Project identification (mandatory)
+    sonar.projectKey = options.projectKey;
+
+    if (options.organization) {
+      sonar.organization = options.organization;
+    }
+
+    // Server connection
+    if (options.region === SonarqubeRegion.US) {
```
mrgrain: The empty-string enum member plus this `=== SonarqubeRegion.US` check feels brittle — adding a region later means touching this code, and I'd double check how an empty-string enum value renders in jsii-generated Java/Python. Simpler: give EU a real value and just do `if (options.region) sonar.region = options.region;` (or drop the EU member entirely since it's "don't emit").

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+
+    // Quality gate (sonar.qualitygate.*)
+    if (options.qualitygate) {
+      const qg: Record<string, any> = {};
+      if (options.qualitygate.wait) {
+        qg.wait = "true";
+      }
+      if (
+        options.qualitygate.timeout !== undefined &&
+        options.qualitygate.timeout !== 300
```
mrgrain: Silently dropping an explicit `timeout: 300` (same for `log.level: INFO` and `scm.exclusions.disabled: false`) loses user intent — if SonarQube ever changes a server-side default, the config no longer says what the user asked for. Meanwhile the presets happily emit defaults, so the philosophy is inconsistent. I'd emit whatever the user explicitly set and only skip `undefined`.

src/sonarqube/sonarqube.ts
```diff
@@ -0,0 +1,844 @@
...
+        "**/test/**",
+        "**/__tests__/**",
+        "**/*.test.ts",
+        "**/*.spec.ts",
+      ],
+      coverage: { exclusions: ["**/test/**", "**/__tests__/**"] },
+      cpd: { exclusions: ["**/test/**/*.json", "**/__tests__/**/*.json"] },
+      javascript: { lcov: { reportPaths: ["coverage/lcov.info"] } },
+      // User options override defaults
+      ...options,
```
mrgrain: Heads up that this spread is shallow, so partial nested options wipe entire default subtrees: `new SonarqubeTypescriptProperties(p, { coverage: {} })` loses the default coverage exclusions, and any `javascript: {...}` override drops the lcov default. Same applies to the other two presets. Either deep-merge or call the replace semantics out loudly in the docs.

mrgrain (commented):
Did a pass over this with my AI agent (comments below were drafted by it on my behalf, I've reviewed them). Really nice addition overall — the typed option tree mirroring the sonar.* namespaces is neat. A few things worth tightening up before this lands, mostly around `flatten` and the `extraProperties` handling.

src/properties-file.ts
eduardomourar: Library eslint-import-resolver-typescript@2.x can't follow package.json "exports" subpaths, so it can't resolve "properties-file/editor" even though it exists. I have bumped eslint-import-resolver-typescript to v4 and it is resolving properly now.

src/sonarqube/sonarqube.ts
eduardomourar: Using deep-merge as suggested.

src/sonarqube/sonarqube.ts
eduardomourar: I gave `EU` real value of `eu`.

src/sonarqube/sonarqube.ts
eduardomourar: Nice catch this one, so modified accordingly.

src/properties-file.ts
```diff
@@ -0,0 +1,144 @@
...
```
mrgrain: We prefer helper functions over private class methods. In this case I think we should even move it into a more generic private utils place since the nested to dot notation conversion is somewhat universal. Will also allow much easier test cases. Sorry I missed that in the first pass.

mrgrain (changes requested):
Almost there! Just a small code org change.

src/properties-file.ts
eduardomourar: As pointed out, I have extracted the nested-to-dot-notation flattening logic out of PropertiesFile as a private static method and into util.ts as a standalone `flattenToDotNotation` function, since the conversion is generic and easier to unit test directly.

mrgrain (approved):

PR #4889, opened 2026-08-22
feat: add support for polaris coverity

eduardomourar:
Adds a new PolarisCoverity component (src/polaris/coverity.ts) that generates a Black Duck Coverity on Polaris configuration file (coverity.yaml) for static analysis, with support for:

- Capture file filtering (include/exclude regex)
- Language filtering for capture
- Standard Polaris configuration options per the official schema (https://docs.blackduck.com/r/cov_polaris/latest/coverity-on-polaris/configuration-file-schema.html)

---


src/polaris/coverity.ts
```diff
@@ -0,0 +1,449 @@
...
+        },
+        languages: capture.languages,
+        "compiler-configuration": capture.compilerConfiguration && {
+          "cov-configure": capture.compilerConfiguration.covConfigure,
+        },
+      },
+      install: this.options.install,
+      analyze: this.options.analyze,
+    };
+  }
```
mrgrain, line 43: There doesn't seem to be any reason for lazy rendering here. We prefer the object to be passed directly.

mrgrain (changes requested):

src/polaris/coverity.ts
eduardomourar: I made the modification to be passed directly, but there are are feel properties that are kebab-case so I still have properly format them.

mrgrain: Okay, but this then tells me this is going to me really hard to maintain. Why do we have different names for options? E.g. our `capture.buildCapture` vs their `capture.build`? We already have a helper to convert object key casing: `kebabCaseKeys` should do it. If we need different case, we can add another helper.

Similar for defaults. There's a `deepMerge` that we can use to set defaults.

Unless there's a strong reason not to, I'd generally prefer these automatic approaches. 


PS: I know this isn't well documented. Feel free to add to the agent docs as you see fit.

mrgrain (changes requested):

src/polaris/coverity.ts
eduardomourar: I found the json schema that can be used to generate this here: https://raw.githubusercontent.com/mongodb/mongo/refs/heads/master/etc/coverity_config_schema.json

mrgrain (approved):

PR #4890, opened 2026-08-22
feat(javascript): add support for nodejs native test runner

eduardomourar:
Adds support for Node.js's built-in test runner (`node --test`) as an alternative to Jest, mirroring the ergonomics of the existing `Jest` component.

- New `JavaScriptTestRunner` abstraction (`src/javascript/javascript-test-runner.ts`) with static factories `JavaScriptTestRunner.useJest(options)` and `JavaScriptTestRunner.useNode(options)`, passed via a new `testRunner` option on `NodeProjectOptions`. It is mutually exclusive with the deprecated `jest`/`jestOptions` options.
 - New `NodeNativeTest` component (`src/javascript/node-native-test.ts`), the Node-backed implementation, mirroring `Jest`'s ergonomics:
    - Wires `test`, `test:watch`, and (when `updateSnapshot: UpdateSnapshot.NEVER`) `test:update` npm scripts.
    - Configuration (coverage, reporters, global setup, module mocks, extra `nodeOptions`, etc.) is written to a generated Node.js config file (`node.config.json` by default) loaded via `--experimental-config-file`, using the `NodeConfigFile`/config-schema support added in #4893.
    - Coverage collection uses `--experimental-test-coverage` with `lcov` and `junit` reporters; since neither reporter creates its output directory, the `test` task now prepends `mkdir -p`.
    - `addTestMatch()` lets callers append test-file glob patterns after construction.
- Extracted `JavaScriptTestRunnerOptions` (`src/javascript/test-runner-options.ts`) — the option surface shared by `Jest` and `NodeNativeTest` (`extraCliOptions`, `updateSnapshot`, `coverageText`, `junitReporting`, `preserveDefaultReporters`, `configFilePath`) — plus a shared `configureTestCommand()` helper that wires the `test`/`test:watch`/`test:update` tasks consistently for both runners.
- `Jest` now also accepts `JavaScriptTestRunnerOptions`, so both runners are configured through a consistent options shape.

---


src/javascript/node-test-runner.ts
```diff
@@ -0,0 +1,254 @@
...
+      "});",
+      "",
+      "// Serialize snapshot matching Jest's pretty-format output.",
+      `${serializeSnapshotValue.toString()}`,
+      "",
+      "snapshot.setDefaultSnapshotSerializers([serializeSnapshotValue]);",
+    ]) {
+      snapshotSetup.addLine(line);
+    }
+    this.project.npmignore?.addPatterns(`/${snapshotSetup.path}`);
```
mrgrain: I don't think we want to maintain this implementation. It feels like there should be 3rd party packages that can be added as `nodeOptions.import` to set up custom snapshotting. If not, I'm inclined to say that we just go with Nodes default format. Users can handle a one-off migration.

src/javascript/node-test-runner.ts
```diff
@@ -0,0 +1,254 @@
...
+          `./${snapshotSetup.path}`,
+        ],
+      },
+    };
+
+    const configFilePath =
+      options.configFilePath ?? "node.test-coverage-config.json";
+    this.file = new JsonFile(this.project, configFilePath, {
+      obj: () => toJson_NodeConfigSchema(this.config),
+    });
```
mrgrain: node config should be its own component and probably best to have as a separate PR

src/javascript/node-test-runner.ts
```diff
@@ -0,0 +1,254 @@
...
+        ...options.nodeOptions,
+        import: [
+          ...(options.nodeOptions?.import ?? []),
+          `./${snapshotSetup.path}`,
+        ],
+      },
+    };
+
+    const configFilePath =
+      options.configFilePath ?? "node.test-coverage-config.json";
```
mrgrain: `test-coverage-config` or just `test-config` or whatever?

src/javascript/node-project.ts
```diff
@@ -312,6 +314,21 @@ export interface NodeProjectOptions
...
+   *
+   * @default false
+   */
+  readonly nodeTestRunner?: boolean;
+
+  /**
+   * Node test runner options
+   * @default - default options
+   */
+  readonly nodeTestRunnerOptions?: NodeTestRunnerOptions;
```
mrgrain: I propose we instead make this a FutureComponent like `ScriptRunner` so that we can eventually also add Vitest and migrate Jest to a single property.


```suggestion
  /**
   * Configure the test runner to execute unit tests. 
   *
   * @default NodeTestRunner.jest()
   */
  readonly testRunner?: INodeTestRunner;
```

Or maybe `IJavaScriptTestRunner` not sure.

src/javascript/test-runner-base.ts
mrgrain, line 1: This feels like it's way too much shared functionality and also carrying over legacy jest features that we probably wouldn't implement like this anymore. I suggest we start and design from scratch with the native test runner and then see what overlap we have with jest.

src/javascript/node-test-runner.ts
mrgrain: E.g. [@simonegianni/node-test-jest-compat](https://www.npmjs.com/package/@simonegianni/node-test-jest-compat) 

Not that I think we should include this by default, but I think if users want compatibility with their existing jest code they can do this themselves.

mrgrain (changes requested):
Some directional changes.

src/javascript/node-test-runner.ts
eduardomourar: For the `setResolveSnapshotPath`, I would agree with you that there is no need for projen to keep it. For the `setDefaultSnapshotSerializers`, no. As of Node.js v24.19, the snapshot content is formatted in a single line which makes it unusable for reviewing code changes (Node core might change that  behavior in the future).

src/javascript/node-test-runner.ts
eduardomourar: I will set it to the default filename `node.config.json` as per documentation [here](https://nodejs.org/docs/latest-v24.x/api/cli.html#--experimental-config-filepath---experimental-config-file).

src/javascript/node-test-runner.ts
eduardomourar: I will create a separate PR to move in its own component. I had to add this as part of this change because we have multiple experimental properties in use and it is hard to keep track of that through the command line arguments. Is there a easy way to use the schema to set the argument names?

src/javascript/node-project.ts
eduardomourar: As suggested, I have created a `IJavaScriptTestRunner` for this property.

src/javascript/test-runner-base.ts
eduardomourar: Removed the abstract class. Just moved some of the shared options to the `JavaScriptTestRunnerOptions` in order to keep them in sync.

src/javascript/node-test-runner.ts
eduardomourar: I tried out [@simonegianni/node-test-jest-compat](https://www.npmjs.com/package/@simonegianni/node-test-jest-compat). Unfortunately, it aligns the paths similar to Jest, but it does not change the serializer.

mrgrain:
FYI I am waiting for the config PR first

src/javascript/node-test-runner.ts
mrgrain: I still don't really want to maintain this kind of code in projen. We can document how to change it or find a different package that does that. Okay to add a mini-shim to add to import.

eduardomourar:
There you go: https://github.com/projen/projen/pull/4893

src/javascript/node-test-runner.ts
eduardomourar: Agreed. So I will actually remove this shim altogether. I have published a library called [pretty-node-snapshot](https://www.npmjs.com/package/pretty-node-snapshot) that will do exactly what people will need when migrating from Jest (like I am doing right now). This will solve my direct need and without being part of projen itself.

src/javascript/jest.ts
mrgrain, line 1: I'm always a bit suspicious on these seemingly unrelated changes. Aside from the options changes, none of this should be needed right? If this is a generic refactor let's move it into a separate PR.

src/javascript/javascript-test-runner.ts
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * The underlying `Jest` component, if this runner is backed by Jest.
+   */
+  readonly jest?: Jest;
+
+  /**
+   * The underlying `NodeNativeTest` component, if this runner is backed by
+   * Node.js' built-in test runner.
+   */
+  readonly nodeNativeTest?: NodeNativeTest;
```
mrgrain: this isn't something that should be in this interface. Instead `Jest` and `NodeNativeTest` should implement `IJavaScriptTestRunner` as needed.

src/javascript/javascript-test-runner.ts
```diff
@@ -0,0 +1,110 @@
...
+ * it (typically `NodeProject` itself, via its `testRunner` option).
+ */
+export class JavaScriptTestRunner
+  extends FutureComponent
+  implements IJavaScriptTestRunner
+{
+  /**
+   * Use Jest to execute unit tests.
+   */
+  public static useJest(options: JestOptions = {}): JavaScriptTestRunner {
```
mrgrain: Why `useJest`?

```suggestion
  public static jest(options: JestOptions = {}): JavaScriptTestRunner {
```

src/javascript/node-project.ts
```diff
@@ -302,16 +302,27 @@ export interface NodeProjectOptions
...
+   *
+   * @deprecated Use `testRunner: JavaScriptTestRunner.useJest(jestOptions)` instead.
    * @default - default options
    */
   readonly jestOptions?: JestOptions;
 
+  /**
+   * The test runner used to execute unit tests.
+   *
+   * @default - `JavaScriptTestRunner.useJest()`, unless `jest` is set to `false`
```
mrgrain:
```suggestion
   * @default - `JavaScriptTestRunner.useJest()`, unless `jest` is set to `false`, then none
```

src/javascript/node-project.ts
```diff
@@ -509,8 +520,15 @@ export class NodeProject extends GitHubProject {
    */
   public readonly runScriptCommand: string;
 
+  /**
+   * The test runner used to execute unit tests.
+   */
+  public readonly testRunner?: JavaScriptTestRunner;
```
mrgrain:
```suggestion
  public readonly testRunner?: IJavaScriptTestRunner;
```

src/javascript/node-project.ts
```diff
@@ -923,6 +948,23 @@ export class NodeProject extends GitHubProject {
...
+   * the deprecated `jest`/`jestOptions` options.
+   */
+  private resolveTestRunner(
+    options: NodeProjectOptions,
+  ): JavaScriptTestRunner | undefined {
+    if (options.testRunner) {
+      return options.testRunner;
+    }
+
+    const jestEnabled = options.jest ?? true;
```
mrgrain, line 965:
```suggestion
    // @deprecated
    const jestEnabled = options.jest ?? true;
```

src/javascript/node-project.ts
```diff
@@ -951,8 +993,11 @@ export class NodeProject extends GitHubProject {
 
   private useCodecov(options: NodeProjectOptions): boolean {
     // Use Codecov when it is enabled or if or a secret token name is passed in
-    // AND jest must be configured
-    return (options.codeCov || options.codeCovTokenSecret) && this.jest?.config;
+    // AND testing must be configured
+    return Boolean(
+      (options.codeCov || options.codeCovTokenSecret) &&
+      this.testRunner?.initialized,
```
mrgrain: we should avoid this and inside this class we always want a concrete component

src/javascript/test-runner-options.ts
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Additional options to pass to the test runner's CLI invocation.
+   *
+   * Each element is passed as a single argument, exactly as given: no shell
+   * parses these, so a flag and its value need separate elements
+   * (`["--foo", "bar"]`, not `["--foo bar"]`).
+   *
+   * @default - no extra options
+   */
+  readonly extraCliOptions?: string[];
```
mrgrain: make this per test runner even if duplicated.

src/javascript/test-runner-options.ts
```diff
@@ -0,0 +1,173 @@
...
+   * @default true
+   */
+  readonly junitReporting?: boolean;
+
+  /**
+   * Preserve the default reporter when additional reporters are added.
+   *
+   * @default true
+   */
+  readonly preserveDefaultReporters?: boolean;
```
mrgrain: these feel very much test runner specific, even if both jest and node native have them

src/javascript/test-runner-options.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Path to the JSON configuration file for the test runner.
+   *
+   * @default - no separate config file
+   */
+  readonly configFilePath?: string;
```
mrgrain: Similar, while I get that pretty much all test runners will have a config having them separate will allow better doc blocks etc.

src/javascript/util.ts
```diff
@@ -2,9 +2,55 @@ import { existsSync, readFileSync } from "fs";
...
 import { findUp } from "../util";
 
+/**
+ * Excludes a set of patterns from both the project's `.gitignore` and
+ * `.npmignore` files (if present).
+ */
+export function excludeFromVcs(project: NodeProject, ...patterns: string[]) {
+  project.gitignore.exclude(...patterns);
+  project.npmignore?.exclude(...patterns);
+}
```
mrgrain: this doesn't do what it says (it's more than eclude from VCS), also excluding from git already npmignores. Then there is also the generic [addPackageIgnore](https://projen.io/docs/api/projen#projen.Project.addPackageIgnore)

test/javascript/node-project.test.ts
```diff
@@ -793,6 +793,17 @@ test("codecov upload not added to github release workflow", () => {
   expect(workflow).not.toContain("uses: codecov/codecov-action@");
 });
 
+test("codecov upload uses token secret when codeCovTokenSecret is provided", () => {
```
mrgrain, line 796: Cool, but unrelated?

src/javascript/private/test-runner-command.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Installs the following npm scripts:
+ *
+ * - `test`, intended for testing locally and in CI. Will update snapshots
+ *   unless `updateSnapshot: UpdateSnapshot.NEVER` is set.
+ * - `test:watch`, intended for automatically rerunning tests when files change.
+ * - `test:update`, intended for testing locally and updating snapshots to
+ *   match the latest unit under test. Only available when
+ *   `updateSnapshot: UpdateSnapshot.NEVER`.
+ */
+export function configureTestCommand(
```
mrgrain: this should go into a private/<name>.ts file so it's not exported.

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    const reporters: string[] = [];
+    const reporterDestinations: string[] = [];
+    const addReporter = (reporter: string, destination: string) => {
+      reporters.push(reporter);
+      reporterDestinations.push(destination);
+    };
+
+    if (this.preserveDefaultReporters && this.coverageText) {
+      addReporter("spec", "stdout");
```
mrgrain: `stdout` is not a documented reporter here: https://nodejs.org/api/test.html#test-reporters

any source for it?

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,271 @@
...
+    const collectCoverage = options.collectCoverage ?? true;
+    const coverageDirectory = options.coverageDirectory ?? "coverage";
+    this.coverageDirectory = collectCoverage ? coverageDirectory : undefined;
+
+    const reporters: string[] = [];
+    const reporterDestinations: string[] = [];
+    const addReporter = (reporter: string, destination: string) => {
+      reporters.push(reporter);
+      reporterDestinations.push(destination);
+    };
```
mrgrain: I think this screams for a dedicated reporters config instead of the current collection of booleans to allow arbitrary reporters.

Plus add/remove/list api to dynamically add/remove them.

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,271 @@
...
+      updateArgs: this.buildUpdateArgs(),
+      watchArgs: this.buildWatchArgs(),
+    });
+  }
+
+  /**
+   * Adds a test match pattern.
+   * @param pattern glob pattern to match for tests
+   */
+  public addTestMatch(pattern: string) {
```
mrgrain: probably also removal

mrgrain (changes requested):
Let's start here. There is some more around the FutureComponent that I don't quite like yet but might have to play with this myself.

src/javascript/node-native-test.ts
eduardomourar: Not in the public documentation. For this one, I had to check Node.js' source code. More specifically these tests: https://github.com/nodejs/node/blob/a382c1c8edc7a43f6555898f4dd6b85cbdce0be6/test/parallel/test-runner-reporters.js#L46-L64

src/javascript/jest.ts
eduardomourar: The refactor here was only necessary to help me visualize the test runner generic part vs jest specific one. As you pointed out that some of the options should not be generic, I guess most of the changes here will go away.

test/javascript/node-project.test.ts
eduardomourar: It is to exercise the change for `NodeProject.useCodecov` private function.

src/javascript/node-project.ts
eduardomourar: Alright. So I will remove the initialized method and always assume it to be concrete

src/javascript/node-native-test.ts
eduardomourar: done

src/javascript/node-native-test.ts
eduardomourar: done

src/javascript/javascript-test-runner.ts
eduardomourar: I have redone the whole `JavaScriptTestRunner` class alignining better with the approach taken by the `TypeScriptScriptRunner`.

eduardomourar:
As discussed, I have created the following PR: #4924

PR #4900, opened 2026-08-28
feat(typescript): add typecheckTests option

mrgrain:
Adds a `typecheckTests` option to `TypeScriptProject`, defaulting to `false`. When enabled, it prepends a type-check step to the `test` task:

```
tsc --noEmit -p <tsconfigDev>
```

Being part of the existing `test` task, the check also gates `build` and CI.

Defaulting to `false` because this could prevent builds from projects that legitimately decided they want this behavior. The causation chain is somewhat complicated though: When `ts-jest` runs with a tsconfig that sets `isolatedModules: true` it automatically switched to transpile-only mode. Type errors in test files then fail neither `jest` nor eslint. This is documented in `ts-jest`, but might not be obvious to users.

`projen` falls into the above case. So we enable the option for projen itself. That surfaced three existing errors in `test/inventory.test.ts`: `throwIfNotRenderable` was typed as taking a `ProjectOption` but reads `simpleType`, which is only declared on `InventoryProjectOption`. All call sites already pass inventory options, so the parameter type is widened accordingly.

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -144,6 +144,7 @@ const project = new TypeScriptProject({
     },
   },
 
+  typecheckTests: true,
   jestOptions: {
     // makes it very hard to iterate with jest --watch
     coverageText: false,
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -564,6 +564,15 @@
       "description": "Run tests",
       "steps": [
         {
+          "name": "Type-check the test suite",
+          "execArgs": [
+            "tsc",
+            "--noEmit",
+            "-p",
+            "test/tsconfig.json"
+          ]
+        },
+        {
           "execArgs": [
             "jest",
             "--passWithNoTests",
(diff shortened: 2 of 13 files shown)


.projenrc.ts
```diff
@@ -296,6 +296,14 @@ project.github
 
 setupCheckLicenses(project);
 
+// We use ts-jest in tests in transpile-only mode, so type errors in test files never fail.
+// Instead add an explicit type-check over the test tree.
+const typecheckTests = project.addTask("typecheck:tests", {
```
iliapolo: Shouldn't we add this then to the default configuration of `TypeScriptProject` project? Our consumers may be in the same position.

iliapolo (changes requested):

iliapolo (approved):

mrgrain:
@mergifyio queue

PR #4901, opened 2026-08-28
feat(typescript): stable transformTypes option for the nodejs runner using amaro

mrgrain:
If you use the native Node.js TypeScript runner with `experimentalTransformTypes`, your projects break on Node.js 26: the `--experimental-transform-types` flag was removed, so the synthesized commands no longer run. This also affects every CDKTN TypeScript app, which used this option by default. The option is already deprecated in a previous release, but until now there was no replacement — and without it, TypeScript-only syntax like enums and namespaces cannot be executed.

This PR adds a stable `transformTypes` option to `TypeScriptRunner.nodejs()` that works on all supported Node.js versions. It uses [amaro](https://github.com/nodejs/amaro), the same TypeScript transformer Node.js uses internally, loaded via `--import=amaro/transform`. You don't need to install anything yourself: the `amaro` package is added to your project automatically. Stack traces keep pointing at your original TypeScript source because the command enables `--enable-source-maps`, as amaro recommends.

To migrate, replace `experimentalTransformTypes: true` with `transformTypes: true`. CDKTN TypeScript apps need no changes at all: the default runner now uses the new option, so the synthesized `cdktf.json` app command becomes `node --enable-source-maps --import=amaro/transform src/main.ts` on the next `projen` run. If you stay on the deprecated option, its behavior is unchanged and keeps working on Node.js versions before 26; if both options are set, `transformTypes` wins.

Tested via unit tests covering the new option, its precedence over the deprecated one, and the updated CDKTN app command, plus a clean jsii compile and lint run.

---

### DIFF
--- a/src/typescript/typescript-runner.ts
+++ b/src/typescript/typescript-runner.ts
@@ -171,6 +171,11 @@ export class TypeScriptRunner extends ScriptRunner {
    * reserved by `constructs.Construct` for the construct's tree node.
    */
   public static nodejs(options: NodeRunnerOptions = {}): TypeScriptRunner {
+    if (options.transformTypes && options.experimentalTransformTypes) {
+      throw new Error(
+        "Only one of 'transformTypes' or 'experimentalTransformTypes' may be specified, not both. Remove the deprecated 'experimentalTransformTypes'.",
+      );
+    }
     return new TypeScriptRunner("node", options);
   }
 
--- a/src/typescript/typescript-runner.ts
+++ b/src/typescript/typescript-runner.ts
@@ -83,10 +83,23 @@ export interface NodeRunnerOptions {
 
   /**
    * Whether to also enable `--experimental-transform-types`.
-   * @deprecated This flag has been removed from Node.js 26
+   * @deprecated This flag has been removed from Node.js 26. Use `transformTypes` instead.
    * @default false
    */
   readonly experimentalTransformTypes?: boolean;
+
+  /**
+   * Whether to enable transformation of TypeScript-only syntax (e.g. enums, namespaces).
+   *
+   * Uses `amaro` (the TypeScript transformer used internally by Node.js) as an
+   * external loader via `--import=amaro/transform`. Adds a dependency on the
+   * `amaro` package and enables `--enable-source-maps` to preserve accurate
+   * stack traces.
+   *
+   * @see https://github.com/nodejs/amaro
+   * @default false
+   */
+  readonly transformTypes?: boolean;
 }
 
 /**
@@ -119,6 +132,7 @@ type RunnerKind = "ts-node" | "tsx" | "node";
 interface RunnerState extends TypeScriptRunnerOptions {
   readonly swc?: boolean;
   readonly experimentalTransformTypes?: boolean;
+  readonly transformTypes?: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/typescript/typescript-runner.ts
```diff
@@ -249,7 +263,11 @@ export class TypeScriptRunner extends ScriptRunner {
     const steps: TaskStep[] = [];
     const dependencies: DependencyRequest[] = [];
 
-    if (this._options.experimentalTransformTypes ?? false) {
+    if (this._options.transformTypes ?? false) {
+      dependencies.push({ name: "amaro" });
```
eduardomourar, line 272: `cdktn` and `constructs` are being added as runtime dependencies and this one is being done as devDependency. Was that on purpose?

eduardomourar (approved):
I have performed the testing using Node.js v24 in current project with CDKTN app. It is working as expected by importing `amaro`. It successfully synthesize and deploys the project.

src/typescript/typescript-runner.ts
mrgrain: Yes, but more because the task runner contract currently just is to add these as devDep. `ts-node` and `tsx` get the same treatment. 

How do`cdktn` and `constructs` come in here?  E.g. `cdk init` puts `tsx` and `aws-cdk` in [devDeps as well](https://github.com/aws/aws-cdk-cli/blob/7a77baf5587a81f1e4aa7133f6c85a148eb64a6d/packages/aws-cdk/lib/init-templates/app/typescript/package.json). Not that this is an authoritative source.

eduardomourar: It makes sense. I did not mean in the `TypeScriptRunner`, but in `CdktnTypeScriptApp` here: https://github.com/projen/projen/blob/370f94c40f8b998797718b0ff4f1c69e56bb448d/src/cdktn/cdktn-app-ts.ts#L71-L74 we are setting the dependencies as RUNTIME.

In the project I am consuming CDKTN component that contains both Lambda function code and IaC, I am having to override those so that `cdktn` and `constructs` are part of devDependencies. I think that is as expected because the `CdktnTypeScriptApp` assumes you are initializing an IaC-only project.

mrgrain: In CDK we made this distinction a long time ago: The app is what production is. You should be able to execute the compiled JS file and for that you need `constructs` and `aws-cdk-lib`. All the variations on how one could also run the app we considered Dev/build time. So that includes `aws-cdk` as well as `tsx`. 

Of course with modern approaches to execute TS directly this is a very arbitrary distinction. I can also see a good argument to say `aws-cdk` binary is an essential part. Because in practice for an app (as opposed to a library) it doesn't really matter that much. Most people will just install both anyway.

Up to you how you define this for cdktn.

eduardomourar: I do think we can follow the same approach as taken in CDK. No issues from my side. It was more curiosity than anything else.

src/typescript/typescript-runner.ts
```diff
@@ -249,7 +263,11 @@ export class TypeScriptRunner extends ScriptRunner {
     const steps: TaskStep[] = [];
     const dependencies: DependencyRequest[] = [];
 
-    if (this._options.experimentalTransformTypes ?? false) {
+    if (this._options.transformTypes ?? false) {
```
iliapolo, line 271: Why not throw when both `transformTypes` and `experimentalTransformTypes` are used? We want to move people away from the deprecated option. Also creates the possibility of:

```ts
{
  transformTypes: true,
  experimentalTransformTypes: false/true
}
```

Somewhat confusing.

iliapolo (changes requested):

iliapolo (approved):

PR #4903, opened 2026-08-29
fix(cli): show fixed args in task inspection

### LINKED ISSUE #3518: Should `args` be returned by the task --inspect CLI command?
If you set up a task with `args` running the CLI command `--inspect` for the task will return the steps but not the args for the task or step. It seems like using `--inspect` leaves a pretty large gap in understanding the task if it doesn't return the args.

https://github.com/projen/projen/blob/6a6ecdbc25f8688c33004c1536865529ae4824ce/src/cli/tasks.ts#L52

Sion612:
Fixes #3518

## Summary

- Include fixed step arguments in task `--inspect` output.
- Serialize argument arrays as JSON so whitespace and shell-significant characters remain unambiguous.
- Keep spawned-step args associated with the spawning step.

## Testing

- `node ./projen.js compile`
- `node ./projen.js test test/cli.test.ts --runInBand`
- `node ./projen.js test test/cli/run-task.test.ts --runInBand`
- `node ./projen.js eslint`
- `git diff --check`

---

### DIFF
--- a/src/cli/cmds/run-task.ts
+++ b/src/cli/cmds/run-task.ts
@@ -68,13 +68,20 @@ export function discoverTaskCommands(runtime: TaskRuntime, ya: yargs.Argv) {
     for (const step of task.steps ?? []) {
       if (step.spawn) {
         writeln(`- ${chalk.bold(step.spawn)}`);
-        inspectTask(step.spawn, indent + 2);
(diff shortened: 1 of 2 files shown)


mrgrain (approved):
Nice, thank you!

PR #4904, opened 2026-08-29
feat(release): control whether a GitHub release is marked as latest

### LINKED ISSUE #3273: Allow github release to be explicitly set as latest or explicitly not be latest
`gh release create` has three different modes for "latest":

- default; not specified - the latest release will be determined by date and version (see https://cli.github.com/manual/gh_release_create)
- `--latest` - this release will be set as the latest release
- `--latest=false` - this release will NOT be set as the latest release (see <https://github.com/cli/cli/issues/6963>)

projen currently only allows the default behavior.

We should change Publisher to accept all 3 variations of it. We should then expose this option via the `Release` component and `releaseBranches`. 

--- 

Use case: I have a package with multiple release branches for older, but still supported versions. Currently releases for these older versions are still marked as "Latest" whenever they occur, despite them definitely not being the latest or recommend version to use.

pttydou:
Fixes #3273

When releasing from multiple branches, GitHub marks whichever release is newest as "latest". So a patch for an older major version ends up labeled as the latest release. This adds `githubReleaseLatest` to the project and to each entry in `releaseBranches` (and a `latest` option on `publishToGitHubReleases()`). Branches can now explicitly opt-out, force "latest" or keep to automatic behavior.

Took the opportunity to refactor the very complex release script, with a simpler `gh release view || gh release create` chain. Since the flags are known at synthesis time, we can inject them as is. Re-running a release remains a no-op if the release already exists.

---


mrgrain (commented):
Hey @pttydou thanks for getting this started! I didn't want us to keep growing this hard to understand, complex and non-portable script further, so I refactor it to something much simpler. Keeping your API though.

mrgrain (approved):

PR #4906, opened 2026-08-30
feat(javascript): support package.json config

### LINKED ISSUE #2973: Support  `config` field of `package.json`
Thanks for creating a very useful library.
It's very useful, but I had a little trouble using it.

I would like to edit the config field in `package.json` but it's impossible.
At glance, I guess the root cause is that there are no properties related to config in `NodePackageOptions`.
Also, it is not possible to configure the options that inherit from NodePackageOptions with respect to config, so it seems that support is needed there as well.

I would like to support for it in projects related to JavaScript and TypeScript.

Sion612:
## Summary

- add `config` to `NodePackageOptions`
- preserve nested JSON-compatible values in the synthesized `package.json`
- expose the option through projects that inherit `NodePackageOptions`

## Testing

- `npm ci`
- `node ./projen.js compile`
- `node ./projen.js test test/javascript/node-package.test.ts --runInBand` (96 tests; local Jest/Corepack caches redirected off the full system drive)
- `node ./projen.js test test/inventory.test.ts --runInBand` (2,539 tests)
- `node ./projen.js compat`
- `node ./projen.js docgen`
- `node ./projen.js eslint`
- `git diff --check`

Fixes #2973

---


mrgrain (approved):

mrgrain:
@mergifyio queue

PR #4924, opened 2026-09-07
feat(javascript): add `NodeNativeTestRunner` component

eduardomourar:
Adds `NodeNativeTestRunner`, a standalone component for running tests via Node's built-in test runner (`node --test`). Self-contained on purpose — no changes to `Jest` or `NodeProject`, some duplication with `Jest` is fine for now. Wiring it up as a `testRunner` option is a separate PR.

Covers the generated config file, coverage/reporters, and the `test`/`test:update`/`test:watch` tasks, plus a `transformTypes` option to use `amaro` for TS-only syntax.

Also a small fix to `NodeConfigFile` so lazily-computed config values (like the live reporters list) get resolved before being written out.

---


src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+    return this._patterns.size > 0
+      ? Array.from(this._patterns)
+      : this._defaultValue;
+  }
+}
+
+export type ReporterKind = "dot" | "junit" | "lcov" | "spec" | "tap";
+
+export type FilePathDestination = string & {};
+export type Destination = "stdout" | "stderr" | FilePathDestination;
```
mrgrain: This needs to be an enum like class

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+   * The name/kind of the reporter.
+   */
+  readonly name: ReporterKind;
+  /**
+   * Where the reporter's output is written.
+   *
+   * @default "stdout"
+   */
+  // As presented here:
+  // https://github.com/nodejs/node/blob/4215cc35e25c44f9f4fea5a4541afc862db7ef0a/test/parallel/test-runner-reporters.js#L46-L77
```
mrgrain:
```suggestion
```
`//` comments cannot be between docblock and code

src/javascript/node-native-test-runner.ts
```diff
@@ -0,0 +1,560 @@
...
+  // https://github.com/nodejs/node/blob/4215cc35e25c44f9f4fea5a4541afc862db7ef0a/test/parallel/test-runner-reporters.js#L46-L77
+  readonly destination: Destination;
+}
+
+/**
+ * Holds the set of reporters configured for a `NodeNativeTest` component,
+ * backing the `test.testReporter`/`test.testReporterDestination` fields of
+ * the generated Node.js configuration file.
+ */
+export class NodeReporters extends Component {
```
mrgrain, line 166: this doesn't do anything aside from holding data, so it should not be a component or construct. But in fact the pattern we have is to have these on the component as `addReporter` etc.

src/javascript/node-native-test-runner.ts
```diff
@@ -0,0 +1,560 @@
...
+  /**
+   * Never update snapshots in "test" task and create a separate "test:update" task.
+   */
+  NEVER = "never",
+}
+
+/**
+ * Glob patterns matching the files that contain tests.
+ */
+class TestMatch {
```
mrgrain, line 73: this doesn't do anything aside from holding data, so it correctly it is not be a component But in fact the pattern we have is to have these on the component as `addPattern` etc. Refer to Jest for naming.

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+   * Each element is passed as a single argument, exactly as given: no shell
+   * parses these, so a flag and its value need separate elements
+   * (`["--foo", "bar"]`, not `["--foo bar"]`).
+   *
+   * @default - no extra options
+   */
+  readonly extraCliOptions?: string[];
+
+  /**
+   * Preserve the default reporter when additional reporters are added.
```
mrgrain: List the default reporters

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+   * Additional entries for the `nodeOptions` section of the generated
+   * configuration file (e.g. `enableSourceMaps`, `disableWarning`).
+   *
+   * @default - no additional node options
+   */
+  readonly nodeOptions?: NodeConfigSchemaNodeOptions;
+
+  /**
+   * Escape hatch to add or override any value in the `test` section of the
+   * generated configuration file.
```
mrgrain: Use the "Additional config" style you use for `nodeOptions`

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+
+/**
+ * Wires up Node's built-in test runner (`node --test`) as a self-contained
+ * component: the generated configuration file (coverage, reporters, global
+ * setup, etc., written via `--experimental-config-file`), the
+ * coverage/report directories' gitignore/npmignore entries, the live
+ * `testMatch`/`reporters` collections that feed into it, and the
+ * "test"/"test:update"/"test:watch" tasks themselves.
+ */
+export class NodeNativeTest extends Component {
```
mrgrain: NodeNativeTest or NodeNativeTestRunner? WDYT?

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+  readonly testConfig?: NodeConfigSchemaTest;
+}
+
+/**
+ * Wires up Node's built-in test runner (`node --test`) as a self-contained
+ * component: the generated configuration file (coverage, reporters, global
+ * setup, etc., written via `--experimental-config-file`), the
+ * coverage/report directories' gitignore/npmignore entries, the live
+ * `testMatch`/`reporters` collections that feed into it, and the
+ * "test"/"test:update"/"test:watch" tasks themselves.
```
mrgrain: Single short sentence, empty line, more details.

don't use terms like "Wires up" don't list details, keep high level.

src/javascript/node-native-test.ts
```diff
@@ -0,0 +1,560 @@
...
+  /**
+   * The directory where Node outputs its coverage files.
+   */
+  public readonly coverageDirectory?: string;
+
+  /**
+   * Whether snapshots are updated in task "test", or in a separate
+   * "test:update" task.
+   */
+  public readonly updateSnapshot: NodeTestUpdateSnapshot;
```
mrgrain: Do these need to be public? (We can change it later if needed)

src/javascript/node-native-test-runner.ts
```diff
@@ -0,0 +1,560 @@
...
+    ];
+
+    // `testMatch` is a live collection (`add`/`remove` can still be called
+    // after construction), so its resolution is deferred with `lazyExecArgs`
+    // rather than read eagerly here.
+    const testMatchArgs = () => this.testMatch.deferred();
+
+    if (this.updateSnapshot === NodeTestUpdateSnapshot.ALWAYS) {
+      this.project.testTask.execArgs(
+        lazyExecArgs(() => [
```
mrgrain, line 529: not really a fan of this? How do other components solve this?

src/javascript/node-config-file.ts
```diff
@@ -52,7 +53,10 @@ export class NodeConfigFile extends Component {
     this.config = config;
 
     this.file = new JsonFile(project, filePath, {
-      obj: () => toJson_NodeConfigSchema(this.config),
+      // resolve first so that any lazily-computed values (e.g. a function
+      // returning the current state of a live collection) in `this.config`
+      // are evaluated before the generated schema mapper reads them.
+      obj: () => toJson_NodeConfigSchema(resolve(this.config)),
```
mrgrain: this smells. Is it strictly needed for this PR? Do other genarated configs have the same issue? If not why? If yes, I'd rather solve this generally.

mrgrain (changes requested):

src/javascript/node-native-test.ts
eduardomourar: This was just a comment that I did not want to be present in the public API page. How can I do that for a property like this one?

src/javascript/node-native-test.ts
eduardomourar: If there is no issue in having a long name for the component (maybe some restrictions from JSII), we can change to `NodeNativeTestRunner`. I actually thought about the longer name first, but I did not remember seeing it within codebase.

eduardomourar: I checked that we have `VsCodeRecommendedExtensions`, so I will change it to `NodeNativeTestRunner` and change the filename to reflect that as well.

src/javascript/node-native-test.ts
eduardomourar: No need to public for now. So I will change that.

src/javascript/node-native-test-runner.ts
eduardomourar: That was changed based on your comment here: https://github.com/projen/projen/pull/4890#discussion_r3880827267. No matter what, I will stop extending `Component`. Do you want me to keep the class, but keep it internal to this file and only expose the `addReporter` and other methods through the component itself?

src/javascript/node-native-test-runner.ts
eduardomourar: I will keep the `TestMatch` class and private to this file. I already followed the Jest naming (we have `addTestMatch` method). Do want me to follow the convention from `IgnoreFile` (with `addPatterns` method, etc), instead?

src/javascript/node-native-test.ts
mrgrain: afterwards probably

mrgrain: also you can use `// @see <URL>`

src/javascript/node-native-test-runner.ts
mrgrain: `addTestMatch` is good 👍🏻

src/javascript/node-native-test-runner.ts
eduardomourar: Same trick `ProjenTaskRunner` uses on itself. https://github.com/projen/projen/blob/4cc1f08d5e5e886c96c52dddc8eff03b147fd10a/src/task-runner.ts#L39-L42
https://github.com/projen/projen/blob/4cc1f08d5e5e886c96c52dddc8eff03b147fd10a/src/tasks.ts#L164-L176
`resolve()` already calls any function it finds mid-tree, that's how those work. No Lazy helper exists in projen to use instead, so casting is the idiom we've got. Can drop the `lazyExecArgs` name and just inline `(() => [...]) as any` if you'd rather match that style.

src/javascript/node-config-file.ts
eduardomourar: Yeah, it's needed. `toJson_NodeConfigSchema` is a dumb generated mapper, it chokes if it gets a raw function instead of resolved data, so resolve has to run before the mapper, not after (which is all JsonFile gives you by default). Turns out `PyprojectTomlFile` hit this exact issue and solved it cleaner: https://github.com/projen/projen/blob/4cc1f08d5e5e886c96c52dddc8eff03b147fd10a/src/python/pyproject-toml-file.ts#L20-L24
I'll switch to that. Not a general JsonFile bug btw. I checked biome/pnpm-workspace/uv, none of them have live mutable state feeding their mappers so they never hit it.

src/javascript/node-native-test-runner.ts
mrgrain: Yes, my comment didn't mention component. Make it a class, make it private, expose only via `NodeNativeTestRunner`

mrgrain (changes requested):
One last thing I believe. (`NodeReporters`)

mrgrain (approved):

